Issue #656, opened 2014-05-20
Divide scalar by a vector


alyst:
AFAIK, at the moment STAN does not support division of a scalar by a vector (i.e. a/X=(a/X[1], a/X[2], .., a/X[N])). Is it possible to enable it?
While this is not a conventional linalg, it's supported in this way in e.g. R, which results in leaner code and better vectorization.

PR #660, opened 2014-05-21
Support scalars in elementwise products

alyst:
Supporting scalars in elementwise products at both positions (but not simultaneously) to address #656.
At the moment the pull request misses scalar support for `.*`.
I will add it, once the code for `./` would get ok after the review.


JohannesBuchner:
"elementwise quotient of \farg{y} and \farg{x}}" -- since this is somewhat ambiguous, could you please add in the documentation that this means y/x (which I find unintuitive).

bob-carpenter:
@JohannesBuchner:  To not lose this comment, I opened an issue in the current doc repo:

https://github.com/stan-dev/docs/issues/142

Issue #1069, opened 2014-10-07
partial initialization


bob-carpenter:
Allow users to provide partial initializations and take the rest randomly.  

This should be done a variable at a time --- that is, we won't allow a multivariate quantity to be partially initialized.

It would also be nice to expose the transforms we use along with their Jacobians to users in applications.

syclik:
@maverickg, it's merged -- feel free to update RStan to handle partial inits.

maverickg:
@syclik  thanks. 

rstan has a branch to use it now.

mathDR:
Was pystan ever updated to use partial inits?  Anyone know?

PR #1082, opened 2014-10-13
Feature/partial inits (issue #1069)

maverickg:
#### Summary:

Allow to generate initial values randomly for some parameters if not all are specified by a user.

#### Intended Effect:

1. In the case of manually providing initial values, previously a user must specify inits for _all_ parameters.  This PR makes it possible to provide inits for some parameters by using random partial inits for other parameters. 

2. Provide an option to disable random partial inits

#### How to Verify:

More unit tests are added to `initialize_state_test.cpp`, which can be run as follows.  In particular, a new mock model mimics the scenarios where 

1. parameters need transformation
2. the number of elements for a parameter in the constrained space is different from that in the unconstrained space 

```
$ ./runTests.py src/test/unit/common/initialize_state_test.cpp
```

#### Side Effects:

This PR would break interfaces since two arguments are needed for the new API:

1. to specify the range when partial inits are in a file as in CmdStan
2. to disable random partial inits

So the interfaces such as (Cmd/Py/R)Stan needs to add options and update call of function `initialize_state`.

#### Documentation:

None in the manual. 

#### Reviewer Suggestions:

Any


maverickg:
Trying to clean up. 

So should I bring this up to date (which should not be hard) and get it merged?  Or should we wait until there are better functionalities in the model?

syclik:
Hi Jiqiang,

I think we should wait... is that ok? If you want it in now, I can go ahead
and get it migrated. But I still think we should wait to do it correctly.


Daniel

maverickg:
I would appreciate it if you get it migrated, rstan users would benefit from this feature.

syclik:
ok. Mike, I'm going to tackle this. We can do it more correctly later.

bob-carpenter:
??  That message was from Jiqiang, not Michael.

What's the motivation for waiting?  As long as the
functionality is right, we can always replace it later
with a better implementation.

- Bob

syclik:
Thanks, Jiqiang! I got caught up this week. I just had to fix some cpplint warnings.

PR #1191, opened 2015-01-02
Separate math/ode from math [POC]

jpritikin:
For models where ODE is not required, it would be nice to avoid including all the ODE support headers. This change is worth about 110k lines of headers (600 header files).

Obviously, this change cannot be merged because I don't know where to add back the include ode.hpp that I remove here. Perhaps a more experienced developer can point out where to add it back?

### DIFF
--- a/src/stan/math.hpp
+++ b/src/stan/math.hpp
@@ -7,7 +7,6 @@
 #include <stan/math/rep_matrix.hpp>
 #include <stan/math/constants.hpp>
 #include <stan/math/functions.hpp>
-#include <stan/math/ode.hpp>
 #include <stan/math/meta.hpp>
 
 #endif


syclik:
Josh, I'm closing the pull request. Can you open an issue instead? 

The higher level includes, by design, include all headers in the folder directly underneath it. 

If we wanted to do this, it will require us moving ode out.

bob-carpenter:
math.hpp already depends on Boost and Eigen.   It think ode.hpp will
be more of an issue when it has its own dependencies, as it will be
soon when we swap in some new integrators.  

We have an ongoing issue to only include what we use in each
header file, which should make it easier going forward to
do more minimal custom includes.  But we're not there yet.

Hopefully this will be easier when math.hpp gets cleaned up a bit,
which I'm working on now --- as is, there are some unnecessary higher-level
includes.  When there's only a few libraries under math, it'll be
easier to just include those.

- Bob

PR #1197, opened 2015-01-03
Feature/issue 778 math includes

bob-carpenter:
#### Summary

Clean up the stan/math.hpp includes, which had files too high up the file hierarchy.

#### Intended Effect:

Rationalize the includes so that including functions, matrix, etc. does the right thing.

#### How to Verify:

The current unit tests and model build tests should suffice.

#### Side Effects:

Anyone using Stan directly from C++ may have to change some includes if they included parts of math.hpp.

#### Documentation:

Nothing user facing.

#### Reviewer Suggestions:

Daniel --- he loves this kind of re-org!


syclik:
It looks like the constants and rep_* tests were removed. I think they should be moved to the right level and included. 

@bob-carpenter, I can take care of it.

bob-carpenter:
I thought I moved them in Git, but they don't seem to have gotten
added.  I just pushed the versions that should have the right names.
I ran all the unit tests and header tests here.

The reason I missed it is that I'm now getting tons of false positives
in git status from .cpp versions in test-models/good.  I'll try to clean
them out by hand and hope they don't come back.

- Bob

Issue #1206, opened 2015-01-06
Upgrade libraries


syclik:
- [x] upgrade boost to version 1.55.0
- [x] remove Eigen 3.2.0 folder (we have already upgraded to 3.2.2)

Issue #1205, opened 2015-01-06
check_multiplicable incorrectly allows Mx0 or 0xN matrices


syclik:
This should be prevented. I'm working on a fix on branch #685.

bob-carpenter:
I guess it wouldn't make sense to multiply (M x 0) and (0 x N)
because there's no way to get an (M x N).  

What's the general policy on M x 0 matrices?  Are they even
constructible in Eigen?

- Bob

Issue #1204, opened 2015-01-05
reader's cholesky_corr not compiling


syclik:
We didn't have a test for this, so this wasn't caught. I'm fixing it with the pull request for #685.

syclik:
Btw, this shouldn't have been an bug in 2.5.0. This was introduced after 2.5.0. It's just been sitting in develop without a test. I added a test and a fix.

PR #1207, opened 2015-01-07
Feature/issue 685 separate errors

### LINKED ISSUE #1201: check_consistent_sizes not correct for 0-length vector-like arguments
If you pass in a scalar and a 0-length vector, check_consistent_sizes fails.

I've got a patch for this on the branch for #685.

syclik:
#### Summary:

This pull request changes the error handling behavior of the math library. Before this pull request, all errors had the same behavior: throw `std::domain_error`. This pull request separates the errors into three different kinds of errors and throws different exceptions for the three errors. 

The motivation is that certain errors are numerical errors and will change between different evaluations of the log probability for a particular Stan program. Other errors are programming errors and will not change, such as accessing elements out of range or providing matrices of incorrect sizes for functions. These last errors should be treated differently than runtime errors that may be corrected with the next evaluation.

Additionally:
- fixes `check_consistent_sizes` to allow for 0-length vectors (#1201)
- fixes `cholesky_corr` function within `stan/io/reader.hpp` so that it compiles (#1204)
- fixes behavior of `check_multiplicable` (#1205)
- implements Eigen NumTraits for `fvar<T>` (#1206)

#### Intended Effect:

1. Rather than all errors in the math library resulting in a `std::domain_error` exception being thrown, the math library throws:
    - `std::out_of_range` when accessing an element of a matrix or vector out of range
    - `std::invalid_argument` for function arguments that just don't work for the function. Most of these are matrices of the wrong dimension.
    - `std::domain_error` for all other errors, which should be runtime errors.
2. Added additional methods for throwing `std::out_of_range` and `std::invalid_argument` exceptions with corresponding tests.
3. Documented all of `src/stan/error_handling` to indicate the condition and the exception thrown.
4. Updated all tests.
5. Fixes #1201, #1204, #1205.

#### How to Verify:

Inspect the code.
Run the unit tests.

#### Side Effects:

None. The sampling code within Stan catches `std::exception`, so the behavior to the user will be the same. At some point in the future, we will want to reject within sampling if it is a `std::domain_error` exception and stop sampling if we see a `std::invalid_argument` or `std::out_of_range` exception.

#### Documentation:

No changes to the manual necessary.

Code level documentation was added for error handling.

#### Reviewer Suggestions:

Anyone. Many files have changed with this pull request, so it may take a little while to review, but a spot check should be sufficient.

#### Fixes:
- [x] Left over FIXME's in error_handling. Fix them.
- [x] fix check_pos_definite so it includes check_square, check_positive_index, check_symmetric
- [x] remove doc about what functions don't do
- [x] check_nonzero_size.hpp: change size_t() to 0
- [x] change types of check_range.hpp to int, not size_t
- [x] rename check_positive_index to check_positive_size
- [x] check_std_vector_index.hpp: drop extra parens
- [x] check_consistent_size.hpp: remove "vectorized" from error message
- [x] src/stan/math/matrix/mdivide_right_spd.hpp: transpose using Eigen
- [x] move `stan::error_handling` namespace to `stan::math` namespace
- [x] remove the type information from error messages
- [x] remove static cast on function calls
- [x] fix check_size_match to have a more useful message


src/stan/error_handling/matrix/check_cov_matrix.hpp
```diff
@@ -16,22 +16,26 @@ namespace stan {
...
+     * @param y Matrix to test
+     *
+     * @return <code>true</code> if the matrix is a valid covariance matrix
+     * @throw <code>std::invalid_argument</code> if the matrix is not square
+     *   or if the matrix is 0x0
+     * @throw <code>std::domain_error</code> if the matrix is not symmetric, 
+     *   if the matrix is not positive definite, 
+     *   or if any element of the matrix is nan
      */
     // FIXME: update warnings
```
bob-carpenter: Is there still something to fix?

src/stan/error_handling/matrix/check_cov_matrix.hpp
```diff
@@ -16,22 +16,26 @@ namespace stan {
...
                                  const std::string& name,
                                  const Eigen::Matrix<T_y,Eigen::Dynamic,Eigen::Dynamic>& y) {
-      check_size_match(function, 
-                       "Rows of covariance matrix", y.rows(),
-                       "columns of covariance matrix", y.cols());
-      check_positive(function, "rows", y.rows());
+      check_square(function, name, y);
+      check_positive_index(function, name, "rows", y.rows());
       check_symmetric(function, name, y);
       check_pos_definite(function, name, y);
```
bob-carpenter: Shouldn't `check_pos_definite` entail all these other checks?  

In most contexts, we'll be doing something else that checks positive definiteness and don't want to call this directly anyway.

src/stan/error_handling/matrix/check_ldlt_factor.hpp
```diff
@@ -28,9 +38,8 @@ namespace stan {
         msg << "is not positive definite. "
             << "last conditional variance is ";
```
bob-carpenter: What does "conditional variance" have to do with positive definiteness?  Is lack of pos defintieness the only reason this could fail?

bob-carpenter:
1. I'd prefer error handling be in `stan::math` so we don't have to do so many qualifications on the names.  

2.  I don't like general functions printing out in format `f(double)` because I think it'll be confused with the signature for the function.

src/stan/error_handling/matrix/check_matching_dims.hpp
```diff
@@ -17,6 +41,12 @@ namespace stan {
                                     const std::string& name2,
                                     const Eigen::Matrix<T2,R2,C2>& y2) {
       check_size_match(function, 
+                       "Templated rows of matrix 1", R1,
```
bob-carpenter: I don't think we should b;e comparing the template parameters, just the sizes.

src/stan/error_handling/matrix/check_column_index.hpp
```diff
@@ -2,37 +2,53 @@
...
-      std::ostringstream msg;
-      msg << ") must be greater than 0 and less than " 
-          << y.cols();
-      dom_err(function, name, i,
-              "(", msg.str());
+      std::stringstream msg;
+      msg << " for columns of " << name;
+      out_of_range(function, 
+                   static_cast<int>(y.cols()),
+                   static_cast<int>(i),
```
bob-carpenter: We don't need static_cast on the function calls.  I think the others ones are just there to avoid the signed/unsigned comparison warnings.

src/stan/error_handling/matrix/check_multiplicable.hpp
```diff
@@ -3,23 +3,46 @@
...
   namespace error_handling {
 
-    // NOTE: this will not throw if y1 or y2 contains nan values.
+    /**
+     * Return <code>true</code> if the matrices can be multiplied.
+     *
+     * This checks the runtime sizes to determine whether the two
+     * matrices are multiplicable. This allows Eigen matrices,
+     * vectors, and row vectors to be checked.  This function does not
+     * check the values within the matrices for <code>NaN</code>.
```
bob-carpenter: I don't think we need to include information about what functions don't do, such as not checking for NaN.

src/stan/error_handling/matrix/check_nonzero_size.hpp
```diff
@@ -14,33 +14,36 @@ namespace stan {
...
-      typedef typename index_type<T_y>::type size_t;
       if (y.size() > 0) 
         return true;
 
-      dom_err(function, name, size_t(), 
-              "has size ", 
-              ", but must have a non-zero size");
+      using stan::math::index_type;
+      typedef typename index_type<T_y>::type size_t;
+      invalid_argument(function, name, size_t(),
```
bob-carpenter: I don't see what `size_t()` is doing.  Can't we put 0 in there explicitly if you want 0 to come out?

src/stan/error_handling/matrix/check_range.hpp
```diff
@@ -3,36 +3,97 @@
...
+     * @param function Function name (for error messages)
+     * @param name Variable name (for error messages)
+     * @param max Maximum size of the variable
+     * @param index Index to check
+     * @param error_msg Additional error message (for error messages)
+     * 
+     * @return <code>true</code> if the index is within range
+     * @throw <code>std::out_of_range</code> if the index is not in range
+     */
+    inline bool check_range(const std::string& function,
```
bob-carpenter: Should we really have an unsigned type here?  If we pass an integer to it, it will get cast to positive and the message will be confusing.

src/stan/error_handling/matrix/check_size_match.hpp
```diff
@@ -16,14 +32,14 @@ namespace stan {
                                  const std::string& name_j, 
                                  T_size2 j) {
       typedef typename boost::common_type<T_size1,T_size2>::type common_type;
-      if (static_cast<common_type>(i) == static_cast<common_type>(j))
+      if (likely(static_cast<common_type>(i) == static_cast<common_type>(j)))
```
bob-carpenter: When do we have this cast to a common type where `==` itself wouldn't be defined?  That is, can we just get rid of this cast?

src/stan/error_handling/matrix/check_spsd_matrix.hpp
```diff
@@ -15,21 +15,25 @@ namespace stan {
...
+     *   or if the matrix is not positive semi-definite
      */
-    // FIXME: update warnings
     template <typename T_y>
     inline bool check_spsd_matrix(const std::string& function, 
                                   const std::string& name,
                                   const Eigen::Matrix<T_y,Eigen::Dynamic,Eigen::Dynamic>& y) {
       check_square(function, name, y);
-      check_positive(function, "rows", y.rows());
+      check_positive_index(function, name, "rows()", y.rows());
```
bob-carpenter: Why is this called `check_positive_index` when what's being check is not an index but a size?

src/stan/error_handling/matrix/check_std_vector_index.hpp
```diff
@@ -11,28 +12,33 @@ namespace stan {
...
-    template <typename T_y>
+    template <typename T>
     inline bool check_std_vector_index(const std::string& function,
                                        const std::string& name,
-                                       const std::vector<T_y>& y,
+                                       const std::vector<T>& y,
                                        size_t i) {
-      if ((i > 0) && (i <= static_cast<size_t>(y.size())))
+      if ((i >= stan::error_index::value) 
+          && (i < y.size() + stan::error_index::value))
```
bob-carpenter: My preference is to go with the language's built-in precedence and drop the parens around the comparisons.

src/stan/error_handling/matrix/check_symmetric.hpp
```diff
@@ -16,14 +16,20 @@ namespace stan {
     /**
      * Return <code>true</code> if the specified matrix is symmetric
      * 
+     * The error message is either 0 or 1 indexed, specified by
+     * <code>stan::error_index::value</code>.
+     *
      * NOTE: squareness is not checked by this function
```
bob-carpenter: How could a matrix be symmetric without being square?  Shouldn't we also test squareness here?

src/stan/error_handling/out_of_range.hpp
```diff
@@ -0,0 +1,97 @@
...
+  namespace error_handling {
+
+    /**
+     * Throw an out_of_range exception with a consistently formatted message.
+     * 
+     * This is an abstraction for all Stan functions to use when throwing
+     * out of range. This will allow us to change the behavior for all
+     * functions at once.
+     *
+     * The message is:
```
bob-carpenter: Cutting and pasting the messages everywhere seems like it'll make things hard to refactor.  Can you just point somewhere to the formatting instead?

src/stan/error_handling/out_of_range.hpp
```diff
@@ -0,0 +1,97 @@
...
+     * @param function Name of the function
+     * @param max Max
+     * @param index Index
+     * @param msg1 Message to print
+     */
+    inline void out_of_range(const std::string& function,
+                             const int max,
+                             const int index,
+                             const std::string& msg1) {
+      out_of_range(function, max, index, msg1, "");
```
bob-carpenter: Wouldn't it be easier to just take the value of `msg1` to default to `""` in the original definition?  This will work because `msg1` is a constant reference.

src/stan/error_handling/scalar/check_consistent_size.hpp
```diff
@@ -2,35 +2,51 @@
...
-      if (!is_vector<T>::value && x_size == 1)
+      if (is_vector<T>::value && expected_size == stan::size_of(x))
         return true;
       
       std::stringstream msg;
-      msg << ", expecting dimension of either 1 or "
-          << "max_size=" << max_size
+      msg << ", expecting dimension = "
+          << expected_size
           << "; a vectorized function was called with arguments of different "
```
bob-carpenter: We're trying to wean ourselves off saying "vectorizd" because it's so ambiguous.  I'd just drop the word here.

src/stan/io/reader.hpp
```diff
@@ -995,8 +995,10 @@ namespace stan {
        * Cholesky factor for a correlation matrix.
        */
       inline matrix_t cholesky_corr(size_t K) {
+        using stan::error_handling::check_cholesky_factor_corr;
         matrix_t y(matrix(K,K));
```
bob-carpenter: Shouldn't this just be `y(K,K)` to avoid creating a matrix then copying it into the new matrix?  The compiler may be smart enough not to do too much work.  I also can't find `matrix` --- is it a typedef being brought in from elsewhere?

src/stan/math/matrix/mdivide_right_spd.hpp
```diff
@@ -27,12 +27,12 @@ namespace stan {
...
-      stan::error_handling::check_symmetric("mdivide_right_spd", "A", A);
-      stan::error_handling::check_pos_definite("mdivide_right_spd", "A", A);
       stan::error_handling::check_square("mdivide_right_spd", "A", A);
       stan::error_handling::check_multiplicable("mdivide_right_spd", 
                                                 "b", b,
                                                 "A", A);
+      stan::error_handling::check_symmetric("mdivide_right_spd", "A", A);
+      stan::error_handling::check_pos_definite("mdivide_right_spd", "A", A);
       // FIXME: This is nice and general but likely slow.
       return transpose(mdivide_left_spd(A,transpose(b)));
```
bob-carpenter: The transposes should be done with Eigen's expression templates, i.e.,
```
return mdivide_left_spd(A,b.transpose()).transpose().
```
That should minimize copying.  The transpose itself will be trivial compared to solving the system.

bob-carpenter:
Overall, this is great.

I code reviewed all the code, but not all the tests.  Some of the comments inline are more general than just where they appear.  

None of my comments were critical issues --- most were matters of style and/or a bit of efficiency.

src/stan/io/reader.hpp
syclik: `matrix` is actually a method defined in the reader that creates a matrix by reading `K * K` elements from the reader and building a matrix from them. The existing code is correct. If anything, we should rename the `matrix` method.

src/stan/math/matrix/mdivide_right_spd.hpp
syclik: `mdivide_left_spd` won't take a `b.transpose()` because it returns a `Transpose` type, not a `Matrix` type. I believe we should be able to change it once we accept `MatrixBase` as a type for all our functions, if we ever get that far.

src/stan/error_handling/out_of_range.hpp
syclik: I left it doc'ed individually because are only a handful of these functions and when I was looking at it, it was easier to use with the doc next to the function. If we end up with a lot more, we can point.

src/stan/error_handling/matrix/check_ldlt_factor.hpp
syclik: I don't know what this means.
@bgoodri or @mbrubake, thoughts?

bob-carpenter: The old messages were often awkward to comply with Boost's
general error dumper.  I think it makes sense to print all of
the matrix or none of it.  Or maybe all of it up to a fixed
max number of entries?

bgoodri: The last diagonal element of D in L * D * L' is the conditional variance of
the last variable given all the preceeding variables. Since this is a
variance, it cannot be negative if the covariance matrix is positive
definite.

src/stan/error_handling/matrix/check_range.hpp
syclik: I switched this to `int`s -- the GitHub diff isn't showing the line where I changed it, but it has been changed to a signed type.

src/stan/error_handling/out_of_range.hpp
syclik: Yes! I completely forgot this can be done with default parameters values. I'm pushing an updated version with a single function that has default parameter values.

syclik:
Bob, here's a link to compare across the last commit you reviewed and the current state of the branch:
https://github.com/stan-dev/stan/compare/f8917644c45569b46e880420b2eaa19eee9cf44f...85d720505489ba4cb30405cda5642ed69ac8825a

Most of the changes are due to changing namespaces of error handling from `stan::error_handling` to `stan::math`. The rest of the changes were targeted and fixed each of the points you raised. It might be easier looking at the individual commits to find which fixes I made.

The one non-trivial behavior change is in the behavior of `assign()`. We are now allowing matrices shaped like vectors to be assigned to vectors. In the current develop branch, this will throw an exception.

bob-carpenter:
Why are matrices shaped like vectors being assigned to 
vectors?

The Stan language will still rule it out.  The reason
I don't like to allow that in general is that I'd like to
have assignment match function invocation so that if you can
assign to type T, you can be a function arg of type T.  

The problem with allowing general casting is that it then
makes some operations ambiguous (e.g., matrix * vector = vector,
matrix * matrix = matrix).  We might be able to finagle
something consistent (or, for instance, take un-cast
versions as having precedence over ones requiring cast,
so that if A and B are matrices and B just happens to be
shaped like a vector, then A*B is still a matrix.

- Bob

syclik:
We have unit tests that tickle this behavior.

I think this has to do with the check of static sizes, but that's just a
guess.

I'll get a fix in on this branch to match the old behavior.

bob-carpenter:
I don't think it'll hurt anything to have the C++ API be
more general than the language.  In fact, I'd prefer the
C++ API to mirror the way Eigen works, only allowing mixed
types.  In the ideal case, the refactor to reference and
base/derived types should generalize everything, at least
in the dense case, which is all we're dealing with now.

I'm still going to try to figure out how to get the operator*
in.  

- Bob

syclik:
Bob, I found out what's going on. I inadvertently changed the behavior of the `assign` code. Here was the old code:
```
    /**
     * Copy the right-hand side's value to the left-hand side
     * variable.  
     *
     * The <code>assign()</code> function is overloaded.  This
     * instance will be called for arguments that are both
     * <code>Eigen::Matrix</code> types, but whose shapes  are not
     * compatible.  The shapes are specified in the row and column
     * template parameters.
     *
     * @tparam LHS Type of left-hand side matrix elements.
     * @tparam RHS Type of right-hand side matrix elements.
     * @tparam R1 Row shape of left-hand side matrix.
     * @tparam C1 Column shape of left-hand side matrix.
     * @tparam R2 Row shape of right-hand side matrix.
     * @tparam C2 Column shape of right-hand side matrix.
     * @param x Left-hand side matrix.
     * @param y Right-hand side matrix.
     * @throw std::domain_error if sizes do not match.
     */
    template <typename LHS, typename RHS, int R1, int C1, int R2, int C2>
    inline void 
    assign(Eigen::Matrix<LHS,R1,C1>& x, 
           const Eigen::Matrix<RHS,R2,C2>& y) {
      std::stringstream ss;
      ss << "error in call to assign(Eigen::Matrix,Eigen::Matrix)"
         << "; shapes must match, but found"
         << " R1=" << R1 
         << "; C1=" << C1
         << "; R2=" << R2
         << "; C2=" << C2;
      throw std::domain_error(ss.str());
    }
```
I changed the behavior -- and I don't remember why. It now looks like this (comments omitted):
```
template <typename LHS, typename RHS, int R1, int C1, int R2, int C2>
inline void assign(Eigen::Matrix<LHS,R1,C1>& x, const Eigen::Matrix<RHS,R2,C2>& y) {
   stan::math::check_matching_dims("assign", "x", x, "y", y);
}
```

My fault. I completely changed the behavior without realizing. I'm changing the behavior back to the original behavior.

syclik:
note: don't merge this yet... I'll need to update it based on #1234.

syclik:
@bob-carpenter and @rtrangucci, I'm going to merge this after the tests pass and I've verified the speed is the same, unless you say otherwise. There are a few too many changes to look through by hand, but this pull request makes most of our error handling consistent.

syclik:
I've verified, by hand, that this doesn't slow down Stan.

Issue #685, opened 2014-06-14
Need to separate out error conditions


syclik:
Stan should stop immediately under certain errors. This is a simple model where there's an illegal assignment of vectors due to a size mismatch.
```
parameters {
  vector[1] A;
}
transformed parameters {
  vector[2] B;
  B <- A;
}
model {
  A ~ normal(0,1);
}
```

When the model is compiled and run using "sample," the model shows this error message 100 times:
```
Informational Message: The current Metropolis proposal is about to be rejected because of the following issue:
Error in function assign(l): Rows of matrix 1Rows of matrix 1 (1) and Rows of matrix 2 (2) must match in size
If this warning occurs sporadically, such as for highly constrained variable types like covariance matrices, then the sampler is fine,
but if this warning occurs often then your model may be either severely ill-conditioned or misspecified.
Rejecting proposed initial value with zero density.
```

It should have quit instead.

PR #1210, opened 2015-01-11
Feature/issue 1209 matrixbase

jpritikin:
Here is a proof-of-concept changeset for conversion to matrixbase.


syclik:
Awesome! It would take a whole lot of work, but I think it's do-able. Would the map be sufficient for testing that it works for all `Eigen::MatrixBase`?

jpritikin:
> Would the map be sufficient for testing that it works for all Eigen::MatrixBase?

Actually, if you look closely at my patch, you'll find that I needed to change less than all the instances of Eigen::MatrixBase. For example, `Eigen::Matrix<typename return_type<T_y, T_loc>::type, Eigen::Dynamic, 1> y_minus_mu(size_y);` stays the same because it doesn't depend on the storage policy of the types passed into the function. So I suggest that our changes should be driven by testing with Eigen::Map and not by blindly grepping for Eigen::Matrix.

Before I make a more serious attempt at this patch, I'm waiting for #1212 and #685 to get merged.

jpritikin:
Can somebody look at how to convert the error checking stuff to handle Eigen::MatrixBase? I fiddled around but it is taking me a long time to understand the template expansion errors.

syclik:
Hi Joshua,

That's a little too vague for me to help. Which test is giving you compiler
errors?

Daniel

jpritikin:
Ah, right. Try `make test/unit/prob/distributions/multivariate/continuous/multi_normal`

bgoodri:
Here is a one-liner that mostly returns instances where `Eigen::Matrix` appears within a function signature (I agree that changing all of them _inside_ a function is less of a priority):
```
git grep -p -n -F "Eigen::Matrix" src/ |  \
grep -v -e "^src.*:\s*Eigen::Matrix" -e "^src.*\.[ch]pp=" -e "using Eigen::Matrix" -e "^src.*:\s*\*" -e "return"
```

bob-carpenter:
The key issue is to change the ones that result in
redundant copies.  That can happen inside functions, too.

- Bob

syclik:
@jpritikin, finally had a moment to look at it. In order to implement this, this will probably be a bit bigger than you envisioned. Although... maybe if you fixed `value_of_rec` this might be fixed. That's where I'd start, if I were you.

jpritikin:
I haven't finished running the integration test, but it's looking promising. How you like like this patch? What is the next step?

syclik:
@jpritikin, I'm a little confused. It looks like the changes you made in the past have disappeared and only changes in value_of_rec are included. Is that what you intended?

Also, why were comments removed? And what happened to the templated version of the function?

jpritikin:
>  I'm a little confused. It looks like the changes you made in the past have disappeared and only changes in value_of_rec are included. Is that what you intended?

Well, you suggested starting with `value_of_rec` so that's where I started. I still have the other changes in a branch, but I figured I'd take 1 small step at a time.

> Also, why were comments removed? And what happened to the templated version of the function?

It seems that the template is not needed? I tried to adjust the documentation to match the code.

syclik:
Well, you suggested starting with value_of_rec so that's where I started. I still have the other changes in a branch, but I figured I'd take 1 small step at a time.

Sorry about the confusion. I was suggesting that would be a step to getting the branch closer to compiling.

But yes, taking it one small step at a time is good.

    It seems that the template is not needed? I tried to adjust the documentation to match the code.

Did you run through the rest of the tests? I would be very surprised if we didn't need the base template function to match something that's not specialized.

jpritikin:
> Did you run through the rest of the tests?

I'm not sure. The makefile help output is not very loquacious on this point. How do I run all the tests?

syclik:
If you have python installed, type:
`./runTests.py src/test/unit`

jpritikin:
All tests passed, as far as I can tell.

jpritikin:
I adjusted get() as well; 100% passing. How does it look?

jpritikin:
Hey, I actually got multi_normal converted!

bob-carpenter:
First of all, thanks for taking this on.  It should in general improve efficiency for our matrix arguments.  

I have quite a few comments, so I thought I'd just enumerate them in one comment rather than trying to sprinkle them inline.

* we don't want any lingering TODO items if they're things that need to be done;  we occassionally drop in a FIXME tag if we've implemented something suboptimally because the better way to do it is too much work

* it looks like this pull request reimplements `VectorView` --- it should just use the existing traits class

* changes like simplifying template params from `T_y` to `T` when there's only one template parameter are great

* for check_symmetric, will this now apply to 1 x 1 row vectors and column vectors?  I suppose that's OK, but it should be in the doc --- we're trying to be better now about the documenting the API than we were in the past, starting with new pull requests

* replace `const double& x` with `double x` --- primitives should be passed by value, not constant reference;  no need for const because C++ is pass by value, so they implicitly won't change the calling value;  so the following `const int x` should just be `int x`.  This isn't super consistent in what we have so far, but I update every time I take a pass over one

* update to `index_type` and `value_type` look great and are in the proper location

* you're right in the `qr_R` funciton that we don't need to static cast to `size_t` --- that was actually a bug as it was because `m.rows()` and `m.cols()` are signed and so is the argument to `check_greater_or_equal`

* I'd rather not use `M` for matrices --- we tend to use upper case for template params and constants (like sizes);  we do use `L` in the code for Cholesky factors and `Sigma` and `Omega` for positive-definite matrices when it follows common statistical notation, so our conventions aren't perfectly consistent even as is

* I like that you switched `value = 0` to `value = false` in `is_vector` --- we should be using the boolean constants wherever the value is logically boolean, as in `is_vector` --- otherwise, our code will look like C, and who wants that?

* the convoluted definition of `get` to finesse signed vs. unsigned types makes me think we might want parallel methods with signed types  (from what I've seen on the Eigen lists, they may be moving to using `ptrdiff_t`, which is the signed equivalent of `size_t`, but they're not there yet, so we'll never be able to get rid of all the warnings arising from Eigen itself)

* `length` has more `const double&` types;  all three of those definitions (`double`, `int` and `Eigen::MatrixXd::Index&` can be generalized.  
    * template the type of `MatrixXd` --- no reason to just go for double values --- can you use the same `MatrixBase` construction as elsewhere?  If not, just scalar type, rows and columns
    * You can suck up all of the arithmetic types using Boost's `enable_if` along with Boost's `is_arithmetic`

* Having said all this, I really don't like the name `length` when used as synonymous with `size`, the reason being that `length` is ambiguous for a vector between number of dimensions (i.e., `size`) and norm (e.g., Euclidean length);  but it doesn't look like this pull request introduced the name, so it probably should go in another issue to fix

* is the argument `Matrix<T,R,C>` redundant for `is_vector_like` when you have `MatrixBase<T>`?

* can `VectorView(const std::vector<scalar_t>& v)` be generalized by templating out the `scalar_t`?  If it needs to be arithmetic, you can use Boost's `enable_if` again;  there's also an `is_integral` if there's a reason to separate floating point from integers

jpritikin:
> it looks like this pull request reimplements VectorView --- it should just use the existing traits class

I'm not sure where you got that impression. I only changed the prototype of a constructor. The rest is the same.

> for check_symmetric, will this now apply to 1 x 1 row vectors and column vectors?

Yes

> I suppose that's OK, but it should be in the doc --- we're trying to be better now about the documenting the API than we were in the past, starting with new pull requests

Good idea. I'll mention it.

> the convoluted definition of get to finesse signed vs. unsigned types makes me think we might want parallel methods with signed types (from what I've seen on the Eigen lists, they may be moving to using ptrdiff_t, which is the signed equivalent of size_t, but they're not there yet, so we'll never be able to get rid of all the warnings arising from Eigen itself)

I'm not sure if you are asking me to make some changes.

> template the type of MatrixXd --- no reason to just go for double values --- can you use the same MatrixBase construction as elsewhere? If not, just scalar type, rows and columns

The reason I used `Eigen::MatrixXd::Index` is to name Eigen's Index type. The type `MatrixXd` is irrelevant, just mentioned to get to Index. Is there a more direct way to name the Index type?

> You can suck up all of the arithmetic types using Boost's enable_if along with Boost's is_arithmetic

I'm not sure how to do this. Can you point to an example?

> can `VectorView(const std::vector<scalar_t>& v)` be generalized by templating out the `scalar_t`?

I'm not sure. I find the definition of `VectorView` rather confusing. I don't think I can assist here. I suggest opening a separate issue if you want to pursue.

- [ ] Check whether the argument `Matrix<T,R,C>` is redundant for `is_vector_like` when I have MatrixBase<T>
- [ ] Simplify arguments like `const double &x` to `double x`
- [ ] Find the correct home for `var`/`fvar` variants of `length` and `get`.

jpritikin:
The argument `Matrix<T,R,C>` is not redundant for `is_vector_like`. I do not understand why, but `test/unit/math/fwd/mat/prob/dirichlet` fails without the `Matrix<T,R,C>` version.

make/tests
```diff
@@ -208,3 +208,7 @@ endif
 # Include the test dependencies
 ##
 include make/test-math-dependencies
+
+.PHONY: test
```
syclik: can you take this target out of the makefile? I don't think there should be a circular dependency here.

makefile
```diff
@@ -120,17 +120,19 @@ endif
 	@echo '  - doxygen        : Builds the API documentation. The documentation is located'
 	@echo '                     doc/api/'
 	@echo '                     (requires doxygen installation)'
-	@echo ' TESTS:'
+	@echo '  TESTS:'
```
syclik: can you undo these spacing changes?

src/stan/io/reader.hpp
```diff
@@ -2,9 +2,9 @@
 #define STAN__IO__READER_HPP
 
 #include <boost/throw_exception.hpp>
+#include <stan/math/prim/scal.hpp>
```
syclik: can you revert this change? Unless there's a specific reason this needs to be included in this order, but even so, this looks like the wrong order.

src/stan/io/writer.hpp
```diff
@@ -1,9 +1,9 @@
 #ifndef STAN__IO__WRITER_HPP
 #define STAN__IO__WRITER_HPP
 
+#include <stan/math/prim/scal.hpp>
```
syclik: same here.

src/stan/math/prim/arr/meta/is_vector_like.hpp
```diff
@@ -0,0 +1,13 @@
+#ifndef STAN__MATH__PRIM__ARR__META__IS_VECTOR_LIKE_HPP
```
syclik: Thanks for adding this in. It looks like this never had tests in develop. I'm going to create an issue to create and test `is_vector_like` separately outside of this pull request.

src/stan/math/prim/mat/meta/get.hpp
```diff
@@ -2,11 +2,12 @@
...
 
 #include <stan/math/prim/mat/fun/Eigen.hpp>
+#include <stan/math/prim/mat/meta/value_type.hpp>
 
 namespace stan {
 
-  template <typename T, int R, int C>
-  inline T get(const Eigen::Matrix<T,R,C>& m, size_t n) {
+  template <typename T>
+  inline typename stan::math::value_type<T>::type get(const Eigen::MatrixBase<T>& m, size_t n) {
```
syclik: is the call through the metaprogram `value_type` necessary here? What type is `T` here? (these are actual questions -- I don't know the answer to this)

src/stan/math/prim/mat/meta/index_type.hpp
```diff
@@ -21,6 +21,16 @@ namespace stan {
       typedef typename Eigen::Matrix<T,R,C>::Index type;
     };
 
+    template <typename T>
+    struct index_type<Eigen::MatrixBase<T> > {
+      typedef typename index_type<T>::type type;
+    };
+
+    template <typename T>
+    struct index_type<Eigen::Map<T> > {
```
syclik: Is `Eigen::Map<T>` different than `Eigen::MatrixBase<T>`?

src/stan/math/prim/mat/prob/multi_normal_rng.hpp
```diff
@@ -20,10 +20,10 @@ namespace stan {
   namespace prob {
     using Eigen::Dynamic;
 
-    template <class RNG>
+    template <class RNG, typename T1, typename T2>
```
syclik: we don't need `T1` and `T2` as template parameters.

src/stan/math/prim/mat/prob/multi_normal_rng.hpp
```diff
@@ -20,10 +20,10 @@ namespace stan {
...
     inline Eigen::VectorXd
-    multi_normal_rng(const Eigen::Matrix<double, Dynamic, 1>& mu,
-                     const Eigen::Matrix<double, Dynamic, Dynamic>& S,
+    multi_normal_rng(const Eigen::MatrixBase<T1>& mu,
```
syclik: change `T1` and `T2` to `double`.

make/tests
jpritikin: > I don't think there should be a circular dependency here.

I am happy to comply, but you realize that the purpose of PHONY is to tell make that the rule does not create a file called "test", right? See https://www.gnu.org/software/make/manual/html_node/Phony-Targets.html

src/stan/math/prim/scal/err/check_bounded.hpp
```diff
@@ -4,6 +4,8 @@
 #include <stan/math/prim/scal/err/domain_error.hpp>
 #include <stan/math/prim/scal/err/domain_error_vec.hpp>
 #include <stan/math/prim/scal/meta/max_size.hpp>
+#include <stan/math/prim/scal/meta/is_vector_like.hpp>
+#include <stan/math/prim/arr/meta/is_vector_like.hpp>
```
syclik: remove this line. headers in `prim/scal` should not include things in `prim/arr`.

src/stan/math/prim/scal/err/check_equal.hpp
```diff
@@ -4,6 +4,7 @@
 #include <stan/math/prim/scal/err/domain_error.hpp>
 #include <stan/math/prim/scal/err/domain_error_vec.hpp>
 #include <stan/math/prim/scal/meta/VectorView.hpp>
+#include <stan/math/prim/arr/meta/is_vector_like.hpp>
```
syclik: remove.

src/stan/math/prim/scal/err/check_finite.hpp
```diff
@@ -5,6 +5,7 @@
 #include <stan/math/prim/scal/err/domain_error_vec.hpp>
 #include <stan/math/prim/scal/meta/length.hpp>
 #include <stan/math/prim/scal/meta/is_vector_like.hpp>
+#include <stan/math/prim/arr/meta/is_vector_like.hpp>
```
syclik: remove.

src/stan/math/prim/scal/err/check_greater.hpp
```diff
@@ -4,6 +4,7 @@
 #include <stan/math/prim/scal/err/domain_error.hpp>
 #include <stan/math/prim/scal/err/domain_error_vec.hpp>
 #include <stan/math/prim/scal/meta/is_vector_like.hpp>
+#include <stan/math/prim/arr/meta/is_vector_like.hpp>
```
syclik: remove.

src/stan/math/prim/scal/err/check_greater_or_equal.hpp
```diff
@@ -4,6 +4,7 @@
 #include <stan/math/prim/scal/meta/length.hpp>
 #include <stan/math/prim/scal/meta/VectorView.hpp>
 #include <stan/math/prim/scal/meta/is_vector_like.hpp>
+#include <stan/math/prim/arr/meta/is_vector_like.hpp>
```
syclik: remove.

src/stan/math/prim/scal/err/check_less.hpp
```diff
@@ -5,6 +5,7 @@
 #include <stan/math/prim/scal/err/domain_error_vec.hpp>
 #include <stan/math/prim/scal/meta/is_vector_like.hpp>
 #include <stan/math/prim/scal/meta/length.hpp>
+#include <stan/math/prim/arr/meta/is_vector_like.hpp>
```
syclik: remove.

src/stan/math/prim/scal/err/check_less_or_equal.hpp
```diff
@@ -6,6 +6,7 @@
 #include <stan/math/prim/scal/meta/length.hpp>
 #include <stan/math/prim/scal/meta/get.hpp>
 #include <stan/math/prim/scal/meta/is_vector_like.hpp>
+#include <stan/math/prim/arr/meta/is_vector_like.hpp>
```
syclik: remove.

src/stan/math/prim/scal/err/check_nonnegative.hpp
```diff
@@ -4,8 +4,10 @@
 #include <stan/math/prim/scal/err/domain_error.hpp>
 #include <stan/math/prim/scal/err/domain_error_vec.hpp>
 #include <stan/math/prim/scal/meta/length.hpp>
+#include <stan/math/prim/arr/meta/length.hpp>
```
syclik: remove.

src/stan/math/prim/scal/err/check_nonnegative.hpp
```diff
@@ -4,8 +4,10 @@
...
 #include <stan/math/prim/scal/meta/value_type.hpp>
 #include <stan/math/prim/scal/meta/is_vector_like.hpp>
+#include <stan/math/prim/arr/meta/is_vector_like.hpp>
```
syclik: remove.

src/stan/math/prim/scal/err/check_not_nan.hpp
```diff
@@ -3,6 +3,10 @@
 
 #include <stan/math/prim/scal/meta/length.hpp>
 #include <stan/math/prim/scal/meta/is_vector_like.hpp>
+#include <stan/math/prim/arr/meta/is_vector_like.hpp>
```
syclik: remove.

src/stan/math/prim/mat/meta/get.hpp
jpritikin: > is the call through the metaprogram value_type necessary here?

Type `T` could be Eigen::Matrix but it could also be Eigen::Map. That's why I use `value_type` to unwind the type down to the Scalar type of the matrix. So yeah, I think it is necessary.

src/stan/math/prim/scal/err/check_not_nan.hpp
```diff
@@ -3,6 +3,10 @@
...
+#include <stan/math/prim/arr/meta/length.hpp>
```
syclik: remove.

src/stan/math/prim/scal/err/check_not_nan.hpp
```diff
@@ -3,6 +3,10 @@
...
+#include <stan/math/prim/arr/meta/get.hpp>
```
syclik: remove.

src/stan/math/prim/scal/err/check_not_nan.hpp
```diff
@@ -3,6 +3,10 @@
...
+#include <stan/math/prim/mat/meta/get.hpp>
```
syclik: remove.

src/stan/math/prim/scal/err/check_positive.hpp
```diff
@@ -7,6 +7,7 @@
 #include <stan/math/prim/scal/meta/length.hpp>
 #include <stan/math/prim/scal/meta/get.hpp>
 #include <stan/math/prim/scal/meta/is_vector_like.hpp>
+#include <stan/math/prim/arr/meta/is_vector_like.hpp>
```
syclik: remove.

src/stan/math/prim/scal/err/domain_error_vec.hpp
```diff
@@ -7,6 +7,8 @@
 #include <stan/math/prim/scal/meta/value_type.hpp>
 #include <stan/math/prim/scal/meta/error_index.hpp>
 #include <stan/math/prim/scal/meta/get.hpp>
+#include <stan/math/prim/arr/meta/get.hpp>
```
syclik: remove.

src/stan/math/prim/scal/err/domain_error_vec.hpp
```diff
@@ -7,6 +7,8 @@
...
+#include <stan/math/prim/mat/meta/get.hpp>
```
syclik: remove.

src/stan/math/prim/scal/fun/value_of_rec.hpp
```diff
@@ -12,33 +12,14 @@ namespace stan {
...
-     * should be specialized.
-     *
-     * @tparam T Type of scalar.
      * @param x Scalar to convert to double.
      * @return Value of scalar cast to a double.
      */
-    template <typename T>
-    inline double value_of_rec(const T x) {
-      return static_cast<double>(x);
+    inline double value_of_rec(double x) {
```
syclik: what's the reason for these changes to `value_of_rec.hpp`?

src/stan/math/prim/mat/meta/index_type.hpp
jpritikin: Yes, `Eigen::Map` signals to Eigen that the memory allocated for the matrix is not controlled by Eigen whereas `Eigen::MatrixBase` is a generic wrapper class that should wrap any Eigen dense matrix where the actual matrix type is `T`. So they do different things.

src/stan/math/prim/scal/meta/VectorView.hpp
```diff
@@ -46,8 +46,8 @@ namespace stan {
 
     VectorView(std::vector<scalar_t>& v) : x_(&v[0]) { }
 
-    template <int R, int C>
-    VectorView(Eigen::Matrix<scalar_t,R,C>& m) : x_(&m(0)) { }
+    template <typename ET>
```
syclik: what does ET mean? Why is this function templated as opposed to just using `scalar_t`?

syclik: I see. So the `Derived` type within the `MatrixBase< Derived >` is an actual `Eigen::Matrix<scalar_t, -1, -1>`, right? I'm not sure what's supposed to happen here, then.

src/stan/math/prim/scal/meta/get.hpp
```diff
@@ -6,8 +6,7 @@
 
 namespace stan {
 
-  template <typename T>
-  inline T get(const T& x, size_t n) {
+  inline double get(double x, size_t n) {
```
syclik: why the change to `get.hpp`?

src/stan/math/prim/scal/meta/is_vector_like.hpp
```diff
@@ -10,7 +10,7 @@ namespace stan {
   // handles scalar, eigen vec, eigen row vec, std vec
   template <typename T>
   struct is_vector_like {
-    enum { value = stan::is_vector<T>::value };  
+    enum { value = false };
```
syclik: why is this being changed? Maybe this is why you needed to define `is_vector_like` for `std::vector`? Does this not work for `MatrixBase` by just adding a template specialization for `MatrixBase`?

src/stan/math/prim/scal/meta/length.hpp
```diff
@@ -2,13 +2,38 @@
 #define STAN__MATH__PRIM__SCAL__META__LENGTH_HPP
 
 #include <cstdlib>
+#include <cstddef>
 
 namespace stan {
 
+  size_t length(double /*x*/) {
```
syclik: what are these changes for?

src/stan/math/prim/scal/meta/length_mvt.hpp
```diff
@@ -9,16 +9,12 @@ namespace stan {
 
   // length_mvt() should only be applied to std vector or Eigen matrix
   template <typename T>
-  size_t length_mvt(const T& ) {
-    throw std::out_of_range("length_mvt passed to an unrecognized type.");
```
syclik: this behavior seems to have been removed by your changes.

src/stan/math/prim/scal/meta/length_mvt.hpp
```diff
@@ -9,16 +9,12 @@ namespace stan {
...
+  size_t length_mvt(const Eigen::MatrixBase<T>& ) {
```
syclik: why are you changing `length_mvt` beyond just adding a specialization for `MatrixBase`?

syclik: Also, this specialization should be inside `prim/mat`, not `prim/scal`

src/stan/math/prim/scal/meta/length_mvt.hpp
```diff
@@ -9,16 +9,12 @@ namespace stan {
...
   }
-  template <typename T, int R, int C>
-  size_t length_mvt(const Eigen::Matrix<T,R,C>& ) {
-    return 1U;
-  }
-  template <typename T, int R, int C>
-  size_t length_mvt(const std::vector<Eigen::Matrix<T,R,C> >& x) {
+
+  template <typename T>
+  size_t length_mvt(const std::vector<T>& x) {
```
syclik: this needs to go in `prim/arr` if it needs to stay.

src/stan/math/prim/scal/meta/scalar_type.hpp
```diff
@@ -48,6 +48,15 @@ namespace stan {
     typedef typename scalar_type<T>::type type;
   };
 
+  template <typename T>
```
syclik: this needs to go to `prim/mat`.

src/stan/math/rev/scal/meta/get.hpp
```diff
@@ -0,0 +1,14 @@
+#ifndef STAN__MATH__REV__SCAL__META__GET_HPP
```
syclik: ??? did this not work in the past before the changes on this branch?

src/stan/model/model_header.hpp
```diff
@@ -8,13 +8,13 @@
 #include <stan/math/prim/mat/fun/value_of_rec.hpp>
 #include <stan/math/prim/mat/meta/get.hpp>
 
+#include <stan/math/prim/scal.hpp>
```
syclik: the order here was intentional. Please change it back to: arr then mat then scal.

src/test/unit/math/prim/mat/fun/value_of_rec_test.cpp
```diff
@@ -26,11 +26,17 @@ TEST(MathMatrix,value_of_rec) {
   
   Eigen::Matrix<double,2,5> a; 
   fill(a_vals, a);
+  Eigen::Map< Eigen::Matrix<double,2,5> > am(a.data(), a.rows(), a.cols());
```
syclik: thanks for adding these tests! can you split this test into its own test case? I'd prefer if it weren't interleaved into the existing test.

syclik: one question: is an `Eigen::Map< MatrixXd >` sufficient for testing `MatrixBase< Derived >`?
(the answers would either be 'yes' or 'no, we should also be testing ___'. I'm not sure if we need to add a couple tests for some of Eigen's expression templates (say `am.transpose()` or if `Eigen::Map` is sufficient)

syclik:
@jpritikin, most of the changes look ok to me. There are lots of places where the existing meta programs changed, which I don't think is wise. Or if so, we should split those out into separate issues.

I assume this is still a proof of concept. Do you have an idea of how much more we need to change to get full coverage? We'll also need to add a whole lot more testing. It's becoming pretty clear that if we were to go through with this, we would want to add tests that make sure that the function compiles with all of Eigen's expression templates. I just don't know how to write tests that would push this.

make/tests
syclik: I think you missed the point.

Remove the `test` target completely. Not just the line with `.PHONY`

src/stan/math/prim/mat/meta/index_type.hpp
syclik: Hmmm.... does that mean we need to have separate function signatures for `MatrixBase` and `Map`? In other words, do we to split every function that has `Eigen::Matrix` in the arguments into `MatrixBase` and `Map`?

src/stan/io/reader.hpp
jpritikin: All these order changes are due to the following problem. Perhaps there is a better way to resolve it?

```
In file included from src/stan/math/prim/arr.hpp:7:0,
                 from src/stan/io/reader.hpp:5,
                 from src/test/unit/io/reader_test.cpp:1:
src/stan/math/prim/arr/meta/is_vector_like.hpp:7:10: error: ‘is_vector_like’ is not a class template
```

src/stan/math/prim/mat/meta/index_type.hpp
jpritikin: > Hmmm.... does that mean we need to have separate function signatures for `MatrixBase` and `Map`?

In general, no, because MatrixBase can wrap Map. However, due to the way template specialization works, I had to add both signatures in some places to get things to compile. I certainly didn't want to add more than MatrixBase, but I felt that I didn't have a choice when the compiler complained. Maybe there is a better way to get things working.

src/stan/math/prim/mat/prob/multi_normal_rng.hpp
jpritikin: How so? How do you make a function accept MatrixBase arguments without using template parameters? I'm baffled.

src/stan/math/prim/mat/prob/multi_normal_rng.hpp
jpritikin: T1, T2 might be MatrixXd or Map<MatrixXd> but they are certainly not a scalar type.

src/stan/math/prim/scal/fun/value_of_rec.hpp
jpritikin: Ah yes, this is a problem that I ran into in a few different places. The problem is that template<typename T> matches when T=Eigen::MatrixBase<T1> even when there is a more specific template<typename T> for Eigen::MatrixBase. I couldn't figure out how to get the template matching to work without removing the most generic template<typename T>. Maybe I am missing something?

src/stan/math/prim/scal/meta/VectorView.hpp
jpritikin: > So the `Derived` type within the `MatrixBase< Derived >` is an actual `Eigen::Matrix<scalar_t, -1, -1>`, right?

Yes.

The ET template argument is just a placeholder to match anything that looks like Eigen::MatrixBase. The type matching for the scalar type is completely orthogonal and is handled by the compiler using the regular type matching mechanisms.

src/stan/math/prim/scal/meta/get.hpp
jpritikin: Yeah, I made similar changes to your other meta-programming functions. Again, the problem is that `template<typename T>` matches `Eigen::MatrixBase<T>` even when there is a more specific template specification for `Eigen::MatrixBase`.

src/stan/math/prim/scal/meta/length_mvt.hpp
jpritikin: Wouldn't it be preferable to get a compile-time error instead of a run-time error?

src/stan/math/prim/mat/meta/index_type.hpp
bob-carpenter: Yes.  I'd suggest looking these things up!  Eigen has
great doc.  Map<T> is the expression template class that
maps contiguous memory to an Eigen type T.

- Bob

bob-carpenter: What were the compiler errors?

- Bob

> On Mar 12, 2015, at 8:43 AM, Joshua Pritikin <notifications@github.com> wrote:
> 
> In src/stan/math/prim/mat/meta/index_type.hpp:
> 
> > @@ -21,6 +21,16 @@ namespace stan {
> >        typedef typename Eigen::Matrix<T,R,C>::Index type;
> >      };
> >  
> > +    template <typename T>
> > +    struct index_type<Eigen::MatrixBase<T> > {
> > +      typedef typename index_type<T>::type type;
> > +    };
> > +
> > +    template <typename T>
> > +    struct index_type<Eigen::Map<T> > {
> 
> Hmmm.... does that mean we need to have separate function signatures for MatrixBase and Map?
> 
> In general, no, because MatrixBase can wrap Map. However, due to the way template specialization works, I had to add both signatures in some places to get things to compile. I certainly didn't want to add more than MatrixBase, but I felt that I didn't have a choice when the compiler complained. Maybe there is a better way to get things working.

What was the compiler error when you didn't have it?

src/stan/math/prim/mat/prob/multi_normal_rng.hpp
bob-carpenter: We only allow RNGs to apply to double values because we
can't autodiff through them, so I'm guessing
Daniel is suggesting to just hard-code double as the
scalar type.

- Bob

src/stan/math/prim/scal/meta/length_mvt.hpp
bob-carpenter: Absolutely better to check at compile-time if possible.

syclik: Agreed, but the problem is that we'll need a generic template function to
pick all the possible instantiations. I'm guessing this works because we're
missing some tests.

src/stan/math/prim/mat/prob/multi_normal_rng.hpp
syclik: Yes.

src/stan/math/prim/mat/meta/index_type.hpp
syclik: Sorry, I get confused with Eigen doc. I still haven't found how to
specialize matrix multiplication.

The question I have is... is there a single templated function we can write
to cover both Eigen::MatrixBase and Eigen::Map?

syclik:
@jpritikin, after looking over your code and thinking about it a bit, I think we need to step back and get the template metaprograms properly. Part of the problem is that we haven't documented how this all ties together, but the other part is that there were multiple changes happening at once which caused a cascade of other changes.

Any suggestions on the best way to work on this?

bob-carpenter:
I agree.  These wholesale changes seem dangerous because
our testing just isn't thorough enough yet.

I'm willing to help out with the template metaprogramming
aspects of this if you can break down something for me to do.

- Bob

jpritikin:
> I think we need to step back and get the template metaprograms properly.

@syclik I heartily agree. Adding all those includes you are asking me to remove took a lot of time because I had to run the test suite (which subjectively takes forever) and add 1 include at a time until it would compile.

> I'm willing to help out with the template metaprogramming aspects of this if you can break down something for me to do.

@bob-carpenter Well, can you take my branch as it is and start with the definition of stan::get. Change it however you prefer and see if you can get everything to still compile and pass the test suite. That's what I did. I don't have a more strategic suggestion. I am not yet a C++ template expert.

src/stan/math/prim/mat/meta/index_type.hpp
bob-carpenter: > On Mar 12, 2015, at 11:25 AM, Daniel Lee <notifications@github.com> wrote:
> 
> ...
> Sorry, I get confused with Eigen doc. I still haven't found how to specialize matrix multiplication. The question I have is... is there a single templated function we can write to cover both Eigen::MatrixBase and Eigen::Map?

The page to read is this:

  http://eigen.tuxfamily.org/dox/TopicFunctionTakingEigenTypes.html

MatrixBase includes Map.  DenseBase includes both MatrixBase and ArrayBase.

The Eigen doc itself isn't very good at saying what class(es)
a given class inherits from.

But now I'm really confused, because what they're saying should work
doesn't, unless I'm missing something fundamental.  Could this be a bug
in their asserts?


#include <gtest/gtest.h>
#include <iostream>
#include <Eigen/Dense>

using Eigen::Map;
using Eigen::Matrix;
using Eigen::MatrixXd;
using Eigen::MatrixBase;
using Eigen::Dynamic;

template <typename Derived>
double foo(const MatrixBase<Derived>& a, int m, int n) {
  return a(m,n);
}

TEST(foo, bar) {

  MatrixXd z(2,3);
  z << 1, 2, 3,
    4, 5, 6;
  EXPECT_FLOAT_EQ(1, foo(z,0,0)); // OK
  EXPECT_FLOAT_EQ(6, foo(z,1,2)); // OK

  std::vector<double> x;
  for (int i = 0; i < 6; ++i)
    x.push_back(i);
  Map<MatrixXd> y(&x[0],2,3);
  EXPECT_FLOAT_EQ(0, foo(y,0,0));  // OK
  EXPECT_FLOAT_EQ(5, foo(y,1,2));  // OK

  MatrixXd v(2,3);
  v << 1, 2, 3, 4, 5, 6;

  EXPECT_FLOAT_EQ(6, foo(v.transpose(),2,1)); // OK

  MatrixXd u(2,2);
  u << 1, 2, 3, 4;

  MatrixXd uv = u * v;
  EXPECT_FLOAT_EQ(33, foo(uv,1,2));  // OK

  std::cout << "(u * v)=" << std::endl
            << (u * v) << std::endl;

  std::cout << "(u * v).rows()=" << ((u * v).rows()) << std::endl;
  std::cout << "(u * v).cols()=" << ((u * v).cols()) << std::endl;

  std::cout << std::endl;

  EXPECT_FLOAT_EQ(9, foo((u * v).transpose(), 0, 0)); // OK

  EXPECT_FLOAT_EQ(9, foo((u * v), 0, 0)); //  FAILS.  WTF?
}


I'm going to write to the Eigen list and see if they have any suggestions.

- Bob

bgoodri: FYI: The Eigen class doc is http://eigen.tuxfamily.org/dox/TopicClassHierarchy.html

syclik: Bob and Ben, thanks for sending those links... they're really helpful and I wish I read them more carefully earlier.

src/stan/math/prim/mat/prob/multi_normal_rng.hpp
jpritikin: But `T1` and `T2` are not scalar types to begin with. You can't constrain the scalar type using the template when you are using `Eigen::MatrixBase<>`

bob-carpenter: To follow Eigen's naming conventions, these should probably
be Derived1 and Derived2. 

- Bob

jpritikin:
I think the crux of it is that `Eigen::MatrixBase<T>` does not constrain type `T` and therefore conflicts with your generic `template <typename T>` that is suppose to match other types (like `double`) that do not have any specialized template.

I'm not going to put any more effort into this issue until you provide clear guidelines about how the metaprogramming stuff should be written. The ball is in your court.

src/stan/math/prim/mat/prob/multi_normal_rng.hpp
betanalpha: Pretty sure you can with enable_if.  For example,

  template<typename DerivedA, typename DerivedB>
  inline typename std::enable_if<
    is_var<typename Eigen::MatrixBase<DerivedA>::Scalar>::value &&
    is_var<typename Eigen::MatrixBase<DerivedB>::Scalar>::value &&
    std::is_same<typename Eigen::MatrixBase<DerivedA>::Scalar,
                 typename Eigen::MatrixBase<DerivedB>::Scalar>::value,
    typename Eigen::Matrix<typename Eigen::MatrixBase<DerivedA>::Scalar,
             Eigen::Dynamic, Eigen::Dynamic>
    >::type
  multiply(const Eigen::MatrixBase<DerivedA>& M1,
           const Eigen::MatrixBase<DerivedB>& M2) {�}

bob-carpenter:
Understood.  Daniel and I need to put our heads together
and hash out what needs to be done.  That probably won't
happen until next week's Stan meeting because we're currently
pretty slammed trying to sort out some issues for 2.6.2. 
Sorry it's going so slowly.  We've been burned by a bunch
of this stuff in the past so are a bit reluctant to make
wholesale changes without trying to explore the consequences.
We'd be in better shape if our tests were more thorough in
this area --- it looks like you've added a bunch, which is great.

Our goal for the last refactoring was to reduce our spaghetti-like
dependencies to something less cyclic (to mix metaphors) and allow
more standalone compilation of autodiff components (like being able
to handle reverse-mode without requiring forward mode).

For Eigen::MatrixBase<Derived>, the Derived template parameter
obviously needs to be instantiated to something that implements
the relevant matrix operations.  I think the "Derived" thing is
just the convention for the CRTP, which Eigen is following. 

- Bob

syclik:
I second Bob's sentiment here.

Sorry it's taking so long, but this template metaprogramming stuff is
tricky. We've gotten this far with a lot of reading and a lot more trial
and error. That and we've been working on making Stan better with bug fixes
and other patches.

Mind leaving your branch as-is so I can look at it for reference? I'll
probably start over with the template metaprogramming and move bottom up
from where I think (or hope) is the bottom.




Daniel

jpritikin:
> Mind leaving your branch as-is so I can look at it for reference?

Sure

> I'll probably start over with the template metaprogramming and move bottom up
from where I think (or hope) is the bottom.

@syclik Sure, as long as you can get across the goal post, that is, test/unit/math/prim/mat/prob/multi_normal from the present branch.

syclik:
Just updating this pull request. `is_vector_like` is fine as-is, see the pull request for #1360. We will still have to extend to `MatrixBase`, but won't need to rewrite the other functions.

syclik:
And back to the expression templates we need to test for `MatrixBase` to be safe, we now know that Eigen's multiplication is a different class of expression template that may terminate in an assertion, so I think we should test:
- Map
- transpose
- multiplication
- block

Any others that I missed that we should test with? One way to deal with the multiplication issue is to call `.eval()` on the expression template on the inside to eagerly evaluate the expression template. I don't know what the speed hit would be on a normal `Matrix`.

jpritikin:
> @syclik won't need to rewrite the other functions

Great! Which pull request can I monitor?

bob-carpenter:
I think we should follow the advice of the Eigen developers.

If we need to access all of the elements of a matrix, use (a * b).eval(),
otherwise use the lazy eval.  I imagine it'll be the former.

There's no speed hit in that it's more efficient to build the evaluation
than to evaluate element by element.  

It won't be able to evaluate further expression templates through
the call, but that looks like the best we can do.

- Bob

jpritikin:
I see issue #1360 got merged. Does that help the present change set?

syclik:
I'm closing this pull request. We've moved the math library to a separate repo and I think we know what we need to do to get the compilation to work.

PR #1219, opened 2015-01-16
Bugfix/issue 1200 optimizer segfault

bob-carpenter:
#### Summary

Added a single line to resize `var_alloc_stack_` during nested memory recoverty so as to not double delete entries.

#### Intended Effect:

Make sure optimizer doesn't segfault deleting `chainable_alloc` entries (like LDLT class used by multi_normal) by resizing stacks keeping track of what needs to be deleted.  This will allow nested memory to be used and then a final `recover_memory()` to be called redundantly without errors.  (That's why it worked for sampling but not optimization---there was no redundant final call to recover memory anywhere in sampling.)

#### How to Verify:

There's a new unit test that fails in 2.5 with a segfault and succeeds in 2.6. 

You can also run the example model with data supplied by Jiqiang for the issue to see that there's no memory leak introduced for optimization or sampling and that sampling now converges.

#### Side Effects:

None --- everything should work as documented now.

#### Documentation:

Nothing user facing.

#### Reviewer Suggestions:

Anyone --- it's a very simple fix.

### DIFF
--- a/src/stan/agrad/rev/var_stack.hpp
+++ b/src/stan/agrad/rev/var_stack.hpp
@@ -76,8 +76,9 @@ namespace stan {
                                " before calling recover_memory()");
       ChainableStack::var_stack_.clear();
       ChainableStack::var_nochain_stack_.clear();
-      for (size_t i = 0; i < ChainableStack::var_alloc_stack_.size(); i++)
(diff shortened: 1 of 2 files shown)


syclik:
Thanks!

PR #1220, opened 2015-01-16
Add out-of-tree make/local.

ido:
#### Summary:

This one-line makefile change adds an out-of-tree makefile in addition to make/local, allowing users to place defines and build configuration (e.g. location of their compiler, additional CXXFLAGS, etc.) in a place outside of the git repository that will persist across otherwise destructive git commands.

#### Intended Effect:

Now, in addition to being able to use `make/local` to put any local build configuration or additional make rules, users can do so in `$HOME/.config/stan/make.local`.  If this file is present, it will be included by GNU make.  If it is not present, nothing should happen.

#### How to Verify:

Try putting a build configuration setting, e.g. setting CXXFLAGS or CXX or CC in ~/.config/stan/make.local and see if it sticks.

#### Side Effects:

If a user forgets that he/she has additional local build customizations in the aforementioned file location, his/her build may include said local build customizations unwittingly.  

#### Documentation:

This is documented in the same fashion as the `make/local` file is currently... :-)

#### Reviewer Suggestions:

Anyone with familiarity with the current (GNUmake-based) Stan build system would be a good reviewer for this.

### DIFF
--- a/makefile
+++ b/makefile
@@ -130,6 +130,7 @@ include make/libstan  # bin/libstan.a bin/libstanc.a
 include make/tests    # tests
 include make/doxygen  # doxygen
 include make/manual   # manual: manual, doc/stan-reference.pdf
+-include $(HOME)/.config/stan/make.local # for local stuff
 -include make/local    # for local stuff
 
 ##


syclik:
@ido:

- Is `$(HOME)/.config` a common place to put local configuration files? 
If so, can you provide a link or reference? It'd be good to know where we can stash local configuration for the future. If it's not common, some justification for why you think it should be there?

- Can you follow our developer process? 
At the very least, fill out the pull request template: https://github.com/stan-dev/stan/wiki/Pull-Request-Template 
It helps us keep track of what gets into the code repository.

ido:
@syclik 
`$HOME/.config` is a common place to put configs.  Many programs do so currently including chromium, evolution, gconf, xfce4, Qt, ImageMagick, CMake, and many many more. Here is a reference: http://standards.freedesktop.org/basedir-spec/basedir-spec-latest.html
(Another alternative is to just call it `$HOME/.stan.make.local` or similarly ad-hoc location as long as said location is outside of `$GIT_ROOT`...  Either way, if the file is not present, nothing changes.  If it is present, it gets included into the makefile.)

As for following the developer process, happy to comply, the ratio of documentation to code is going to be high on this one, here we go... :-)

#### Summary:

This one-line makefile change adds an out-of-tree makefile in addition to make/local, allowing users to place defines and build configuration (e.g. location of their compiler, additional CXXFLAGS, etc.) in a place outside of the git repository that will persist across otherwise destructive git commands.

#### Intended Effect:

Now, in addition to being able to use `make/local` to put any local build configuration or additional make rules, users can do so in `$HOME/.config/stan/make.local`.  If this file is present, it will be included by GNU make.  If it is not present, nothing should happen.

#### How to Verify:

Try putting a build configuration setting, e.g. setting CXXFLAGS or CXX or CC in ~/.config/stan/make.local and see if it sticks.

#### Side Effects:

If a user forgets that he/she has additional local build customizations in the aforementioned file location, his/her build may include said local build customizations unwittingly.  

#### Documentation:

This is documented in the same fashion as the `make/local` file is currently... :-)

#### Reviewer Suggestions:

Anyone with familiarity with the current (GNUmake-based) Stan build system would be a good reviewer for this.

syclik:
@ido, that link wasn't a reference, but a spec for a particular program. Do you know of any reference that discusses why a package would use `$HOME/.config/` as opposed to anything else? (I'd rather do something principled or conventional before putting things into arbitrary places.)

Regarding the other programs that use `$HOME/.config`, I don't have any of those, so I can't verify. What sorts of files end up there? On the Jenkins machine, I don't have a `$HOME/.config` directory. On my laptop, it exists and I have subfolders for `configstore`, `git`, and `www.na-mic.org`.  Do you have subfolders for each of the packages you've listed inside that folder? 

I'm asking a lot of questions cause I don't know the answers and googling hasn't turned up anything useful. I would have thought searching for `.config` and any one of those packages would have brought up similar documentation as XDG, but it doesn't. Searching stackoverflow for config doesn't help either, so I was wondering if it was a suggestion or a convention.

ido:
@syclik 
The link I provided is a reference / standard, but not for a single program.  XDG is not a program, it is an informal standard group from freedesktop.org for interoperability between X Windows apps and among different desktop environments.  (Basically, a de facto standard for how to organize `$HOME` so that other programs will not step on your toes.)   

From http://www.freedesktop.org/wiki/ : "freedesktop.org is not a formal standards organization, though some see a need for one that covers some of the areas we are working on. For Linux operating system standards, look at the Linux Standard Base project. The X.Org Foundation and the IETF are other groups that do formal standards. The Free Standards Group is one group that publishes "de jure" standards for free software; freedesktop.org is loosely affiliated with the FSG."  (The LSB Filesystem Hierarchy Standard does not specify anything about the internal structure of `/home/$USER/`.)

Yes, I have folders for each of the software packages I listed under my `$HOME/.config`.

Ask as many questions as you'd like.  StackOverflow is not likely to be helpful in this area.  The link I sent provides the convention.  The convention is loose enough that many programs just do whatever they want in `$HOME`.  For example, if you think `$HOME/.stan/make.local` or `$HOME/.stan.make.local` is more appropriate, please feel free to use those instead.  I am not tied to any particular location. :-)

syclik:
Thanks for the info. It's useful to know that I haven't missed anything. (Btw, the other XDG specs don't help me out -- don't really have time to wade through other doc for something that I'm not knowledgeable about. If you could point me to specific places, I'd have a better shot.)

@bob-carpenter, where do you think we should put it? `$HOME/.config/stan/make.local`? Let me know where you think it'll be easy to find.

bob-carpenter:
I have no opinion whatsoever.

- Bob

syclik:
@ido, I'm going to close this pull request and recreate it. The makefile has shifted around and we now have a conflict. It's easy to recreate your patch.

PR #1221, opened 2015-01-17
Feature/parse index

bgoodri:
Add some Bash-scripting to make/manual so that when the manual is created and put under doc/ also parse stan-reference.idx into doc/stan-functions.txt that looks like

```
goodrich@T540p:/opt/stan/doc$ cat stan-functions.txt
# This file is semicolon delimited
StanFunction; Arguments; ReturnType; Page
abs;(int x);int;272
abs;(real x);real;281
acosh;(real x);real;286
acos;(real x);real;285
append_col;(matrix x, matrix y);matrix;314
append_col;(matrix x, vector y);matrix;314
append_col;(row_vector x, row_vector y);row_vector;315
append_col;(vector x, matrix y);matrix;315
...
von_mises_rng;(reals mu, reals kappa);real;359
weibull_ccdf_log;(reals y, reals alpha, reals sigma);real;352
weibull_cdf_log;(reals y, reals alpha, reals sigma);real;352
weibull_cdf;(reals y, reals alpha, reals sigma);real;352
weibull_log;(reals y, reals alpha, reals sigma);real;352
weibull;~;real;352
weibull_rng;(real alpha, real sigma);real;352
wishart_log;(matrix W, real nu, matrix Sigma);real;373
wishart;~;real;373
wishart_rng;(real nu, matrix Sigma);matrix;373
```
Such a file can easily be read by the interfaces to Stan and augmented with a mapping from a Stan function to a similar native function in the interface (e.g., for R, normal  -> dnorm). This will make it easier for Stanimals to find the Stan functions they need to use.

You only have to run `make manual` to verify that the PR works.


make/manual
```diff
@@ -8,6 +8,25 @@ VERSION_STRING := $(MAJOR_VERSION).$(MINOR_VERSION).$(PATCH_VERSION)
 manual: $(STANAPI_HOME)src/docs/stan-reference/stan-reference.pdf
 	mkdir -p doc/
 	$(foreach artifact, $^, cp $(artifact) doc/$(basename $(notdir $(artifact)))-$(VERSION_STRING).pdf;)
+	echo '# This file is semicolon delimited' > doc/stan-functions.txt
```
syclik: @bgoodri, mind if we put `@` before each of these lines to suppress the output to screen?

bgoodri: Fine

syclik:
@bgoodri, can we do a couple things with this?
1. place this set of commands in a different make target. I'd rather separate this from building the manual.
2. can we wrap it in a script that will return a non-zero exit code so we can tell if any future builds of the manual will break this functionality?

I'm not good enough with bash scripting to make the second one work. I can handle the first thing if you want.

bgoodri:
It already should do 2 since the -e flag at the top will cause it to
terminate if any of the commands do not return an exit code of zero. We
could also put it into a different make target but my thinking was that
(once the script works) if the script does not work, then the index is not
generated by make manual in the expected way.

syclik:
Thanks for clearing that up.

I'm trying to separate the building of the manual from the building of this
index file because it seems like they're separate things. I don't mind
tying them together with a make target, but they seem to be doing different
things and it'd be good to know what breaks when.

syclik:
@bgoodri, how do these changes look? 
- `make manual` still builds both the manual and the text file
- the text file is built in the `src/docs/stan-reference` directory, then copied into the `doc` directory with the version number in the file name.

bgoodri:
Looks fine. You don't have to run the whole test suite to verify it is
working.

syclik:
Cool. I just tagged it with 'code review complete'. I'll get it merged after it gets tested.

Yes, I agree that this *shouldn't* affect anything else, but there have been plenty of things that interact in ways that we don't understand. I'd rather just wait for almost every pull request. I'll change my tune once we have more pull requests than we can test. For the time being, we have plenty of free cycles.

syclik:
I take it back. I sshed into the machine and tested it. It's good to go.

Issue #822, opened 2014-08-02
(multivariate) normal operating on sufficient statistics


bob-carpenter:
Joshua N. Pritikin mentioned on stan-users that the <a href="http://openmx.psyc.virginia.edu">OpenMx package</a> has a multivariate normal that works on sufficient statistics:

<blockquote>
In OpenMx, multi_normal has two implementations. One implementation  handles data in the form of 1 case per row. The other implementation handles data as a covariance matrix (and means). These two approaches are equivalent for certain data (no missingness, etc).
</blockquote>

I would think the second implementation would be much more efficient because the sufficient statistics could be computed just once.

Does anyone know what the density over the sufficient stats would look like?

syclik:
This issue was moved to stan-dev/math#38

PR #1203, opened 2015-01-05
attempt at multi-normal-sufficient-log

jpritikin:
This compiles but explodes with a run-time error in boost::math::tools::promote_args. I guess I don't really understand the purpose of promote_args. What is the purpose of promote_args? How am I using it wrong? See any other obvious problems?

PR #1226, opened 2015-01-21
Adds univariate normal distribution on sufficient statistics

randommm:
#### Summary:
Adds univariate normal distribution distribution.

#### Intended Effect:
Adds the univariate normal distribution and partially fix issue #822 (since we'll still lack the multivariate version).

#### How to Verify:
Code check.

#### Side Effects:
None.

#### Documentation:
Included.


bob-carpenter:
Marco and Joshua:  Is this pull request (#1226, for issue #822) related to pull request #1203 (also for issue #822).  Could you two coordinate on who's going to finish this pull request by coordinating with Daniel on testing?  Thanks, Bob.

randommm:
It's related but independent AFAIK. This one is for univariate normal, Joshua's one is for multivariate normal.

src/test/prob/normal/normal_sufficient_test.hpp
```diff
@@ -0,0 +1,117 @@
+// Arguments: Doubles, Doubles, Ints, Doubles, Doubles
```
rtrangucci: Small comment: Shouldn't this test be in src/test/prob/normal_sufficient/ rather than in src/test/prob/normal/?

randommm: I don't know: poisson_log and poisson_log_log tests are in the same folder, binomial_log and binomial_logit_log too, but neg_binomial_log and neg_binomial_2_log aren't.

bob-carpenter: We're on the fence with this decision.  Daniel's been
splitting functions out, but we've kept all of each
distribution together (density, cdfs, rngs).  

I think the various parameterizations of neg_binomial should
go in their own files.  That's the direction we're heading.

- Bob

Issue #1081, opened 2014-10-12
next manual 2.6.0


bob-carpenter:
This is where issues for the manual after 2.5.0 go.

bob-carpenter:
* [ ] add process description chapter

http://www.r-project.org/doc/R-FDA.pdf

Something along the lines of the SDLC section would be great to have in your user-manual. Basically this describes the software life-cycle as how Stan is programmed, released and managed.

blindglobe:
regarding the "process description chapter", I think you mean, "Process for Software Development and Release", which would be wonderful for supporting corporate IT computer systems validation work by providing justification that "STAN is developed in a way which makes it fit for the purpose of the process of performing a Bayesian statistical analyses".   Sure it might seem obvious, but for dotting the i's and crossing the t's for a critical review, it makes things that much easier if the process for software development is described and followed.  I'd be happy to review and comment if you want before you release (speaking as co-ghost-writer of the R-FDA document cited).

bob-carpenter:
* [ ] replace `lambda/sqrt(pi)` with `lambda/2` in `exp_mod_normal` doc (the implementation is correct, but our doc was wrong)

* [ ] thank Andrew Ellis in the manual for pointing it out

bob-carpenter:
* [ ] add clarification that inverse Wishart and Wishart both take the scale matrix S as a parameter
* [ ] compare to BDA's notation `Sigma ~ Inv-Wishart(inv(S))`

bob-carpenter:
*[ ] Fix this one (from Ben on stan-users)

Note that the next sentence of the manual is wrong. It should say

```
At $\nu = 1$, the LKJ correlation distribution reduces to the uniform distribution over correlation matrices of order $K$.
```

bob-carpenter:
Contributed by Gokcen Eraslan via a patch to master #1227 that we haven't merged because it was to master:

* [ ] remove duplicate "externally" in programming.tex after soft k-means
* [ ] thank Gokcen in the acknowledgments

bob-carpenter:
John Sutton mentioned on stan-users:

* [ ] append_row needs to have its own description, not just a cut-and-paste of append_col
* [ ] thank John in the acknowledgments

bob-carpenter:
* [ ] undo the fix that went in for Andrew on cbind and rbind so as not to make the index confusing as to what is a function and what isn't

bob-carpenter:
[...]

PR #1227, opened 2015-01-21
Typo fix.

gokceneraslan:


bob-carpenter:
Thanks.  This is a good fix, but we don't take merges into master and I'm not good enough at git to figure out how to change this to merge into develop.

bob-carpenter:
I added a note to #1081 to fix this, so you don't have to do anything else.  Thanks for pointing it out.

gokceneraslan:
OK, sorry for the wrong branch.

PR #1228, opened 2015-01-21
Remove git conflict marker

gokceneraslan:


bob-carpenter:
We don't accept merges to master.  I mistakenly merged this one and now have to figure out how to undo it.  If you want changes to the doc, it's easiest to just leave me a note in the open doc issue and I'll do it.  Otherwise, if you want to submit a patch, please branch from develop.  There's a wiki page on our developer process, and we try to follow, usually when I blow it by doing something like merging into master.

PR #1229, opened 2015-01-21
Typo fix

gokceneraslan:


bob-carpenter:
Thanks for the patch.  We're always happy to take suggestions, especially in the form of patches.

The form "were" of the verb "to be" is the past subjunctive, and what I intended there. See:

http://en.wikipedia.org/wiki/Subjunctive_mood

Does it sound too archaic for contemporary English?

gokceneraslan:
Ah, my fault then. Sorry.

syclik:
Looks like this was sorted out. I'm closing the pull request.

PR #1231, opened 2015-01-21
Update all.bib

aadler:
Updated Hoffman & Gelman as it is actually published. Changed URL from Andrew's repository to JMLR abstract

### DIFF
--- a/src/docs/bibtex/all.bib
+++ b/src/docs/bibtex/all.bib
@@ -526,9 +526,10 @@
   Title = {{T}he {N}o-{U}-{T}urn {S}ampler: {A}daptively {S}etting {P}ath {L}engths in {H}amiltonian {M}onte {C}arlo},
   Author = {Hoffman, Matthew D. and Gelman, Andrew},
   Journal = {Journal of Machine Learning Research},
-  Year = {2013},
-  Volume = {in press},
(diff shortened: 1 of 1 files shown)


syclik:
@bob-carpenter, unfortunately, we need to let these pull requests go to testing. This fix to all.bib doesn't compile. I'll fix it on my current branch, but I think we should just go through with the full testing on even little pull requests like this one.

aadler:
The fix is instead of

\url{{http://jmlr.org/papers/v15/hoffman14a.html}}

it needs

{\url{http://jmlr.org/papers/v15/hoffman14a.html}}

Finger-slip; sorry

syclik:
Yes, it's already been fixed on the branch I'm working on.

bob-carpenter:
I thought that URL looked fishy, but I figured if someone
entered it, they'd tested it.  Bad assumption, as always.

And sorry --- I just merged another one before testing.
I forgot they broke the build when the doc breaks.  I was
just going to fix everything later, but that obviously won't work.

Maybe we can more strongly suggest that people just note
issues on the issue tracker ather than sending us patches for
simple one line doc or bib fixes.  I find them harder to review
than just do and as you can see, I'm very error prone on the
manual merges.

- Bob

syclik:
No worries. That was really easy to find and fix. Just do what's
comfortable for you. I'll code around it.

PR #1235, opened 2015-01-22
Fixes the speed regression from v2.5.0

syclik:
#### Summary:

Changes `const std::string&` back to `const char*`. This was causing Stan to be slow in the current develop branch.

#### Intended Effect:

Behavior is the same, but speed is much faster.

#### How to Verify:

Run any model. I ran the logistic model in the example models repository.

In v2.5.0, this model runs in ~2.3 seconds and ~4.1 in develop. With this fix, it runs in ~2.3.

#### Side Effects:

Gets back to the old speed.

#### Documentation:

None required.

#### Reviewer Suggestions: 

Anyone.

### DIFF
--- a/src/stan/error_handling/matrix/check_corr_matrix.hpp
+++ b/src/stan/error_handling/matrix/check_corr_matrix.hpp
@@ -34,8 +34,8 @@ namespace stan {
      */
     // FIXME: update warnings
     template <typename T_y>
-    inline bool check_corr_matrix(const std::string& function,
-                                  const std::string& name,
+    inline bool check_corr_matrix(const char* function,
+                                  const char* name,
                                   const Eigen::Matrix<T_y,Eigen::Dynamic,Eigen::Dynamic>& y) {
       using Eigen::Dynamic;
       using Eigen::Matrix;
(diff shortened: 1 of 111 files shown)


src/stan/error_handling/matrix/check_pos_definite.hpp
```diff
@@ -46,8 +47,9 @@ namespace stan {
         std::ostringstream msg;
         msg << "is not positive definite. " 
                 << name << "(0,0) is ";
+        std::string msg_str(msg.str());
```
bob-carpenter: There's no need to define an intermediate msg_str here that I can see.  Not that it's a big deal.

bob-carpenter:
Looks good.  Heroic amount of re-refactoring.

src/stan/error_handling/matrix/check_corr_matrix.hpp
```diff
@@ -58,8 +58,9 @@ namespace stan {
               << name << "(" << stan::error_index::value + k 
               << "," << stan::error_index::value + k 
               << ") is "; ;
```
bob-carpenter: duplicate colons --- shouldn't hurt anything.

syclik: I'll fix it on the other branch I have open that touches most of this code.

Issue #1223, opened 2015-01-21
erroneous partial for pareto_type_2_log


rtrangucci:
Line 109 in src/stan/prob/distributions/univariate/continuous/pareto_type_2.hpp is:

``` const T_partials_return sum_dbl = lambda_dbl + y_dbl + mu_dbl; ```

but should be:

``` const T_partials_return sum_dbl = lambda_dbl + y_dbl - mu_dbl; ```

Tickling the error requires modifying lines in 
src/test/unit-distribution/univariate/continuous/pareto_type_2/pareto_type_2_test.hpp:

change line 15 from:

```    param[1] = 0;           // mu```

to:

```     param[1] = 0.1;           // mu ```

and line 19 from:

``` log_prob.push_back(-3.753417975251507474525379613285); // expected log_prob```

to: 

``` log_prob.push_back(-3.548244797701304964476776149240); // expected log_prob```

Run the following script after modifying the pareto_type_2_test.hpp file:

./runTests.py -j2 src/test/unit-distribution/univariate/continuous/pareto_type_2/pareto_type_2_00000_generated_v_test.cpp

Which will result in failures in the finite_diff portion of the tests for test 32 and 36.

PR #1237, opened 2015-01-22
Bugfix/issue 1223 pareto type 2 log bad partials

rtrangucci:
#### Summary:

Fixes the improper calculation of partial derivatives of the pareto_type_2_log with respect to the first three parameters. Addresses #1223 

#### Intended Effect:

Finite diff tests for pareto_type_2_log should no longer fail at non-zero entries for the location parameter, mu.

#### How to Verify:

Verify that the valid_value method for AgradDistributionParetoType2 in src/test/unit-distribution/univariate/continuous/pareto_type_2/pareto_type_2_test.hpp has a non-zero entry in parameters[0][1] which corresponds to mu, and run pareto_type_2_00000_generated_v_test.cpp. This test should run without any failures. 

#### Side Effects:

None.

#### Documentation:

None required.

#### Reviewer Suggestions: 

Anyone.


syclik:
Jenkins, retest this please.

syclik:
Looks good.

PR #1242, opened 2015-01-25
Bugfix/issue 1241 append row return

bob-carpenter:
#### Summary:

Fixes return type of `append_row(vector,vector)` to be `vector`

#### Intended Effect:

See summary.

#### How to Verify:

Unit test added to check return type.  Failed in 2.5, passes in patch.

#### Side Effects:

Nothing user facing.

#### Documentation:

The behavior is being patched to match the doc.

#### Reviewer Suggestions:

Anyone.  This is a very tiny fix.

### DIFF
--- a/src/stan/math/matrix/append_row.hpp
+++ b/src/stan/math/matrix/append_row.hpp
@@ -42,22 +42,6 @@ namespace stan {
       return result;
     }
        
-    //vector append_row(vector, vector)
-    template <typename T1, typename T2, int R1, int R2>
-    inline Matrix<typename return_type<T1, T2>::type, Dynamic, 1>
-    append_row(const Matrix<T1, R1, 1> & A,
-               const Matrix<T2, R1, 1> & B) {          
-      int Asize = A.size();
-      int Bsize = B.size();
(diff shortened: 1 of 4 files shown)


src/stan/math/matrix/append_row.hpp
```diff
@@ -46,7 +46,7 @@ namespace stan {
     template <typename T1, typename T2, int R1, int R2>
     inline Matrix<typename return_type<T1, T2>::type, Dynamic, 1>
     append_row(const Matrix<T1, R1, 1> & A,
-          const Matrix<T2, R1, 1> & B) {          
+               const Matrix<T2, R1, 1> & B) {
```
syclik: @bob-carpenter, should this line have `R2` instead of `R1`?

bob-carpenter: That signature will never get called because
there's no way to infer R2.  The whole thing should
have been removed.  I'll do that and push.

- Bob

randommm:
Hi, sorry for this bug. I just pushed a commit that does for append_col exactly what you've done for append_row.

syclik:
This actually passed -- there was a configuration error that prevented the
performance tests from running.

PR #1243, opened 2015-01-26
Adding a logistic model as a performance test.

syclik:
#### Summary:

Adds a logistic model as a performance test.

#### Intended Effect:

Just adds a test to be used in benchmarking Stan's speed. It's not meant to catch everything.

#### How to Verify:

Run: 
`> ./runTests.py src/test/performance/logistic_test.cpp`

This will create a csv file in `test/performance/performance.csv` that records the model run 100 times, whether the numbers match the last tagged values, and whether all the values are the same.

#### Side Effects:

None. This only touches tests.

#### Documentation:

None.

#### Reviewer Suggestions: 

Anyone. It should be merged soon so I can get Jenkins to display these numbers going forward.


syclik:
I'm going to break protocol and merge this so I can start getting Jenkins to run this. Can someone take a look and let me know if it looks alright? If anything seems off, I'll patch it with a pull request later.

src/test/performance/logistic_test.cpp
```diff
@@ -0,0 +1,234 @@
...
+    time_start = microsec_clock::universal_time(); // start timer
+    stan::common::command<stan_model>(argc, argv);
+    milliseconds_per_run[n] 
+      = (microsec_clock::universal_time() - time_start).total_milliseconds();
+    last_draws_per_run[n] = get_last_iteration_from_file("test/performance/logistic_output.csv");
+  }
+  SUCCEED();
+}
+
+bool close_enough(double expected, double val) {
```
bob-carpenter: Replace `if (cond) return true else return false;`  with `return cond;`

src/test/performance/logistic_test.cpp
```diff
@@ -0,0 +1,234 @@
...
+  static void SetUpTestCase() {
+    N = 100;
+    milliseconds_per_run.resize(N);
+    last_draws_per_run.resize(N);
+    matches_tagged_version = false;
+    all_values_same = false;
+  }
+
+  static int N;
+  static Eigen::Matrix<long, -1, 1> milliseconds_per_run;
```
bob-carpenter: Why Eigen types here rather than std::vector?  I also prefer `Eigen::Dynamic` to `-1`, but it's no big deal.

src/test/performance/logistic_test.cpp
```diff
@@ -0,0 +1,234 @@
...
+template <typename T>
+std::string quote(const T& val) {
+  std::stringstream quoted_val;
+  quoted_val << "\""
+             << val
+             << "\"";
+  return quoted_val.str();
+}
+ 
+std::string get_git_hash() {
```
bob-carpenter: Does this mean our tests will only run if someone has Git installed?  Kind of hard to imagine someone running without git, but I also don't see why it should be a dependency here.

bob-carpenter:
I added some comments inline. 

My biggest request, given that we're now testing all this,
is that we have some spec for the output so we know what's
expected and what can be changed without breaking all the tests.

Will this test break if we change RNG or seed?  I'm still very
nervous about maintaining tests like this going forward
that just freeze existing behavior.  Checking the fit of the
model is one thing, but testing we get to the same adaptation
seems dangerous.

Do you expect this test to be updated each release?  
It has comments that are 2.5.0 specific as "last tagged 
release".

I was confused by parse_command_output.  Could you doc
what the format is that it's parsing?  Is it just
a bunch of key=value pairs, one per line?  

- Bob

syclik:
@bob-carpenter, thanks. I'll doc it all and create a pull request.

PR #1278, opened 2015-02-05
Fixes #1277. Add out-of-tree make/local.

### LINKED ISSUE #1277: Add out-of-tree make/local
Add an additional makefile to store local variables.

syclik:
[skip ci]

### DIFF
--- a/makefile
+++ b/makefile
@@ -39,7 +39,8 @@ LDLIBS_STANC = -Lbin -lstanc
 EXE = 
 WINE =
 
--include make/local    # for local stuff
+-include $(HOME)/.config/stan/make.local  # define local variables
+-include make/local                       # overwrite local variables
 
 ##
 # Get information about the compiler used.


syclik:
From @ido:

#### Summary:

This one-line makefile change adds an out-of-tree makefile in addition to make/local, allowing users to place defines and build configuration (e.g. location of their compiler, additional CXXFLAGS, etc.) in a place outside of the git repository that will persist across otherwise destructive git commands.

#### Intended Effect:

Now, in addition to being able to use `make/local` to put any local build configuration or additional make rules, users can do so in `$HOME/.config/stan/make.local`.  If this file is present, it will be included by GNU make.  If it is not present, nothing should happen.

#### How to Verify:

Try putting a build configuration setting, e.g. setting CXXFLAGS or CXX or CC in ~/.config/stan/make.local and see if it sticks.

#### Side Effects:

If a user forgets that he/she has additional local build customizations in the aforementioned file location, his/her build may include said local build customizations unwittingly.  

#### Documentation:

This is documented in the same fashion as the `make/local` file is currently... :-)

#### Reviewer Suggestions:

Anyone with familiarity with the current (GNUmake-based) Stan build system would be a good reviewer for this.

syclik:
@ido, thanks for the patch. I'm merging this. It doesn't need the full testing. I've verified that it works on my machine.

PR #1281, opened 2015-02-06
Feature/issue 330 parameter error messages

syclik:
#### Summary:

Changes `operator<<` behavior for `var` and `fvar` to only print the value. The current develop version prints both the value and the adjoint, recursively.

The `operator<<` behavior for `vari` (the var implementation pointer within `var`) still prints both the value and the adjoint. (I don't think the original code worked.)

#### Intended Effect:

Makes the print easier to understand for our typical use case: error messages.

#### How to Verify:

Run the unit tests for `var` and `fvar`. There are also print tests in the code.

#### Side Effects:

This affects the output of our `check_*` methods. These will now only print the value of parameters, not the value and the adjoint. 

#### Documentation:

None required. I searched for examples of the rejection messages, but didn't find any in the manual. 

#### Reviewer Suggestions: 
Anyone.


syclik:
Jenkins, retest this please.

src/test/unit/agrad/fwd/matrix/stan_print_test.cpp
```diff
@@ -24,28 +24,28 @@ TEST(AgradFwdMatrixStanPrint,fvar_double) {
   b.push_back(a);
   b.push_back(a);
   stan_print(&output, b);
-  EXPECT_EQ("[1:2,1:2,1:2]",output.str());
+  EXPECT_EQ("[1,1,1]",output.str());
```
rtrangucci: Are we enforcing google-style guide rules for tests? If so, spaces before output.str() throughout the file. If not, ignore.

rtrangucci:
@syclik code looks good

syclik:
Regarding style for Google test, no. That's more work than I'm willing to
take on now. We have 16k warnings in our code base. Once we get it down to
something manageable, we can kill warnings in our tests.

syclik:
I just verified. This patch doesn't introduce any new warnings.

Issue #1297, opened 2015-02-14
return type should be vector for append_row(vector,vector) with mixed double/var inputs


bob-carpenter:
The fix in 2.6 wasn't sufficient for all types and the test was masked by an over general assignment operator that compiles then fails at run time (I'm filing another issue for that).

The three signatures are:

```
template <typename T1, typename T2, int R1, int C1, int R2, int C2>
inline Matrix<typename return_type<T1, T2>::type, Dynamic, Dynamic>
append_row(const Matrix<T1, R1, C1> & A, const Matrix<T2, R2, C2> & B);

template <typename T, int R1, int C1, int R2, int C2>
inline Matrix<T, Dynamic, Dynamic>
append_row(const Matrix<T, R1, C1> & A,  const Matrix<T, R2, C2> & B);

template <typename T>
inline Matrix<T, Dynamic, 1>
append_row(const Matrix<T, Dynamic, 1> & A,  const Matrix<T, Dynamic, 1> & B);
```

Everything works fine for double/double or var/var inputs, but when you append a double and var, only the first pattern matches and the return type's wrong.

My suggested fix is to

* [ ] remove the middle signature; it is subsumed by the first
* [ ] generalize the last signature to mixed types and reimplement to follow the first pattern
* [ ] add more tests on return type  
* [ ] do the same thing for `append_col`
    * [ ] remove middle sig
    * [ ] generalize last sig
    * [ ] add more tests    

This would leave two signatures:

```
template <typename T1, typename T2>
inline Matrix<typename return_type<T1, T2>::type, Dynamic, 1>
append_row(const Matrix<T1, Dynamic, 1> & A, const Matrix<T2, Dynamic, 1> & B);

template <typename T1, typename T2, int C1, int C2>
inline Matrix<typename return_type<T1, T2>::type, Dynamic, Dynamic>
append_row(const Matrix<T1, Dynamic, C1> & A, const Matrix<T2, Dynamic, C2> & B);
```

This is slightly less general from the C++ perspective because we're not supporting fixed sizes for Eigen, but it's all we should need for Stan's language.

I think it may even be possible to collapse to a single signature:

```
template <typename T1, typename T2, int C1, int C2>
inline 
Matrix<typename return_type<T1, T2>::type, Dynamic, std::max(C1,C2)>
append_row(const Matrix<T1, Dynamic, C1> & A, const Matrix<T2, Dynamic, C2> & B);
```

[...]

PR #1298, opened 2015-02-14
Fix append_col and append_row return types + extra tests

randommm:
#### Summary:
Fix issue #1297. append_col and append_row should be returning the correct types now.

#### Intended Effect:
In Stan v2.5.0, append_col was correct and append_row had a bug caused by a typo (R1 instead of R2, as Daniel Lee reported, and in another place). So I thought rolling back to the old version and fixing this typo would be wiser (the old version included specialized versions for matrices of same type. Although they are not necessary, they should make the operation slightly faster, so I see no reason not to have them).

#### How to Verify:
Old + new tests for return type.

#### Side Effects:
None.

#### Documentation:
Not applicable.

#### Reviewer Suggestions:
Bob Carpenter


randommm:
To compare these with old version:

    git fetch -a #to update your branches
    git diff origin/feature/issue-1297-fix-append_row-append_col v2.5.0 -- src/stan/math/matrix/append_col.hpp
    git diff origin/feature/issue-1297-fix-append_row-append_col v2.5.0 -- src/stan/math/matrix/append_row.hpp

bob-carpenter:
Wow --- thanks for fixing this right away.

I'm swamped getting ready for my ML summer school class,
but I can look it over closely next week.  

- Bob

bob-carpenter:
This is basically OK other than the using statement at the namespace level, which we can't have.  I'll fix it and push and then if it passes tests, it's OK by me. 

I find that `<<` notation super-confusing, and have no idea how they made it work the way they do, but it apparently works.  I have no idea if it's more efficient.  

To answer your question in the pull request, the reason not to have multiple definitions, even if one's slightly more efficient, is that it adds maintenance overhead downstream and comprehension overhead for someone like me trying to understand the code and which template function is getting called.

PR #1314, opened 2015-02-26
Feature/issue 1244 parser expects

bob-carpenter:
#### Summary:

Clean up error messages in the parser to make them more understandable for users, remove duplicate output, and to provide errors and reasons in the places where they were missing.

#### Intended Effect:

Make it easier for users to debug Stan programs syntactically.

#### How to Verify:

There are a bunch of new test models that make sure the parser produces the right error.  To get the long form, use stanc to parse the models.  

#### Side Effects:

Error messages change text, which may affect unit tests for CmdStan or other interfaces.

#### Documentation:

Nothing user facing.

#### Reviewer Suggestions:

Anyone.


bob-carpenter:
So how do I figure out what went wrong with CmdStan?

And why are we testing CmdStan for a Stan pull request?

syclik:
HI Bob,

There was some merge problem that wasn't taken care of automatically. I'm
hunting it down. I'll review the pull request today.

There were a couple reasons: 1. I just wanted to know when a change to Stan
broke CmdStan. The idea was to include RStan, PyStan, *Stan, but it never
got implemented. 2. CmdStan used to be our end-to-end test. We now have
tests inside src/test/performance that do pretty much what we need.

syclik:
Awesome!

There are a couple things that need to be fixed:
1. I think a test was removed accidentally: `src/test/unit/agrad/rev/var_stack_test.cpp` I can help recover that.
2. Mind killing some of the cpplint warnings in src/stan/lang? (http://d1m1s1b1.stat.columbia.edu:8080/job/Stan%20Pull%20Request%20-%20CppLint/34/warnings10Result/new/) Not all the "new" ones are actually coming from this pull request -- it's just new relative to the last time it ran.

bob-carpenter:
OK --- I'm out the rest of today, but can get to it over
the weekend.  

Have you actually looked at these?  Some of them are
complaining about things I intentionally did for readability,
such as writing grammar rules as:

  a %= b
     | c
     | d
     ;

I could pull that ";" up to get rid of a "high priority"
issue.

I also put spaces inside and around [ ] because of spacing.
What goes inside is a lazy semantic Boost Phoenix action and
smushing it all together will lead to lines longer than 80
characters and also what I think is unreadable code.  I tried
to make the grammar non-terminals easy to scan, so I wind up doing
this:

  a %= b
       [ semantic action for b ]
     | c
       [ semantic action for c ]
     ;

Let me know what you think I should do syntactically after
looking at the actual files where the warnings come up.

Thanks.

- Bob

syclik:
I'll take a look and let you know. We can also tell cpplint to either
ignore the warning type or annotate the code to skip the warning for that
but of code.

bob-carpenter:
You might want to wait until after I've taken a pass to
clean up the ones I think need to be cleaned up.

How do I generate a new report to make sure I've gotten
things?  Is there a make task?  Do I need to install more
software?  Could you put an answer in the Wiki or point me
to the answer if it's already there?

Thanks,

- Bob

syclik:
Ok. I'll wait until you take a pass.

You'll get a new report if you push. That's a bit unsatisfactory and I
haven't written up how to do it yet, so I'll get that up in the next couple
days. I'll put it up on the wiki. I should also create a make target for it.

In short, cpplint.py is a python script. We already ignore a set of
warnings: I'll get that either documented or tracked or something. Right
now, Jenkins is the only place where the settings are known.

bob-carpenter:
Pushing and waiting several hours isn't going to
be a productive way for me to get rid of the cpplint
errors, because I'm not sure what they're targeting
as the "right" answer.  Also, I have no idea
how to suppress errors, nor am I sure I want to clog
up our code with annotations to get a prettiness checker
to shut up.  

I'll take a single pass through everything now, give
the fix my best shot, then I think we should merge.
I need to start working on another branch that depends
heavily on this one.  If I start reformatting every other
line to get the cpplint program to stop griping (and that's
what it looks like I'll need from the reports), it's going
to be merge hell down the road.

I like pretty code and have strong opinions about it, but
that includes thinking it's a second-order consideration.
I'm willing to reformat where it makes sense, but
I don't think it will for the oddball edge case of
the grammars, which involve overloading standard C++
syntax with lazy semantic evaluations that are huge compared
to the usual indexing inside [] along with single statements
that span dozens of lines (grammar rules with | disjunctions).

- Bob

syclik:
I completely agree. Let's get it merged. I'll add a make target so we can
run it locally without much effort.

syclik:
Quick question: are you going to push a change or should we merge now?

bob-carpenter:
Nope --- I'm ready for it to merge at this point if there
aren't issues other than lint related formatting.

It'd be great to be able to run the lint tool from make.

I'm going to work on some of the other issues that just popped
up and I just created issues for, then if I can get in front 
of those,  I'll try to add line numbers to run-time errors.

- Bob

PR #1318, opened 2015-02-28
Feature/issue 1316 remove long long

bob-carpenter:
#### Summary:

Removes use of `long long` types in `agrad::var` constructor and `io::dump` writer function.  The version in `dump` is generalized to arbitrary types and `agrad::var` just has it removed.

#### Intended Effect:

Allow compilation in older parsers not supporting `long long`, which is technically a C++11 extension.

#### How to Verify:

Run tests, several new ones have been added.

#### Side Effects:

Will no longer support `var(long long)` and `var(unsigned long long)` constructors, which means it won't support `var(size_t)` in 64-bit Windows.  This will not affect Stan programs but could potentially affect uses of the C++ API we don't know about.  The correct fix there will be to static cast to `double`.   

There is a more general solution,

```
template <typename T>
var(T x, typename boost::enable_if<boost::is_arithmetic<T> >::type* /*dummy*/ = 0)
: vi_(new vari(static_cast<double>(x))) {
} 
```

but I'm worried the extra argument may cause unwanted overhead (though I didn't see any in quick tests).
#### Reviewer Suggestions:

Anyone.


bob-carpenter:
Arghh!  This one's going to take me forever to fix
because I don't have a Windows machine so will have to
keep pushing to find the errors in our tests one by one.
And it's going to spam everyone with failed tests all
the while --- please ignore them for this issue.

An example of the problem is this:

```
  vector_v vk(DpKm1);
  for (size_t k = 0; k < DpKm1; ++k)
    vk[k] = (k + 1) * (k + 2);
```

The variable k needs to be declared as size_t in order to index vk 
without the compiler griping, but then the right-hand side is going to need to be 
static cast to an integer or long value in order to act as an input, 
because the fix I put in is such that the var() constructor no long accepts 
unsigned long long, which is what Windows uses for size_t.

I could go with the more complicated enable_if version, which
I describe in the pull request.  I believe that would make everything
work, but then I'm not sure how to test finely enough to make sure
it doesn't hurt performance.

- Bob

syclik:
Bob, how can I help? Do you want me to set up an account on our Windows test box so you can run on it directly?

bob-carpenter:
Thanks.  It may come down to that, but right now, the
current failure has me completely stumped. 

Unfortunately, the enable-if solution that works just
fine on the Mac fails utterly on Windows.  So I reverted
to the list of types, which excludes size_t.  But that
is somehow failing on fvar<fvar<var>> determinant calculations.

I have no idea how long long is getting implicated.

I don't actually think there are as many instances of
the problem as I thought (trying to construct a var with
a size_t on Windows, where size_t is not equal to unsigned
long, but rather equal to the illegal until C++11 unsigned 
long long).

- Bob

bob-carpenter:
Is there some way to ifdef for Windows in the code?

Do we still need to support 32-bit OSes?  If not, I 
may just be able to finagle this thing to working.
But that'll probably mess up some users and, of course,
CRAN.

- Bob

syclik:
Yes, you can. See src/test/unit/agrad/rev/matrix/LDLT_alloc_test.cpp:
#ifndef _WIN32

bob-carpenter:
We tried and failed last time to fix this, which is why
there were long long left in the code.  

I'm pretty sure the size situation is this:

      /*                 uint   ulong     size_t    AMBIGUOUS
       * Windows 32      32        32         32    ulong / size_t
       * Windows 64      32        32         64    
       * Linux/Mac 32    32        64         32
       * Linux/Mac 64    32        64         64    ulong / size_t
       */

What I need is a set of constructors for var() that can handle
any size int and any size unsigned int across platform.

I tried the Boost built-in fixed size types and the C built-in
fixed size types, but I can't get anything to work without causing
ambiguities in constructor resolution on either Windows or the Mac.
I tried Boost's size ifdefs, and couldn't get that to work.

If I try to get away with just 

  var(double)

because whatever the arg is it gets cast to double, it creates
ambiguities with 

  explicit var(vari*)

even though it's explicit when trying to do:

  var x(3L);

If I just use { uint, size_t }, I miss 64-bit longs on 32-bit Linux/Mac.

If I use just { uint, ulong }, I miss 64-bit longs on Windows. 

C++'s messed up sizing just makes me want to weep.

- Bob

syclik:
just poking around in our repo, I see msinttypes.h sitting in
src/msinttypes.h. It's included in stack_alloc.hpp. Not sure if it helps.
Looking at it makes me dizzy.

bob-carpenter:
OK, this is a minor issue because Windows uses a different endline.

I think I may be close.

- Bob

bob-carpenter:
This one's going to be necessary for the long long removal
in order to not invoke any C++11 features that would run
afoul of CRAN or other pedantic compilation requirements. 
long long missed the C++03 boat, but it's been in C for ages,
so all the compilers supported it, which is why it hasn't
been an issue previously.

I suggest we do a 2.6.1 release with this fix, and whatever
else has gone in (I've rewritten most of the parser error
messages and added quite a few more and removed redundant
error reporting [or at least tried to]).

This one's going to be a bit tricky to review, so I'm adding
some hints.  And I now realize that I have some commented
out code lingering because I was playing with bunches of
different options --- that'll have to go before the merge,
but I don't think any behavior will change.  And I'll add
doc.  

I wound up going with a relatively simple solution to the
var() ctors:

      var() : vi_(static_cast<vari*>(0U)) { }

      template <typename T>
      explicit var(T* vi) : vi_(vi) {  }

      template <typename T>
      var(T x) : vi_(new vari(x)) { }

The var(T x) signature applies to arithmetic types and the
var(T* x) signature to vari* and classes that inherit from vari*.
If I just used var(vari* x), any extension of vari would wind up
being ambiguous between var(vari*) and var(T x).  Despite about
10 hours of trying, I couldn't figure out how to explicitly cover
the bases for 32/64 bit and Windows/Unix with explicit constructors
or with enable_if.  I could get the enable_if stuff working on
the Mac, but it blew up on Windows.

I had to then go through the API and change a rather large number
of other ambiguous calls, including a fair number of what wound up
being ambiguous calls to matrix arithmetic, mostly instances of 
double * Matrix<var,..> that needed to be replaced with our multiply()
implementation.  That's good, because what was happening before would
have been casting the double to a var with the implicit var(double x)
constructor.

- Bob

syclik:
This looks good to me. We'll get this merged and we can plan for the 2.6.1 release.

Issue #1312, opened 2015-02-25
prim/mat/meta/is_vector_like.hpp weirdness through check_finite.hpp


rtrangucci:
matrix specialization for is_vector_like doesn't work as it should. In order for is_vector_like to work correctly, mat/meta/is_vector.hpp must be included, which shouldn't be the case.

bob-carpenter:
Do you need help with this, and if so, could you create a branch with a test that fails that you think should pass?

bob-carpenter:
Also, whenever creating an issue, please give it a milestone and labels, and ideally an assignee.

syclik:
This issue was moved to stan-dev/math#79

PR #1320, opened 2015-02-28
Feature/issue 1222 autodiff reorg

rtrangucci:
#### Summary:

Reorganization of math library in order to break cross dependencies within headers and tests.

#### Intended Effect:

There are several intentions with the reorganization:

* Break interdependencies between headers and unit tests along two axes: container type, and scalar type
  * Allows C++ users access to primitive-only, reverse-mode-only, fwd-mode-only versions of Stan's math library
  * Allows C++ users access to Eigen-free and std::vector-free versions of Stan's math library    
* Allow for intuitive navigation of file directories
* Move each function to its own header file

Organization proceeded according to the map:

[scalar-type]/
  [container-type]/
    [functional specification]/

stan/
  math/
    fwd/
      arr/      
      mat/
      scal/
        err/
        fun/
        functor/
        meta/
        prob/
      
    rev/
    prim/

#### How to Verify:

`make test-math-dependencies | wc -l`

should return 72. The remaining math dependencies almost exclusively lie within stan/math/scal/meta/, where VectorView and OperandsAndPartials proved to be too complex to tear apart at this stage. However, this a major step in the right direction. 

#### Side Effects:

In the current branch, there are order dependencies in some of the template meta-programs that did not exist previously (which was function of how tangled the headers were, rather than a sign of scrupulous layout).

See issue #1312 

#### Documentation:

New doxygen doc needs to be written, however it will be handled in a separate issue.

#### Reviewer Suggestions:

Bob, Daniel, Rob


rtrangucci:
Jenkins, retest this please!

rtrangucci:
Jenkins, retest this please

rtrangucci:
Jenkins, retest this please

bob-carpenter:
Tests should automatically get kicked off when you merge.
I think the polite retest command is just to nudge Jenkins
after it goes down for some reason.  Daniel will know for sure.

rtrangucci:
Absolutely, understood. The problem is that the workspace needs to be wiped out before testing a radical new reorg so the auto merge build is flawed and almost certainly will fail. Need to manually clear the workspace and kick off again.  

Rob

syclik:
Jenkins, retest this please.

syclik:
Jenkins, retest this please.

syclik:
I have no idea what went wrong. It said it ran out of disk space. I'll just put it back on the testing queue. Hopefully it'll work next time.

PR #1328, opened 2015-03-03
Bugfix/issue 1327

bgoodri:
Two-line fix allows stan-functions*.txt to show up with a return type of `T[]` instead of `[`. You only have to do make manual to verify it is working now.


syclik:
Looks good. We'll merge after verifying.

PR #1338, opened 2015-03-04
Bugfix/issue 1337 var ctor

bob-carpenter:
#### Summary:

Going back to explicitly typed var constructors to make the implicit promotions all work without ambiguity.  Requires an ifdef for `_WIN64` to add `var(size_t)` 

#### Intended Effect:

Allow all our models and tests to compile without C++11.

#### How to Verify:

Run tests.

#### Side Effects:

None.

#### Documentation:

Nothing user facing here.

#### Reviewer Suggestions:

Anyone.

### DIFF
--- a/src/stan/agrad/rev/var.hpp
+++ b/src/stan/agrad/rev/var.hpp
@@ -183,6 +183,7 @@ namespace stan {
       // in WIN64, size_t = unsigned long long, which is a C++11
       // feature, so use this ctor to enable it
       var(size_t x) : vi_(new vari(static_cast<double>(x))) { }
+      var(ptrdiff_t x) : vi_(new vari(static_cast<double>(x))) { }
 #endif
 
       /**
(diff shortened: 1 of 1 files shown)


bob-carpenter:
As far as reviewing goes, this version has the same `var()` implementations as before, but without `long long` for Mac/Linux.   I don't think we need `long long`.

syclik:
Looks good. Hopefully this doesn't wreak havoc on Windows 32 bit machines, but we'll take care of that if users have issues.

bob-carpenter:
It shouldn't be a problem for Win32.  (More
famous last words.)

In Win32, size_t is unsigned int and ptrdiff_t is
int, so we already have that covered in the basic
constructors. 

- Bob

syclik:
Great.

Want me to put that on a plaque and put it up in our office?

bob-carpenter:
Too bad we don't have one of those brutalist buildings
beloved of computer science departments --- then we could
just carve it into our cinder block walls.

- Bob

PR #1343, opened 2015-03-05
Bugfix/issue 1342 get base index except

bob-carpenter:
#### Summary:

Fix bug being thrown by matrix double-indexing by changing check_range to use cols instead of rows for second index.

#### Intended Effect:

Fix this bug!

#### How to Verify:

I added a unit test that failed in 2.6 but now passes for the entire range, including all in-range and out-of-range inputs.

#### Side Effects:

None.

#### Documentation:

Behavior now matches doc.

#### Reviewer Suggestions:

Daniel.


syclik:
@bob-carpenter, thanks! It looks like I introduced the bug. Thanks for fixing it and getting a test in.

syclik:
If the distribution tests pass, this should be merged. (Jenkins is going to incorrectly mark this build as a failed build.)

PR #1350, opened 2015-03-08
Bugfix/corr free

### LINKED ISSUE #1349: corr_cholesky_free and corr_cholesky_constrain are missing some includes
Have fix now

bgoodri:
This fixes #1349 and allows models to compile


bob-carpenter:
Thanks, this looks good.  I have no idea why I was able to compile Rob's model with the last fix I put in for `var()`.

syclik:
Do these headers need to be includes here? It didn't look like they were used directly.

Did some model fail to build?

Issue #1363, opened 2015-03-12
__float128 ifdef for gcc issues (lkj_cholesky)


bob-carpenter:
Joshua Pritkin reports a failure on stan-users where the Eigen lib is trying to bring in the GCC-specific `__float128` data type in testing `lkj_cholesky`.   I hope this doesn't mean Eigen's actually trying to use (rather than just parse or maybe compile) it.

PR #1364, opened 2015-03-12
added ifdefed var(__float128) ctor to solve #1363

syclik:
#### Summary:

Adds ifdefed var(__float128) constructor for issues with g++-4.9.

#### Intended Effect:

Allows src/test/unit/math/mix/mat/prob/lkj_corr_test.cpp to be run on linux with g++-4.9.

#### How to Verify:

Run that test.

#### Side Effects:

This might end up using the templated version of boost's functions instead of the `stan::agrad` version of functions, but if this is triggered, it has already been happening implicitly.

#### Documentation:

None.

#### Reviewer Suggestions: 

Daniel. (Bob submitted the actual pull request.)


syclik:
@bob-carpenter, thanks! Looks good. I'll check on a virtual machine and if it works there, I'll tag it as reviewed.

@bgoodri, mind double checking this patch. If it fails, comment and take away the review complete label (if it's been marked).

maverickg:
It seems that this does not pass for g++ 4.9.2 on Mac

./runTests.py src/test/unit/math/mix/mat/prob/lkj_corr_test.cpp

maverickg:
Adding `#include <stan/math/rev/scal/fun/digamma.hpp>` fixes it.

bgoodri:
Ironically, it does now pass on g++-4.9 on Linux.

bob-carpenter:
Thanks so much for testing.  This variability across compilers is
crazy-making. 

- Bob

maverickg:
I might made a mistake when checking out the branch.  I thought I did, but maybe I didn't.

maverickg:
It passes after I correctly checked out this branch.  Sorry.

syclik:
Great news! Thanks!

How did you install 4.9 on Mac? Just so I know.

maverickg:
> How did you install 4.9 on Mac? 

1. install brew (see http://brew.sh)
2. run `brew tap homebrew/versions`, and then run `brew install gcc49`

syclik:
Thanks!

betanalpha:
I found MacPorts to be great for installing and managing all of the
various Clang and gcc compilers while brew kept conflicting with
the built-in compiler.

syclik:
That's also good to know. I wasn't planning on installing a new compiler
just yet. I just wanted to know how to do it when I needed to.

So either brew or MacPorts. But brew has gcc 4.9 for sure.

syclik:
for some reason, Windows is affected by this fix:
http://d1m1s1b1.stat.columbia.edu:8080/job/Stan%20Pull%20Request%20-%20Windows%20-%20Tests%20-%20Unit/89/

bgoodri:
Works on Linux with both clang++ and g++-4.9.

bob-carpenter:
I'd guess that whatever is causing issues with the lkj_corr
density is the same root cause as what's going on in 
mdivide_right_spd.  I don't have any other clue as to what's
going on.

Here's the relevant bit of output:

```
Running main() from gtest_main.cc
...
[----------] 14 tests from AgradFwdMatrixMdivideRightSPD
...
[ RUN      ] AgradFwdMatrixMdivideRightSPD.ffd_exceptions
src/test/unit/math/fwd/mat/fun/mdivide_right_spd_test.cpp:409: Failure
Expected: mdivide_right_spd(rvf1, fd1) throws an exception of type std::domain_error.
  Actual: it throws nothing.
[  FAILED  ] AgradFwdMatrixMdivideRightSPD.ffd_exceptions (1 ms)
[----------] 14 tests from AgradFwdMatrixMdivideRightSPD (3 ms total)
```

- Bob

bob-carpenter:
Could someone try adding a !WINDOWS specification to that include.
Because WINDOWS worked before adding the var(__float128) constructor,
so it doesn't seem to be necessary.  I can try if it's OK to just
keep pushing and letting Jenkins sort out Windows.

- Bob

syclik:
I'll log into the Jenkins machine and try it out directly.

syclik:
I ran it locally and it didn't have any problems. I just merged `develop` back into the branch. Hopefully that solves the problem.

syclik:
Looks like it'll pass without issues. We'll tag after this gets in.

bob-carpenter:
Great news.  And thanks to everyone for all the help in
making this work.  The new code just has some ifdefs for Windows
with its unusual sizes for long int (32 bits) and for g++ 4.9
with its nonstandard type (__float128).  

This is the problem with vague standards that don't fix sizes
and non-standard extensions. 

- Bob

syclik:
Yes! Fixed... we just needed to clear thew workspace and run it again.

Issue #1326, opened 2015-03-03
print errors arising during initialization


bob-carpenter:
We need to print the errors in order to debug models such as the following:

```
> fit <- stan(model_code="parameters { vector[100] u; } model { u ~ gamma(1,1); }")

TRANSLATING MODEL 'anon_model' FROM Stan CODE TO C++ CODE NOW.
COMPILING THE C++ CODE FOR MODEL 'anon_model' NOW.

SAMPLING FOR MODEL 'anon_model' NOW (CHAIN 1).
Error : 

Initialization between (-2, 2) failed after 100 attempts. 
 Try specifying initial values, reducing ranges of constrained values, or reparameterizing the model.

error occurred during calling the sampler; sampling not done
```

syclik:
Bob, I just created a pull request for this. I copied the error message from the other initialization methods. We might want to do better since we're looking at it.

PR #1378, opened 2015-03-15
Print error messages in initialization

syclik:
#### Summary:

Print errors in random initialization.

#### Intended Effect:

These were accidentally omitted. This should help make debugging models a little easier.

#### How to Verify:

Run the model in #1326. This should now print error messages.

#### Side Effects:

None.

#### Documentation:

None.

#### Reviewer Suggestions: 

Bob.

### DIFF
--- a/src/stan/services/init/initialize_state.hpp
+++ b/src/stan/services/init/initialize_state.hpp
@@ -161,11 +161,22 @@ namespace stan {
               << std::endl;
             init_log_prob = -std::numeric_limits<double>::infinity();
           }
-          if (!boost::math::isfinite(init_log_prob))
+          if (!boost::math::isfinite(init_log_prob)) {
+            if (output)
+              *output << "Rejecting initialization at zero "
+                      << "because of vanishing density."
+                      << std::endl;
             continue;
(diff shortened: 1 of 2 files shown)


bob-carpenter:
This looks OK, but is there some way to actually print whatever caused the problem?  Or did that get fixed at some point?

syclik:
Bob, it's been fixed. If you change the model to:
```
model {
  reject("foo");
}
```
The output looks like:
```
...
Informational Message: The current Metropolis proposal is about to be rejected because of the following issue:
foo
If this warning occurs sporadically, such as for highly constrained variable types like covariance matrices, then the sampler is fine,
but if this warning occurs often then your model may be either severely ill-conditioned or misspecified.
Rejecting proposed initial value with zero density.
Rejecting initialization at zero because of vanishing density.
...
```

Of course, that spits out 100 times, but it does get reported.

bob-carpenter:
Great.

I should've just tested myself --- it looked like that
fragment was just going to swallow the error.

syclik:
Fixes #1326.

PR #1382, opened 2015-03-15
Bugfix/issue 1381

### LINKED ISSUE #1381: variable length array in hypergeometric
Fix coming
```
src/stan/math/prim/scal/prob/hypergeometric_rng.hpp:45:19: warning: variable length arrays are a C99 feature [-Wvla-extension]
```

bgoodri:
This fixes #1381 
The hypergeometric distribution headers had variable length arrays, which is technically a C99 feature according to the warning, which I have changed to std::vectors.


bob-carpenter:
Looks good.  We shouldn't have been using arrays anyway.

syclik:
This failed because the file uses std::vector. I can either add another
exception to the rule or we can think about how to write it differently. I
think adding another exception might be reasonable.

bgoodri:
Why does using std::vector need an exception? We use it all the time. It
was replacing a variable-length array, which we shouldn't be using ever.

syclik:
after splitting apart the math library, we now have these levels within
src/stan/math/{prim, rev, fwd, mix}:
scal: scalar depends on nothing, DOES NOT INCLUDE vector or Eigen.
arr: can depend on scal, can include vector, DOES NOT INCLUDE Eigen
mat: can depend on scal and arr.

syclik:
the exception is to the test we have to check for these dependencies.

bob-carpenter:
Ben:  The error's because Daniel put a check in place to make
sure we don't violate our include rules and create more spaghetti.
And I think the "exception" meant exception to our validation rules,
not a C++ exception.  

Everyone:  I think we should leave the rules in place.  This
one doesn't even need the std::vector.  See below.

I'll rewrite this to be more efficient.  And to fix the overincluding.
And to fix all the bad spacing.  Give me a few minutes.

I may have a final question before I'm done about how cdf got into
the distribution.

- Bob

bob-carpenter:
I fixed it by having the cdf() evaluated lazily.  And I cleaned
up the includes.  We shouldn't just be bringing in all the distributions
for a single usage!

A side benefit is that the hypergeometric_rng should have 
expected log(N) calls to the CDF instead of N.

As I was going, I realized that we're still defining the
probability functions in namespace stan::prob.  Too bad we didn't change
that to stan::math in the last refactor. 

IMPORTANT
================================================================
I think we have to be *much* more careful now that we're expecting
people to use the C++ API.  That massive refactor completely broke
backward compatibility at the API level.  At this point, if we try
to rename stan::math to stan::prob, that will also break backward 
compatibility, so I don't think we should do this until Stan 3.

- Bob

bgoodri:
If we need those rules, fine by me. But are we really expecting people to use the C++ API now? Since we broke everything on a 2.6.0 -> 2.6.2 transition I can't imagine anyone trusts us not to rebreak it for 2.7.0.

bob-carpenter:
> On Mar 16, 2015, at 4:34 PM, bgoodri <notifications@github.com> wrote:
> 
> If we need those rules, fine by me.

"need" is a perhaps too strong.  

The reason they're there is to catch any inadvertent mistakes
we might make with include dependencies in our own code.
I would very much like to remove the spaghetti-like dependencies
we created over time in the math lib, and I think the current
re-org is a step in the right direction.

I just wrote another e-mail to the stan-dev group taking up
the issue of when the dust might settle on the remodeling.

- Bob

PR #1385, opened 2015-03-16
added check of positive definite matrix to cholesky decompose

syclik:
### TO DO
- [ ] add checks for negative definite and semi positive definite matrices for LDLT or LLT if necessary
- [ ] add tests for new functions
- [ ] replace existing checks with new check functions
- [ ] remove check_ldlt_factor in favor of check_pos_definite

#### Summary:

Adds a positive definite check to `cholesky_decompose`.

#### Intended Effect:

Helps trap errors.

#### How to Verify:

Run the tests.

#### Side Effects:

This will slow down calls to `cholesky_decompose`, but will make it easier to track errors.

#### Documentation:

Fixed api doc.

#### Reviewer Suggestions: 

@betanalpha or @bgoodri


bgoodri:
This does not make sense to me.  The `check_pos_definite` does pretty much the same thing as `cholesky_decompose` except that it is LDLT instead of LLT. Why not just copy these lines
```
      if (cholesky.info() != Eigen::Success
          || !cholesky.isPositive()
          || (cholesky.vectorD().array() <= CONSTRAINT_TOLERANCE).any())
        domain_error(function, name, y, "is not positive definite:\n");
```
from `check_pos_definite` to `cholesky_decompose` (except instead of checking vectorD() check the diagonal of L against the square root of CONSTRAINT_TOLERANCE).

syclik:
I think the right thing to do would be to add a check function that does exactly what you suggest, after the decomposition. 

What would that be called? Can we write another check function for `check_pos_definite`? How does this look?
```
    template <typename T_y>
    inline bool
    check_pos_definite(const char* function,
                       const char* name,
                       const Eigen::LLT<Eigen::Matrix<T,Eigen::Dynamic,Eigen::Dynamic> >& cholesky) {
                if (cholesky.info() != Eigen::Success
          || !cholesky.isPositive()
          || (cholesky.vectorD().array() <= CONSTRAINT_TOLERANCE).any())
        domain_error(function, name, y, "is not positive definite:\n");
      check_not_nan(function, name, y);
      return true;
    }
```

bgoodri:
On Sun, Mar 15, 2015 at 11:40 PM, Daniel Lee <notifications@github.com>
wrote:

Should be

>     template <typename T_y>
>     inline bool
>     check_pos_definite(const char* function,
>                        const char* name,
>                        const Eigen::LLT<Eigen::Matrix<T,Eigen::Dynamic,Eigen::Dynamic> >& cholesky) {
>                 if (cholesky.info() != Eigen::Success
>           || !cholesky.isPositive()
>           || (cholesky.matrixL().diagnonal().array() <= 0).any())
>         domain_error(function, name, y, "is not positive definite:\n");
>       return true;
>     }
>
> I think the check_nan is not necessary because the LLT will fail these
checks if there are any NaN in the lower triangle. I guess there could be
NaN in the upper triangle but then it would have failed the symmetry check.
Also, the CONSTRAINT_TOLERANCE of 1e-8 was way too big; it is valid (albeit
numerically suspect) for any positive diagonal values.

bob-carpenter:
I just checked and <= creates an expression template, so the
last line should be efficient.

You don't want to indent that first line like that.  

The T_y doesn't match T, and overall, I think you can probably
simplify to use just:

  template <typename Derived>
  ... LLT<Derived>& cholesky

Is there a case where there are NaN entries and the decomposition
returns Eigen::Success and .isPositive()?  If not, get rid of
the redundant check.

- Bob

bob-carpenter:
Awesome --- I think that constraint tolerance may be one of
the things messing up our multivariates.

- Bob

syclik:
sorry... I had just done a crude cut and paste job late at night. It sounds like what I sketched was the right way to go, so I'll get that implemented.

rtrangucci:
@syclik Should we close this pull and open a new issue to write a check_pos_definite that accepts an Eigen::LLT type?

syclik:
No. I'll just fix this branch.

syclik:
@bgoodri, for an `Eigen::LLT<Derived>& cholesky`, I can't do this:

- `cholesky.isPositive()`: no member called `isPositive()`; can I just ignore?
- `cholesky.matrixL().diagnonal().array() <= 0).any()`; no `.diagonal()` available for triangular view. What can I do to make this work?

bgoodri:
I think you can ignore the lack of an isPositive() method since we are
checking the diagonal entries for positiveness. You might see if
matrixLLT().diagonal().array() works, although their documentation says
that its layout is undocumented. If that doesn't work or is in the wrong
layout, then I guess you have to assign the .matrixL() to a dense matrix
before you can evaluate the .diagonal() of that.

syclik:
It definitely doesn't compile. What's the best way of assigning to a dense
matrix.... I'm lost with the templating, I might need to find a way to get
the scalar type of the Eigen derived type?

bgoodri:
Works for me

goodrich@T540p:/tmp$ cat test_LLT.cpp
#include <Eigen/Dense>
#include <iostream>

int main() {
  Eigen::MatrixXd A(3,3);
  A << 4,-1,2, -1,6,0, 2,0,5;
  std:: cout << "The matrix A is" << std::endl << A << std::endl;
  Eigen::LLT<Eigen::MatrixXd> lltOfA(A); // compute the Cholesky
decomposition of A
  Eigen::MatrixXd L = lltOfA.matrixL(); // retrieve factor L in the
decomposition
  std::cout << "The Cholesky factor L is" << std::endl << L << std::endl;
  std::cout << "Its diagonal is" <<
lltOfA.matrixLLT().diagonal().transpose() << std::endl;
  return 0;
}

goodrich@T540p:/tmp$ clang++-3.6 -I/usr/include/eigen3/ test_LLT.cpp
goodrich@T540p:/tmp$ ./a.out
The matrix A is
 4 -1  2
-1  6  0
 2  0  5
The Cholesky factor L is
       2        0        0
    -0.5  2.39792        0
       1 0.208514   1.9891
Its diagonal is      2 2.39792  1.9891

syclik:
Yeah... but the problem is the signature is going to be:
```
    template <typename Derived>
    inline bool
    check_pos_definite(const char* function,
                       const char* name,
                       const Eigen::LLT<Derived>& cholesky)
```

Presumably `Derived` can be anything, but in general, can I replace your
`MatrixXd` with `Derived`?

And... you switched to using `.matrixLLT()` instead of `.matrixL()`: is
that what I want?

bgoodri:
I think you can switch to Derived. That is what I was speculating that the
.matrixLLT() method returns an expression for the internal matrix, rather
than the triangular view. So, you can extract the diagonal. But the
.matrixL() method returns an expression for the lower triangular view, so
that is what should be (and is) used in the density.

bob-carpenter:
We shouldn't need to worry about scalar type since our
comparisons are all overloaded.  Just make sure not to 
use the numeric_limits versions that add extra vars for is_nan().

And no reason to test that the diagonal is positive if we
test it's above a threshold.  

- Bob

bgoodri:
The threshold to be tested should be zero. It is 1e-8 currently for no good
reason.

bob-carpenter:
OK.  You know way more about this than me.  I don't know
what issues will show up with severely ill-conditioned matrices.

- Bob

src/test/unit/math/prim/mat/err/check_pos_definite_test.cpp
```diff
@@ -68,6 +96,80 @@ TEST_F(ErrorHandlingMatrix, checkPosDefinite_non_symmetric) {
...
+  expected_msg1_llt << "function: Cholesky decomposition of y failed";
+  Eigen::LLT<Eigen::Matrix<double,Eigen::Dynamic,Eigen::Dynamic> > llt_err1(y); 
+  EXPECT_THROW_MSG(check_pos_definite(function, "y", llt_err1),
+                   std::domain_error,
+                   expected_msg1_llt.str());
+  expected_msg1_ldlt << "function: LDLT decomposition of y failed";
+  Eigen::LDLT<Eigen::Matrix<double,Eigen::Dynamic,Eigen::Dynamic> > ldlt_err1 = y.ldlt(); 
+  EXPECT_THROW_MSG(check_pos_definite(function, "y", ldlt_err1),
+                   std::domain_error,
+									 expected_msg1_ldlt.str());
```
syclik: Tabs? Get rid of them.

src/stan/math/prim/mat/err/check_pos_definite.hpp
```diff
@@ -12,11 +12,12 @@
 #include <stan/math/prim/scal/err/check_positive_size.hpp>
 #include <stan/math/prim/mat/fun/Eigen.hpp>
 #include <stan/math/prim/mat/fun/value_of_rec.hpp>
-
+#include <cstdio>
```
syclik: Can we safely remove this include to cstdio?

rtrangucci:
Addressed your code review comments

rtrangucci:
I'm looking into this


Rob

syclik:
Jenkins, retest this please.

rtrangucci:
@syclik I'm not sure what's happening here. The tests that fail here pass on my machine, and it says I'm up-to-date with origin/feature/issue-1076-cholesky_decompose

syclik:
I'll take a look.

syclik:
Yeah--I have no idea why it failed. I tried taking develop and merging in the changes and it's fine. I really don't know how it could have gotten into that state.

syclik:
Jenkins, retest this please.

PR #1396, opened 2015-03-18
Feature/issue 1195 runtime error lines

bob-carpenter:
#### Summary:

Adds line numbers from source Stan program to error message output.

#### Intended Effect:

Help users debug programs (and discourage them from including their Stan source as strings in R code).

As an extra bonus, I refactored the reject unit tests that inspect the content of rejection error messages.  And fixed all the spacing/indenting in the generated code.

What's actually going on under the hood is that the parser now stores source line numbers in the statement nodes of the abstract syntax tree.  This was a monster of Spirit Qi programming and why I couldn't figure it out until recently.  It's black magic to create an old-style rule using `raw[...]`, then reset the LHS `value_` using semantic actions.

These are then used as follows in every block where statements are executed:

```
line_number__ = -1;
try {
   line_number__ = 3;
   ... code for statement at line 3...
  line_number__ = 4;
  ... code for statement at line 4...
 ...
  create and throw new exception of same runtime type 
  as e including values of `e.what()` and `line_number__`
}
```

Line numbers are only incremented for start of statement.  And not for statement sequences, which are made up of smaller sequences and would lead to N^2 updates if done naively.

Creating and throwing the new exception is pretty simple.  My original strategy of trying to use nested exceptions Java-style won't work. Instead, I use RTTI to inspect the type of exception that is thrown and then throw a new exception of the same type with a line number and the original `what()` message.  I initially thought I'd be able to use nested exceptions Java-style, but that won't be possible until C++11, and maybe even then won't do much more than Boost exceptions, which require the nested exceptions to inherit from a special class.

#### How to Verify:

You can run buggy models and see what happens.  Or just follow the revised unit tests for rejection, which now test that the line numbers are being properly reported, too.

I verified there was no noticeable slowdown on programs.  As in zero noticeable effect.  The reviewer may want to reverify.

#### Side Effects:

Error messages change.  

#### Documentation:

There's no existing doc on all the error messages, so nothing has to change.

#### Reviewer Suggestions:

Daniel should make sure the right differeing behavior is still available.  If necessary, the new exception adapter in `stan/lang`can be updated.


bob-carpenter:
Everything's passing but CppLint, which reports
14,232 CppLint "errors" in stan/services then craps out. 
I would've expected many pieces of lint to be found in
stan/lang.  

What do I need to do from here?

- Bob

syclik:
Bob, I'm good with it. Thanks for adding in tests.

PR #1404, opened 2015-03-24
updating error messages

syclik:
#### Summary:

Updates error messages on initialization.

#### Intended Effect:

Trying to make the 3 error messages clearer:

--------

Old:
```
Rejecting proposed initial value using method user because of gradient failure.
Rejecting proposed initial value using method random because of gradient failure.
Rejecting proposed initial value using method zero because of gradient failure.
Rejecting proposed initial value using method user partially because of gradient failure.
```
New:
```
Rejecting initial value:
  Error evaluating the log probability at the initial value.
```

------------

Old:
```
Rejecting proposed initial value using method user because of vanishing density.
Rejecting proposed initial value using method random because of vanishing density.
Rejecting proposed initial value using method zero because of vanishing density.
Rejecting proposed initial value using method user partially because of vanishing density.
```
New:
```
Rejecting initial value:
 Log probability evaluates to log(0), i.e. negative infinity.
 Stan can't start sampling from this initial value.
```

---

Old:
```
Rejecting proposed initial value using method user because of divergent gradient.
Rejecting proposed initial value using method random because of divergent gradient.
Rejecting proposed initial value using method zero because of divergent gradient.
Rejecting proposed initial value using method user partially because of divergent gradient.
```
New:
```
Rejecting initial value:
  Gradient evaluated at the initial value is not finite.
  Stan can't start sampling from this initial value.
```

---

New tests added checking the error message coming out of the initialize methods.

#### How to Verify:

Look at code. Run tests.

#### Side Effects:

This pull request will force CmdStan's tests to change.

#### Documentation:

None required.

#### Reviewer Suggestions: 

Anyone.


syclik:
@bob-carpenter, @rtrangucci, @jgabry, @maverickg: any thoughts on these error messages? It would be easy to change now. I omitted the type of initialization -- it didn't seem important in the error message.

bob-carpenter:
"can not" -> "cannot" or "can't" or "is unable"

- Bob

jgabry:
I think they're better than the current error messages and probability as good as you can do unless there's a way to provide the user with useful suggestions for troubleshooting (but that's hard).

maverickg:
I think the function name `initialize_state_value` can be changed too as what it's really doing is some kind of validation.  I kind of didn't realized that when I consolidate all those checks into one function.  Also if `init_how` is not used, we can leave it out.

syclik:
@bob-carpenter: changed to "can't"
@maverickg: I removed `init_how` and removed the redundant validation (only in one of the methods)

syclik:
Are we good with this error message?

rtrangucci:
Looks good to me

Issue #1341, opened 2015-03-05
Making X * b go fast for large sparse X


sakrejda:
After much prodding from Bob, this is an issue summarizing a conversation strewn over a few stan-users exchanges...

Since Stan programs benefit from pre-computing as much as possible outside of the parameters/transformed parameters blocks, models can often be simplified and sped up by generating a model matrix and reducing the in-Stan computation to $X \times b$-type matrix multiplication.  This is applicable to a wide variety of hierarchical models as well as spline models with truncated basis functions (my current use case).  Stan does not yet have sparse data types, but it is possible to represent a sparse X matrix using a set of vectors and (for example) Eigen's compressed sparse column format.  I implemented a simple Stan function for doing this multiplication (https://groups.google.com/d/msg/stan-users/mtF3udlP3PM/nCLhXXBTlNIJ) and am in the process of re-doing an implementation that relies on Eigen's format as described at (http://eigen.tuxfamily.org/dox/group__TutorialSparse.html).  The new implementation with more explicit documentation is here: https://github.com/sakrejda/sparse-stan

From some informal tests I see that you can get all of the expected memory benefit and a small speed gain from a Stan function ignoring the gradient calculation. The gradient is simple and everybody says that's where the real speed gains are.  

To-do's:
- [x] finish implementing Stan-level function
- [x] find a similar C++ function to work from
- [x] follow the instructions for adding a C++ level function (stan-users and the static page).
- [ ] custom gradients??

bob-carpenter:
> On Mar 5, 2015, at 1:54 PM, Krzysztof Sakrejda <notifications@github.com> wrote:
> 
> After much prodding from Bob,

The prodding was to have the discussion on stan-dev.

> this is an issue summarizing a conversation strewn over a few stan-users exchanges...
> 
> Since Stan programs benefit from pre-computing as much as possible outside of the parameters/transformed parameters blocks, models can often be simplified and sped up by generating a model matrix and reducing the in-Stan computation to $X \times b$-type matrix multiplication. 

Is there an example of where this kind of thing is faster?  My intuition
is that it'd be slower than just doing the custom multiplication in
the model.  

The advantage of (X * beta) is that we can do custom gradients.  
We should be doing them, but I'm not sure they're plugged
in everywhere, hence issue #1339.

[...]

PR #1405, opened 2015-03-24
Feature/sparse multiply

sakrejda:
This pull request adds two functions to Stan---sparse_multiply_csc and sparse_multiply_csr, which implement sparse matrix-by-vector multiplication.  The goal is to allow the common X * b pattern for model specification to be used in Stan efficiently, since the model matrix X is often sparse.  Lacking a sparse matrix type in Stan, the matrix must be passed as four vectors.  I also added extractors as C++ functions which can be used to turn a matrix into these vectors in either compressed (sparse) column (CSC) or compressed (sparse) row (CSR) format for use with the appropriately-named function.  For the CSR format I followed Marcus' suggestion to rely on dot_product for most of the computation and benefit from the associated gradients.  

I have a branch for rstan at least where I have added an R wrapper for the extractors for use with either a dense model matrix or R's sparse matrix package ("Matrix").  

Issue 1341 goes with this branch (at https://github.com/stan-dev/stan/issues/1341) as does a stan-dev thread (https://groups.google.com/forum/#!topic/stan-dev/X45dGQkaags). 

In one of the thread posts (https://groups.google.com/d/msg/stan-dev/X45dGQkaags/k_A9eZ3J8zwJ) Bob lays out what the benefits of doing the calculations this way should be and in two relatively big models so far (X about 200kx50, 90% sparse, lots of other computations) I've seen roughly a 2x improvement in speed.  The memory improvement is as-expected---you only store non-zero entries.  Performance-wise I think it's a worthwhile improvement and I'm using it for my own models  with no trouble.

In terms of inclusion into Stan I'm ambivalent.  This is awkward to document and I am concerned about generating a lot of confusion from users trying to create the sparse matrices without using the included extractors.  I could see adding only one sparse format and one set of extractors (CSR, since that includes dot_product gradients) and a very prescriptive doc about implementation for a mixed model in the manual.  A second concern seems to be that the only API I found in Eigen for the low-level access I need uses pointers and pointer arithmetic.  This isn't an issue if the included (and tested) extractors are used to generate the sparse matrix representation but it's certainly easy to construct a sparse matrix representation that will make Stan access outside the allocated vector and crash.


syclik:
Looks like there are sign comparison warnings that should be taken care of:
http://d1m1s1b1.stat.columbia.edu:8080/job/Stan%20Pull%20Request%20-%20Tests%20-%20Unit/270/warnings18Result/new/

syclik:
@sakrejda, I'm seeing some really odd behavior with your test, but we can start with simple:

- [ ] can you remove tabs
- [ ] split the function into two files
- [ ] split the test into two files

So... the weird behavior I'm getting is that if I try to kill the warnings by casting the size to an int within `_csr`, the `_csc` test fails due to calculated numbers being different. My guess is that we've gone over some memory boundary somewhere, but I can't be sure yet.

syclik:
@sakrejda, found it. You MUST initialize the Eigen Matrix with 0. You can't expect it to be 0. It'll be dependent on hardware, run, bits it grabs to start.

sakrejda:
Looks like there are a couple places I forgot to do setZero, not just in the tests.  I'll go through your comments and try to get this in shape.

sakrejda:
@syclik Oops---this test is going to fail.  I forgot to add the piece about casting size to an int and didn't realize it would trigger the test just when I pushed to my branch.

bob-carpenter:
You shouldn't be casting unsigned to signed or vice-versa 
anywhere.  Just use the right signedness from the start.

- Bob

sakrejda:
@bob-carpenter On the one hand I have 'long' types from Stan (I don't think those can be declared unsigned in Stan) and on the other I have scalar_t/unsigned_long from the .size function of a std::vector.... not sure how to avoid a cast here.

bob-carpenter:
There are no long types declared in Stan.

If a is an array (std::vector), use:

  for (size_t i = 0; i < a.size(); ++i)
    ...

If v is a vector (Eigen::Matrix), use:

  for (int i = 0; i < v.size(); ++i)
    ...

You can use the i for indexing either one without the
compiler griping.

This has been a huge pain from the get-go ever since
we started trying to stomp out all the pedantic error
messages.  I still think unsized types were a colossal 
mistake in C++.  It's not like we need the extra bit of
resolution and they lead to all sorts of potential bugs
which are hard to catch.

- Bob

sakrejda:
The warnings are not from loops, they're from sending an int and a size_t to check_equal.  I think the only options are to cast or to drop the checks but this is not something I have a lot of familiarity with:
      stan::math::check_equal("sparse_multiply_csc","n/b", n, b.size());
      stan::math::check_equal("sparse_multiply_csc","n/u", n, u.size()-1);
      stan::math::check_equal("sparse_multiply_csc","n/z", n, z.size()  );
      stan::math::check_equal("sparse_multiply_csc","w/v", w.size(), v.size());
      stan::math::check_equal("sparse_multiply_csc","u/z/v", u[n-1] + z[n-1]-1, v.size());

bob-carpenter:
All of the `const int&` types should be replaced with just `int`.  It's a smaller type to pass and also C++ is pass-by-value, so it's implicitly constant from the client code perspective anyway.

bob-carpenter:
If you want to comply with the new style regulations,

* spaces after commas in function calls and template parameter declarations
* spaces around operators such as + and - (the exception being * and / applied to variables or literals)

bob-carpenter:
The size comparisons are not a problem with your code, it's
in check_equal, which should be defined for int as well as size_t.

Eigen should be returning int sizes and your sizes coming in are int.  I'll create another issue, which hopefully Daniel or Rob can sort out before your code gets merged.  The check code's already so complex that I can't easily see how to fix it.

syclik:
Bob, the issue is that check_equals is templated on both arguments and one
of the sizes is called on an Eigen type (int) and the other is called on
std::vector (size_t).

Could we just write an overload that takes two ints?

syclik:
@sakrejda, use check_size_match instead of check_equal: src/stan/math/prim/scal/err/check_size_match.hpp.

That should reduce most of the warnings.

sakrejda:
Will do, thanks for looking into it.

sakrejda:
@syclik Now I just don't know what's up---is this me doing something screwy or Jenkins?

bob-carpenter:
Nah --- this is a GitHub issue --- they've aparently
been suffering a denial-of-service attack all day.

- Bob

sakrejda:
Ok, I'll just wait then. Krzysztof


Krzysztof Sakrejda

Department of Biostatistics and Epidemiology
University of Massachusetts, Amherst
319 Morrill Science Center South
611 N. Pleasant Street
Amherst, MA 01003

work #: 413-325-6555
email: sakrejda@umass.edu

syclik:
Jenkins, retest this please.

syclik:
@sakrejda, the cpplint tests failed -- too many new warnings.

I'll try to take a look at your updates this weekend.

sakrejda:
@syclik If you guys have specific ways of dealing with cpplint I'd appreciate pointers but I can probably squash these myself, I've dealt with lint-ing before...

syclik:
http://d1m1s1b1.stat.columbia.edu:8080/job/Stan%20Pull%20Request%20-%20CppLint/93/warnings11Result/new/

syclik:
take a look at that link -- those are all the new warnings that were
introduced in the code. Click on the "Details" tab to see info on how to
fix.

sakrejda:
Do you guys have a specific setup you use with cpplint? It's easier to do this if I can run it locally.

syclik:
yes. it's in `make/cpplint`. from the command line, type:
    make cpplint

That'll give you all the warnings. We're trying to cut down.

sakrejda:
Took care of most of the lint errors but there are some that I don't understand (header order) and some that I disagree with completely (no spaces allowed for vertical alignment).  Reading tests is bad enough as it is... anyway, if it fails, I'd appreciate some help deciding what to do with the rest.

syclik:
Sure. I'll make a pass at the rest when I review the code.

syclik:
Ignore that failure for now. It's unrelated. (It's weird and I think I know
what it is, but it's not due to your pull request.)

sakrejda:
Ok.

syclik:
@sakrejda, very good start. It's not quite ready to go, but I think we can outline steps to get this functionality in. I'll make more specific code comments using GitHub.

__General comments__

* Inclusion into Stan:
    * I am supportive of including the `sparse_multiply_csr` and `sparse_multipy_csc` functions into Stan.
    * I am also supportive of deprecating the functions in the future in favor of a built-in type.
* `extract` functions:
    * It makes sense to expose some version of these functions to the users. Even if we don't have a built-in data type now, it would be useful to read in a dense matrix, then represent it using a sparse matrix. This might only make sense within the transformed data block.
    * I would like to expose some version of this to the users. Even if we don't have a built-in data type now, it would be useful to read in a dense matrix, then represent a sparse matrix using helper functions, but perhaps this should only apply to data?
    * Currently, the `extract` functions only accept `Eigen::SparseMatrix`, so these shouldn't be exposed to the users.
* Thoughts about storage format? ([wikipedia link for sparse matrix](http://en.wikipedia.org/wiki/Sparse_matrix#Storing_a_sparse_matrix)) 
    * I understand that CSR / CSC allows for efficient operators, but is this natural for the most number of users? (why isn't there agreement on the acroymn? it switches between CSR and CRS...)
    * Should we expose just one of the two?
    * It looks like dictionary of keys (DOK) is the most intuitive.

 
__Fixes__

* [ ] Both `sparse_multiply_csr` and `sparse_multiply_csc` need to be hardened. Right now, these are unsafe.
    * [ ] `v`: check that each element is between 1 and `n`
    * [ ] `u`: check that it's ordered, each element is greater than or equal to 1, that every element is less than or equal to length of `w`
    * [ ] `z`: check that each number is greater than or equal to 0. Sum is equal to length of `w`.
* [ ] Add additional function signatures for `sparse_multiply_csr` (currently only sparse x dense)
    * [ ] sparse x sparse
    * [ ] dense x sparse
* [ ] Add additional function signatures for `sparse_multiply_csc` (currently only sparse x dense)
    * [ ] sparse x sparse
    * [ ] dense x sparse
* [ ] change explicit 1-index to `stan::error_index::value` -- this is `#define`d as 1, but can be changed to 0
* [ ] Add a constructor of a dense matrix using the sparse format.
* [ ] Move `extract` functions to `src/test` directory -- they aren't really useful as Stan source, but as test utilities.

sakrejda:
Thanks for taking a more thorough look at the code.  I'm willing to keep
pushing this but I think it needs some more discussion.

A couple of issues around formats:
- I went with CSR/CSC because that's what Eigen uses and for me this is
primarily a way of thinking about how to properly do sparse stuff in Stan.
- DOK, referred to as "triplets" in Eigen, is intuitive but unless you
impose an additional constraint like ordered indexes I'm having trouble
seeing how do do an efficient multiplication.
- I think it makes sense to just use CSR internally, that would make it
less work to expand the signatures and test everything properly.  CSR is
the format that lets us use dot_product internally to get custom gradients
and provides us with the greatest speed benefit.
- If we expand the signatures to include sparse x sparse and dense x sparse
we would need a separate format for the vector---I'm pretty sure the CSR
code would work with minor modification if the sparse vector were just a
pair for each non-zero value (index and value).
- My only concern with expanding the signatures is that this is a temporary
hack and should therefore have a minimal footprint but I'm ok doing the
full set for one format.

A couple of issues around the extract functions:
- I would be fine with exposing extractors in Stan.  I didn't go there
initially because for some large dense matrices rstan is unable to either
a) pass them to Stan or b) dump them using stan_rdump so I think the usage
is limited.
- Instead I patched a version of rstan to expose the extractors in R so
that only the sparse format matrix would need to be passed.
- Would the extractors be available to rstan if they were in src/test?

To move on the two issues I'm unsure of are:
1) Is just using CSR for matrices and index/value pairs for sparse vectors
ok?
2) Where should the extractors go?  The only use cases I'm really attached
to are: 1) exposed in the interfaces; and 2) available for running tests.

Krzysztof




Krzysztof Sakrejda

Department of Biostatistics and Epidemiology
University of Massachusetts, Amherst
319 Morrill Science Center South
611 N. Pleasant Street
Amherst, MA 01003

work #: 413-325-6555
email: sakrejda@umass.edu

syclik:
Thanks for taking a more thorough look at the code.  I'm willing to keep
    pushing this but I think it needs some more discussion.

Thanks for getting this far with it. We really need to update our wiki documentation. Feel free to tackle some of it yourself, if you'd like.
    
    A couple of issues around formats:
    - I went with CSR/CSC because that's what Eigen uses and for me this is
    primarily a way of thinking about how to properly do sparse stuff in Stan.

Ok, that makes sense.

    - DOK, referred to as "triplets" in Eigen, is intuitive but unless you
    impose an additional constraint like ordered indexes I'm having trouble
    seeing how do do an efficient multiplication.

You're right. We should just stick to a format that has a bit more structure.

    - I think it makes sense to just use CSR internally, that would make it
    less work to expand the signatures and test everything properly.  CSR is
    the format that lets us use dot_product internally to get custom gradients
    and provides us with the greatest speed benefit.

Are you suggesting not supporting both? I'm ok with that.

    - If we expand the signatures to include sparse x sparse and dense x sparse
    we would need a separate format for the vector---I'm pretty sure the CSR
    code would work with minor modification if the sparse vector were just a
    pair for each non-zero value (index and value).

Do you think we'd really want to support sparse vectors as opposed to sparse matrices with 1 column?

    - My only concern with expanding the signatures is that this is a temporary
    hack and should therefore have a minimal footprint but I'm ok doing the
    full set for one format.

Good point. I don't know if we have a reasonable proposal for a sparse data type, so this temporary hack may exist much longer than we anticipate.

    
    A couple of issues around the extract functions:
    - I would be fine with exposing extractors in Stan.  I didn't go there
    initially because for some large dense matrices rstan is unable to either
    a) pass them to Stan or b) dump them using stan_rdump so I think the usage
    is limited.

Ok. Let's not expose them to the language. I think the extractors should move to test code.

    - Instead I patched a version of rstan to expose the extractors in R so
    that only the sparse format matrix would need to be passed.
    - Would the extractors be available to rstan if they were in src/test?

No.

    
    To move on the two issues I'm unsure of are:
    1) Is just using CSR for matrices and index/value pairs for sparse vectors
    ok?

I would suggest we don't support sparse vectors yet. Is there a way to specify a sparse vector vs sparse row_vector?

    2) Where should the extractors go?  The only use cases I'm really attached
    to are: 1) exposed in the interfaces; and 2) available for running tests.

If the interfaces need it, it should be in the source directory, but if it doesn't, then it should be in the test directory. How would this be exposed to the interfaces? If we were doing the full fix, I don't think we'd have to worry about this.

sakrejda:
I'm going to summarize a little here so we don't get lost but I'll also
answer questions below.  The minimal "hack" as I see it is to do sparse
matrix by dense vector multiplication so that X*b is a viable
model-specification strategy for complex models. The goal is to get around
the fact that a model matrix is typically sparse, sometimes very sparse,
and sometimes very large.

This leads us to use the CSR format for the sparse matrix with four
components, so the signature for the function is:

sparse_multiply(X_w, X_v, X_u, X_z, b)

In CSR each row is represented explicitly, and the storage format only
becomes efficient as the matrix becomes wider.  The CSR format also allows
us to use dot_product internally for multiplication which is why I favor
using only this storage format.

Because CSR represents each row explicitly it is completely inefficient for
storing a column vector, even for a row vector X_v is superfluous (only w,
u, and z are needed).  I resisted including something like sparse x sparse
multiplication because the efficient signature would be something like

sparse_multiply(X_w, X_v, X_u, X_z, b_w, b_u, b_z) or sparse_multiply(X_w,
X_v, X_u, X_z, b_w, b_i)

I think this is so confusing it's just asking for trouble from users.
Maybe we could rename the function to something like

Xb(X_w, X_v, X_u, X_z, b) or sparse_mm(X_w, X_v, X_u, X_z, b)

so that it doesn't sound like we're making promises about doing sparse
multiplication in any generality.  Would that be ok for just sticking to
the single operation, rather than trying for a complete set of signatures?

It's really not a calculation issue, at this point I've read enough I could
implement any of these, it's not having the types available that's killing
us here.

The last piece is the interface-side stuff.  To expose extractors in rstan
I added one function in rstan/src and one in rstan/R (see below).  These
functions let the user generate a dense matrix or a sparse matrix in the R
format and pass them to the "extract_sparse_parts" function which returns a
list with the components (w, v, u, z).  These can then go in as data to a
stan_rdump() or stan() call.

If we keep the complexity down by just doing sparse X by dense b I could
provide the same sort of thing for pystan/julia.

Krzysztof



rstan/src:

#include <stan/math/prim/mat/fun/sparse_extractors.hpp>

#include <Rcpp.h>
#include <RcppEigen.h>

RcppExport SEXP extract_sparse_components(SEXP A) {
  Eigen::SparseMatrix<double> AA(Rcpp::as<Eigen::SparseMatrix<double> >(A));
  Eigen::Matrix<double, Eigen::Dynamic,1> w = stan::math::extract_w(AA);
  std::vector<double> ww(w.size());
  for (unsigned int i=0; i < w.size(); ++i)
    ww[i] = w(i);

  std::vector<int> v = stan::math::extract_v(AA);
  std::vector<int> u = stan::math::extract_u(AA);
  std::vector<int> z = stan::math::extract_z(AA);

  return Rcpp::List::create(
      Rcpp::Named("w") = ww,
      Rcpp::Named("v") = v,
      Rcpp::Named("u") = u,
      Rcpp::Named("z") = z
  );

}


rstan/R:

extract_sparse_parts <- function(A, type='CSC') {
  if (class(A) == 'matrix') {
    A <- Matrix(A, sparse=TRUE)
  }
  if (type == 'CSR') {
    A <- t(A)
  }
  if (class(A) == 'dgCMatrix') {
    o <- .Call("extract_sparse_components",A)
  } else {
    stop("Argument 'x' must either be of class 'matrix' or 'dgCMatrix'.")
  }
  o[['type']] <- type
  return(o)
}




On Thu, Apr 2, 2015 at 1:23 AM, Daniel Lee <notifications@github.com> wrote:

> Thanks for taking a more thorough look at the code.  I'm willing to keep
> pushing this but I think it needs some more discussion.
>
> Thanks for getting this far with it. We really need to update our wiki
> documentation. Feel free to tackle some of it yourself, if you'd like.
>
> A couple of issues around formats:
> - I went with CSR/CSC because that's what Eigen uses and for me this is
> primarily a way of thinking about how to properly do sparse stuff in Stan.
>
> Ok, that makes sense.
>
> - DOK, referred to as "triplets" in Eigen, is intuitive but unless you
> impose an additional constraint like ordered indexes I'm having trouble
> seeing how do do an efficient multiplication.
>
> You're right. We should just stick to a format that has a bit more
> structure.
>
> - I think it makes sense to just use CSR internally, that would make it
> less work to expand the signatures and test everything properly.  CSR is
> the format that lets us use dot_product internally to get custom gradients
> and provides us with the greatest speed benefit.
>
> Are you suggesting not supporting both? I'm ok with that.
>
Agreed, I think just CSR.


> - If we expand the signatures to include sparse x sparse and dense x sparse
> we would need a separate format for the vector---I'm pretty sure the CSR
> code would work with minor modification if the sparse vector were just a
> pair for each non-zero value (index and value).
>
> Do you think we'd really want to support sparse vectors as opposed to
> sparse matrices with 1 column?
>
If we stick to just CSR, a m-1 "sparse matrix" takes NNZ + NNZ + m + m
entries to store.  It's the wrong sparse
format for it.


> - My only concern with expanding the signatures is that this is a temporary
> hack and should therefore have a minimal footprint but I'm ok doing the
> full set for one format.
>
> Good point. I don't know if we have a reasonable proposal for a sparse
> data type, so this temporary hack may exist much longer than we anticipate.
>
Repeating myself from the summary, but could we just keep this to a model
matrix * parameter vector operation, call it
something like Xb(), and limit the hack to it's one known good use case?  I
would rather spend my time figuring out a
proposal for how to include sparse types right than trying to turn this
hack into something fully functional (and really ugly).


> A couple of issues around the extract functions:
> - I would be fine with exposing extractors in Stan.  I didn't go there
> initially because for some large dense matrices rstan is unable to either
> a) pass them to Stan or b) dump them using stan_rdump so I think the usage
> is limited.
>
> Ok. Let's not expose them to the language. I think the extractors should
> move to test code.
>
> - Instead I patched a version of rstan to expose the extractors in R so
> that only the sparse format matrix would need to be passed.
> - Would the extractors be available to rstan if they were in src/test?
>
> No.
>
> To move on the two issues I'm unsure of are:
> 1) Is just using CSR for matrices and index/value pairs for sparse vectors
> ok?
>
> I would suggest we don't support sparse vectors yet. Is there a way to
> specify a sparse vector vs sparse row_vector?
>
> 2) Where should the extractors go?  The only use cases I'm really attached
> to are: 1) exposed in the interfaces; and 2) available for running tests.
>
> If the interfaces need it, it should be in the source directory, but if it
> doesn't, then it should be in the test directory. How would this be exposed
> to the interfaces? If we were doing the full fix, I don't think we'd have
> to worry about this.
>
OK, so I'd like to keep it in src, for rstan all I had to add was a quick
wrapper for extractors in rstan/src and an R function in rstan/R (see above)

bob-carpenter:
"CSR" and "CRS" both stand for "compressed row storage":

http://en.wikipedia.org/wiki/Sparse_matrix#Compressed_row_Storage_.28CRS_or_CSR.29

For the language, I'm working on ragged arrays first.  Sparse arrays
and matrices are going to have to wait.  I thought it was decided to
go ahead with this limited functionality and then maybe deprecate it later
when we have sparse data structures built in?  If not, this is all a moot
point.

I do not like using "Xb" as a function name, nor "sparse_mm".  If you think
"sparse_multiply" is too confusing.  I'd prefer the long form of 
"sparse_matrix_times_vector()" because then people reading the code know
what it does without looking in the manual (even if they don't know the encoding).
And maybe we need "csr" in the name?

- Bob

sakrejda:
So between all the comments I can see a way through this.  @syclik , let me know if this is acceptable.  Based on the exchange with you and Bob I am dropping the CSC format and keeping the 'extractor' functions in the Stan directory so that interfaces can access them.  I will add a dense matrix constructor as you suggested to make round-tripping within Stan easier.  

__Fixes__
* [ ] Make just one function---named "csr_matrix_times_vector"
* [ ] Both `csr_matrix_times_vector` needs to be hardened. Right now, it is unsafe.
    * [ ] `v`: check that each element is between 1 and `n`
    * [ ] `u`: check that it's ordered, each element is greater than or equal to 1, that every element is less than or equal to length of `w`
    * [ ] `z`: check that each number is greater than or equal to 0. Sum is equal to length of `w`.
* [ ] change explicit 1-index to `stan::error_index::value` -- this is `#define`d as 1, but can be changed to 0
* [ ] Add a constructor of a dense matrix using the sparse format.

syclik:
@bgoodri, @bob-carpenter, can you weigh in on this?

@sakrejda, that looks ok to me, but I want to make sure Bob and Ben are ok with it.

bob-carpenter:
I've already commented on it.  As I said before, I'd much rather
get everything in, but if people find this function useful, I
don't think it's a big deal.  If we start getting dozens (or even
handfuls) of these, I'm going to worry.


- Bob

bgoodri:
I'm good with this too. We have to start somewhere with the sparse stuff and if we have to change it later, then fine.

PR #1406, opened 2015-03-24
Removing TO-DO.txt and README.txt

syclik:
#### Summary:

Removing unused text files.

#### Intended Effect:

TO-DO.txt is online in our wiki here: https://github.com/stan-dev/stan/wiki/To-Do-List
I've updated the online one (minimally).

README.txt is out of date and I don't think it's necessary. If anyone thinks we need to have a version of this around, we can add it back in.

#### How to Verify:

See the diff in github.

#### Side Effects:

None.

#### Documentation:

Just removing out-of-date doc.

#### Reviewer Suggestions: 

@bob-carpenter


syclik:
Just FYI, there's a mention of `README.md` in the manual, but that file was left alone. I removed `README.txt`.

bob-carpenter:
It seems odd not to have a README with the distribution, but it's OK by me.

bob-carpenter:
This doesn't really need a merge build, does it?

syclik:
We still have the README.md.

And... no, we don't.

syclik:
The alternative is to include a README.txt. What do you want in it? The old
one hadn't been touched since v1.3.0.

bob-carpenter:
Oh, right --- the .md actually goes out with the distribution.
I keep thinking of it as something that shows up on the GitHub
page only.

Presumably nobody's going to be confused as to what a .md file
is and it's human readable as text.

So unless others have comments, I suggest just merging.

- Bob

PR #1409, opened 2015-03-25
removing mdivide_left_spd and mdivide_right_spd

syclik:
#### Summary:

Removes `mdivide_left_spd` and `mdivide_right_spd` from Stan. These functions were never exposed to users and are not used within Stan.

#### Intended Effect:

Code cleanup. The implementation looked a bit off and at the very least, the error checking wasn't efficient. Just removing code that we aren't actively using / maintaining.

#### How to Verify:

Check the code diff.

#### Side Effects:

None.

#### Documentation:

None.

#### Reviewer Suggestions: 

@bgoodri. Can you tell me if we need to keep this around for any reason?


bgoodri:
My guess is that these need to be fixed up and exposed rather than deleted.
Not having efficient ways to solve Ax = b for x (unless A is triangular) is
a bug.

bob-carpenter:
Yes, we should expose these, not delete them. 

Not exposing them was a lack of an important feature.
I'd reserve the term "bug" for when behavior doesn't match 
the doc (it can then be a bug on either the doc or
code side or both).

- Bob

bgoodri:
I would include "accidentally unexposed" under bug :)

bob-carpenter:
To paragraphse Andrew, when you poke a former semanticist's buttons, 
he'll beep.

The Wikipedia has the usual definition as I understand
it:

    A software bug is an error, flaw, failure, or 
    fault in a computer program or system that causes it 
    to produce an incorrect or unexpected result, or to 
    behave in unintended ways.

It doesn't include things like "I sure wish they had banded
matrices built into Stan."

Reminds me of the quote that's so popular among semantics
Ph.D. students who put quotes at the beginning of chapters:

   "When I use a word," Humpty Dumpty said, in rather a scornful tone, 
   "it means just what I choose it to mean --- neither more nor less."

In fact, I think "Humpty Dumpty" would be a good alternative nickname
for Andrew, and having one person playing fast and loose with conventional
meanings in a project is already confusing enough!

- Bob

syclik:
Ok. I'll close this pull request.

@bgoodri, I'm going to assign the issue original to you. I'll outline what
I need from you to get this done.

PR #1413, opened 2015-03-29
Adds cpplint to lib

syclik:
#### Summary:

Adds cpplint.py and the Google C++ Style guide to our distribution within lib, with license information.
Updates the `make cpplint` target.

#### Intended Effect:

Allows developers to run cpplint without having to set up anything else.

#### How to Verify:

Run:
`make cpplint`

from the command line.

#### Side Effects:

None.

#### Documentation:

Already updated the makefile. That's where it's documented.

#### Reviewer Suggestions: 

Anyone.


bob-carpenter:
Works for me.  Did this address Allen's issue about Python 2.x vs. Python 3.y?

syclik:
Unfortunately, no. It looks like cpplint only works with python 2. It
wasn't really documented anywhere.

PR #1414, opened 2015-03-30
cpplint warnings: removed all from src/stan/math

syclik:
#### Summary:

Removes all cpplint warnings from `src/stan/math`.

#### Intended Effect:

**Note: this does not change code behavior.**

Removes cpplint warnings by:
- changing casts
- changing whitespace
- adding braces where necessary (`{` and `}`)
- in a very few places, added `// NOLINT` to disable cpplint rules

#### How to Verify:

The tests haven't changed, so passing all tests show that it works as intended.

The way to verify this is by eye, but it should be good if it passes tests. Someone should spot-check some of the changes by hand, though.

#### Side Effects:

This may affect existing pull requests. If this is ok, I'll handle the merges.

#### Documentation:

None.

#### Reviewer Suggestions: 

Anyone.


syclik:
Jenkins, retest this please.

syclik:
The failure comes from the `test-math-dependencies` having added a few more issues for:
```
#include <vector>
```
in places where it was missing. I'm going to relax the test-math-dependencies test because I think this is fine for this pull request and we still need to think about how to break apart the dependencies in another effort.

syclik:
Jenkins, retest this please.

bob-carpenter:
OK by me.

I don't think std::vector is such a big deal if used
internally --- it's the API and dependencies on 
Eigen that we should try to tease apart.  At least in
my opinion.

- Bob

PR #1417, opened 2015-04-02
Update all directory includes.

syclik:
#### Summary:

Added missing directory includes. Updated all directory includes.

#### Intended Effect:

Adds include header file for each directory.
Also made the `#define` directive more consistent.

#### How to Verify:

Inspect. This doesn't affect anything else.

#### Side Effects:

None.

#### Documentation:

None.

#### Reviewer Suggestions: 

@bob-carpenter. This shouldn't take more than a quick inspection of the files. I used a script to generate them.


bob-carpenter:
We don't want to use double underscores anywhere, right?  I know that's what I did originally, but the C++ spec reserves them for system internals.  So ALL of our double underscores need to be eradicated.  It's on me to do the ones in the language.  A slightly different name mangling for the generated models is on me, though --- it shouldn't be part of this pull request.  But maybe we could get them out of all the header guards.

bob-carpenter:
Other than the double underscores, this looks great.

Then we can think about an include directory if we need utility includes that don't follow the directory structure.  

Thanks for taking this on.

syclik:
I forgot about the double underscores.

Bob, I took a sweep over ALL the `#define` statements so they do not have double underscores. The last commit, d5d53b2, is just that, no additional changes. And this is just over our source, not our tests.

It changes almost every single header file, so the diff is going to be harder to read, but merges should still be straightforward. If tests pass without a hitch, I'll merge.

bob-carpenter:
Sounds good.  Thanks.

betanalpha:
We’re just talking about _leading_ double underscores, right?

bob-carpenter:
Following are reserved:

  * all double underscores
  * leading underscore followed by capital letter
  * leading underscore in global namespace

See:

http://stackoverflow.com/questions/228783/what-are-the-rules-about-using-an-underscore-in-a-c-identifier

which quotes the spec.  

- Bob

PR #1419, opened 2015-04-06
Feature/fix 1399

### LINKED ISSUE #1399: replace std::cout in initialization
see L156 and L246 of https://github.com/stan-dev/stan/blob/develop/src/stan/services/init/initialize_state.hpp

Potentially, this could cause complain from CRAN.

bgoodri:
Eliminate `std::cout` from src/stan/math/prim/mat/fun/cholesky_corr_constrain.hpp because it may be called in rstan. I think you can then close #1399 .


syclik:
This looks fine by me.

PR #1421, opened 2015-04-11
Feature/bbvb

akucukelbir:
#### Summary:

Automatic Differentiation Variational Inference (ADVI).

#### Intended Effect:

Implements ADVI as an approximate inference procedure within Stan.

#### How to Verify:

Run your favorite model. Run the tests. 

```
$ ./runTests.py src/test/unit/variational
```

#### Side Effects:

Currently, we only output the mean of the variational approximation. We might want to implement a "sample, transform, report" module for ADVI and other future inference techniques.

#### Documentation:

- [x] write documentation in cmdstan manual
- [x] write documentation in stan manual


akucukelbir:
i guess this is telling me that something went wrong on windows? how should we proceed?

bob-carpenter:
Go to the issue conversation page:

https://github.com/stan-dev/stan/pull/1421#issuecomment-92108928

Click on the "details" link next to the failed merge build
to go to Jenkins.

If you don't have a login, get one from Daniel.

Then you'll see the unit tests failed and can troll through
the console output, where you'll find this:

clang++ -I src -isystem lib/eigen_3.2.4 -isystem lib/boost_1.55.0 -Wall -DBOOST_RESULT_OF_USE_TR1 -DBOOST_NO_DECLTYPE -DBOOST_DISABLE_ASSERTS -pipe -Wno-unused-function -Wno-tautological-compare -Wno-c++11-long-long -Wsign-compare -ftemplate-depth=256    -O0 lib/gtest_1.7.0/src/gtest_main.cc test/unit/io/stan_csv_reader.o -DGTEST_USE_OWN_TR1_TUPLE -Wc++11-extensions -Wno-c++11-long-long -isystem lib/gtest_1.7.0/include -isystem lib/gtest_1.7.0 -o test/unit/io/stan_csv_reader test/libgtest.a 
In file included from src/test/unit/io/mcmc_writer_test.cpp:2:
In file included from src/test/test-models/good/io_example.hpp:3:
In file included from src/stan/model/model_header.hpp:19:
In file included from src/stan/services/command.hpp:30:
src/stan/variational/advi.hpp:272:36: error: use of undeclared identifier 'function'
        stan::math::check_positive(function,
                                   ^
src/stan/variational/advi.hpp:275:36: error: use of undeclared identifier 'function'
        stan::math::check_positive(function,
                                   ^
src/stan/variational/advi.hpp:421:36: error: use of undeclared identifier 'function'
        stan::math::check_positive(function,
                                   ^
src/stan/variational/advi.hpp:424:36: error: use of undeclared identifier 'function'
        stan::math::check_positive(function,
                                   ^
4 errors generated.

- Bob

akucukelbir:
Jenkins, retest this please.

akucukelbir:
hi @syclik @bob-carpenter @betanalpha, 

i could easily see that the first failure was my fault. i fixed the bug.

but this one is confusing. i dug a bit into this failure and it makes very little sense to me. some sort of services test is failing i think?

syclik:
@akucukelbir, I'll check on this today.

syclik:
So, this is the test that fails. You can run this from the command line:
```
./runTests.py src/test/unit/services/init/command_init_test.cpp 
```

I'll need to hunt this down, but it looks like some part of the command structure isn't correct and we're trying to print `stan::agrad::var` without it being imported properly.

akucukelbir:
thanks for identifying this dan. i'll take a look at this shortly.

akucukelbir:
i think this has something to do with the order in which Stan libraries are imported. (@syclik similar to when my model tests wouldn't compile and you changed the ordering.)

if i remove all of the code i added to `command.hpp` (lines 662-775) but keep the following include statement
```
#include <stan/variational/advi.hpp>
```
then the error still persists. this is the only include i added to `command.hpp`.

if i comment out the include statement, the error goes away.

so something that i'm including within `stan/variational/advi.hpp` is causing trouble.

i think it's `#include <stan/math/prim.hpp>`. so if i comment out the advi include and instead include `stan/math/prim.hpp` the same error occurs. i think it has something to do with `value_of_rec` again?

syclik:
Thanks. I'll take a look.

syclik:
Alp, so good news, it's running. Bad news, your tests are searching for
exact strings -- I think I mentioned this to you earlier, but on Windows
machines, `nan` gets printed as `1.#QNAN`, so this won't pass as-is.

http://d1m1s1b1.stat.columbia.edu:8080/job/Stan%20Pull%20Request%20-%20Windows%20-%20Tests%20-%20Unit/150/testReport/%28root%29/advi_params_fullrank_test/mean_vector/

akucukelbir:
agh damn.

i thought your helper script took care of the windows thing?

ok, how do i proceed? is it just the nan that's a problem?

akucukelbir:
ok, i've identified the culprits. (see bottom of message.)

can i replace these with `EXPECT_THROW( nasty_function_call(), std::invalid_argument);` ?

```
unit/variational/advi_params_fullrank_test.cpp
40:  double nan = std::numeric_limits<double>::quiet_NaN();
41:  Eigen::Vector3d mu_nan = Eigen::VectorXd::Constant(3, nan);
44:                      "Mean vector is nan, but must not be nan!";
45:  EXPECT_THROW_MSG(stan::variational::advi_params_fullrank my_advi_params_fullrank_nan(mu_nan, L);,
49:          "Input vector is nan, but must not be nan!";
50:  EXPECT_THROW_MSG(my_advi_params_fullrank.set_mu(mu_nan);,
53:  Eigen::MatrixXd L_nan = Eigen::MatrixXd::Constant(3,3,nan);
55:          "Cholesky factor is not lower triangular; Cholesky factor[1,2]=nan";
56:  EXPECT_THROW_MSG(stan::variational::advi_params_fullrank my_advi_params_fullrank_nan(mu, L_nan);,
81:  double nan = std::numeric_limits<double>::quiet_NaN();
82:  Eigen::MatrixXd L_nan = Eigen::MatrixXd::Constant(3,3,nan);
84:          "Cholesky factor is not lower triangular; Cholesky factor[1,2]=nan";
85:  EXPECT_THROW_MSG(my_advi_params_fullrank.set_L_chol(L_nan),
134:  double nan = std::numeric_limits<double>::quiet_NaN();
135:  Eigen::Vector3d x_nan = Eigen::VectorXd::Constant(3, nan);
138:          "Input vector is nan, but must not be nan!";
139:  EXPECT_THROW_MSG(my_advi_params_fullrank.to_unconstrained(x_nan);,

unit/variational/advi_params_meanfield_test.cpp
36:  double nan = std::numeric_limits<double>::quiet_NaN();
37:  Eigen::Vector3d mu_nan = Eigen::VectorXd::Constant(3, nan);
40:                      "Mean vector is nan, but must not be nan!";
41:  EXPECT_THROW_MSG(stan::variational::advi_params_meanfield my_advi_params_meanfield_nan(mu_nan, sigma_tilde);,
45:          "Input vector is nan, but must not be nan!";
46:  EXPECT_THROW_MSG(my_advi_params_meanfield.set_mu(mu_nan);,
49:  Eigen::Vector3d sigma_tilde_nan = Eigen::VectorXd::Constant(3,nan);
51:          "Sigma tilde vector is nan, but must not be nan!";
52:  EXPECT_THROW_MSG(stan::variational::advi_params_meanfield my_advi_params_meanfield_nan(mu, sigma_tilde_nan);,
72:  double nan = std::numeric_limits<double>::quiet_NaN();
73:  Eigen::Vector3d sigma_tilde_nan = Eigen::VectorXd::Constant(3, nan);
76:          "Input vector is nan, but must not be nan!";
77:  EXPECT_THROW_MSG(my_advi_params_meanfield.set_sigma_tilde(sigma_tilde_nan);,
123:  double nan = std::numeric_limits<double>::quiet_NaN();
124:  Eigen::Vector3d x_nan = Eigen::VectorXd::Constant(3, nan);
127:          "Input vector is nan, but must not be nan!";
128:  EXPECT_THROW_MSG(my_advi_params_meanfield.to_unconstrained(x_nan);,
```

bob-carpenter:
Yes, that's the right thing to do, and you can also catch the
exception explicitly and use string comparison to make sure it
has the right bits in it other than NaN, which will be stable
across platform.

try {
  foo();
  FAIL() << "should have thrown exception";
catch (exception you expect e) {
  EXPECT_TRUE(test e.what() for string);
}

- Bob

syclik:
Alternatively, you can build the string with std::stringstream, which will
match the output.

(Bob, the EXPECT_THROW_MSG is a macro I wrote that has one additional
argument, the string to search for, and does what your code does by
searching for the existence of the string in the error message.)

bob-carpenter:
> On Apr 15, 2015, at 1:53 PM, Daniel Lee <notifications@github.com> wrote:
> 
> ...

> (Bob, the EXPECT_THROW_MSG is a macro I wrote that has one additional
> argument, the string to search for, and does what your code does by
> searching for the existence of the string in the error message.)

That's a great idea, but I still can't make heads or tails of the syntax
in these examples.  Were they well formed? 

Also, what needs to get included to turn this on?

> > > 50: EXPECT_THROW_MSG(my_advi_params_fullrank.set_mu(mu_nan);,
> > > 53: Eigen::MatrixXd L_nan = Eigen::MatrixXd::Constant(3,3,nan);
> > > 55: "Cholesky factor is not lower triangular; Cholesky factor[1,2]=nan";


> > > 56: EXPECT_THROW_MSG(stan::variational::advi_params_fullrank
> > my_advi_params_fullrank_nan(mu, L_nan);,
> > > 81: double nan = std::numeric_limits<double>::quiet_NaN();
> > > 82: Eigen::MatrixXd L_nan = Eigen::MatrixXd::Constant(3,3,nan);
> > > 84: "Cholesky factor is not lower triangular; Cholesky factor[1,2]=nan";

...

- Bob

syclik:
On Wed, Apr 15, 2015 at 12:03 AM, Bob Carpenter <notifications@github.com>
wrote:

> > On Apr 15, 2015, at 1:53 PM, Daniel Lee <notifications@github.com>
> wrote:
> >
> > ...
>
> > (Bob, the EXPECT_THROW_MSG is a macro I wrote that has one additional
> > argument, the string to search for, and does what your code does by
> > searching for the existence of the string in the error message.)
>
> That's a great idea, but I still can't make heads or tails of the syntax
> in these examples. Were they well formed?
>

Sorry -- not sure what you mean. But here's what you need to do:

```
#include <test/unit/util.hpp>   // I should really move it

TEST(foo, bar) {
  EXPECT_THROW_MSG(foo(), std::domain_error, "I wanted to find this in the
error message");
}
```



> Also, what needs to get included to turn this on?
>

#include <test/unit/util.hpp>

bob-carpenter:
test/unit/util.hpp is fine with me.  It's easy to find.

Waht I meant was that in the examples below, there's no close
paren and multiple commas, so I couldn't tell what the syntax was.
I always find macros confusing.  

- Bob

akucukelbir:
damn, you guys run a tight ship... i'll look into cpplint stuff now.

bob-carpenter:
Nah, we have a ship that's a mess that we're trying to
clean up.  We'd be drummed out of the Royal Navy for
our slovenly lubberliness :-)

Daniel has the CppLint cranked up to 11 --- you should
really coordinate with him on what's absolutely necessary.

- Bob

syclik:
Alp already figured it out!
Action is now on us to review.

bob-carpenter:
Great.  I was hoping you'd be able to get it to
the point where it wouldn't be producing false
positives.  

What was the conclusion about moving it to some
kind of "experimental" path?  I'm OK with just
leaving it where it makes most sense and then just
putting warning in the interfaces that we haven't
tested it extensively on a range of problems.  I'd
also be OK with not even doing that.  We can use our
users as testers like every other open-source project.

- Bob

akucukelbir:
splendid! i sacrificed a lamb yesterday and
it seems to have worked. (a causal link, surely.)

i'm happy to adapt to whatever you think is
best. i haven't written up any of the documentation
yet, but i'm faster at maths/latex than i am at c++. 

happy to discuss this at any point (skype or google)

src/stan/variational/advi_params_fullrank.hpp
```diff
@@ -0,0 +1,129 @@
...
+      //   natural_params << (Sigma_LDLT.solve(mu_)).array(),
+      //                     (Sigma_LDLT.solve(
+      //                       Eigen::MatrixXd::Identity(dimension_,dimension_))
+      //                     ).array();
+
+      //   return natural_params;
+      // }
+
+      // Implements f^{-1}(\check{z}) = L\check{z} + \mu
+      Eigen::VectorXd to_unconstrained(const Eigen::VectorXd& z_check) const {
```
rtrangucci: @akucukelbir I think the function name 'to_unconstrained' doesn't comport with Stan's definition of unconstrained. unconstrained is about the support of the parameters, and it looks like this function just pushes MVN(0, I) to MVN(mu, Cov).

src/stan/variational/advi.hpp
```diff
@@ -0,0 +1,665 @@
...
+
+        // Initialize ELBO and convergence tracking variables
+        double elbo(0.0);
+        double elbo_prev = std::numeric_limits<double>::min();
+        double delta_elbo = std::numeric_limits<double>::max();
+        double delta_elbo_ave = std::numeric_limits<double>::max();
+        double delta_elbo_med = std::numeric_limits<double>::max();
+
+        // Heuristic to estimate how far to look back in rolling window
+        int cb_size = (int)
```
rtrangucci: @akucukelbir Change c-style cast to c++-style cast

src/stan/variational/advi.hpp
```diff
@@ -0,0 +1,665 @@
...
+        Eigen::MatrixXd L  = Eigen::MatrixXd::Identity(model_.num_params_r(),
+                                                       model_.num_params_r());
+        advi_params_fullrank muL = advi_params_fullrank(mu, L);
+
+        // Robbins Monro ADAgrad
+        do_robbins_monro_adagrad(muL, tol_rel_obj, max_iterations);
+
+        cont_params_ = muL.mu();
+
+        // *print_stream_
```
rtrangucci: @akucukelbir Do you need these comments here?

src/stan/variational/advi_params_fullrank.hpp
```diff
@@ -0,0 +1,129 @@
...
+        stan::math::check_size_match(function,
+                               "Dimension of mean vector",     dimension_,
+                               "Dimension of Cholesky factor", L_chol_.rows() );
+        for (int i = 0; i < dimension_; ++i)
+          stan::math::check_not_nan(function, "Mean vector", mu_(i));
+        stan::math::check_lower_triangular(function,
+                               "Cholesky factor", L_chol_);
+      }
+
+      virtual ~advi_params_fullrank() {}  // No-op
```
rtrangucci: @akucukelbir Do you need this no-op constructor?

src/stan/variational/advi_params_meanfield.hpp
```diff
@@ -0,0 +1,114 @@
...
+      // Implement f^{-1}(\check{z}) = sigma * \check{z} + \mu
+      Eigen::VectorXd to_unconstrained(const Eigen::VectorXd& z_check) const {
+        static const char* function = "stan::variational::advi_params_meanfield"
+                                      "::to_unconstrained";
+
+        stan::math::check_size_match(function,
+                         "Dimension of mean vector", dimension_,
+                         "Dimension of input vector", z_check.size() );
+        for (int i = 0; i < dimension_; ++i)
+          stan::math::check_not_nan(function, "Input vector", z_check(i));
```
syclik: I believe `check_not_nan` was written enough generality to not require this loop. So you should be able to write:
```
stan::math::check_not_nan(function, "Input vector", z_check);
```

src/stan/variational/advi.hpp
```diff
@@ -0,0 +1,665 @@
...
+       * Calculates the "blackbox" gradient with respect to BOTH the location
+       * vector (mu) and the cholesky factor of the scale matrix (L) in
+       * parallel. It uses the same gradient computed from a set of Monte Carlo
+       * samples
+       *
+       * @param muL     mean and cholesky factor of affine transform
+       * @param mu_grad gradient of location vector parameter
+       * @param L_grad  gradient of scale matrix parameter
+       */
+      void calc_combined_grad(const advi_params_fullrank& muL,
```
rtrangucci: @akucukelbir Should be a member function of advi_paramas_fullrank, doesn't make sense as a standalone.

src/stan/variational/advi.hpp
```diff
@@ -0,0 +1,665 @@
...
+       * Calculates the "blackbox" gradient with respect to BOTH the location
+       * vector (mu) and the variance vector (sigma^2) in parallel.
+       * It uses the same gradient computed from a set of Monte Carlo
+       * samples
+       *
+       * @param musigmatilde      mean and log-std vector of affine transform
+       * @param mu_grad           gradient of mean vector parameter
+       * @param sigma_tilde_grad  gradient of log-std vector parameter
+       */
+      void calc_combined_grad(const advi_params_meanfield& musigmatilde,
```
rtrangucci: @akucukelbir Same comment as for full-rank calc_combined_grad

src/stan/variational/advi.hpp
```diff
@@ -0,0 +1,665 @@
...
+        // writer.write("mu", muL.mu());
+        // s << "\n";
+        // writer.write("L_chol", muL.L_chol());
+        // std::string written = s.str();
+        // *print_stream_ << written << std::endl;
+
+        return;
+      }
+
+      void run_meanfield(double tol_rel_obj, int max_iterations) {
```
rtrangucci: @akucukelbir Do we need these functions? It doesn't look like this or run_fullrank do very much aside from calling two child functions.

src/stan/variational/advi.hpp
```diff
@@ -0,0 +1,665 @@
...
+          if (iter_counter == max_iterations) {
+            *print_stream_ << "MAX ITERATIONS REACHED" << std::endl;
+            do_more_iterations = false;
+          }
+
+          ++iter_counter;
+        }
+      }
+
+      void run_fullrank(double tol_rel_obj, int max_iterations) {
```
rtrangucci: @akucukelbir Do we need this function? Doesn't seem to do much aside from calling two other functions.

src/stan/variational/advi.hpp
```diff
@@ -0,0 +1,665 @@
...
+            eta_stepsize_ * mu_grad.array()
+            / (tau + mu_s.array().sqrt()));
+          musigmatilde.set_sigma_tilde(
+            musigmatilde.sigma_tilde().array()  +
+            eta_stepsize_ * sigma_tilde_grad.array()
+            / (tau + sigma_tilde_s.array().sqrt()));
+
+          // Check for convergence every "refresh_"th iteration
+          if (iter_counter % refresh_ == 0) {
+            elbo_prev = elbo;
```
rtrangucci: @akucukelbir Given heuristics of convergence, it might make sense to put the following 5-10 lines into a function that gets called internally to the fullrank and meanfield run_robbins_munro_adagrad. That way if we decide to change the heuristics, we can do it once easily.

src/stan/variational/advi.hpp
bob-carpenter: You should assume people reading the code know C++. 

In particular, you can assume they know mu = and L = ...
are assignments.

And you named the function "do_robbins_monro_adagrad", so you
don't need a comment that repeats the function name.

(In general, I'm opposed to function names that have "do" in
the title just to make them verbs, but I can live with it.)

Also, I'm not seeing the rest of the code, but is it really
necessary to create an identity matrix?  That's a quadratic
operation in the number of params and identity matrices aren't
usually good for much.

- Bob

src/stan/variational/advi_params_fullrank.hpp
bob-carpenter: It's a no-op destructor.  And you don't need
a comment that says "no-op" because it's clear from
the function body being empty.

Also, does it need to be virtual?  Are there extensions
planned for this class?  If not, don't make it virtual.
At which point, it'll be defined implicitly and you don't
need it at all.

- Bob

src/stan/variational/advi.hpp
bob-carpenter: Tastes differ on whether to prefer standalone functions
or members.  I prefer members, but standalone's not so bad.

But that comment "FULL-RANK GRADIENTS" is bad --- the first
sentence should succinctly describe what the function takes
as inputs and what it returns. 

- Bob

src/stan/variational/advi.hpp
bob-carpenter: Also, there shouldn't be any commented-out code left
around.  It's just confusing to people reading the code.

- Bob

src/stan/variational/advi.hpp
bob-carpenter: Is the CppLint thing not griping about this?  It doesn't match
the recommended style, which only has you start new lines in case
things won't fit in 80 characters of width.  

Also, start new expressions with the operator, so it's

   a
   + b

not

   a + 
   b

It's just like standard math typesetting style.

And again, you can assume readers will know that a % b == 0
means do something every b-th version of a.

Feel free to pull functionality like this out into inline
functions with reasonable names and drop comments.

- Bob

src/stan/variational/advi.hpp
```diff
@@ -0,0 +1,665 @@
...
+          // Draw from standard normal and transform to unconstrained space
+          for (int d = 0; d < dim; ++d) {
+            z_check(d) = stan::prob::normal_rng(0, 1, rng_);
+          }
+          z_tilde = muL.to_unconstrained(z_check);
+
+          // Compute gradient step in unconstrained space
+          stan::model::gradient(model_, z_tilde, tmp_lp, tmp_mu_grad,
+                                print_stream_);
+          stan::agrad::recover_memory();
```
rtrangucci: @bob-carpenter do we need this? model::gradient calls agrad::gradient which calls start_memory_nested() and recover_memory_nested().

akucukelbir: i put this in after a bug-hunt with daniel. without it, 
there is a memory leak that grows across iterations.

hope this helps? and thanks for all the cleanup work!

rtrangucci: @akucukelbir Sounds good, I left it in there because I wasn't sure. That might be worth investigating more deeply because I'm not sure why the thing would be leaking.

bob-carpenter: Nope --- the gradient() call will recover memory itself,
so you shouldn't have that second call (it shouldn't cause
problems, but it's confusing).

Why is a standard normal being translated to unconstrained
support?  I'd think it'd already be set up to be unconstrained.


- Bob

bob-carpenter: > On May 1, 2015, at 8:41 AM, Alp Kucukelbir <notifications@github.com> wrote:
> 
> In src/stan/variational/advi.hpp:
> 
> > +        Eigen::VectorXd tmp_mu_grad = Eigen::VectorXd::Zero(dim);
> > +        Eigen::VectorXd z_check = Eigen::VectorXd::Zero(dim);
> > +        Eigen::VectorXd z_tilde = Eigen::VectorXd::Zero(dim);
> > +
> > +        // Naive Monte Carlo integration
> > +        for (int i = 0; i < n_monte_carlo_grad_; ++i) {
> > +          // Draw from standard normal and transform to unconstrained space
> > +          for (int d = 0; d < dim; ++d) {
> > +            z_check(d) = stan::prob::normal_rng(0, 1, rng_);
> > +          }
> > +          z_tilde = muL.to_unconstrained(z_check);
> > +
> > +          // Compute gradient step in unconstrained space
> > +          stan::model::gradient(model_, z_tilde, tmp_lp, tmp_mu_grad,
> > +                                print_stream_);
> > +          stan::agrad::recover_memory();
> 
> i put this in after a bug-hunt with daniel. without it, there is a memory leak that grows across iterations. hope this helps?

Could you check that it still leaks without recover_memory()?  If so,
it's a deeper bug.  You should *not* need to call recover_memory() after
calling this gradient() function.

- Bob

P.S.  stan/model/util.hpp   for stan::model::gradient wrapper of functional

      stan/math/rev/mat/functor/gradient.hpp   for functional definition

akucukelbir: @syclik i can't remember the explanation you gave me when we added this recovery_memory business in there. has something changed between now and then? can we remove it? (i can't test this myself until later this week.)

rtrangucci: @akucukelbir Nothing's changed, but the problem is that we shouldn't need this recover_memory call in the code, so there is something fishy if you're getting a memory leak.

akucukelbir:
ahoy, i'm back now. @rtrangucci could we skype sometime today to bring me back up to speed?

rtrangucci:
@akucukelbir @bob-carpenter @syclik removing the call to recover_memory() didn't cause any memory leaks so I left them out.

akucukelbir:
can you try for like 10k iterations? i'll remove it on my end and give it a shot too. it could be because of the ELBO computation... i used to have to do this trick where i would compute the normalizing constant using the vars version but then compute the ELBO using the double version. but then it turns out i didn't need to do that and could just use the double version. so perhaps that was the issue? (still doesn't make sense to me though.)

i'll confirm here whether removing recover_memory() works on my end

akucukelbir:
@rtrangucci @bob-carpenter @syclik ok, it works fine on my end too. so... huzzah?

rtrangucci:
@akucukelbir That's great! FYI, the tests don't compile anymore because of the constructor change to advi. Want me to update the tests? Also, what're some recommended settings for refresh, and the number of ELBO / gradient draws?

akucukelbir:
@rtrangucci ah damn... no worries. i'll update them and restart the tests. jenkins can then test overnight. "refresh" is now "eval_elbo" and 100 seems ok. the defaults for number of elbo (100) and number of gradient (10) draws also seem to work well with all of the models i've tried up to now.

syclik:
I had it skip the distribution tests because we know this doesn't affect it. There are still some cpplint warnings to fix -- I can help out with that.

@rtrangucci, is it pretty much good to go?

syclik:
that's my bad -- I'm messing with the Windows machine right now; I'm trying
to get CVODE working, but it's not going so well. I'll get this back on the
queue.

syclik:
Jenkins, retest this please.

rtrangucci:
@syclik I want to take a look at the code that Alp added yesterday to sample from the multivariate normal approximation to the unconstrained posterior and push those samples back to the constrained space. As long as that looks good, I think we (you / Bob) can merge it later today.

akucukelbir:
@rtrangucci sounds great. all the changes are in `advi.hpp`. i added a function called `draw_posterior_sample( ... )` and modified `run_fullrank(...)` and `run_meanfield(...)`

bob-carpenter:
I think there are too many functions in this:

src/stan/variational/advi.hpp

It would be nice if we could extract the transform bits of
draw_posterior_sample() thing into a more general library.
It's probably a job for the next person, though.  

It'd be tough given the way the code's currently written to 
disentangle the blackboard architecture used for everything in 
advi, which encapsulate a global state for the command
execution. 

- Bob

bob-carpenter:
I'm looking at the code now.  It's super clean for what it's doing.

What's with using this:

+      template <typename T>
+      Eigen::VectorXd draw_posterior_sample(const T& advi_params) {

pattern everywhere?  Is there no base class you can use to give
us poor readers a hint as to what can instantiate T?

There needs to be doc at least as to what the T is supposed to
be instantiated to.  The doc comment is 

  @tparam T ...what can T be ...

And that's not fixed in the advi instantiation?

I'd also very much like to see everything going into our code base now
have doc.  As in this mondo-constructor:


 template <class M, class BaseRNG>
+    class advi {
+    public:
+      advi(M& m,
+           Eigen::VectorXd& cont_params,
+           int n_monte_carlo_grad,
+           int n_monte_carlo_elbo,
+           double eta_stepsize,
+           BaseRNG& rng,
+           int eval_elbo,
+           int n_posterior_samples,
+           std::ostream* print_stream,
+           std::ostream* output_stream,
+           std::ostream* diagnostic_stream):


It's not clear from looking at this what all the variables
should be, at least to me.  And if eval_elbo is boolean, it
should be declared as bool.

You don't need to static-cast here:

+        elbo /= static_cast<double>(n_monte_carlo_elbo_);

C++ has built-in promotion so the result of double OP int is double.

- Bob

akucukelbir:
@bob-carpenter i made the changes you recommended. i don't think we need to rerun all the tests, since i changed so little. (but i'll leave that up to you.)

rtrangucci:
Took a look at the code and the output csv and there aren't any column names. Is that something that  advi needs to handle?

Rob

bob-carpenter:
Yes --- there needs to be column names in the CSV output.
That can be done the same way it's done for sampling or for
optimization.

- Bob

bob-carpenter:
We always rerun all the tests.  It's better to be safe than
sorry (i.e., breaking the builds for others).

- Bob

akucukelbir:
@bob-carpenter @rtrangucci what do you mean no column names? i see the column names. e.g. lp,theta.1,theta.2,mu.1.1,mu.2.1,mu.1.2,mu.2.2,sigma.1.1.1,sigma.2.1.1,sigma.1.2.1,sigma.2.2.1,sigma.1.1.2,sigma.2.1.2,sigma.1.2.2,sigma.2.2.2. if you use the cmdStan print function on the output, everything prints out nicely on my end.

akucukelbir:
@bob-carpenter @rtrangucci if it helps, you can see the column names written out in `stan/services/command.hpp` on lines `714` and `738`.

rtrangucci:
@akucukelbir Sorry Alp, I see it now. @bob-carpenter @syclik All is OK on this branch from my standpoint.

akucukelbir:
@rtrangucci no worries. i was just confused; i thought i had to place the column names elsewhere too!

HerraHuu:
Great to see VB being added to Stan. I was thinking about doing it myself, but I guess now there is no need for that.

I have a couple of questions about the implementation. So the goal of VB inference is to optimize evidence lower bound (ELBO). This includes basically two parts:

- stochastic optimization
 - current implementation: stochastic gradient ascent with ADAgrad/RMSprop

- gradient estimation
 - current implementation: mcmc integration

Would it make sense to make some general code for both of these objectives, and not just VB spesific? I think that would be useful to other methods (for example EP) and if also VB would be written in more general way, it would be possible to try different optimization algorithms etc (for example stochastic BFGS, http://arxiv.org/pdf/1401.7020). Does this make sense? I guess/hope it wouldn't be too difficult thing to do, huge amount of work has already been done.

bob-carpenter:
I agree that we want to pull a bunch of the pieces that were
hardwired into VB-specific classes into more general 
implementations.  

The question is whether to insist on doing that before the
existing VB branch gets merged or after.  I think it makes sense 
to wait on extracting reusable pieces until we have an intended 
reuse that can guide how we do it.  We'll need enough tests in place
that we're confident we're not breaking VB when we do it, too.

- Bob

syclik:
A few things to do:
- [ ] @rtrangucci or @akucukelbir  remove `std::cout` from tests.
- [ ] @syclik: verify nothing is being streamed out through `std::cout`. (use the utilities you built)
- [ ] add a `SUCCESS() << "<some message here>"` to the end of the tests to explain why there are asserts in the test.

syclik:
@akucukelbir, did you write any doc for the Stan manual?

akucukelbir:
@syclik i wrote doc for the cmdStan manual. i never pushed that branch to github. i need to bring it up to speed, with the `experimental` keyword and such. 

i can write a draft for the Stan manual. as bob mentioned, this is likely going to be an iterative process.

do these need to coincide this particular pull request?

syclik:
Yes. Or else it won't get done despite all good intentions.

akucukelbir:
roger that. then could @syclik or @rtrangucci help out with the tests (whenever you can)? i'll work on the manuals.

syclik:
Yes.

syclik:
Some more stuff is coming up.
- [ ] change `run_*` methods so they return either success or failure
- [ ] print streams are definitely not safe -- no checking if they're 0

syclik:
@akucukelbir, I just added a test that's going to fail: the code isn't safe for streams that are NULL. I'll continue to fix the existing tests. Focus on the doc for now.

akucukelbir:
@syclik no worries. thank you for catching this. i don't need to touch the code at the moment. i'll just work on the manuals.

dustinvtran:
I mentioned this to @akucukelbir, but just in case: cmdstan upgraded to Boost 1.58 in sync with stan dev branch's update. This branch needs the upgrade to Boost 1.58 too or else it fails to compile.

src/stan/variational/advi.hpp
```diff
@@ -0,0 +1,753 @@
+#ifndef STAN_VARIATIONAL_ADVI_HPP
+#define STAN_VARIATIONAL_ADVI_HPP
+
+#include <stan/math/prim.hpp>
```
syclik: change to #include <stan/math.hpp>

src/stan/variational/advi.hpp
```diff
@@ -0,0 +1,753 @@
...
+#include <stan/math/prim/scal/err/check_positive.hpp>
```
syclik: remove

src/stan/variational/advi_params_fullrank.hpp
```diff
@@ -0,0 +1,112 @@
+#ifndef STAN_VARIATIONAL_ADVI_PARAMS_FULLRANK__HPP
+#define STAN_VARIATIONAL_ADVI_PARAMS_FULLRANK__HPP
+
+#include <stan/math/prim/mat/fun/Eigen.hpp>
```
syclik: change to #include <stan/math.hpp>
remove all other #include <stan/math/...>

akucukelbir: i'm sorry, i didn't exactly follow what to do dan.

syclik: Those were sort of notes to self.

The math library is closer to being stand alone. A lot of the folder level
includes have been removed.

akucukelbir: oh i see.

i think i also need to make some updates to the variational tests. i'll push those soon too.

akucukelbir:
ok so i push the change, but `./runTests.py` is no longer working on my end.

akucukelbir:
ok documentation is done. Stan manual edits are on this branch. edits for the cmdStan manual are at https://github.com/stan-dev/cmdstan/tree/feature/advi-documentation

syclik:
@akucukelbir, @dustinvtran, this pull request is going to get shaken up a bit with stan-dev/stan#1502. `command.hpp` is moving to CmdStan. It won't be too much work to transfer over, I think.

dustinvtran:
Yup, just looked into the merge conflicts—four files which have minor changes. I can work on it once those two outstanding requests (cmdstan and stan respectively) are merged into develop.

akucukelbir:
@dustinvtran thanks dustin.

@syclik @dustinvtran so is the plan to merge this branch into develop, merge stan-dev/stan#1502 into develop, and merge ??? into cmdstan, and then fix the interface?

akucukelbir:
@dustinvtran cpplint is failing. no problem. i will take care of this before i look at https://github.com/stan-dev/stan/pull/1505

src/stan/variational/advi.hpp
```diff
@@ -0,0 +1,368 @@
...
+        }
+
+        // draw more samples from posterior and write on subsequent lines
+        if (out_stream_) {
+          for (int n = 0; n < n_posterior_samples_; ++n) {
+            cont_params_ = variational.sample(rng_);
+            for (int i = 0; i < cont_params_.size(); ++i)
+              cont_vector.at(i) = cont_params_(i);
+            services::io::write_iteration(*out_stream_, model_, rng_,
+                          0.0, cont_vector, disc_vector, &std::cout);
```
syclik: should be `print_stream_` instead of `&std::cout`, right? we've removed uses of `std::cout` from code.

akucukelbir:
thanks @syclik !

brendan-R:
💃

PR #1440, opened 2015-05-05
changing src/stan/math.hpp to the uber-include for the Stan Math Library

### LINKED ISSUE #811: include what you use in models
Alexey Stukalov reports that on a distributed filesystem, Stan models take 2 minutes to build, whereas on a local system, the same model takes 13 seconds.

To cut down these builds, our best strategy is probably to figure out how to include only what we need rather than the entire model header include.

To make this work, we won't be able to include all of Eigen or large swaths of Boost, either, because the number of Stan files is minimal compared to those libs.

If a function uses a matrix argument, it needs the relevant matrix libs.  If it doesn't, then you don't need to bring those in.  But all the matrix lib gets imported for densities anyway.  And we need most of the matrix libs now for running Stan because MCMC and optimization moved to Eigen for parameter vectors.

syclik:
#### Summary:

Changes the function of `src/stan/math.hpp` to be the uber-include for the Stan Math Library.

#### Intended Effect:

To use the Stan Math Library, will only need to include `src/stan/math.hpp`. This includes all of reverse mode.

#### How to Verify:

Run the test -- that was the thing I thought didn't compile due to include order, but it seems to work under develop now.

#### Side Effects:

`src/stan/math.hpp` is different than all other folder level header files. Will need to fix #811.

#### Documentation:

In the wiki. Will update once this is merged.

#### Reviewer Suggestions: 

@bob-carpenter


syclik:
Jenkins, retest this.

syclik:
This was all failing due to a weird Windows path issue after restart of the machine. Hopefully it's all fixed now.

syclik:
Jenkins, retest this please.

syclik:
After the merging of the multiple namespaces gets into the code base:

- [ ] DL: add tests using all examples in the stan paper.

syclik:
Bob, I'm trying to go through your examples in the paper and the first one fails to compile:
```
#include <cmath>
#include <boost/math/constants/constants.hpp>
#include <stan/math.hpp>

int main() {
  using std::pow;
  double y = 1.3;
  stan::math::var mu = 0.5, sigma = 1.2;
    
  stan::math::var lp = 0;
  lp -= pow(2 * boost::math::constants::pi<double>(), -0.5);
  lp -= log(sigma);
  lp -= 0.5 * pow((y - mu) / sigma, 2);
  std::cout << "f(mu,sigma)= " << lp.val() << std::endl;

  lp.grad();
  std::cout << " d.f / d.mu = " << mu.adj()
            << " d.f / d.sigma = " << sigma.adj() << std::endl;
}
```

The `stan::math::var` class does not have a `.grad()` with no arguments, so it can't compile this part: `lp.grad()`. Options:

1. we create the `grad()` function that has the right behavior.
2. we replace `lp.grad()` with `stan::math::grad(lp.vi_)`. I think this is the weakest option
3. we create a new `stan::math::grad(var&)` function, then replace `lp.grad()` with `stan::math::grad(lp)`

Thoughts?

syclik:
bump. @bob-carpenter, want to take a look?

bob-carpenter:
Looks good --- do you want to merge it so I don't mess it up?

syclik:
Ok. Will do.

PR #1444, opened 2015-05-07
Bugfix/issue 1443 zero size array sum

bob-carpenter:
#### Summary:

Fix issue with sum() segfaulting when given array of size zero.

Improve efficiency of fwd-mode sum by calling sum() for value and tangent calculations.

#### Intended Effect:

Fix bug and speed up.

#### How to Verify:

Many more unit tests added, including tests for gradients.

#### Side Effects:

Includes get a little more complicated as there are now even more sum() implementations flying around (six files named sum.hpp, I believe)

#### Documentation:

Added in code.  No user-facing doc necessary.

#### Reviewer Suggestions:

Anyone.


syclik:
Looks good. We'll just wait for tests to pass.

PR #1460, opened 2015-05-22
Feature/issue 1457 boost 1.58

bob-carpenter:
#### Summary:

Update Stan to use Boost 1.58.

#### Intended Effect:

Allow use of RStan with the latest release of R package BH (Boost headers), which is Boost 1.58.

#### How to Verify:

Run tests. 

Note the updated Wishart RNG tests that now include a seed.  They weren't passing by a small numerical margin, but look to be OK.

#### Side Effects: 

Will no longer run with Boost versions lower than 1.56

#### Documentation:

Nothing user facing. (the license doc in the language manual no longer mentions versions)

#### Reviewer Suggestions:

Anyone.


bob-carpenter:
Hint:  other than Wishart tests, the changes are all in `lang/grammars/*_def.hpp` and are all the same

bgoodri:
This is a FAIL for me too on Linux

src/test/unit/io/dump_test.cpp:73: Failure

bob-carpenter:
I'm trying to get a patch for Boost 1.58 ASAP so that
we can upgrade RStan so that everyone's installs will start
working again.  This is my only priority at the moment and
I've spent the better part of three days on this already.

The last commit but one worked on Mac.  The most recent
commit breaks everything and needs to be backed out --- I
inadvertently brought in a C++11 dependency for expm1.  

Windows is still broken.

Linux is uncertain.

So far, I've fixed the semantic actions in the parser due
to changes in Boost's lazy evaluation, fixed the dump parser 
due to changes in Boost's underflow exception handling just on
Windows, and now need to figure out why Boost's expm1() is
failing on Windows --- I think it's boundary conditions
again on overflow/underflow, which we're carefully testing
and hence failing on when things change.

- Bob

bgoodri:
The expm1 error is happening on Linux when C++11=false. If true, it seems
to be okay.

syclik:
Ben, how do I turn off C++11 in clang++? I've searched for the flag for the
last hour and I can't find it.

bgoodri:
Is C++11 the default for clang++ on a Mac these days? If so, you would
probably have to put -std=g++99 or something into make/os_mac.

syclik:
I can't get the -std flag to prevent C++11 features. I've tried all the
standards I can find, which is really tough to do, btw:
http://stackoverflow.com/questions/21870283/how-to-get-list-of-supported-standards-in-clang

syclik:
Do you know of any other flags that will force clang++ to fail to compile
if C++11 is used?

betanalpha:
Which feature are you trying to remove?  Remember that there
is the standard (-std) and then the standard library (-stdlib).
Even if you force, say, -std=c++03, you might still call the
implementations in the newer standard library.

syclik:
I'm trying to compile with strict C++98 compliance. Windows doesn't have
std::expm1, that's been introduced as of C++11. It's a lot easier testing
locally than it is getting to the Windows machine and back for compliance.

betanalpha:
Have you tried forcing the older standard library implementation, -stdlib=libstdc++?

syclik:
there's no -stdlib option specified in the compilation

betanalpha:
Sure, which should then resort to the default which will be -stdlib=libc++
on new versions of clang.  Explicitly setting -stdlib=libstdc++ should force
the compiler to link to the older (pre C++11) implementation of the
standard library that doesn’t include the C++11 functions.

syclik:
Thanks, Bob. I walked through the changes and they look good. We'll merge once it passes tests.

syclik:
This is fine. I fixed the performance test.

Issue #1436, opened 2015-05-01
Stan creates code that does not compile when max() used in parameter dimension


ksvanhorn:
When I run rstan::stan_model on a file containing the following model code, the generated C++ code fails to compile. If I replace max(1, K0_psevere) and max(1, Kb_psevere) with K0_psevere + 1 and Kb_psevere + 1, the problem goes away.

    data {
      int<lower=0, upper=1> use_psevere;

      int<lower=1> K;
      int<lower=2> B;
      int<lower=1, upper=K> K0_lambda;
      int<lower=1*use_psevere, upper=K*use_psevere> K0_psevere;
      int<lower=0, upper=K> Kb_lambda;
      int<lower=0, upper=K*use_psevere> Kb_psevere;
    }
    transformed data {
    }
    parameters {
      vector[K0_lambda] beta0_lambda;
      vector[max(1,K0_psevere)] beta0_psevere;

      matrix[Kb_lambda, B] pre_betab_lambda;
      matrix[max(1,Kb_psevere), B] pre_betab_psevere;
    }
    transformed parameters {
    }
    model {
      beta0_lambda ~ normal(0,1);
      if (use_psevere)
        beta0_psevere ~ normal(0,1);

      to_vector(pre_betab_lambda) ~ normal(0, 1);
      if (use_psevere)
        to_vector(pre_betab_psevere) ~ normal(0, 1);
    }

bob-carpenter:
Confirmed.  In clang, here's the error message:

```
In file included from <built-in>:326:
In file included from <command line>:5:
In file included from /Users/carp/temp2/comp.hpp:3:
In file included from stan/src/stan/model/model_header.hpp:5:
In file included from stan/src/stan/math/prim/mat/meta.hpp:4:
In file included from stan/src/stan/math/prim/mat/meta/get.hpp:4:
In file included from stan/src/stan/math/prim/mat/fun/Eigen.hpp:4:
In file included from stan/lib/eigen_3.2.4/Eigen/Dense:1:
In file included from stan/lib/eigen_3.2.4/Eigen/Core:293:
stan/lib/eigen_3.2.4/Eigen/src/Core/PlainObjectBase.h:629:27: error: no member named
      'FLOATING_POINT_ARGUMENT_PASSED__INTEGER_WAS_EXPECTED' in
      'Eigen::internal::static_assertion<false>'
                          FLOATING_POINT_ARGUMENT_PASSED__INTEGER_WAS_EXPECTED)
[...]

PR #1468, opened 2015-06-01
Bugfix/issue 1436 int max bounds

bob-carpenter:
#### Summary:

* Adds `using std::max;` and `using std::min;` to generated model code.  

* Removed includes of C-style `<math.h>` (namespace `::`) in favor of C++ `<cmath>` (namespace `std::`) where possible.  

* Removed lint in math library code.

#### Intended Effect:

Fix issue that prevented `max(int,int)` being used in sizes for array and matrix variable declarations.

#### How to Verify:

There is a new model in the test suite, `src/test/test-models/good/min-max-types.stan`, which fails in the current develop version and Stan 2.6.2, but now works.

#### Side Effects:

None.

#### Documentation:

This brings behavior in line with doc.

#### Reviewer Suggestions:

Anyone.


syclik:
Thanks, Bob. This looks great.

syclik:
Fixes #1436.

PR #1470, opened 2015-06-01
Bugfix/issue 1435 size 0 vector params

bob-carpenter:
#### Summary:

Test for size 0 containers in `io/reader.hpp` and return empty containers without underlying access.

#### Intended Effect:

Allow size 0 containers to be the last parameters declared in the parameters block.

#### How to Verify:

There's a new unit test, `zeroSizeVecs`, in `test/unit/io/reader_test.cpp` that fails in 2.6.3 but passes now.

#### Side Effects:

None.

#### Documentation:

This brings behavior in line with docs.

#### Reviewer Suggestions:

Anyone.


syclik:
Thanks, Bob.

PR #1480, opened 2015-06-04
Feature/issue 1465 math lib: moving the Stan Math Library to the lib folder

syclik:
#### Summary:

Preps the Stan Math Library for release.
- moved the existing Stan math library to `lib/stan_math_2.7.0/`
- the git history is preserved. Merging the existing pull requests should be easy.
- I had to copy a lot of the makefiles (and trimmed where appropriate) to the new location of the Stan Math library for the tests
- removed all folder level header files
- updated the travis yml files so that all the unit tests still run, including the math unit tests.

#### Intended Effect:

Moves the Stan Math library closer to release. We'll move it to its own repository once the existing pull requests are updated.

One thing that I didn't take care of was removing `src/stan/math/rev/core.hpp` and `src/stan/math/fwd/core.hpp`. These two headers are implicated in many files and tests and couldn't be removed easily.

#### How to Verify:

1. Inspect the code. The `./src/stan/math` folder has moved to `lib/stan_math_2.7.0/src/stan/math`. The tests have also moved -- both unit tests and prob tests.
2. Run the tests from the main Stan folder.
3. Run the math tests from the `lib/stan_math_2.7.0` folder. `> ./runTests.py src/test/unit/math` and `> ./runTests.py src/test/prob`

#### Side Effects:

Everything works as it used to work.

#### Documentation:

I updated some of the doxygen. We'll might consider documenting the Stan Math Library as part of another pull request.

#### Reviewer Suggestions: 

Bob.


syclik:
Jenkins has not been updated for this -- can someone review this pull request? If it's good, I'll update the Jenkins jobs appropriately so all the jobs run.

bob-carpenter:
Are you thinking that in 2.7 we'll have a standalone math lib?

Will this need to be coordinated with the R StanHeaders package?

- Bob

syclik:
With this pull request, the math library will be in the lib folder and it
can start being treated as a separate library. I don't think we will be
able to separate out the repos in time for a 2.7.0 release. We really want
to wait for the outstanding pull requests before splitting it out just to
avoid duplication of work.

Yes, it will need to be coordinated with the R StanHeaders package. This
shouldn't be a problem. (Sorry, I forgot about it.) That's a quick fix.

bob-carpenter:
This all looks great. 

Is there any way we can change the path of Stan in the math lib to parallel Boost and Eigen, so that it looks as follows?

```
stan/lib/stan_math_2.7.0/stan
```

I see that what you've done is put everything you think is going along with the math library to its new repo.  When we use Boost and Eigen, we cut out just the header files we need and don't include anything else.  I'm hoping we can do that with Stan Math, too, because it's header only.

syclik:
Yes -- I should have thought about that. We can make that happen. I'll adjust the makefiles accordingly.

syclik:
I've updated the math library's location of source and test files, the makefiles, and everything else. Bob, want to take another look? 

The tests will fail on Jenkins for now. If it looks ok, I'll update Jenkins and run through the tests again.

bob-carpenter:
Looks good to me.

syclik:
Jenkins, retest this please.

syclik:
Jenkins, retest this please.

PR #1482, opened 2015-06-04
Bugfix/issue 1198 missing fvar derivs

rtrangucci:
#### Summary:

Fixed tangent calculations for `fvar<T> operator*=()` and `fvar<T> operator/=()` 
`fvar<T> operator*()` and `fvar<T> operator/()` are less general

#### Intended Effect:

Operators return correct derivatives.

#### How to Verify:

./runTests.py src/test/unit/math/fwd/core/operator_multiply_test.cpp
./runTests.py src/test/unit/math/mix/core/operator_multiply_test.cpp
./runTests.py src/test/unit/math/fwd/core/operator_divide_test.cpp
./runTests.py src/test/unit/math/mix/core/operator_divide_test.cpp
./runTests.py src/test/unit/math/fwd/core/operator_multiply_equal_test.cpp
./runTests.py src/test/unit/math/fwd/core/operator_divide_equal_test.cpp
./runTests.py src/test/unit/math/mix/core/operator_multiply_equal_test.cpp
./runTests.py src/test/unit/math/mix/core/operator_divide_equal_test.cpp

#### Side Effects:

None.  

#### Documentation:

None.

#### Reviewer Suggestions: 

Daniel or Bob.


syclik:
Jenkins, retest this please.

syclik:
Jenkins, retest this please.

bob-carpenter:
This is the most exciting pull request in our queue.  I'd 
be overjoyed to see this in the first packaged version of 
the math library.  

Then we have to write the fvar version of the autodiff
paper, which is all sorts of fun on the functional front,
has a spiffy generalization of OperandsAndPartials and 
the distribution code, and a finite diff testing framework
that matches the functionals.  Maybe we can sneak in a general
framework for vectorization.  It's going to be another 100 pager.
Maybe Rob could write it in between the three or four other 
projects he's juggling.

bob-carpenter:
I should've read more closely --- this is just patching bugs in fvar derivatives.  What I'm excited about is getting the three cases Rob identified fixed so that higher-order autodiff will compile.

bob-carpenter:
Looks good to me.  We probably want to follow the Google code guide and instead of writing

```
inline
foo
bar(baz x) {
  ...
```

to

```
inline foo bar(baz x) { 
  ...
```

but it's not necessary for this pull request.

PR #1483, opened 2015-06-04
Bugfix/issue 1474 dirichlet rng tiny

bob-carpenter:
#### Summary:

Add Marsaglia and Tsang PRNG for Dirichlet with alpha < 1

#### Intended Effect:

Avoid returning NaN values from Dirichlet RNG due to underflow of all gamma variates.

#### How to Verify:

New unit tests.

#### Side Effects:

None.

#### Documentation:

Already exists.

#### Reviewer Suggestions:

Anyone.

### DIFF
--- a/src/stan/math/prim/mat/prob/dirichlet_rng.hpp
+++ b/src/stan/math/prim/mat/prob/dirichlet_rng.hpp
@@ -27,9 +27,10 @@ namespace stan {
      * For prior counts greater than zero, the usual algorithm that
      * draws gamma variates and normalizes is used.  
      * 
-     * For prior counts less than zero, a log-scale version of the
-     * following algorithm is used to deal with underflow:
-     * 
+     * For prior counts less than zero (i.e., parameters with value
+     * less than one), a log-scale version of the following algorithm
+     * is used to deal with underflow:
+     *
      * <blockquote>
      * G. Marsaglia and W. Tsang. A simple method for generating gamma
      * variables. ACM Transactions on Mathematical Software.
@@ -47,7 +48,6 @@ namespace stan {
       using boost::variate_generator;
       using boost::gamma_distribution;
       using boost::random::uniform_real_distribution;
-      using stan::math::log_sum_exp;
       using Eigen::VectorXd;
       using std::exp;
       using std::log;
--- a/src/stan/math/prim/mat/prob/dirichlet_rng.hpp
+++ b/src/stan/math/prim/mat/prob/dirichlet_rng.hpp
@@ -65,27 +65,21 @@ namespace stan {
         }
         double log_sum_y = log_sum_exp(log_y);
         VectorXd theta(alpha.size());
-        for (int i = 0; i < alpha.size(); ++i) 
+        for (int i = 0; i < alpha.size(); ++i)
           theta(i) = exp(log_y(i) - log_sum_y);
         return theta;
       }
 
       // standard normalized gamma algorithm
-      double sum = 0;
       Eigen::VectorXd y(alpha.rows());
       for (int i = 0; i < alpha.rows(); i++) {
         variate_generator<RNG&, gamma_distribution<> >
           gamma_rng(rng, gamma_distribution<>(alpha(i, 0), 1e-7));
         y(i) = gamma_rng();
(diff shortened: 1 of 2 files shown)


syclik:
Jenkins, retest this please.

syclik:
Jenkins, retest this please.

src/stan/math/prim/mat/prob/dirichlet_rng.hpp
```diff
@@ -3,38 +3,83 @@
...
+     * @param rng Pseudo-random number generator.
+     */
     template <class RNG>
     inline Eigen::VectorXd
     dirichlet_rng(const Eigen::Matrix<double, Eigen::Dynamic, 1>& alpha,
                   RNG& rng) {
       using boost::variate_generator;
       using boost::gamma_distribution;
+      using boost::random::uniform_real_distribution;
+      using stan::math::log_sum_exp;
```
syclik: remove `using stan::math::log_sum_exp` statement.

src/stan/math/prim/mat/prob/dirichlet_rng.hpp
```diff
@@ -3,38 +3,83 @@
...
-          gamma_rng(rng, gamma_distribution<>(alpha(i, 0), 1));
+          gamma_rng(rng, gamma_distribution<>(alpha(i, 0), 1e-7));
         y(i) = gamma_rng();
-        sum += y(i);
       }
-
-      for (int i = 0; i < alpha.rows(); i++)
-        y(i) /= sum;
-      return y;
+      return y / y.sum();
```
syclik: any reason we don't replace with `return y.mean()`?

src/stan/math/prim/mat/prob/dirichlet_rng.hpp
```diff
@@ -3,38 +3,83 @@
...
+
 namespace stan {
 
   namespace math {
 
+    /**
+     * Return a draw from a Dirichlet distribution with specified
+     * parameters and pseudo-random number generator.
+     *
+     * For prior counts greater than zero, the usual algorithm that
```
syclik: Should `zero` be `one`?

src/stan/math/prim/mat/prob/dirichlet_rng.hpp
```diff
@@ -3,38 +3,83 @@
...
+        return theta;
+      }
 
-      double sum = 0;
+      // standard normalized gamma algorithm
       Eigen::VectorXd y(alpha.rows());
       for (int i = 0; i < alpha.rows(); i++) {
         variate_generator<RNG&, gamma_distribution<> >
-          gamma_rng(rng, gamma_distribution<>(alpha(i, 0), 1));
+          gamma_rng(rng, gamma_distribution<>(alpha(i, 0), 1e-7));
```
syclik: Is the `gamma_rng` function better behaved at `1e-7` vs `1`? Also, do we need to normalize if we're using a different coefficient other than 1?

bob-carpenter:
I'm going to reply all at once so as not to spam people.

1.  removed unnecessary include of `stan::math::log_sum_exp`

2.  the doc is right as written --- the prior count boundary is zero, because the parameter is prior count plus one --- I added a clarifiation note in the doc because this point confuses everyone (not that anyone reads our function doc).  

3.  no, gamma is not better behaved at 1e-7 and I do not believe more normalization is necessary. 

4.  we can't use `y.mean()` in place of `y / y.sum()` because the former is a scalar and the latter is a vector.

syclik:
Thanks!

syclik:
I'm merging this without running through all the tests -- I'm messing with the Jenkins config to handle the math lib change, so it'll fail.

I saw the changes made and it's ok.

src/stan/math/prim/mat/prob/dirichlet_rng.hpp
bob-carpenter: Didn't I answer this?  It's because we want to return
a simplex, not a scalar.

- Bob

syclik: Yes, you did.

Issue #1471, opened 2015-06-01
neg_binomial_2_log_rng error messages


bob-carpenter:
As of 2.6.3, `neg_binomial_2_log_rng` doesn't do enough of its own error checking and allows invalid values to be passed internally so that errors are raised in calls it makes internally to other rngs.

* [ ] check bounds at the input level for `neg_binomial_2_log_rng` rather than letting the underlying messages fail and get through;  if necessary, catch exceptions raised from internal calls to `gamma_rng` and `poisson_rng`

There are probably other cases of this for other functions, which don't need to be fixed as part of this issue, but can be logged into another issue if found.

randommm:
@bob-carpenter do you mind taking a look at https://github.com/stan-dev/stan/compare/feature/issue-1471-neg_binomial_2_log_rng-error-messages to check if that's what you think it should be?

bob-carpenter:
I left a sentence dangling that wasn't going anywhere (so I ended
it), but otherwise it looks right.  If you don't want to do it,
just unassign yourself.  

There are lots more functions we'd like to implement.  And many many
more open issues for 2.7.0++ we'd like to track down.  I think we're
going to roll out 2.7.0 ASAP to fix the issues with R builds and BH. 

- Bob

PR #1484, opened 2015-06-08
Fixes in error messages of neg_binomial_rng, neg_binomial_2_rng and neg_binomial_2_log_rng

randommm:
#### Summary:
Fixes in erros messages of neg_binomial_rng, neg_binomial_2_rng and neg_binomial_2_log_rng. Fix issue #1471

#### Intended Effect:
Do all checks inside the rng function itself, without calling other functions like stan::math::gamma_rng (this would output error messages of gamma_rng instead of neg_binomial_rng which would confound users).

#### How to Verify:
Check changes made. All current unit tests shall pass.

#### Side Effects:
None.

#### Documentation:
Not applicable.

#### Reviewer Suggestions:
Anyone.


src/stan/math/prim/scal/prob/neg_binomial_2_log_rng.hpp
```diff
@@ -31,20 +31,45 @@ namespace stan {
...
       using stan::math::check_finite;
       using stan::math::check_positive_finite;
+      using stan::math::check_not_nan;
+      using stan::math::check_nonnegative;
+      using stan::math::check_less;
 
       check_finite(function, "Log-location parameter", eta);
       check_positive_finite(function, "Precision parameter", phi);
 
+      double exp_eta_div_phi = std::exp(eta)/phi;
```
rtrangucci: @randommm why is this exp(eta) / phi instead of phi / exp(eta) like we had before?

randommm: Because we are calling boost::gamma_distribution directly instead of stan::math::gamma_rng. See file src/stan/math/prim/scal/prob/gamma_rng.hpp:


      /*
        the boost gamma distribution is defined by
        shape and scale, whereas the stan one is defined
        by shape and rate
      */
      variate_generator<RNG&, gamma_distribution<> >
        gamma_rng(rng, gamma_distribution<>(alpha, 1.0 / beta));
      return gamma_rng();

rtrangucci: @randommm nvm, I see that, yes, Stan's parameterization of gamma is different than boost's.

src/stan/math/prim/scal/prob/neg_binomial_2_log_rng.hpp
```diff
@@ -31,20 +31,45 @@ namespace stan {
                            RNG& rng) {
       using boost::variate_generator;
       using boost::random::negative_binomial_distribution;
+      using boost::random::poisson_distribution;
+      using boost::gamma_distribution;
 
       static const char* function("stan::math::neg_binomial_2_log_rng");
+      static const double POISSON_MAX_RATE = std::pow(2.0, 30);
 
       using stan::math::check_finite;
```
syclik: we can remove all these `using stan::math::check_*` statements: this is all in the same namespace now.

src/stan/math/prim/scal/prob/neg_binomial_2_rng.hpp
```diff
@@ -31,17 +31,44 @@ namespace stan {
                        RNG& rng) {
       using boost::variate_generator;
       using boost::random::negative_binomial_distribution;
+      using boost::random::poisson_distribution;
+      using boost::gamma_distribution;
 
       static const char* function("stan::math::neg_binomial_2_rng");
+      static const double POISSON_MAX_RATE = std::pow(2.0, 30);
 
       using stan::math::check_positive_finite;
```
syclik: remove these using statements

src/stan/math/prim/scal/prob/neg_binomial_2_rng.hpp
```diff
@@ -31,17 +31,44 @@ namespace stan {
                        RNG& rng) {
       using boost::variate_generator;
       using boost::random::negative_binomial_distribution;
+      using boost::random::poisson_distribution;
+      using boost::gamma_distribution;
 
       static const char* function("stan::math::neg_binomial_2_rng");
+      static const double POISSON_MAX_RATE = std::pow(2.0, 30);
```
syclik: We should move `POISSON_MAX_RATE` to `src/stan/math/prim/scal/fun/constants.hpp`. Right now, it's defined in 3 places.

src/stan/math/prim/scal/prob/neg_binomial_rng.hpp
```diff
@@ -32,16 +32,38 @@ namespace stan {
                      RNG& rng) {
       using boost::variate_generator;
       using boost::random::negative_binomial_distribution;
+      using boost::random::poisson_distribution;
+      using boost::gamma_distribution;
 
       static const char* function("stan::math::neg_binomial_rng");
+      static const double POISSON_MAX_RATE = std::pow(2.0, 30);
```
syclik: move to `constants.hpp`

src/stan/math/prim/scal/prob/neg_binomial_rng.hpp
```diff
@@ -32,16 +32,38 @@ namespace stan {
...
 
       using stan::math::check_positive_finite;
```
syclik: remove using statements

syclik:
Thanks, Marco. I don't know why it failed... I'll just start it again.

syclik:
Jenkins, retest this please.

syclik:
Are there any reasonable tests that could be written for this? I can see why it was written this way. It's a subtle change to the behavior of the RNG's, not just a shift in where the error message appears.

randommm:
Good point, but tests to what specifically? The behaviour shouldn't change since the checks are (or should be) exactly the same as the old ones, the only difference being the errors messages the user would face.

bob-carpenter:
I have tests that check the error messages being thrown have
certain phrases in them.  Otherwise, it's hard to check the
right error's being triggered in some of our interfaces because
they throw the same type of exception.

- Bob

randommm:
Where can I find one of those tests?

bob-carpenter:
I used a utility in test/unit/lang/utility.hpp called test_throws(),
but it boils down to this:

 try {
    ... what you want to test ...
    FAIL() << "should have thrown" << ...;
    if (std::string(e.what()).find(msg) != std::string::npos)
      FAIL() << "should have found msg inside error" << ...;
    SUCCEED();
 }

You can check for a more specific exception in the catch
if you want and add a second catch for std::exception& with
another failure saying you were expecting a different type of
exception.

- Bob

syclik:
Unfortunately (or fortunately) the behavior has changed. There are direct
calls to boost instead of relying on other stan math rngs. I verified this
by hand, although I might have messed the check up. I think the new code is
more restrictive, meaning it throws where the current state of develop
doesn't.

I'm ok with changing behavior; I just want tests that demonstrate that the
change was intentional.

Issue #1458, opened 2015-05-22
missing poisson_log_rng()


bob-carpenter:
As Bee pointed out on stan-users, the poisson_log_rng() function is not defined.

* [ ] add `poisson_log_rng()` --- implement as `poisson_rng(exp(...))`.
* [ ] add tests --- derive from `poisson_rng` tests

randommm:
The function poisson_log_rng() and its tests exist, I think it's just a matter of adding to function_signatures.h

If so, this could be included in 2.7.0

bob-carpenter:
No, it's there, it's just that what happens is that functions
call it with extreme parameters, then exceptions get raised in
calls it makes internally to gamma rng, so it reports errors
back to the user that there were errors in the gamma rng, not
that there was an error in the call to poisson_log_rng.

So we want to either trap the errors before the calls to the internal
RNG if we can or if not, just wrap everythnig in a try/catch where
the catch returns a more meaningful error that includes the info that
the error arose in poisson_log_rng.

randommm:
I think you are making a confusion with the other issue (about neg_binomial_2_log_rng error messages) which doesn't have anything to do with this one. poisson_log_rng() does not call gamma_rng.

bob-carpenter:
You're right --- sorry about that.

- Bob

PR #1488, opened 2015-06-09
Feature/issue 1458 missing poisson log rng

randommm:
This one is simple, fix issue  #1458: expose poisson_log_rng() which already has the tests and documentation.


bob-carpenter:
Looks good.  Thanks for adding this.

PR #1490, opened 2015-06-11
Feature/issue 1425 signature err msg

bob-carpenter:
#### Intended Effect:

Change error messages for unknown arguments/functions so that it distinguishes unknown functions from illegal argumetns and sampling statement from function usage.

#### How to Verify:

Four new test models with a new test program in `src/test/unit/lang/parser/function_sig_errors_test.cpp` with expected output.

#### Side Effects:

None other than that the message texts change.

And not quite a side effect, but I took care of a lot of the lint in the AST classes in `lang/` and the statement grammar.

#### Documentation:

None required.  This just makes error messages clearer to users.

#### Reviewer Suggestions:

Anyone.

Other than lint fixes, the only changes are in the two grammar files and the AST file, which adds a new function that changes behavior depending on context, with a default to the old behavior.


syclik:
Awesome!

PR #1492, opened 2015-06-15
Bugfix/issue 1295 no local var constrain

bob-carpenter:
#### Intended Effect:

Raise error in parser if local vector, row vector, or matrix types are declared with lower or upper bound constraints.

#### How to Verify:

Three new unit tests that fail in previous version.

#### Side Effects:

May stop some existing models from working, but those constraints weren't being checked and were disallowed by the spec.

#### Documentation:

Existing doc is correct.

#### Reviewer Suggestions:

Anyone.


syclik:
@bob-carpenter, thanks. This looks good. We should make sure to put a note that existing models may fail to compile if they have used constraints in local vars.

bob-carpenter:
Where should such a note go?  The doc says there can't be
constraints on local vars already.  

A warning in the release notes or to the list, perhaps?

syclik:
If it's in the manual, I think a note in the release notes is sufficient.

PR #1493, opened 2015-06-15
Bugfix/issue 1100 assignment error clearer

bob-carpenter:
#### Intended Effect:

Add new non-terminal with better name for error reporting for unassignable expression on right-hand side of assignment.

#### How to Verify:

There's a new test model with the a test for the message saying variable is unassignable.

#### Side Effects:

Error messages are slightly different.

#### Documentation:

None --- behavior is unchanged other than clarifying text of error messages.

#### Reviewer Suggestions:

Anyone.


syclik:
Looks good.

PR #1494, opened 2015-06-15
Bugfix/issue 1152 ccdf sigs

bob-carpenter:
#### Intended Effect:

1. Raise warning from Stan parser when attempting to use CDF or known non-distribution with `_log` suffix with sampling notation.

2.  Soften the "Jacobian-required" message to a warning conditioned on checking non-linearity (same file, so I couldn't pass it up).

3.  Fix bug in warnings tests in parser test utiliy leading to false negatives on warnings test (still no idea why the old way didn't work --- maybe a macro resolution syntax issue?).

#### How to Verify:

1. New models tickling new warnings in parser unit test for statement grammar.

2. Unit test for warning in statement grammar.

3.  Revised test for Jacobian messages in statement grammar unit test will fail as before (but not as the test was written before), but the revised test will pass.

#### Side Effects:

None (models wouldn't compile before and will fail with warning message now)

#### Documentation:

None --- this brings behavior in line with doc.

#### Reviewer Suggestions:

Anyone.

### DIFF
--- a/src/stan/lang/grammars/statement_grammar_def.hpp
+++ b/src/stan/lang/grammars/statement_grammar_def.hpp
@@ -248,6 +248,26 @@ namespace stan {
         std::string function_name(s.dist_.family_);
         std::string internal_function_name = function_name + "_log";
 
+        if ((internal_function_name.find("multiply_log") 
+             != std::string::npos)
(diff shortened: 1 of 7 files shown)


syclik:
Looks good.

PR #1495, opened 2015-06-16
Feature/issue 445 deprecate model write csv

bob-carpenter:
#### Summary:

Remove `write_csv` methods from generated model code.

Remove lots of lint from `generator.hpp`

#### Intended Effect:

Make code easier to maintain and remove buggy code that wrote in the wrong order and was confusing people looking at the program C++ code.

#### How to Verify:

Look at generated model code.  

#### Side Effects:

Any code that might have dependenced on `write_csv()`.  But that code probably would've been broken because the header writer and value writer weren't matched.  These can be replaced with the newer array-based methods.

#### Documentation:

None.

#### Reviewer Suggestions:

Anyone.


syclik:
Looks good.

PR #1497, opened 2015-06-17
Feature/issue 1383 versions

syclik:
#### Summary:

Adds the Stan version as a macro.
* [x] define `STAN_MAJOR`, `STAN_MINOR`, `STAN_PATCH`
* [x] define `STAN_MATH_MAJOR`, `STAN_MATH_MINOR`, `STAN_MATH_PATCH`
* [x] add it to the language parser to check for reserved words
* [x] add to the manual's list of reserved words.

Also added `fvar` as a reserved word.

#### Intended Effect:

Adds versions as macros. The original `const std::string`s are still available.

#### How to Verify:

New unit test:
```
> ./runTests.py src/test/unit/lang/parser/reserved_words_test.cpp
```

#### Side Effects:

Added a few more reserved words.

#### Documentation:

`language.tex` was updated with the keywords.

#### Reviewer Suggestions: 

@bob-carpenter. Can you make sure I categorized the macros correctly in the manual?

### DIFF
--- a/lib/stan_math_2.6.3/.gitignore
+++ b/lib/stan_math_2.6.3/.gitignore
@@ -1,5 +1,3 @@
-
-
 # /
 /*.dSYM
 
@@ -38,6 +36,7 @@ stan.kdev4
 !/test/*.hpp
 !/test/*.cpp
 !/test/*.stan
+
 ## generated tests
 /test/prob/*/*_generated_*_test.cpp
 ## lib files
--- a/lib/stan_math_2.6.3/stan/math/version.hpp
+++ b/lib/stan_math_2.6.3/stan/math/version.hpp
@@ -3,12 +3,12 @@
 
 #include <string>
 
-#ifndef STRING_EXPAND
-#define STRING_EXPAND(s) #s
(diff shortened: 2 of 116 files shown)


bob-carpenter:
I think defining a property called `STRING` could be dangerous in terms of conflicts.  Can we call that `STAN_STRING` or something?  And do we really need two levels of indirection here in the macro or can those two string macros be combined into one?

Otherwise, everything looks great.  I particularly love your reserve in leaving versions at 2.6.3 until the release update!

syclik:
Sure, we can call it `STAN_STRING`. We need two levels of indirection --
macros are a nasty business. I tried to do it with one; it didn't work.

(for the manual branch -- there's no need to update version numbers. I
search through the manual to update versions, so no need to do that by
hand.)
​

bob-carpenter:
> On Jun 17, 2015, at 12:22 AM, Daniel Lee <notifications@github.com> wrote:
> 
> Sure, we can call it `STAN_STRING`.

Thanks.

> We need two levels of indirection --
> macros are a nasty business. I tried to do it with one; it didn't work.

OK.

PR #1505, opened 2015-06-20
Feature/bbvb family

dustinvtran:
#### Summary:
Allows one to specify an arbitrary class of variational distribution in order to perform variational inference with. I built a base class `base_family.hpp` from which all variational distributions should inherit from and build methods for, c.f., `variational/families` directory, which makes the two Gaussian families we currently have as special cases.

#### Intended Effect:
No effect on the user/outputs. Removes a ton of hard coded nastiness and makes it easier to do black box variational inference with whatever desired variational distributions.

#### How to Verify:
Run tests, run models. Fix a seed and check that the estimates are exactly the same as in the original codebase.

#### Side Effects:
None

#### Documentation:
None

#### Reviewer Suggestions:
@akucukelbir


dustinvtran:
I think this is a good enough separation of the raw inference algorithm from the variational families used, i.e., the raw inference algorithm simply calls upon generic methods for a variational family such as `variational.sample()` and `variational.mean()`.

Ideally the gradient of the ELBO should exist as a function inside `advi.hpp` and not in the variational family classes. At the moment it's so directly tied to the variational families that for now I'm leaving it as `variational.calc_grad()`.

bob-carpenter:
@dustinvtran Thanks!

@syclick Do you understand this code well enough to review it?  I missed the overview and haven't been caught up yet.

akucukelbir:
@bob-carpenter this is a branch off of feature/bbvb. i can review it and merge it into feature/bbvb. so nothing here is touching develop at the moment.

bob-carpenter:
@dustinvtran It's only failing in the cpplint style filter, which enforces the Google C++ style.  The errors are usually very easy to fix.  You can run it locally using `make cpplint`.  See:

https://github.com/stan-dev/stan/wiki/Coding-Style-and-Idioms

dustinvtran:
Great, thanks. I went ahead and fixed it on both branches.

akucukelbir:
you're moving too fast for me to react! :+1: thanks dustin.

PR #1509, opened 2015-06-22
Bugfix/issue 1506 intel 14 parser

bob-carpenter:
#### Summary:

* Remove all uses of lazy evaluation of assignment (operator=) from semantic actions in the grammar

* ifdef out the offending bit of code in the Boost lib for the Intel compiler (yes, this involved changing Boost);  see  https://software.intel.com/en-us/articles/boost-compiler-error-class-has-no-member-type

* remove all the remaining lint from the grammars

#### Intended Effect:

Allow Stan to be compiled with Intel 14 compiler.

#### How to Verify:

Verifying the intended effect requires an Intel compiler, but it should be enough to verify that it passes unit tests.  No behavior should have changed --- this is purely refactoring.

#### Side Effects:

That definition of `operator=` is now invisible for clients using our copy of Boost.  This is very unlikely to be a problem.

#### Documentation:

Nothing user facing.

#### Reviewer Suggestions:

Anyone.

### DIFF
--- a/lib/boost_1.58.0/boost/phoenix/core/actor.hpp
+++ b/lib/boost_1.58.0/boost/phoenix/core/actor.hpp
@@ -209,6 +209,12 @@ namespace boost { namespace phoenix
             return *this;
         }
 
+#ifdef __INTEL_COMPILER 
+/* Intel compiler cannot currently compile return type on assignment
+ * operator declared below.  See:
+ * https://software.intel.com/en-us/articles/boost-compiler-error-class-has-no-member-type
+ */ 
+#else 
         template <typename A0>
         typename proto::result_of::make_expr<
             proto::tag::assign
@@ -220,6 +226,7 @@ namespace boost { namespace phoenix
         {
             return proto::make_expr<proto::tag::assign, phoenix_domain>(this->proto_expr_, a0);
         }
+#endif
 
         template <typename A0>
         typename proto::result_of::make_expr<
--- a/lib/boost_1.58.0/boost/phoenix/core/actor.hpp
+++ b/lib/boost_1.58.0/boost/phoenix/core/actor.hpp
@@ -209,17 +209,17 @@ namespace boost { namespace phoenix
             return *this;
         }
 
-        template <typename A0>
-        typename proto::result_of::make_expr<
-            proto::tag::assign
-          , phoenix_domain
-          , proto_base_expr
-          , A0
-        >::type const
-        operator=(A0 const & a0) const
-        {
-            return proto::make_expr<proto::tag::assign, phoenix_domain>(this->proto_expr_, a0);
-        }
+        // template <typename A0>
+        // typename proto::result_of::make_expr<
+        //     proto::tag::assign
+        //   , phoenix_domain
+        //   , proto_base_expr
+        //   , A0
+        // >::type const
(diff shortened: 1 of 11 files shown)


bob-carpenter:
That failure looks unrelated --- it looks more like an issue about output streams in CmdStan.

syclik:
yeah -- sorry, I was trying to fix it, but haven't had time. I'll get it
working again.

syclik:
Jenkins, retest this please.

syclik:
@bob-carpenter, everything looks good, but the change made to the boost library won't help Intel compile users with RStan, right? That will still rely on the BH package, which hasn't been patched.

bob-carpenter:
Correct.  

Ben's going to patch RStan so that it uses the same Boost
header.  As long as we make sure to search ours first, the header
guard will silence the one from Boost.  At least I think that's
Ben's plan.

- Bob

bgoodri:
I now have an icpc version 15. I am getting a test failure on this branch at
```
[ RUN      ] io_dump.reader_big_doubles
unknown file: Failure
C++ exception with description "data b value 2.22507e-308 beyond numeric range" thrown in the test body.
```
The rstan package installs but when you try to run a model, you get
```
/usr/local/lib/R/site-library/rstan/include/rstan/stan_fit.hpp(947): error: no default constructor exists for class "stan::mcmc::fixed_param_sampler"
          stan::mcmc::fixed_param_sampler sampler;
```
followed by an error novel. This is without doing the fixed_param algorithm, which is weird.

bgoodri:
Never mind. After another git pull and a reinstall of rstan, it works with icpc. But the bignumbers unit test failure is still there.

bob-carpenter:
You mean the failure above of `reader_big_doubles`?   With the most recent Intel compiler?

If that's not working, it looks like the Intel compiler messed up either its `std::numeric_limits` or its I/O.  

I wonder if it's an issue with normalization?  It looks like `min()` is normalized but not `max()`:

http://www.cplusplus.com/reference/limits/numeric_limits/

bgoodri:
Yes, changing dmin to lowest() passes in C++11 mode, as does changing dmin
to -dmax without C++11.

bob-carpenter:
I'm confused --- where did you get dmin and dmax?

- Bob

bob-carpenter:
Ben --- can you go ahead and make those fixes so we test the normalized min and max?  Or give us a hint as to where to find dmin and dmax so we can do it?

bgoodri:
Right now, it is
```
  double dmax = std::numeric_limits<double>::max();
  double dmin = std::numeric_limits<double>::min();
```
which fails, but changing dmin to either
```
  double dmin = std::numeric_limits<double>::lower();
```
or
```
  double dmin = -dmax;
```
passes. So, I think it does have something to do with the normalization.

bgoodri:
OK. This passes:
```
double dmin = 2.225075e-308;
```
whereas
```
double dmin = std::numeric_limits<double>::min(); // about 2.22507_4_e-308
```
fails. So, it looks as if icpc can't handle that edge case.

bob-carpenter:
Thanks for tracking this down.  

I think it's my fault for not understanding the tests 
I put in there to begin with.  I just have a habit of trying
to test boundary conditions.

- Bob

PR #1513, opened 2015-06-22
Feature/issue 1512 unconstrained param access

bob-carpenter:
#### Summary:

Exposed `params_r__` in model block for users in a way parallel to `lp__`, but with even sterner warnings.

#### Intended Effect:

Allow Andrew to implement EP the way he wants.

#### How to Verify:

There is a grammar that uses it where legal in the model block that is now part of the good model tests and in the parser test.

If you run this model

```
parameters {
  real mu[2];
  real<lower=0> sigma;
}
model {
  params_r__[1] ~ normal(0, 1);
  params_r__[2] ~ normal(10, 100);
  params_r__[3] ~ normal(1, 0.25);
}
```

in one chain (hey, it's CmdStan) you will get (roughly) this:

```
                 Mean     MCSE   StdDev    5%    50%   95%  N_Eff  N_Eff/s    R_hat
lp__            -0.47  5.9e-02  1.2e+00  -2.8  -0.20  0.83    425     2511  1.0e+00
accept_stat__    0.94  2.4e-03  7.5e-02  0.78   0.97   1.0   1000     5910  1.0e+00
stepsize__       0.70  2.2e-15  1.6e-15  0.70   0.70  0.70   0.50      3.0  1.0e+00
treedepth__       2.4  1.9e-02  5.7e-01   2.0    2.0   3.0    954     5640  1.0e+00
n_leapfrog__      4.5  6.6e-02  2.1e+00   3.0    3.0   7.0    966     5711  1.0e+00
n_divergent__    0.00  0.0e+00  0.0e+00  0.00   0.00  0.00   1000     5910      nan
mu[1]           0.048  3.9e-02  9.7e-01  -1.6  0.023   1.7    627     3708  1.0e+00
mu[2]              11  4.6e+00  1.0e+02  -157    9.9   187    513     3031  1.0e+00
sigma             2.9  3.1e-02  7.2e-01   1.9    2.8   4.2    555     3280  1.0e+00
```

#### Side Effects:

None.

#### Documentation:

Intentionally no user-facing doc.  This one's just intended for Andrew.

#### Reviewer Suggestions:

Anyone.


bgoodri:
Don't you need to turn off the Jacobian adjustment for `sigma`?

bob-carpenter:
[new title to get Andrew's attention]

Having just thought about this more, I think you're right.
Maybe this is what Michael was worried about.

But there's no way to do that programatically with sampling,
so maybe we should just drop this feature or at least wait until
we have a way to plug it in.  I have no idea what
Andrew's doing now --- presumably undoing the Jacobian if there
are constraints or just using all unconstrained params.

If we have to modify the way the commands can be called
and compilation happens, that's going to be a big change
down through all the calls to the log density from the 
MCMC code.

- Bob

betanalpha:
I think you want the Jacobian there — the point is to add an additional
prior to the existing model which would require the Jacobian.  In other
words, you’re not putting a Gaussian prior on the unconstrained space
is isolation, you’re adding it to an existing model.  Here that existing model
would be uniform on all of the parameters, requiring the Jacobian.

syclik:
Looks great. Just needs a merge with develop.

bgoodri:
Did Andrew actually say that he was thinking along the lines of what Michael said? I suppose you could think about it that way, but I would guess Andrew wasn't.

PR #1532, opened 2015-07-09
Bugfix fix full rank dimension when calculating gradient

dustinvtran:
#### Summary:
This fixes the failed unit test that @syclik noted in a [cmdstan pull request](https://github.com/stan-dev/cmdstan/pull/117), caused by dimensions not matching when calculating the gradient of the ELBO.

#### Intended Effect:
Unit test passes.

#### How to Verify:
In cmdstan:
```{bash}
make test/interface/services/argument_configuration_test
test/interface/services/argument_configuration_test
```

#### Side Effects:
None

#### Documentation:
None

#### Reviewer Suggestions:
@syclik

### DIFF
--- a/src/stan/variational/families/normal_fullrank.hpp
+++ b/src/stan/variational/families/normal_fullrank.hpp
@@ -251,7 +251,7 @@ namespace stan {
 
         // Initialize everything to zero
         Eigen::VectorXd mu_grad = Eigen::VectorXd::Zero(dimension_);
-        Eigen::VectorXd L_grad  = Eigen::MatrixXd::Zero(dimension_, dimension_);
+        Eigen::MatrixXd L_grad  = Eigen::MatrixXd::Zero(dimension_, dimension_);
         double tmp_lp = 0.0;
         Eigen::VectorXd tmp_mu_grad = Eigen::VectorXd::Zero(dimension_);
         Eigen::VectorXd eta = Eigen::VectorXd::Zero(dimension_);


src/stan/variational/families/normal_fullrank.hpp
```diff
@@ -251,7 +251,7 @@ namespace stan {
```
syclik: @dustinvtran / @akucukelbir: can you add at least one test for the calc_grad function? I just looked and none of the tests in Stan actually test it directly.

There were only tests that it failed, not that ran.

bob-carpenter:
There are a bunch of tips in this comment, none of which are urgent, but all of which are moderately important in the long run for maintaining our code base.

The code looks fantastic, so much so that I think there are way too many comments.  You don't need to have comments like `Initialize everything to zero` when the code is obviously doing that.

You also don't need to start a class with a shouted comment like `MULTIVARIATE NORMAL DISTRIBUTION FULL-RANK` when the name of the class is `normal_fullrank`.  

The main goal is to make the code easy to read so the in-code comments aren't necessary.  The two examples above are great, because you wrote the code the right way in the first place, so you don't need the comments. 

Also, you don't document `@param` on classes (at least I don't think you do if doxygen is following Javadoc in this).   I think those should be comments on the variables themselves if you think they're necessary (which I don't, because they're well named).

We sort of take `L` to imply a Cholesky decomposition, so `L_chol` seems a bit redundant.  When in doubt, shorter names are better.  For instance, I wouldn't write

```
normal_fullrank operator+(double scalar, normal_fullrank rhs) {
      return rhs += scalar;
```

because calling out that a double is a scalar is redundant,  so I'd just make that

```
normal_fullrank operator+(double x, normal_fullrank rhs) {
      return rhs += x;
```



/*
     * MULTIVARIATE NORMAL DISTRIBUTION FULL-RANK
     *
     * Variational family as full-rank multivariate normal distribution
     *
     * @param  mu     mean vector
     * @param  L_chol cholesky factor of covariance (\Sigma = L_chol *
     *                L_chol.transpose())
     */
    class normal_fullrank : public base_family {
    private:
      Eigen::VectorXd mu_;
      Eigen::MatrixXd L_chol_;
      int dimension_;
```

with the following, though given the names you choose for the member variables, I really don't think you need the variable doc:

    /**
     *
     * Variational family is full-rank multivariate normal.
     */
    class normal_fullrank : public base_family {
    private:

      /** mean vector */
      Eigen::VectorXd mu_;
 
      /**  cholesky factor of covariance */
      Eigen::MatrixXd L_chol_;

      /** I don't think this needs to a name */
      int dimension_;
```

Then you absolutely, no questions asked, should get rid of the comment `Constructors` before the constructors.  Remember the golden rule of doc from the Google style guide:  assume the person reading the code will know the language better than you.  But then I would document each constructor's intent (which isn't obvious) and the parameters and constraints on them and when it'll throw an exception in the face of bad inputs.  This constructor is an external interface, so it's something you want to doc directly so that someone can read how to use the class from the external doc.  

Exact same comment for `Accessors` and `Mutators` --- document the methods, but you don't need to doc blocks of them.  

I would also strongly strongly encourage you to think in terms of immutable classes that do not allow any mutators, but is rather constructed in a safe state and immutable thereafter.  The reason to do this is that it's infinitely easier to understand how a class works in context if its behavior is fixed and consistent from the get go.  And it's immediately thread safe, not that that's likely to be an issue here.

Note that your constructor does NOT check for consistency, whereas the mutator does.  The constructor should always test for consistency if it is external client-facing code (here that's a bit of a fine line since the client is just more of Stan).  If this is just something getting called internally, then you need to make the test higher up, not repeat it endlessly through chains of calls.  One way to make sure this is all consistent is to have the constructors call the mutators.  Then if you don't want the class to acutally be mutable, have the accessors return constants, which is what it's doing.

Nice use of const declarations, by the way!

In things like this,

```
        Eigen::VectorXd eta = Eigen::VectorXd::Zero(dimension_);

        // Draw from standard normal and transform to real-coordinate space
        for (int d = 0; d < dimension_; ++d) {
          eta(d) = stan::math::normal_rng(0, 1, rng);
        }
```

You don't need to set to zero and then reassign --- it's just a bunch of extra copies.  Instead, replace the first line with

```
        Eigen::VectorXd eta(dimension_);
```

As you can now anticipate, I don't think you need that comment.  

You would conventionally use `eta.size()` as the loop bound --- locals are almost always better to access than member variables.  And it's clearer that it's filling the whole thing.

I'd also rewrite this

```
 double entropy() const {
        double tmp = 0.0;
        double result = 0.5 * static_cast<double>(dimension_) *
          (1.0 + stan::math::LOG_TWO_PI);
        for (int d = 0; d < dimension_; ++d) {
          tmp = fabs(L_chol_(d, d));
          if (tmp != 0.0) {
            result += log(tmp);
          }
        }
        return result;
      }
```

to

```
double entropy() const {
  return 0.5 * dimension_ * (1 + stan::math::LOG_TWO_PI)
    + L_chol_.diagonal().log().sum();
}
```

under the assumption that if the Cholesky factor isn't actually a Cholesky factor of a covariance matrix, you're in trouble.  Or do you anticipate having non full-rank Cholesky factors?  Or ones with negative entries?  It seems like just sweeping the problem under the rug in the definition of `entropy()` is not a good idea.  I could be way off base here, though...

Also, C++ automatically promotes integers to doubles in mixed operations, so you don't need all that static casting and `.0` business.  

Finally, I think you want to make all the class-external arithmetic operators inline.  And by now, I hope that redundant doc that says `Arithmetic operators` is bugging you as much as it bugs me.

syclik:
I'm merging now. We'll get it released. We can fix these with another pull request.

dustinvtran:
Thanks for the comments Bob! Making the changes now (and will submit another pull request when it's done). :)

bob-carpenter:
Great --- no rush at all, of course --- they're mostly just
cosmetic changes anyway.  

- Bob

PR #1539, opened 2015-07-12
Feature/advi lp

dustinvtran:
#### Summary:
Calculates log probabilities when writing to output for variational inference, in order to be consistent with sample and optimize (for now, before any refactoring which will happen after 2.7.0).

#### Intended Effect:
N/A

#### How to Verify:
Run models by building them, then running them, e.g. `make examples/bernoulli/bernoulli && cd examples/bernoulli && ./bernoulli variational data file=bernoulli.data.R`.

#### Side Effects:
N/A

#### Documentation:
See [cmdstan pull request](https://github.com/stan-dev/cmdstan/pull/125)

#### Reviewer Suggestions:
N/A

### DIFF
--- a/src/stan/variational/advi.hpp
+++ b/src/stan/variational/advi.hpp
@@ -97,20 +97,16 @@ namespace stan {
         double elbo = 0.0;
         int dim = variational.dimension();
 
-        Eigen::VectorXd zeta = Eigen::VectorXd::Zero(dim);
+        Eigen::VectorXd zeta(dim);
 
         for (int i = 0; i < n_monte_carlo_elbo_; ++i) {
-          // Draw from variational distribution
           zeta = variational.sample(rng_);
-
           // Accumulate log probability
-          elbo += (model_.template log_prob<false, true>(zeta, print_stream_));
+          elbo += model_.template log_prob<false, true>(zeta, print_stream_);
         }
 
         // Divide to get Monte Carlo integral estimate
         elbo /= n_monte_carlo_elbo_;
-
-        // Add entropy term
         elbo += variational.entropy();
 
         return elbo;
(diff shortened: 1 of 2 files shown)


src/stan/variational/advi.hpp
```diff
@@ -313,7 +313,7 @@ namespace stan {
 
         if (out_stream_) {
           services::io::write_iteration(*out_stream_, model_, rng_,
-                                        0.0, cont_vector, disc_vector,
+                                        lp, cont_vector, disc_vector,
```
syclik: @dustinvtran, I think it's about time we found a way to test for output in the variational inference code. We should have a test that verifies this is actually getting out to the `out_stream_`.
How are you currently verifying that this change 1) doesn't break and 2) correctly streams things out? We should just get that into a repeatable test so we don't regress in the future.

syclik:
@dustinvtran, just added a comment. Let's get a test in first. The rest of it looks ok, but that's just based on inspecting the code.

dustinvtran:
Okay cool, just pushed a unit test on the cmdstan repo: https://github.com/stan-dev/cmdstan/pull/125

syclik:
It's good that the test is in cmdstan, but I think we'd want a similar test in stan itself. We won't be able to tell if the code gets broken in Stan until we test CmdStan with the test only sitting out there.

dustinvtran:
FYI I added it to the current unit test which runs ADVI on the hierarchical logistic model. It seems natural since checking output in Stan itself requires running ADVI on an actual model.

syclik:
Just added a test to check the `lp` value. Just making sure it's not exactly 0. Let me know if you think there's a better check that we should do here.

dustinvtran:
Yup, that looks fine to me.

PR #1546, opened 2015-07-15
Feature/issue 1361 writer callback

betanalpha:
#### Summary:

This pull request sets the framework for the new Stan3 API, beginning with callbacks and writers.  This is intended to be a long-term pull request as lots of work is needed to clean everything up.

#### Intended Effect:

Reorganize the internal library into services, which provide functions for interfaces to call, and callbacks, which interfaces must implement for the API.

#### How to Verify:

Unit tests compile and all but McmcChains pass.  

#### Side Effects:

All interfaces will break.

#### Documentation:

None -- I can't seem to find the original wiki page I wrote documenting the new organization.

#### Reviewer Suggestions: 

Everyone.  This is a chance to add the hundreds of tests we need.

@syclik What do you want tested?  

@rtrangucci @bob-carpenter Comments on general design.  See https://github.com/stan-dev/cmdstan/blob/develop/src/cmdstan/command.hpp for example use.

@ariddell @bgoodri @maverickg Can you implement an interrupt callback and a writer callback in (Py/R)Stan?

@akucukelbir @dustinvtran The variational config and execution code is now in src/stan/services/variational/variational.hpp, but it needs to be updated to use writers instead of iostreams.  See src/stan/services/sample/sample.hpp for a working example.  Also, we should use a verb for the services -- I recommend "variational_optimize".


bob-carpenter:
I'm totally slammed until the end of our course on Tuesday.
I want to take a look at it then.

What's a "long-term pull request"?  Do you intend to keep
it open rather than merge it?  

- Bob

betanalpha:
Long-term as in it’s not yet complete — the pull
request is there not to signal a final proposal,
but rather to give us a venue to discuss and
document necessary updates.  Note also that’s
it’s merging into Stan3 and not dev.

src/stan/io/stan_csv_reader.hpp
```diff
@@ -256,14 +256,14 @@ namespace stan {
             break;
 
           if (comment_line) {
-            if (line.find("(Warm-up)") != std::string::npos) {
-              int left = 17;
-              int right = line.find(" seconds");
+            if (line.find("Warmup") != std::string::npos) {
```
syclik: @betanalpha, can we split this into its own pull request so it can patch `develop` now? This seems useful.

betanalpha: This is just an update to the csv_reader because the
writers change the CmdStan output slightly.  Not sure
what the general utility would be.

Really this needs to get pulled into CmdStan in a following
update.

syclik: Ok, then let's split that out into its own pull request.

syclik:
@betanalpha, let's ditch the folder-level include files. It's cluttering up the diff.

src/stan/model/util.hpp
```diff
@@ -263,24 +263,32 @@ namespace stan {
      * Jacobian determinant of inverse parameter transforms is added to the
      * log probability.
      * @tparam M Class of model.
+     * @tparam Writer Class of writer callback
      * @param model Model.
      * @param params_r Real-valued parameter vector.
      * @param params_i Integer-valued parameter vector.
+<<<<<<< HEAD
```
syclik: the git diffs got added.

syclik:
I just removed the folder-level include. There was just one.

I don't know why the diff isn't tracking through file moves. I'm also going to pull some of these changes into separate pull requests.

betanalpha:
I removed the rest of the folder-level includes.

bob-carpenter:
Daniel and I were trying to look through here, but the way it's organized makes it very hard.  Our (really Daniel's) suggestion is to break this down into three pieces:

#### Refactor

(1)  use git mv to move old files to new locations.  if you don't do this, we can't look at changes easily and it just looks like 80 new files to review.  Can this be done so that all tests pass?  If so, this would
be step 1.  If not, it'll have to be combined with (2).

(2)  update everything so that it uses the new structure, but produces the same output as before.  This way we don't have to untangle changes to behavior with refactoring.

#### Change Behavior

(3)  After the refactor is done (1 and 2), change the behavior in a new pull request.

Also, is there a description somewhere of how the behavior is intended to change at a high level other than just a unit test?  Daniel tells me it's changed, but I don't know how.

syclik:
@betanalpha, I came across some issues and I'm glad I'm splitting this apart. I don't think I would have caught it otherwise and we'd be in a whole lot of hurt.

- a lot of the writers were broken: `stringstream`, `cerr`, `cout`. I'm not sure about the others. They don't handle the printing correctly. I've fixed that on a local branch.
- do we need this many functions for our writers?
  1. `void operator()(const std::string& key, double value)`
  2. `void operator()(const std::string& key, const std::string& value)`
  3. `void operator()(const std::string& key, const double* values, int n_values)`
  4. `void operator()(const std::string& key, const double* values, int n_rows, int n_cols)`
  5. `void operator()(const std::vector<std::string>& names)`
  6. `void operator()(const std::vector<double>& state)`
  7. `void operator()()`
  8. `void operator()(const std::string& message)`

Could we get away with:
  1. `void operator()(const std::string& message)`
  2. `void operator()`
and provide a few helper functions that take the rest of the arguments and provide strings? I was thinking that only other useful function I would include is `comment(const std::string& message)`, but I don't think that's necessary.

I think that would reduce a lot of the complexity of the writers. Thoughts?

betanalpha:
What do you mean broken?  Some of the output changed intentionally but the classes themselves should have all compiled and run.

The _whole point_ of the refactor is to not burden the interfaces with string manipulation.  By passing key-value pairs the interfaces can now access timing and adaptation parameters naturally.  Same for the vectors, not to mention enabling advanced options like streaming aggregators.

syclik:
@betanalpha, I just removed the old tests that were in there for all the writer stuff.

I think all of that needs to be tested. I'll make inline comments.

src/stan/interface_callbacks/writer/cerr.hpp
```diff
@@ -0,0 +1,66 @@
+#ifndef STAN_INTERFACE_CALLBACKS_WRITER_CERR_HPP
```
syclik: add tests. (delete comment when added.)

src/stan/interface_callbacks/writer/cerr.hpp
```diff
@@ -0,0 +1,66 @@
...
+
+        void operator()(const std::string& key,
+                        const double* values,
+                        int n_rows, int n_cols) {
+          if (n_rows == 0 || n_cols == 0) return;
+
+          std::cerr << key << ":" << std::endl;
+
+          for (int i = 0; i < n_rows; ++i) {
+            std::cerr << "," << values[i];
```
syclik: bug. rewrite loop.

src/stan/interface_callbacks/writer/cout.hpp
```diff
@@ -0,0 +1,65 @@
+#ifndef STAN_INTERFACE_CALLBACKS_WRITER_COUT_HPP
```
syclik: add tests. (delete comment when added.)

src/stan/interface_callbacks/writer/fstream_csv.hpp
```diff
@@ -0,0 +1,94 @@
+#ifndef STAN_INTERFACE_CALLBACKS_WRITER_FSTREAM_CSV_HPP
```
syclik: add tests. (delete comment when added.)

src/stan/interface_callbacks/writer/noop.hpp
```diff
@@ -11,11 +11,18 @@ namespace stan {
```
syclik: add tests. (delete comment when added.)

src/stan/interface_callbacks/writer/stringstream.hpp
```diff
@@ -0,0 +1,101 @@
+#ifndef STAN_INTERFACE_CALLBACKS_WRITER_STRINGSTREAM_HPP
```
syclik: add tests. (delete comment when added.)

src/stan/io/stan_csv_reader.hpp
syclik: I'd still like to see the change in behavior as a separate pull request.

bob-carpenter: What in the world is there a hard coded 17 in here for?  Isn't
that going to make things rather brittle?

src/stan/interface_callbacks/writer/fstream_csv.hpp
```diff
@@ -0,0 +1,94 @@
...
+
+          std::vector<double>::const_iterator last = state.end();
+          --last;
+
+          for (std::vector<double>::const_iterator it = state.begin();
+               it != last; ++it)
+            output << *it << ",";
+          output << state.back() << std::endl;
+        }
+
```
bob-carpenter: I prefer

``
void operator()(const std::vector<double>& state) {
  for (std::vector<double>::const_iterator it = state.begin();
       it != state.end(); ++it) {
    if (it != state.begin()) output << ',';
    output << *it;
  }
}
```

And I'd replace `output` with `output__` to follow our member variable naming convention.

bob-carpenter:
Overall, I'm OK with all of this.  There are a lot of little details that follow, but please don't feel like I expect them all to be done.  Just my opinion on style and idioms and I know you (Michael) have very different opinions.

I like that the interrupts get their own base class.  

If there are performance issues downstream, we could replace the OO with the CRTP (like Eigen or Boost's random package uses).  But that would require templating rather than interfaces based on base classes.

Daniel just commented "bug." in cerr.hpp and I don't see what he means looking at the code.

Why make the flush operation `operator()()` in base_writer.hpp?  That seems very confusing compared to calling it `flush()`.  

What is `is_writing()` for?  Is that like a logger pattern where we want to avoid calling complex string manipulations if something's not going to get logged?  I also don't see it implemented in cerr.hpp, so I don't see how that class could be instantiated.

I don't see any reason not to replace `writer::cerr` and `writer::cout` with `writer::ostream_writer` that stores a member variable for the stream to write to.   But maybe I missed some difference in behavior.

I see that `writer::fstream_csv` writes in a different format.  Is that the same output as `writer::stringstream`?  What are these for?  And what's the method `fstream_csv::good()` for --- it won't be callable from anything that just takes a `base_writer` and I don't see the need to call it on the outside because it delegates to a stream. 

Rather than `stringstream::clear()`?  I would prefer to just create a new `writer::stringstream` unless this thing's going to get used in a tight loop.  Also no big deal, but it removes the possible error of forgetting to clear a stream before reuse. 

I can see what `stringstream::contents()` is for---that wouldn't be called by something taking a generic writer, so that's OK.

If the `writer::messages` class is for different kinds of output, then I'd use a different base class.  

I also strongly prefer the `no_op` class to the `is_writing` thing with pointers --- it'll even be more efficient because there's no branching.  That'll let us replace the pointer with references in the regular `writer::messages` class.

I don't know that you need this deep a namespace nesting --- `interface_callbacks` should be enough.

In `base_mcmc.hpp`, What is `base_mcmc::clear_buffers()` for?  I also don't see what the flush operations do given that they return empty strings.  If you really care about efficiency (don't know that it matters), you could make a static empty string and just return it by reference everywhere.  Which reminds me that the returns can probably be constant for the strings---otherwise, just returning a static string wouldn't be safe.   

I also don't like the pointers for the streams, but see this is going on everywhere.  Are you going to try to pass 0 in some cases?  I realize this has been in the code for a while.  It looks like you've moved to more mutability, which I really don't like.  It's just hard to guarantee invariants like a class always being in a usable state.

Finally, why is there a protected `name_`?   I see no reason to have a mutable class that stores only a single mutable name---can't you pass it in through the constructor and store it as a constant?  Then you could have `name()` return a constant string reference rather than copying each time.

For `msg.str(std::string());` wouldn't it be easier to understand if you just wrote `msg << std::endl`?

I don't understand the `argument_parser`, though I'm not keen on the move from immutable constructor based config to more setters and getters.  

I really appreciate the use of `std::vector.at()` for range safety.

I don't like `unsigned int` as a type --- in general, I think we should just be using `int` and testing signs on construction.  It's too easy to get sign errors with unsigned, and it's not like we care about 31 vs. 32 bits of resolution.

Now that I see it in action, I find the all-operator approach difficult to read.  I'd prefer named methods for readability from the client side, but I assume I'm not going to get my way, so no big deal.  I do like it much much better than the pointers plus tests for 0 and explicit endlines.

I find things like `print_help(info, 0, false);` hard to read, at least out of context.  What's the 0 for?  And should there be an overloaded version with the 0 and/or false built-in?  I see that 0 argument all over the place.  I also see lots of cases where there used to be an `output` argument and there's now a 0 argument --- are all those zeroes supposed to be null pointers?  If so, could we use `NULL` at least?

On the same topic, why `io::write_error_msg(writer, e);` rather than having a method on the writer?

The rest looks fine --- it's mainly replacing output streams with the new writer abstraction and `typename` with `class` and `M` with `Model`.  In general, I think it'd be easier all around if the superficial changes went into their own pull request so there aren't so many potential conflicts and so that the meaningful changes are easier to see.

Most of the rest of this is too hard to review because there are no diffs and thousands of lines of code.  But I think I got the basic gist of the writer refactor, and most of the comments on the other stuff I made the last time you wrote it.

There's going to be a ton of lint hitting the fan on this one, but I'll let cpplint sort that out.  I hope it catches the `if(` and requires `if (` --- otherwise Stan's C++ will look like R.

src/stan/io/stan_csv_reader.hpp
betanalpha: Bob, you�re looking at the wrong branch.  This is how it�s currently implemented
in develop � in feature/issue-1361-writer_callback this is done much more robustly.

betanalpha:
Bob, make sure you�re looking at feature/issue-1361-writer_callback.

> Daniel just commented "bug." in cerr.hpp and I don't see what he means looking at the code.

There�s an extra print statement for each column.

> Why make the flush operation operator()() in base_writer.hpp? That seems very confusing compared to calling it flush().

Because it needed not be a flush, that�s just how it�s implemented in the iostream writers.

> What is is_writing() for? Is that like a logger pattern where we want to avoid calling complex string manipulations if something's not going to get logged? I also don't see it implemented in cerr.hpp, so I don't see how that class could be instantiated.

That�s an artifact of develop, again make sure you�re on the right branch.

> I don't see any reason not to replace writer::cerr and writer::cout with writer::ostream_writer that stores a member variable for the stream to write to. But maybe I missed some difference in behavior.

Yes, these could be combined into one class. 

> Rather than stringstream::clear()? I would prefer to just create a new writer::stringstream unless this thing's going to get used in a tight loop. Also no big deal, but it removes the possible error of forgetting to clear a stream before reuse.
> 
> I can see what stringstream::contents() is for---that wouldn't be called by something taking a generic writer, so that's OK.

These are to facilitate testing.  Really the stringstream implementation is just for testing.

> If the writer::messages class is for different kinds of output, then I'd use a different base class.

Those are carryovers from the previous implementation and are being removed.

> In base_mcmc.hpp, What is base_mcmc::clear_buffers() for? I also don't see what the flush operations do given that they return empty strings. If you really care about efficiency (don't know that it matters), you could make a static empty string and just return it by reference everywhere. Which reminds me that the returns can probably be constant for the strings---otherwise, just returning a static string wouldn't be safe.

Wrong branch � feature/issue-1361-writer_callback removes all streams/writer
dependencies from base_mcmc.

> For msg.str(std::string()); wouldn't it be easier to understand if you just wrote msg << std::endl?

No � msg.str(std::string()) also resets some flags.  Took me a while to figure out
how to do that correctly.

bob-carpenter:
Argh!  I just read the pull request.  Would you mind deleting
the no-longer-relevant pull request and/or branches?

Has it all changed so that I have to go through the whole
thing again?  I spent hours going through the first version
on the pull request and am not so keen to do it all again, but
I will if that's the only alternative.

Why would stream writers flush when you call operator()()?
Shouldn't it just be a no-op if that's the intention?  I
just don't like the conflation of behavior --- it's not that
flushing is intrinsically bad (though it does slow things
down in cases where you aren't writing to the screen --- buffering
is much better for anything that writes to a file).  I don't know
enough about the built-in C++ I/O, though I could look into
how to buffer properly.

If there's a new branch, would it be possible to:

(a) use git move rather than delete/add so I can get the
diffs?

(b) break the pull request down into smaller chunks so
it's easier to get my head around?  There were all kinds of
things like naming changes built into this request that
took up screens and screens I had to wade through.

- Bob

betanalpha:
> Argh! I just read the pull request. Would you mind deleting
> the no-longer-relevant pull request and/or branches?

To quote Han Solo �It�s not my fault!�.  Dan merged develop
into feature/issue-1361-writer_callback, inadvertently adding
back all of the old writer implementations and causing all of
this confusion.

> Has it all changed so that I have to go through the whole
> thing again? I spent hours going through the first version
> on the pull request and am not so keen to do it all again, but
> I will if that's the only alternative.

I cleaned things up dramatically.  In addition to cleaning up
a large number of cpplint errors I simplified all of the stream
writers into one template class.  So now there are only a few
new callbacks to review.

On stan-dev/stan:feature/issue-1361-writer_callback

src/stan/interface_callbacks/writerbase_writer.hpp			
src/stan/interface_callbacks/writernoop_writer.hpp			
src/stan/interface_callbacks/writerstream_writer.hpp
src/stan/interface_callbacks/writerstream_writer_typedefs.hpp

src/test/unit/interface_callbacks/writer/noop_writer_test.cpp	
src/test/unit/interface_callbacks/writer/sstream_writer_test.cpp
(Any more tests needed for different template instantiations?)

Then there are the places were these are used

On stan-dev/stan:feature/issue-1361-writer_callback

src/stan/mcmc/hmc/base_hmc.hpp
src/stan/mcmc/hmc/hamiltonians/dense_e_point.hpp
src/stan/mcmc/hmc/hamiltonians/diag_e_point.hpp
src/stan/mcmc/hmc/hamiltonians/ps_point.hpp
src/stan/mcmc/hmc/hamiltonians/unit_e_point.hpp
src/stan/mcmc/hmc/integrators/base_leapfrog.hpp
src/stan/model/util.hpp
src/stan/services/diagnose/diagnose.hpp
src/stan/services/init/initialize_state.hpp
src/stan/services/io/write_error_msg.hpp
src/stan/services/io/write_iteration.hpp
src/stan/services/io/write_model.hpp
src/stan/services/io/write_stan.hpp
src/stan/services/optimize/do_bfgs_optimize.hpp
src/stan/services/optimize/optimize.hpp
src/stan/services/sample/generate_transitions.hpp
src/stan/services/sample/init_adapt.hpp
src/stan/services/sample/init_windowed_adapt.hpp
src/stan/services/sample/mcmc_writer.hpp
src/stan/services/sample/run_adaptive_sampler.hpp
src/stan/services/sample/run_sampler.hpp
src/stan/services/sample/sample.hpp

On stan-dev/cmdstan:feature/update_to_writer_callback

src/cmdstan/command.hpp

> Why would stream writers flush when you call operator()()?
> Shouldn't it just be a no-op if that's the intention? I
> just don't like the conflation of behavior --- it's not that
> flushing is intrinsically bad (though it does slow things
> down in cases where you aren't writing to the screen --- buffering
> is much better for anything that writes to a file). I don't know
> enough about the built-in C++ I/O, though I could look into
> how to buffer properly.

I�m trying to avoid the overhead of creating an empty string.
The idea is that the interfaces have operator()() to use as they
please and the stream implementation of the callback uses it
to write a newline without creating any temporary strings.

syclik:
My fault on that merge. Do you want to just revert the merge with develop? I didn't realize it would have that effect. (to be fair, it would have happened with a future merge anyway... we need to 1) move quicker on this stuff and 2) create smaller pull requests.)

To be explicit, this branch changes the output format significantly enough for the reader to crash. Of course, we can fix this by being better with the readers. I hacked that together in a weekend and it was when I was still figuring out C++, Eigen, etc, so the code is really bad. I say we rip that out / apart and reimplement it correctly. That'll fix a lot of this.

betanalpha:
> My fault on that merge. Do you want to just revert the merge with develop? I didn't realize it would have that effect. (to be fair, it would have happened with a future merge anyway... we need to 1) move quicker on this stuff and 2) create smaller pull requests.)
> 
I spent a good chunk of my plane flight back putting everything back
into a reasonable state.  The branches stan-dev/stan:feature/issue-1361-writer_callback
and stan-dev/cmdstan:feature/update_to_writer_callback are now up to date with develop,
except for the variational inference which for some reason I could not get to compile.
> To be explicit, this branch changes the output format significantly enough for the reader to crash. Of course, we can fix this by being better with the readers. I hacked that together in a weekend and it was when I was still figuring out C++, Eigen, etc, so the code is really bad. I say we rip that out / apart and reimplement it correctly. That'll fix a lot of this.
> 
Was this before or after the merge?  The new writers perturbed only three parts
of the CmdStan output — the adaptation step size and metric, and timing — 
and I had originally updated the reader to be compatible with these changes.

I agree that we should rewrite it, but I’d hold off until we figure out the arguments/
configuration and we have a well-defiend CmdStan output spec.

bob-carpenter:
As I've said, I'm OK with the design.  Let me know if you
want me to look at details.

- Bob

syclik:
Jenkins, retest this please.

bob-carpenter:
@betanalpha and @syclik  This isn't active any more, is it?  If not, could one of you close it?  Thanks.

PR #1553, opened 2015-07-18
Removing a lot of CppLint warnings

syclik:
#### Summary:

Removes a lot of CppLint warnings.
There are a handful that take a little more care that should come in a different pull request.

#### Intended Effect:

Clean up warnings.

#### How to Verify:

All tests should still run without any issues.

#### Side Effects:

None.

#### Documentation:

None.

#### Reviewer Suggestions: 

Bob. If you don't like the way I broke apart lines in `src/stan/lang/generator.hpp`, just revert `e445439081c`. (`> git revert e445439081c`) I left all the changes to that file as one commit.

Almost everything is whitespace related. There were a few includes and extraneous semicolons.

### DIFF
--- a/src/stan/interface/callback/noop_callback.hpp
+++ b/src/stan/interface/callback/noop_callback.hpp
@@ -10,7 +10,7 @@ namespace stan {
       class noop_callback: public callback {
       public:
         noop_callback() {}
-        void operator()() { };
+        void operator()() { }
       };
 
     }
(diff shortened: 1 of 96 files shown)


syclik:
Jenkins, retest this please.

bob-carpenter:
That looks great.  Thanks so much for doing this.  I've been
meaning to tackle the generator.hpp lint, but it always looked
too daunting.

bob-carpenter:
Whoa, down to 69! 

I think we need those implicit ctors for the grammar
interface, but I'll see how many I can get rid of when
I get a chance and then I suppose we have to delint the rest.

I didn't have good luck on Windows trying to convert to fixed
sizes and cover all the possible types without violating
no C++11 rules, so maybe we should just turn that warning off?

- Bob

syclik:
I'm happy to turn off the warnings for implicit ctors and the fixed sizes. If you think the ctors can be fixed, then we can turn off the warnings for the fixed sizes.

PR #1568, opened 2015-07-28
Feature/issue 1567 move math

syclik:
#### Summary:

Moves `lib/stan_math_2.7.0` to `lib/stan_math`.

#### Intended Effect:

Makes it clearer that it is no longer the tagged version of the library.

#### How to Verify:

Check the git log.

#### Side Effects:

None. The build process is the same; the makefile has been updated appropriately.

#### Documentation:

None.

#### Reviewer Suggestions: 

@betanalpha: should take a minute to review.


betanalpha:
Looks good to me.

Issue #1576, opened 2015-07-31
better doc for Wiener distribution


bob-carpenter:
Original question from Andrew Ellis on stan-dev:

<blockquote>
According to the manual (p. 396), Stan returns the first passage time of the accumulation process over the upper boundary only. How do I deal with both decision boundaries, if I cannot pass in lower-boundary response times as negative numbers?
</blockquote>

We need better doc in the manual for the Wiener distribution.  A reference isn't enough

At the very least, there needs to be 

* [ ] a clear articulation of the constraints on the arguments.

and ideally

* [ ] a mathematical description in our doc of what the density is.

It'd be really great if someone can

* [ ] add a chapter (or section) to the manual with an example;  you can look at the diff eq section for the kind of minimal example I'm talking about

And also,

* [ ] thank Andrew Ellis for bringing this up on the mailing list

Finally, here's some more from Andrew Ellis from stan-dev:

<hr />
this seems to work:

lower boundary responses should no longer be coded as negative response times; instead the response times and responses should be passed to stan as data:

```
library(rstan)
library(RWiener)

set.seed(0)
dat <- rwiener(n = 5000, alpha = 1, tau = 0.2, beta = 0.75, delta = 0.5)
standata <- list(y = dat$q,
                 resp = ifelse(dat$resp == "upper", 1, 0),
                 N = length(dat$resp))
```

and in the stan model, `y ~ wiener(a, t, b, v)` if the "upper" response was selected, and `y ~ wiener(a, t, 1-b, -v)` if the "lower" response was selected.

```
data {
  int<lower=0> N;
  real y[N];
  int<lower=0, upper=1> resp[N];
}
parameters {
  real<lower=0> a;
  real<lower=0> t;
  real<lower=0, upper=1> b;
  real v;
}
model {
  v ~ normal(0, 2);
  a ~ gamma(3, 3);
  t ~ gamma(4, 10);
  b ~ beta(5, 5);
  for (n in 1:N) {
      if (resp[n] == 1) {
          y[n] ~ wiener(a, t, b, v);
      }
      else {
          y[n] ~ wiener(a, t, 1-b, -v);
      }
  }
}
```

Full code example:

```
library(rstan)
library(RWiener)


set.seed(0)
dat <- rwiener(n = 5000, alpha = 1, tau = 0.2, beta = 0.75, delta = 0.5)

## stan model ----
wiener_model <- "
data {
    int<lower=0> N;
    real y[N];
[...]

PR #1578, opened 2015-08-02
Improvements in Wiener documentation

randommm:
#### Summary:
Improves the Wiener distribution documentation with the parameter constraints and the PDF description.

#### Intended Effect:
Fix issue #1576

#### How to Verify:
Documentation only.

#### Side Effects:
None.

#### Documentation:
Documentation only.

#### Reviewer Suggestions:
Bob Carpenter


randommm:
Notes:

* My editor ate a lot of whitespaces at the end of line, sorry. You can view the diff without them with:


        git diff --ignore-space-at-eol origin/feature/issue-1576-better-doc-for-Wiener-distribution develop
       
        or
        
        git diff --ignore-space-at-eol origin/feature/issue-1576-better-doc-for-Wiener-distribution 19c7ee66


* I didn't write an example because it isn't really my area of study.

* Andrew Ellis is already in acknowledgements.tex

src/docs/stan-reference/distributions.tex
```diff
@@ -1641,27 +1641,43 @@ \subsubsection{Stan Functions}
...
+\[
+\distro{Wiener}(y|\alpha, \tau, \beta, \delta)
+=
+\frac{\alpha}{(y-\tau)^3/2}
+\exp \left(- \delta \alpha \beta - \frac{\delta^2(y-\tau)}{2}\right)
+\sum_{k = - \infty}^{\infty} (2k + \beta) \varphi \left(\frac{2k + \alpha \beta}{\sqrt{y - \tau}}\right)
+\!.
+\]
+
+where $\varphi(x)$ denotes the standard
```
syclik: @randommm, is $\varphi(x)$ standard notation? I've typically seen it as $\phi(x)$. If you're seeing the standard normal density as $\varphi(x)$ feel free to keep it.

randommm: I just used the way it were in the article of Blurton2012, but you are right $\phi$ is more common.

syclik: thanks. I'll just switch it to $\phi$ and we'll get it in.

syclik:
Jenkins, retest this please.

bob-carpenter:
Jenkins, retest this please.

bob-carpenter:
I don't seem to be able to run the tests and the last ones seem
to have broken on a Jenkins issue.

- Bob

syclik:
Looks ok to me.

syclik:
The documentation isn't fully accurate (I'll fix these):
- y has to be greater than 0 (not greater than or equal).
- y has to be less than tau.

I just went to the code. I can't figure out how this lines up with the doc. 2 questions:
1. is the doc right? (if so, could we not code it this way?)
2. is the code right?

randommm:
I think the code is correct:

We have:

      check_positive(function, "Random variable", y);

And:

      for (size_t i = 0; i < N; i++)
        if (y_vec[i] < tau_vec[i]) {
          lp = stan::math::negative_infinity();
          return lp;
        }

bob-carpenter:
Our general convention for probabilty functions is to raise an
exception.  So rather than the comparisons, this should use
another check function.

And there's never any reason to set a value and then immediately
return it.  It could be written as

  return stan::math::negative_infinity();

But I really think it should be replaced with something that
throws an informative exception --- the idea is to trap errors
as early as possible and let the user know where they arise.

- Bob

syclik:
Sorry, I was having trouble matching the math and the implementation. My
series math has gotten rusty. I just couldn't see how the written out
equation was the same as the code. (It could be... I might just need to
work at it a bit longer.)

src/docs/bibtex/all.bib
```diff
@@ -362,6 +362,17 @@ @article{BleiLafferty:2007
   pages = {17--37}
 }
 
+@article{Blurton2012,
+	author = {Blurton, Steven P. and Kesselmeier, Miriam and Gondan, Matthias},
+	journal = {Journal of Mathematical Psychology},
+	number = {6},
+	pages = {470-475},
+	title = {Fast and accurate calculations for cumulative first-passage time distributions in Wiener diffusion models},
```
bob-carpenter: This needs to be {Wiener} to get it capitalized in the output and the pages need to be -- rather than whatever is there.  You need to check the output.

syclik:
@randommm, can you confirm that the code in the implementation matches the probability distribution function you've described?

randommm:
Hi, I took it from the article "Fast and accurate calculations for cumulative first-passage time distributions in Wiener diffusion models" which is referenced by the author of the code in his article "The RWiener Package: an R Package Providing Distribution Functions for the Wiener Diffusion Model". I sent an email the author Joachim Vandekerckhove right now.

syclik:
@randommm, I didn't follow closely enough to know if the math described there was actually the same math that Joachim implemented. It didn't look like it was. Think it is?

randommm:
He said it is, he just wasn't sure if the pdf (an infinite series) described in the manual is exactly equal or if it is still an approximation even if we take the sum to infinity (of course, computationally it is an approximation, since we cut the series at a finite number). I tried to contact the author he mentioned, but got no answer.

If you wish, I could take away the pdf description in the manual and just point people towards Joachim's article for the description maybe?

syclik:
Ah... I missed that.

My series math isn't strong enough to see that just based on the series.
But... if the series is approximating a closed form solution, shouldn't we
just implement using the closed form? That's where I got confused.

If you're convinced that the doc is right and that the implementation
matches, then we can merge as is. If you're not convinced, we can point
towards the article with the approximation.

randommm:
On 2015-10-20 10:29 PM, Daniel Lee wrote:
> if the series is approximating a closed form solution, shouldn't we
> just implement using the closed form? That's where I got confused.
Maybe the infinite series described in the manual is exactly the pdf. 
But he wasn't sure it is.


It also seems, according to what he said, that if the infinite series is 
not exactly the pdf (but an approximation), than there is no known 
analytical closed form that is exactly the pdf.

syclik:
I thought your doc was closed form? Did I misread it?

randommm:
On 2015-10-20 10:56 PM, Daniel Lee wrote:
> I thought your doc was closed form? Did I misread it?
>
No, it's an infinite series.

syclik:
ah.... ok. I'll take a look again.

randommm:
All these whitespaces got the diff polluted, I just discovered you can ignore them with ?w=1

https://github.com/stan-dev/stan/pull/1578/files?w=1

syclik:
=). Thanks.

syclik:
Ok. I think if I squint, I see how it's the cdf of the normal distribution
in there.

I think separately we should think about simplifying the code to use the
Math library's Phi function to simplify the code.

bob-carpenter:
I know this issue's closed, but in general we want to separate 
a formula for what's being computed and whatever back end 
algorithm's used to compute it.  We usually don't doc the latter.
For instance, you don't see the gamma function doc anywhere,
and it's only defined through an intractable integral.

- Bob

mshvartsman:
Sorry to comment on an old issue, but there's still a mismatch between docs and code, at least in the spirit of @bob-carpenter's comment that one wants to separate what's being computed and what the algorithm is. Here is what is being computed: 

> If alpha in R+, tau in R+, beta in [0,1], delta in R, and dx = delta dy + dW, with absorbing boundaries {0, alpha}, x(0)=alpha * beta and y' = y-tau, then wiener(y | alpha, tau, beta, delta) = inf_y'{y' | x(y') = 0}. 

Feller (1968) gives two series expansions for this first passage time, one of which is in the doc (and lines 194-196 of `wiener_log.hpp`), and the other of which is lines 203-205. Both converge to the same thing, but their error bounds as a function of y and the number of terms used are different, so the algo in the file (which looks like what's given by Navarro & Fuss 2009?) picks which one to use based on whichever requires fewer terms. In light of this, I think the most correct description of what's being computed is the first passage time expression, not the two series. A second option is to include both series, but I think both are a better option than including only one series. 

Additionally, the docs say y \in (0, tau) which also seems incorrect (if y < tau then I believe the density is 0). 

Finally, as a side note, the density as it stands now is a pretty domain-specific parameterization (the domain being the Ratcliff tradition of mathematical psychology). I think that other fields use WFPTs (quant finance? physics) and even not all neuroscientists / psychologists use this formulation, so I wonder if it would be better to parameterize it in a more domain-neutral way and let people do domain-specific parameterizations in the transformed parameters block.

PR #1584, opened 2015-08-13
Bugfix/issue 1547 fun overload

bob-carpenter:
#### Summary:

Disallow overloaded user-defined functions.

#### Intended Effect:

Fix bug whereby they would be parsed, code would be generated, then it would fail in compilation.

#### How to Verify:

New unit tests which fail in previous versions and pass now.

#### Side Effects:

None other than the intended one.

#### Documentation:

Manual updated to match behavior.

#### Reviewer Suggestions:

Anyone.


syclik:
Thanks, Bob.

PR #1586, opened 2015-08-16
Feature/issue 1542 array index print

bob-carpenter:
#### Summary:

Updated generator to allow `user_facing` argument to print indexing in user notation.  Pushed through grammars where expressions are printed.

#### Intended Effect:

Make error messages easier to understand.

#### How to Verify:

Unit tests.

#### Side Effects:

Not unless someone was keying off exact earlier error messages.

#### Documentation:

No.

#### Reviewer Suggestions:

Anyone.


syclik:
looks good.

PR #1589, opened 2015-08-17
feature/issue 1587 increase max dropped gradients

dustinvtran:
#### Summary:
Sets the max number to be 5*grad_samples—the intuition being that we allow 5 dropped gradients per grad_sample, and so that it still scales with the size of the mini batch.

#### Intended Effect:
Allows dropped gradients by default; prior to this with `grad_samples=1` (default) no dropped gradients are allowed.

#### How to Verify:
Run any model that ends up dropping a gradient with the default argument (`grad_samples=1`), c.f., unit test `gradient_warn_test.cpp`.

#### Side Effects:
None

#### Documentation:
None

#### Reviewer Suggestions:
None


akucukelbir:
this is a good catch and a straightforward change.

Issue #1575, opened 2015-07-31
Refactor stochastic gradient things in ADVI


dustinvtran:
It would be worth having something generic for all things related to stochastic gradient descent, to be separated from variational inference itself. E.g., a sgd class to have different stochastic gradient methods available, a learning rate class for testing various learning rates, a subsampling class, etc. This will eventually be necessary as we start working on more research tracks, e.g., [Mandt and Blei (2014)](http://www.cs.columbia.edu/~blei/papers/MandtBlei2014.pdf), [Theis and Hoffman (2015)](http://jmlr.org/proceedings/papers/v37/theis15.pdf), [Tran et al. (2015)](http://arxiv.org/pdf/1506.03159v1.pdf).

It should also be applicable for computing the penalized MLE, so that the optimization interface of Stan also has SGD available for users.

bob-carpenter:
I completely agree.  Same for L-BFGS --- it should be abstracted
from the application as much as possible for re-use.

Is there SGD in ADVI now?  There shouldn't be given that
there's no way for users to run it yet --- things should live
on branches until they're ready to go.

We made the mistake of putting in higher-order autodiff and
discrete sampling infrastructure before either were ready and
it's just been a huge burden.

Thanks for the refs!

- Bob

syclik:
Branch from feature/issue-1751-service-methods if you're going to work on this soon.

PR #1545, opened 2015-07-15
Feature/issue 1544 include gradient and elbo exceptions

dustinvtran:
#### Summary:
Check for NaN's or non-finite numbers when calling methods from the model, specifically the gradient and log probability. From [stan-users](https://groups.google.com/forum/#!topic/stan-users/E7gBBpPtAoc) mailing list.

#### Intended Effect:
If either of the above happens, ADVI should throw a warning and drop that sampled gradient/ELBO evaluation and continue.

#### How to Verify:
Run models by building them, then running them, e.g. `make examples/bernoulli/bernoulli && cd examples/bernoulli && ./bernoulli variational data file=bernoulli.data.R`.

#### Side Effects:
N/A

#### Documentation:
N/A

#### Reviewer Suggestions:
N/A

PR #1590, opened 2015-08-17
feature/advi-adapt-eta

akucukelbir:
## Blueprint of new pull requests (last updated: 10/13)
This will be separated into multiple pull requests as follows:
- #1639 https://github.com/stan-dev/cmdstan/pull/167 Add approximate Bayesian inference to README.md
- #1638 Change argument of eta_adagrad to eta
- #1640 Add more informational messages
- #1684 `feature/advi-convergence`
  - scale `eta` by 1/sqrt(iteration_counter), which makes convergence more reliable.
  - increase minimum size of `elbo_diff` to `2`.
- `feature/advi-internal-changes` `stan-dev/cmdstan feature/advi-internal-changes`
  - follow everything written in https://github.com/stan-dev/stan/pull/1545 by Bob, e.g., 
    - initialize certain variables to empty instead of to zeros
    - remove lots of comments
    - remove doxygen comments on classes
  - refactors the main algorithm into a bunch of general functions
  - changes argument order of advi
  - move `base_family.hpp` inside `families/` directory, following the convention in MCMC code
  - rename a lot of internal variables
  - loops instead of whiles
  - _tests:_ unit tests for everything
- `feature/buffer-diff` (for stan_math)
  - moves `rel_diff` and `circ-median-buffer` to stan_math
  - _tests:_ for the two functions
- `feature/csv-reader`
  - _tests:_ everything previously still works
- `feature/advi-tune`
  - tuning
  - _tests:_ new model testing, tune testing
- `feature/sgd`
  - issue #1575
- other changes to be made: data subsampling, 10 instead of 5 drops, new initialization of `omega` and `L`

#### Summary:
Implements a "warmup" phase for ADVI that autotunes the `eta_adagrad` parameter. If the user defines this parameter, the "warmup" phase is skipped. To avoid confusion with the warmup phase with sampling, we call it the tuning phase.

The default parameter for `eta` is now equal to `0.0`, which means autotune. It uses a sequence of `eta` values to try: 1.0, 0.5, 0.01, 0.05, and 0.01. 

Other features
- Have `eta` default to a certain value (not zero) if warmup is not done, similar to how MCMC does it
- Reports the tuned `eta` in output
- User can specify the # of tuning iterations

#### Intended Effect:
Automatic variational inference is now more automatic.

#### How to Verify:
New unit test for tunings.

#### Side Effects:
None

#### Documentation:
CmdStan and Stan updated.

#### Reviewer Suggestions:
None


dustinvtran:
Since we're making `eta_adagrad` a function of the iteration number, I think we should generalize and call our learning rate something else. We were also running RMSprop before which is not technically a case of AdaGrad. Perhaps at this point it would make sense to start constructing learning rate classes to implement various things, e.g., this bespoke one, RMSprop, and AdaGrad separately.

akucukelbir:
good point. shall we call it just `eta`? or maybe be more explicit and just call it `advi_learning_rate`? 

i'm less keen on constructing many learning rate classes, at least at this moment. we seem to be converging on a learning rate that "just works" (TM).

dustinvtran:
I suggest: `eta_adagrad` -> `eta`, `eta_adagrad_sequence` -> `eta_adagrad`, `robbins_monro_adagrad` -> `robbins_monro`, `params_adagrad` -> `learning_rate`

akucukelbir:
how about something like:

`eta_adagrad` -> `eta`
`eta_adagrad_scaled` -> `eta_scaled`
`eta_adagrad_sequence` -> `eta_sequence`
`robbins_monro_adagrad` -> `stochastic_gradient_ascent`
`params_adagrad` -> `propagated_params`

dustinvtran:
Sure those are fine with me. 

Another note: `eta_adagrad` shouldn't be 0 in the metadata for the output file if the warmup thing is done. I'm not sure how this is dealt with for MCMC.

akucukelbir:
another rare-case bug is that if `eval_elbo` is too large, it becomes (naturally) quite difficult to assess convergence. perhaps we want to limit it, or at least compare it to `max_iterations` and intervene if necessary?

dustinvtran:
I called it the wrong thing; the problem is with `max_iterations` and `tuning_iterations`, not `max_iterations` and `eval_ELBO`. But yeah, we need some constraint where `tuning_iterations < max_iterations`. I put it on the todolist

akucukelbir:
i don't think the `elbo_old` thing does what we want it to do. 

if we're only going to use a single value, then we need to set it immediately after taking the first step for each value of `eta` that we're trying out in the tuning phase.

re the typo: good catch!

dustinvtran:
1863ead implements checks for two things: (1) divergence, comparing `elbo` to `elbo_init`, where `elbo_init=calc_ELBO(variational_init)`; (2) it goes down the list of eta's and compares `elbo` to `elbo_best`, where `elbo_best` stores the current best ELBO. 

We can discuss tomorrow in person whether we should do (2) or not. At the moment, I'm a bit worried with the fact that we're storing so many copies of `Q`: one for `variational`, one for `elbo_grad`, one for `params_prop`, and now two more for `variational_init` and `variational_best`.

bob-carpenter:
What is Q?

- Bob

dustinvtran:
`Q` is the templated class of the variational distribution. Each `Q` object stores a set of parameters which parametrize the variational distribution. It turns out that to do efficient parameter updates, we need to use this class for storing parameters in general and which aren't necessarily used as a variational distribution, e.g., for storing gradients over the variational parameters (`elbo_grad`), learning rate parameters (`params_prop`) and initial parameter values (`variational_init`).

bob-carpenter:
Storing five sets of parameters should be no big deal --- it's
only 8 bytes per parameter!  I don't know enough about the code
to know why you need to store them in an object rather than
in a vector, but there's not much overhead.  The alternative is
to store them outside of a class and pass them by reference to
a function.  (Sorry if this is all obvious --- I never know how much
coding people know.)

- Bob

dustinvtran:
Right, the overhead is negligible in the scheme of things. Our ambitions for ADVI is to store as little which scales with the number of latent variables as possible. Whether there's bigger problems, like storing the full data and/or the model in Stan at the moment, is another concern.

I've gotten it down to 4 copies right now, which is good (the starting set of parameters, the actual variational distribution, the object for storing the gradients, and the object for storing the learning rate parameters) without much drop in speed.

bob-carpenter:
When you say "without much drop in speed", what % of memory did
you save and what was the slowdown?  I wouldn't think, say, a 10%
drop in memory would be worth a 10% drop in speed, but maybe I'm
thinking about this the wrong way.

If you really need to scale (and even 10 copies of 10M parameters is
going to be under 1GB), you'll probably have to think
about moving out of memory and using something like memory
mapping.  But we're going to be talking about somehting like a factor of 2,
because we need roughly 32 bytes per parameter plus 40 bytes per unary
expression and 48 bytes per binary operator expression for autodiff. 
So it seems reducing from 5 to 4 copies of the parameters isn't going
to save more than a few % in overall memory required.  So you're usually down
for at least the equivalent of 10 copies of the parameters.  

Stan itself requires O(log M) copies of the parameters for NUTS, where
M is the tree depth.

So how big in terms of number of parameters do you think you can
make this?  If the parameters are associated with data items (like
in a large IRT model or a huge hierarchical model or even LDA), then
there probably needs to be a different approach that swaps them in
and out of memory along with the data.

- Bob

src/stan/services/arguments/arg_variational_eta.hpp
```diff
@@ -0,0 +1,27 @@
...
+        _name = "eta";
+        _description = "Stepsize scaling parameter for variational inference";
+        _validity = "0 < eta <= 1.0";
+        _default = "automatically tuned";
+        _default_value = "automatically tuned";
+        _constrained = false;
+        _good_value = "good";
+        _value = _default_value;
+      }
+      // bool is_valid(double value) { return value >= 0 && value <= 1.0; }
```
syclik: either:
- remove this comment or
- put this function back in.

src/stan/services/variational/print_progress.hpp
```diff
@@ -0,0 +1,41 @@
...
+#include <stan/services/io/do_print.hpp>
+#include <cmath>
+#include <iomanip>
+#include <iostream>
+#include <string>
+
+namespace stan {
+  namespace services {
+    namespace variational {
+
```
syclik: add function-level C++ doxygen documentation

src/stan/services/variational/print_progress.hpp
```diff
@@ -0,0 +1,41 @@
...
+      void print_progress(const int m,
```
syclik: I'm channeling @bob-carpenter. The arguments for `int`, `bool`, and other primitive types don't need `const`. Just replace `const int m` with `int m`.

src/stan/services/variational/print_progress.hpp
```diff
@@ -0,0 +1,41 @@
...
+                          const int start,
+                          const int finish,
+                          const int refresh,
+                          const bool tune,
+                          const std::string prefix,
```
syclik: change this to a constant reference. `const std::string& prefix`.

src/stan/services/variational/print_progress.hpp
```diff
@@ -0,0 +1,41 @@
...
+                          const std::string suffix,
+                          std::ostream& o) {
+        int it_print_width = std::ceil(std::log10(static_cast<double>(finish)));
```
syclik: Add a unit test for this function. Since there's no error checking here and no doc, what are valid values of `start`, `finish`, `refresh`, `tune`, etc. that should be tested?

src/stan/services/variational/print_progress.hpp
dustinvtran: As an FYI, I wrote this function while changing as few lines as necessary from the equivalent `print_progress()` function in `src/stan/services/mcmc/print_progress.hpp`. This is because that function should really be refactored and this is a short term solution until all such things get generalized. So I recommend I/Alp make your suggested changes in both MCMC and variational's `print_progress()` functions or we don't touch it.

src/stan/variational/advi.hpp
```diff
@@ -150,18 +150,143 @@ namespace stan {
       }
 
       /**
-       * Runs stochastic gradient ascent with Adagrad.
+       * Adaptively set hyperparameters for ADVI.
+       *
+       * @param variational variational distribution
+       * @param tuning_iter number of tuning iterations
```
syclik: what's valid for `tuning_iter`? should the function check it?

src/stan/variational/advi.hpp
```diff
@@ -150,18 +150,143 @@ namespace stan {
...
+       * @return optimal value of eta in a coarse grid
```
syclik: what does coarse mean?

src/stan/variational/advi.hpp
```diff
@@ -285,26 +414,48 @@ namespace stan {
...
 
-          ++iter_counter;
+          ++iter_main;
         }
       }
 
       /**
        * Runs the algorithm and writes to output.
        *
+       * @param  eta            eta parameter for stepsize scaling
```
syclik: valid values of these things?

src/stan/variational/advi.hpp
```diff
@@ -285,26 +414,48 @@ namespace stan {
...
        * @param  tol_rel_obj    relative tolerance parameter for convergence
        * @param  max_iterations max number of iterations to run algorithm
+       * @param  tuning_iter    number of iterations for hyperparameter tuning
        */
-      int run(double tol_rel_obj, int max_iterations) const {
+      int run(std::string eta, double tol_rel_obj,
```
syclik: first argument should be `const std::string& eta`

src/stan/variational/advi.hpp
```diff
@@ -334,6 +520,14 @@ namespace stan {
 
         // draw more samples from posterior and write on subsequent lines
         if (out_stream_) {
```
syclik: I feel like I'm missing something: why is there an `out_stream_` and `print_stream_` here for this if statement? I could understand one or the other.

bob-carpenter: This is all going to get refactored, so don't sweat it too much at this point.
Usually when we have two streams, one is for output to a CSV file and one is
printing from the model and exceptions like warnings that pop up.

src/stan/services/variational/print_progress.hpp
syclik: I'd suggest making changes to this one and opening up a new issue if you think it exists elsewhere. We've got to move our code to getting better, not just letting things slide because other parts are broken.

src/stan/variational/advi.hpp
```diff
@@ -150,18 +150,143 @@ namespace stan {
...
+
+          for (iter_tune = 1; iter_tune <= tuning_iter; ++iter_tune) {
+            m = (eta_sequence_size - eta_sequence.size() - 1) *
+              tuning_iter + iter_tune; // # of total tuning iterations
+            stan::services::variational::print_progress(
+              m, 0, tuning_iter*eta_sequence_size,
+              tuning_iter, true, "", "", *print_stream_);
+
+            // Compute gradient of ELBO
+            calc_ELBO_grad(variational, elbo_grad);
```
syclik: question: is this the right pattern? what does `calc_ELBO_grad` actually calculate? it doesn't return anything, so I'm not sure what it does.
(no need to take any action... just want to know.)

src/stan/variational/advi.hpp
```diff
@@ -150,18 +150,143 @@ namespace stan {
...
+
+        int iter_tune;
+        bool do_more_tuning = true;
+        while (do_more_tuning) {
+          // Try next eta
+          eta = eta_sequence.front();
+          eta_sequence.pop();
+
+          for (iter_tune = 1; iter_tune <= tuning_iter; ++iter_tune) {
+            m = (eta_sequence_size - eta_sequence.size() - 1) *
```
syclik: move the `*` to the next line. So it looks like:
```
m = (eta_sequence_size - eta_sequence.size() - 1)
       * tuning_iter + iter_tune;
```

src/stan/variational/advi.hpp
```diff
@@ -150,18 +150,143 @@ namespace stan {
...
+              tuning_iter, true, "", "", *print_stream_);
+
+            // Compute gradient of ELBO
+            calc_ELBO_grad(variational, elbo_grad);
+
+            // Update learning rate parameters
+            if (iter_tune == 1) {
+              params_prop += elbo_grad.square();
+            } else {
+              params_prop = pre_factor * params_prop +
```
syclik: same here... move `+` to next line.

src/stan/variational/advi.hpp
```diff
@@ -150,18 +150,143 @@ namespace stan {
       }
 
       /**
-       * Runs stochastic gradient ascent with Adagrad.
+       * Adaptively set hyperparameters for ADVI.
+       *
+       * @param variational variational distribution
```
syclik: is the `variational` parameter actually modified through this method? If so, maybe mark it as both an in and out parameter in doxygen:
`@param[in, out]`

src/stan/variational/advi.hpp
syclik: nm... I get it.

src/stan/variational/advi.hpp
```diff
@@ -150,18 +150,143 @@ namespace stan {
...
+            // Reset
+            params_prop = Q(model_.num_params_r());
+          }
+          variational = Q(cont_params_);
+        }
+        return eta_best;
+      }
+
+      /**
+       * Runs stochastic gradient ascent with an adaptive stepsize.
```
syclik: can you describe the input and output more? it's hard to tell that the thing that this function is updating is the `variational` parameter. What's expected to happen at the end of this function?

src/stan/variational/advi.hpp
```diff
@@ -285,26 +414,48 @@ namespace stan {
...
         }
       }
 
       /**
        * Runs the algorithm and writes to output.
        *
+       * @param  eta            eta parameter for stepsize scaling
        * @param  tol_rel_obj    relative tolerance parameter for convergence
        * @param  max_iterations max number of iterations to run algorithm
+       * @param  tuning_iter    number of iterations for hyperparameter tuning
```
syclik: what does this function return?

src/stan/variational/advi.hpp
syclik: There's a double conditional statement here... I think the `if (out_stream_)` should be removed. @akucukelbir / @dustinvtran agree?

src/stan/variational/advi.hpp
```diff
@@ -150,18 +150,143 @@ namespace stan {
...
 
       /**
-       * Runs stochastic gradient ascent with Adagrad.
+       * Adaptively set hyperparameters for ADVI.
+       *
+       * @param variational variational distribution
+       * @param tuning_iter number of tuning iterations
+       * @return optimal value of eta in a coarse grid
+       */
+      double tune(Q& variational, int tuning_iter) const {
```
syclik: can you add a unit test for this function? I think it's clear now that ADVI is solidly in the code base, so it should be treated as such.

In the test, I'm looking for
1. a way to instantiate the `tune` function easily.
2. that it actually tests the behavior. I think `variational` changes, so actually test that it changes to the right thing.
3. boundary conditions? perhaps `tuning_iter` at 0?

src/stan/variational/advi.hpp
```diff
@@ -150,18 +150,143 @@ namespace stan {
...
        *
        * @param  variational    variational distribution
+       * @param  eta            eta parameter for stepsize scaling
        * @param  tol_rel_obj    relative tolerance parameter for convergence
        * @param  max_iterations max number of iterations to run algorithm
        */
-      void robbins_monro_adagrad(Q& variational,
-                                 double tol_rel_obj,
-                                 int max_iterations) const {
+      void stochastic_gradient_ascent(Q& variational,
```
syclik: add unit test for `stochastic_gradient_ascent`

src/stan/variational/advi.hpp
syclik: it looks hard to write a test for `run`, but if you can figure it out, that'd be great. If not, I think it'll be ok.

src/stan/variational/advi.hpp
```diff
@@ -374,13 +553,29 @@ namespace stan {
...
       int n_monte_carlo_grad_;
       int n_monte_carlo_elbo_;
-      double eta_adagrad_;
       int eval_elbo_;
       int n_posterior_samples_;
       std::ostream* print_stream_;
       std::ostream* out_stream_;
       std::ostream* diag_stream_;
 
+      double circ_buff_median_(const boost::circular_buffer<double>& cb) const {
```
syclik: It looks like this function is duplicated. Can you:
1. consolidate the cut and paste.
2. test it as a unit test? what happens when cb is length 0?

src/stan/variational/advi.hpp
```diff
@@ -374,13 +553,29 @@ namespace stan {
...
+              i != cb.end(); ++i) {
+          v.push_back(*i);
+        }
+
+        size_t n = v.size() / 2;
+        std::nth_element(v.begin(), v.begin()+n, v.end());
+        return v[n];
+      }
+
+      double rel_difference_(double prev, double curr) const {
```
syclik: do we really not have this function implemented already?

syclik: I can't find it. can you add tests for this? obviously, there are issues when `prev` is 0. what do you expect to happen there? what if both `prev` and `curr` are 0?

src/test/unit/variational/eta_adapt_test.cpp
```diff
@@ -0,0 +1,51 @@
...
+                                                     100,
+                                                     1,
+                                                     &print_stream_,
+                                                     &output_stream_,
+                                                     &diagnostic_stream_);
+
+  // ADVI should choose a relatively small eta and
+  // should converge in < 2e4 iterations
+  test_advi.run("automatically tuned", 0.01, 2e4, 50);
+}
```
syclik: can you add a test here? this runs, but you don't check to see if it:
1. runs as you expect
2. fails as you expect
3. does anything besides for running

If the test you end up writing is too brittle, it might be because the functions aren't broken apart appropriately where you can test the pieces easily. If you need help, talk to me.

src/test/unit/variational/gradient_warn_test.cpp
```diff
@@ -65,11 +65,11 @@ class advi_test : public ::testing::Test {
 
 
 TEST_F(advi_test, gradient_warn_meanfield) {
-  EXPECT_EQ(0, advi_meanfield_->run(0.01, 10000));
+  EXPECT_EQ(0, advi_meanfield_->run("0.1", 0.01, 10000, 50));
   SUCCEED() << "expecting it to compile and run without problems";
```
syclik: is there anything else you can / want to test? This just verifies that it runs.

src/test/unit/variational/hier_logistic_cp_test.cpp
```diff
@@ -29,15 +37,14 @@ TEST(advi_test, hier_logistic_cp_constraint_meanfield) {
...
                                                      1,
-                                                     &std::cout,
-                                                     &std::cout,
-                                                     &std::cout);
+                                                     &print_stream_,
+                                                     &output_stream_,
+                                                     &diagnostic_stream_);
 
-  test_advi.run(1,2e4);
+  test_advi.run("0.1", 0.1, 1e4, 50);
```
syclik: same here... make sure it actually does something.

This really allows someone to come in and mess up the `run` function and none of these tests will tell the person that something bad has happened.

src/test/unit/variational/hier_logistic_test.cpp
```diff
@@ -89,7 +89,7 @@ class advi_test : public ::testing::Test {
 };
 
 TEST_F(advi_test, hier_logistic_constraint_meanfield) {
-  EXPECT_EQ(0, advi_->run(1, 2e4));
+  EXPECT_EQ(0, advi_->run("0.01", 1, 2e4, 50));
   SUCCEED() << "expecting it to compile and run without problems";
   EXPECT_NE("", output_stream_.str());
```
syclik: beyond just this `EXPECT_NE`, are you expecting to see anything else? 

You might want to test the lower level function for other behavior and leave this as an `EXPECT_NE`, but I don't see any of those test around.

src/stan/services/arguments/arg_variational_eta.hpp
akucukelbir: removed.

src/stan/variational/advi.hpp
akucukelbir: should be positive. done.

src/stan/variational/advi.hpp
akucukelbir: i clarified this.

src/stan/variational/advi.hpp
akucukelbir: these are checked in `stochastic_gradient_ascent` which makes more sense.

src/stan/variational/advi.hpp
akucukelbir: i added `const`. 

testing for `run`... yeah. we can talk.

src/stan/variational/advi.hpp
akucukelbir: i don't think so. we need to check for `out_stream_` otherwise there's no csv file to write posterior samples to. we need to check for `print_stream_` before spitting things out to the terminal.

src/stan/variational/advi.hpp
akucukelbir: roger.

src/stan/variational/advi.hpp
akucukelbir: done.

src/stan/variational/advi.hpp
akucukelbir: done.

src/stan/variational/advi.hpp
akucukelbir: done.

src/stan/variational/advi.hpp
akucukelbir: done.

src/stan/variational/advi.hpp
akucukelbir: updated.

src/stan/services/variational/print_progress.hpp
akucukelbir: done.

src/stan/services/variational/print_progress.hpp
akucukelbir: we'll fix MCMC separately. done.

src/stan/services/variational/print_progress.hpp
akucukelbir: done.

src/stan/services/variational/print_progress.hpp
akucukelbir: i added error checking. this function will likely go away with the writer refactor anyhow.

src/stan/variational/advi.hpp
bob-carpenter: Also, don't be shy about replacing this

```
bool flag = true;
while (flag) {
  if (cond) {
    ...
    flag = false;
  } else {
    ...
    flag = true;
  }
}
```

with

```
while (true) {
  if (cond) {
     ...
     break;
  } else {
     ...
  }
}
```

I find code with flags very hard to follow because you need to track 
through the whole function to figure out where and how the flag gets reset.
Here, I think there's probably also needless call to the constructor
at the end when the flag is set to false before exit.

Also, because params_prop is only used in that loop, you should declare
it close to the loop.  The general style rule is to declare variables as
close to where they are used as possible.  

Also, this sequence:

```

        eta_sequence.push(1.00);
        eta_sequence.push(0.50);  // 1/2 of previous value
        eta_sequence.push(0.10);  // 1/5 of previous value
        eta_sequence.push(0.05);  // 1/2 of previous
        eta_sequence.push(0.01);  // 1/5 of previous
```

isn't evenly spaced on the log scale, so I have no idea why you
chose those numbers other than they match American currency denominations!
(So I'm missing the 0.25!).  


There's also this

```

        double eta;
        double eta_best;
```

which is bad for several reasons.  There are no initial values and
it's not declared by where it's used and it's not used in the scope
where it's declared.  Instead, you should replace this

```

          eta = eta_sequence.front();
```

with

```
          double eta = eta_sequence.front();
```

Also, I don't like that whole loop.  You should just iterate through
eta values declared as a static const array, or as a vector if you must
rather than popping a sequence.  It makes it clearer what happens when you run 
out of values.  So this:

```
        while (do_more_tuning) {
          // Try next eta
          eta = eta_sequence.front();
          eta_sequence.pop();
```

should be rewritten as

```
  for (size_t i = 0; i < eta_sequence.size(); ++i)
    eta = eta_sequence[i];
```

or use iterators if you're comfortable with that idiom.  I'm not clear
what the intended behavior is in the current code for when you pop the last
value off the eta_sequence.

- Bob

src/stan/variational/advi.hpp
bob-carpenter: I'd like to see it pulled out as its own standalone tested function
that doesn't depend on the variational algorithm.  And why isn't Q a const
here?  Are you resetting it in the algorithm?

src/stan/variational/advi.hpp
bob-carpenter: If not, it's not critical here --- but do create a pull request to move
this into the math lib.

src/stan/variational/advi.hpp
bob-carpenter: This should be `const std::string& eta` --- if you don't
make it a reference, it allocates more memory and copies on
every call.

src/stan/services/variational/print_progress.hpp
bob-carpenter: Just to clutter up the issues more, thanks for taking care of all
this.  It makes a big difference to those of us in the maintenance
department!

akucukelbir: maintenance makes the world go around. my pleasure.

src/stan/variational/advi.hpp
akucukelbir: 1. where is it duplicated?
2. let's move these functions to stan-math in a separate pull request after merging.

src/stan/variational/advi.hpp
akucukelbir: let's move these functions to stan-math in a separate pull request after merging

src/stan/variational/advi.hpp
akucukelbir: good catch. thanks.

src/stan/variational/advi.hpp
akucukelbir: yes. i modified the doxygen to reflect this. 

i'll think about how to write a unit test for this.

src/stan/variational/advi.hpp
bob-carpenter: It's better to allocate ahead of time and assign if you know the size
rather than repeatedly calling push_back.

I didn't know about that n_th_element function --- nifty!

- Bob

src/stan/variational/advi.hpp
akucukelbir: using the `break` paradigm is messier because we need to reset the variational distribution Q before trying an element of `eta_sequence` and before returning `eta_best`.

a queue made sense to me because we might want to chance `eta_sequence` down the road (add/remove values) and it makes sense that we are working our way down a queue in the procedure. i could change this if you feel strongly though.

i've implemented the other changes you suggested.

bob-carpenter: > On Aug 26, 2015, at 1:26 PM, Alp Kucukelbir <notifications@github.com> wrote:
> 
> In src/stan/variational/advi.hpp:
> 
> > +                const char* msg1 = "failed. Your model may be either "
> > +                  "severely ill-conditioned or misspecified.";
> > +                stan::math::domain_error(function, name, "", msg1);
> > +                return 0;
> > +              }
> > +            }
> > +            // Reset
> > +            params_prop = Q(model_.num_params_r());
> > +          }
> > +          variational = Q(cont_params_);
> > +        }
> > +        return eta_best;
> > +      }
> > +
> > +      /**
> > +       * Runs stochastic gradient ascent with an adaptive stepsize.
> 
> using the break paradigm is messier because we need to reset the variational distribution Q before trying an element of eta_sequence and before returning eta_best.

OK --- I think I misread the end of the loop --- it's hard when the
loops are this nested.  A higher-level comment would be to try to
find subroutines to call so the loop structure's clearer.  Then they
can be given names and the overall structure can be read directly
rather than inferred.  I'm not saying you have to do that here, just
that it's a good idea in general for code readability.

Also, that last 

  } else {
    if (cond) {
      ...
    } else {
      if (cond2) {
        ...
      } else {
        ...
      }
  }

can (and probably should) be simplified to

  } else if (cond) {
    ...
  } else if (cond2) {
    ...
  } else {
    ...
  }

Programs are always easier to read with less nesting. 

> a queue made sense to me because we might want to chance eta_sequence down the road (add/remove values) and it makes sense that we are working our way down a queue in the procedure. i could change this if you feel strongly though.

It's no big deal --- it's much more efficient to not use mutable
containers (they require memory allocations on the heap), but I'm sure
it's no big deal in this algorithmic context. 

> i've implemented the other changes you suggested.

Thanks.

- Bob

dustinvtran: On why we're calling `variational = Q(cont_params_);` even when the flag is turned off: I thought about this carefully and unfortunately I couldn't find out a better way to write this messy control flow. The last constructor call is necessary because we need to reset `variational` after having found the tuned parameter. The object is recarried over to then do the variational inference in `stochastic_gradient_ascent()`.

src/stan/variational/advi.hpp
```diff
@@ -285,26 +420,49 @@ namespace stan {
...
+                << "converged."
+                << std::endl
+                << "Values from this variational approximation are not "
+                << "guaranteed to be meaningful."
+                << std::endl;
             do_more_iterations = false;
           }
 
-          ++iter_counter;
+          ++iter_main;
```
bob-carpenter: You can just make this whole thing a for loop and the structure will be clearer:

```
for (int iter_main = 1; do_more_iterations; ++iter_main) {
  ...
}
```

Two general principles:  keep variables defined in the scope in which they're used and use standard idioms like for loops when it's a for-loop structure.

We can go through and refactor some of this code later.  I don't want to postpone this merge indefinitely if the code's well tested.

akucukelbir: done.

akucukelbir:
@syclik i can't figure out how to test a protected member. can you help? the file is `stan/src/test/unit/variational/protected_functions_test.cpp`. i was trying to follow what they suggest at (https://code.google.com/p/googletest/wiki/FAQ#How_do_I_test_private_class_members_without_writing_FRIEND_TEST)

bob-carpenter:
You can write a subclass that makes it accessible through a method.
Or you can not worry about making it protected.  The only reason to
make it protected instead of private would be if a subclass needed to
use it directly (rather than through getter or setter methods, that might,
say, be set up to check validity).

- Bob

bgoodri:
In CmdStan, the help-all output currently says
```
  eta=<string>
    Stepsize scaling parameter for variational inference
    Valid values: 0 < eta <= 1.0
    Defaults to automatically tuned
```
Should eta be a `double` rather than a `string`? Either way, I think it should say 0 <= eta <= 1.0 and the default is 0, which implies autotuning.

bgoodri:
Also, the code and doc require `tuning_iter > 0`. Would it make sense to permit `tuning_iter == 0` to disable autotuning? I guess autotuning can be disabled by specifying `eta`, but that seems a little indirect.

dustinvtran:
@bgoodri Setting 0 as specifying autotuning was the original procedure but we thought that may be confusing to users. The tuned parameter if performing autotuning is also placed in the output.

Ideally we'd like to have it follow the adaptation hierarchy that MCMC does. It seems like a lot of that structure is going to be flattened though, so we settled with this solution. Along those lines `tuning_iter` is a hierarchical argument that (should) only exist if one does autotuning.

bgoodri:
OK, so to call the thing from R I should pass the character string `"automatically tuned"` by default or the string representation of a double on (0,1]? But in the latter case, it is still okay to pass the default integer value of 50 for `tuning_iter`?

akucukelbir:
@bgoodri yes, please pass in the string `"automatically tuned"` by default. if a string representation of a double is passed in, automatic tuning will get bypassed. in either case, it is okay to pass the default integer value for `tuning_iter`

bgoodri:
@bob-carpenter @maverickg Can you help me? I have a new feature/adapt-eta branch in the rstan repo that tries to sync up with the advi-adapt-eta branch in the stan repo. But now that `eta` (formerly known as `eta_adagrad` and formerly a double) is a string, it is causing problems when calling it from R. When it tries to build a model (even if it is not using ADVI), it says 
```
In file included from /tmp/RtmpjMl9sQ/Rinsta6119486b07/rstan/include/rstan/rstaninc.hpp:3:
In file included from /tmp/RtmpjMl9sQ/Rinsta6119486b07/rstan/include/rstan/stan_fit.hpp:84:
/tmp/RtmpjMl9sQ/Rinsta6119486b07/rstan/include/rstan/stan_args.hpp:274:5: error: call to implicitly-deleted default constructor of 'union (anonymous union at /tmp/RtmpjMl9sQ/Rinsta6119486b07/rstan/include/rstan/stan_args.hpp:104:5)'
    stan_args(const Rcpp::List& in) : init_list(R_NilValue) {
    ^
/tmp/RtmpjMl9sQ/Rinsta6119486b07/rstan/include/rstan/stan_args.hpp:151:9: note: default constructor of '' is implicitly deleted because variant field 'variational' has a non-trivial default constructor
      } variational;
        ^
```
I am pretty sure that the fact that `eta` is now a std::string is what makes the default constructor non-trivial because I don't get that build error if I comment `eta` out of the variational union. But if I comment `eta` out of the variational union, then I obviously get build errors when trying to pass `eta` to the `.run()` method for class `advi`.

bob-carpenter:
Can't you just add in a default constructor
by hand?  Do the unary ones need to be implicit?

I think Michael might be able to help because this is
in the args thing.

- Bob

bgoodri:
Maybe. We are actually talking about the `stan_args` class at
https://github.com/stan-dev/rstan/blob/feature/adapt-eta/rstan/rstan/inst/include/rstan/stan_args.hpp#L90
which holds everything prior to actually calling anything in Stan. Would writing a default constructor entail specifying values for all possible arguments? In the other cases where there is a string argument, the number of possible strings is finite so there is an `enum` for each of them.

maverickg:
I will take a look tonight.  I think we can fix it.  And it doesn't need to
be a string when it's passed from R to c++ code.

bgoodri:
That is what I was thinking: Pass eta as a double and if it is 0.0, then
pass "automatically tuned" to the first argument of the `.run()` method and
if it is > 0.0, static_cast<std::string> it as it is passed to the `.run()`
method.

src/stan/io/stan_csv_reader.hpp
```diff
@@ -194,6 +194,11 @@ namespace stan {
         }
         ss.seekg(std::ios_base::beg);
 
+        // All the code below is specific to NUTS
+        // which has at least 4 lines of adaptation output
+        if (lines < 4)
```
syclik: @akucukelbir, this needs a test. I didn't see a change to the `stan_csv_reader_test.cpp` and I didn't see an appropriate test anywhere else, but could have missed it.

src/stan/variational/advi.hpp
```diff
@@ -85,8 +87,8 @@ namespace stan {
        * the variational distribution and then evaluating the log joint,
        * adjusted by the entropy term of the variational distribution.
        *
-       * @tparam Q class of variational distribution
-       * @return   evidence lower bound (elbo)
+       * @param variational variational distribution
+       * @return evidence lower bound
```
syclik: add `@throw` statement and under what condition.

syclik: that's for the doxygen doc.

src/stan/services/variational/print_progress.hpp
```diff
@@ -0,0 +1,70 @@
...
+#include <string>
+
+namespace stan {
+  namespace services {
+    namespace variational {
+
+      /**
+       * Helper function for printing progress for variational inference
+       *
+       * @param m       total number of iterations
```
syclik: document what's valid for these things. Or perhaps what's invalid.

src/stan/variational/advi.hpp
syclik: Looks like the duplicate was removed.
Still needs tests.

syclik:
@akucukelbir and @dustinvtran, you've got to start testing. This helps with multiple things:
1. Design. Testable code is a lot easier to work with.
2. Maintainability. Having a few tests in place for everything makes it so much easier to approach the code if it ever needs to be fixed. Based on how this is going, it looks like it will need to be handled. If you're not going to be around to fix it immediately, we need it to be up to standards where someone else can approach it.
3. Peace of mind. Code that doesn't have a single test calling it scares me. It tells me that the person that wrote it didn't even bother to run it once. I don't expect that to be ok releasing to the wild.
4. Making sure no one breaks your code. If this were a perfect world, we wouldn't have to worry about interactions with code. But this is an imperfect world where we do what we can. We won't know something is broken until it's too late if there are no tests to make sure something stays correct.

bgoodri:
I got the rstan branch to build by specifying eta as a double on the R side.

bob-carpenter:
It would need to be capped to prevent infinite loops.

In fact, the whole algorithm should be instrumented for
interrupt callbacks, too, if it isn't already.

syclik:
@akucukelbir and @dustinvtran, sorry for taking so long to respond. I'm going to try to break down what needs to happen in order for this work to get into the Stan development branch.

Some general points: 
- everything needs to be tested. At least once. Just show that whatever you've written actually behaves as you expect. It'd be much better if you tested all boundary conditions, etc, but we haven't really required that everywhere, so I'm not going to ask you to be the first to start.
- split apart concerns. In this pull request, there is a change to the output format, changes to argument names, some tuning. Separate that out into different issues and different pull requests. Let's get the easier things to reason through into the code base sooner. It's also easier to determine the side-effects of changes to code when they are split apart.

1. New issue, pull request, and tests for the change to `src/stan/io/stan_csv_reader.hpp`. I realize that code is super-brittle, but lots of things hinge on it, so changing functionality should have additional tests. Besides, you've added more complexity to it and there should be some reason for adding that complexity.
2. New issue and pull request for changing the names of arguments. I'm talking about moving `src/stan/services/arguments/arg_variational_eta.hpp` to `src/stan/services/arguments/arg_variational_eta_adagrad.hpp`. Sorry, but you can't just change things because you want to anymore. It's out in the open with users dependent on it. This breaks backwards compatibility, so depending on how we treat this, this may need to sit on its own until 3.0. (By the way, this change alone disqualifies this pull request from getting into develop now.) We probably should have forced you to think about the impact of naming when this first got in, but we were all excited and let it slide before we understood the consequences. Now we know. So... the right thing to do here is to split this into a new issue with corresponding pull request that's independent of the behavioral change.
3. New issue and pull request for the tuning. It should just touch `advi.hpp`, a test for `advi.hpp`, and possibly the new argument. It's missing a lot of testing. Do you realize you don't test `tune()`? At all? There's a lot to be done here. Once this is split out from the rest of the non-related changes, we should be able to tell what needs to be done further.
4. @bgoodri has claimed that ADVI isn't stable enough for certain models. I've asked for 5 models with tolerances for the inferences that would show that ADVI is stable enough. @bgoodri, action is on you to provide these models. And I'm going to keep hounding you to provide them until they are out and it's some sort of threshold before we allow algorithms into Stan and released through ALL the interfaces we control (CmdStan, RStan, PyStan). Just beware, I'm going to hold all algorithms to the same standard. This includes optimization and if that doesn't pass the threshold, I'm going to either demand that be removed from RStan or this included. I just wanted you to understand what this condition means. That said, @akucukelbir and @dustinvtran, I think what @bgoodri is asking for is perfectly reasonable, albeit really late. There's still so much work required on these changes that I think he has plenty of time to provide it before you get there.

Hopefully that's clear. If you have questions, ask.

andrewgelman:
Thanks all.  If we can get this working smoothly and giving reasonable answers on Ben’s suite of examples, I think this will be huge.
A

dustinvtran:
Perfect, thanks for all the help guys. Alp and I promise that a little after the AISTATS deadline, which is October 9, we will focus our energies purely on getting this (and initialization) merged into develop.

syclik:
I'll help out too. You're not an island.

bob-carpenter:
I'm closing this as it's no longer a pending pull request (it won't be erased, it just won't show up on my to-do list!)

PR #1592, opened 2015-08-17
Adding more writer functionailty.

syclik:
#### Summary:

Updates the writer to have the same interface as Michael's writer interface.

This should require none of the interfaces to change.

#### Intended Effect:

Expanding the writer capability without changing any existing behavior.

#### How to Verify:

Look at the code. Run the tests.

#### Side Effects:

None.

#### Documentation:

Updated C++ doc.

#### Reviewer Suggestions: 

@betanalpha, this ok for now? The next one will be more substantive.

### DIFF
--- a/src/stan/interface_callbacks/writer/base_writer.hpp
+++ b/src/stan/interface_callbacks/writer/base_writer.hpp
@@ -2,20 +2,147 @@
 #define STAN_INTERFACE_CALLBACKS_WRITER_BASE_WRITER_HPP
 
 #include <string>
+#include <sstream>
 #include <vector>
 
 namespace stan {
   namespace interface_callbacks {
     namespace writer {
 
+      /**
+       * Base writer class.
+       *
+       * This is a abstract base class. The only functions
+       * an implementing class needs to implement are
+       * - void operator()(const std::string& message)
+       * - void operator()()
+       * - bool is_writing()
+       */
       class base_writer {
       public:
+        /**
+         * Virtual destructor.
+         */
         virtual ~base_writer() {}
-        // Can't enforce this method with a pure virtual function
-        // template <class T>
-        // virtual void operator()(const std::vector<T>& x) = 0;
-        virtual void operator()(const std::string x) = 0;
+
+        /**
+         * Writes a key, value pair.
+         *
+         * @param key The key
+         * @param value The value
+         */
+        virtual void operator()(const std::string& key,
+                                double value) {
+          std::stringstream ss;
+          ss << key << " = " << value;
+          this->operator()(ss.str());
+        }
+
+        /**
+         * Writes a key, value pair.
+         *
+         * @param key The key
+         * @param value The value
+         */
+        virtual void operator()(const std::string& key,
+                                const std::string& value) {
+          this->operator()(key + " = " + value);
+        }
+
+        /**
+         * Writes a key and an array of values.
+         *
+         * @param key The key
+         * @param values The values
+         * @param n_values The number of values
+         */
+        virtual void operator()(const std::string& key,
+                                const double* values,
+                                int n_values)  {
+          std::stringstream ss;
+          ss << key << " = ";
+          if (n_values > 0)
+            ss << values[0];
+          for (int n = 1; n < n_values; n++)
+            ss << "," << values[n];
+          this->operator()(ss.str());
+        }
+
+        /**
+         * Writes a key and a 2-d array of values.
+         *
+         * @param key The key
+         * @param values The values stored in row-major order
+         * @param n_rows The number of rows
+         * @param n_cols The number of columns
+         */
+        virtual void operator()(const std::string& key,
+                                const double* values,
+                                int n_rows, int n_cols) {
+          std::stringstream ss;
+          ss << key << " = [";
+          if (n_rows > 0 && n_cols > 0) {
+            for (int i = 0; i < n_rows; i++) {
+              ss << values[i * n_cols];
+              for (int j = 0; j < n_cols; j++)
+                ss << "," << values[i * n_cols + j];
+              if (i != n_rows - 1)
+                ss << std::endl;
+            }
+          }
+          ss << "]";
+          this->operator()(ss.str());
+        }
+
+        /**
+         * Writes a vector of names.
+         *
+         * @param names Names to write
+         */
+        virtual void operator()(const std::vector<std::string>& names) {
+          std::stringstream ss;
+          if (names.size() > 0)
+            ss << names[0];
+          for (std::vector<std::string>::const_iterator it = names.begin()+1;
+               it != names.end(); it++)
+            ss << "," << *it;
+          this->operator()(ss.str());
+        }
+
+        /**
+         * Writes a vector of states.
+         *
+         * @param state State to write
+         */
+        virtual void operator()(const std::vector<double>& state) {
+          std::stringstream ss;
+          if (state.size() > 0)
+            ss << state[0];
+          for (std::vector<double>::const_iterator it = state.begin()+1;
+               it != state.end(); it++)
+            ss << "," << *it;
+          this->operator()(ss.str());
+        }
+
+        /**
+         * Writes a message.
+         *
+         * @param message Message to write
+         */
+        virtual void operator()(const std::string& message) = 0;
+
+        /**
+         * Writes nothing.
+         *
+         * An implementation may choose to treat this as a flush.
+         */
         virtual void operator()() = 0;
+
+        /**
+         * Indicates whether this writer is writing.
+         *
+         * This will be removed.
+         */
         virtual bool is_writing() const = 0;
       };
 
--- a/src/stan/interface_callbacks/writer/base_writer.hpp
+++ b/src/stan/interface_callbacks/writer/base_writer.hpp
@@ -83,7 +83,7 @@ namespace stan {
           if (n_rows > 0 && n_cols > 0) {
             for (int i = 0; i < n_rows; i++) {
               ss << values[i * n_cols];
-              for (int j = 0; j < n_cols; j++)
+              for (int j = 1; j < n_cols; j++)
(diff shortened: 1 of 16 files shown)


betanalpha:
Do not remove is_writing — the check prevents unnecessary flushes
that can substantially slow down I/O.

Similarly, I am very much against this implementation of the base_writer.
base_writer should be an abstract base class with no implementation
(or really a hybrid abstract base class/policy here).  In particular the
stream writer should remain a separate implementation.  I have no idea
why you decided to do this.

The comments are not particularly informative and I don’t see their
utility.

syclik:
Umm... the removal of is_writing is an artifact of your branch. I was happy
leaving it in, but put in work getting to match your pull request. I'll add
it back in.

Some points about the implementation:
- I did this to prevent duplicated code and repeated bugs. If you look at
the original branch there were bugs that were duplicated.
- I'd prefer a simpler interface, but I think you were also opposed to
that. It would have the same effect as this code: simplify the complexity.
- I think we fundamentally disagree about what we prefer. If it were
correct and tested the other way, that would have been fine, but it wasn't,
so I'm bringing it up to a place where I think someone other that you and I
can follow along. If a hypothetical person can't, then this pull request
needs more work.

Regarding documentation, yes, but now there's doxygen doc.

betanalpha:
> Umm... the removal of is_writing is an artifact of your branch. I was happy
> leaving it in, but put in work getting to match your pull request. I'll add
> it back in.

Right, sorry.  I forgot the noon () implementation will get compiled
away and avoid any unnecessary overhead.

> Some points about the implementation:
> - I did this to prevent duplicated code and repeated bugs. If you look at
> the original branch there were bugs that were duplicated.

What bugs?  You never gave these in any detail despite my request.

> - I'd prefer a simpler interface, but I think you were also opposed to
> that. It would have the same effect as this code: simplify the complexity.

What simplification?  Firstly the key/value pair methods were proposed
on stan-dev a long time ago and _nobody_ made any comments against 
them.  The writer is unifying multiple types of output and cannot be
simplified without burdening the interfaces dramatically.

> - I think we fundamentally disagree about what we prefer. If it were
> correct and tested the other way, that would have been fine, but it wasn't,
> so I'm bringing it up to a place where I think someone other that you and I
> can follow along. If a hypothetical person can't, then this pull request
> needs more work.

Again, you need to explain what you mean by incorrect.  The refactor
changes the output per the proposed interface but worked according
to that spec.  All you’ve done is convolved files together and broken the
OO abstraction.  How does this make it any easier for someone to use?

syclik:
On Monday, August 17, 2015, Michael Betancourt <notifications@github.com
<javascript:_e(%7B%7D,'cvml','notifications@github.com');>> wrote:

>
> > Umm... the removal of is_writing is an artifact of your branch. I was
> happy
> > leaving it in, but put in work getting to match your pull request. I'll
> add
> > it back in.
>
> Right, sorry. I forgot the noon () implementation will get compiled
> away and avoid any unnecessary overhead.
>

Is that your final answer? I'm ok with it, just wanted to make sure.



> > Some points about the implementation:
> > - I did this to prevent duplicated code and repeated bugs. If you look at
> > the original branch there were bugs that were duplicated.
>
> What bugs? You never gave these in any detail despite my request.


I felt like I sent you to this line:

https://github.com/stan-dev/stan/pull/1546/files#diff-8ef860bae9dd720a1bce85652e02862eR41

But, it's possible I didn't send it out. The elements aren't printed in the
order you intended. A simple bug, but one that's duplicated in a few places.

(There were a couple more that I can't remember: I just tend to put in
tests and move on; I'm not trying to be critical, just want this to be done
safely.)


>
> > - I'd prefer a simpler interface, but I think you were also opposed to
> > that. It would have the same effect as this code: simplify the
> complexity.
>
> What simplification? Firstly the key/value pair methods were proposed
> on stan-dev a long time ago and _nobody_ made any comments against
> them. The writer is unifying multiple types of output and cannot be
> simplified without burdening the interfaces dramatically.


Personally, I would simplify to an api with either 2 calls or 1 call:
1. (Both) void operator()(const std::string& message)
2. void operator()(const std::string& key, const std::string& value)

The transformations from (double*, int) -> std::string would be handled by
utility functions outside the individual writers. Same for (double*, int,
int) -> std::string. The goal would be to reduce the chance of writer
error.

Of course, I see what you're trying to do and am ok with it if we make it
less error prone by providing a base implementation that is tested. That's
what I tried to do to keep your api as-is because I believe you e touched
on some use cases where this is easier. I can still imagine accomplishing
the same thing with a simplified api, but I was trying to stick to your
interface.

Regarding looking at it on the list: I did and didn't have an objection at
the time. Now that I see how it's implemented and how it's programmer error
prone, I don't think it's the greatest idea, but can live with it with
changes to the implementation.



>
> > - I think we fundamentally disagree about what we prefer. If it were
> > correct and tested the other way, that would have been fine, but it
> wasn't,
> > so I'm bringing it up to a place where I think someone other that you
> and I
> > can follow along. If a hypothetical person can't, then this pull request
> > needs more work.
>
> Again, you need to explain what you mean by incorrect. The refactor
> changes the output per the proposed interface but worked according
> to that spec. All you’ve done is convolved files together and broken the
> OO abstraction. How does this make it any easier for someone to use?
>


Incorrect = bugs.

The OO abstraction is still there, and in fact it's a little more OO than
you had it, but let's not talk about the semantics. Let's move to something
more productive: why is this wrong? I don't understand where you think I've
convolved files. What does that mean?

In terms of easier to use: right now, it's exactly the same as your pull
request. No easier, no harder. The API is the same.


P.S. I'm really trying to get your design into the code base with minimal
changes to your code that will allow for maintenance other than yourself.
Im taking it a step at a time so we can actually look over the code and the
changes and make sure the little piece is right. The problem with larger
pull requests is that there's too much to digest and that often leads to us
making mistakes. I'm just proactively trying to protect against that.

betanalpha:
> > Right, sorry. I forgot the noon () implementation will get compiled
> > away and avoid any unnecessary overhead.
> >
> 
> Is that your final answer? I'm ok with it, just wanted to make sure.

Yes

> https://github.com/stan-dev/stan/pull/1546/files#diff-8ef860bae9dd720a1bce85652e02862eR41
> 
> But, it's possible I didn't send it out. The elements aren't printed in the
> order you intended. A simple bug, but one that's duplicated in a few places.
> 
> (There were a couple more that I can't remember: I just tend to put in
> tests and move on; I'm not trying to be critical, just want this to be done
> safely.)


You mean 

          for (int i = 0; i < n_rows; ++i) {
+            std::cerr << "," << values[i];  <� This line?
+            for (int j = 1; j < n_cols; ++j)
+              std::cerr << "," << values[i * n_cols + j];
+            std::cerr << std::endl;
+          }

Yes, that�s a bug from copying and pasting.  Remove the line and everything
works as its supposed to.

But how the hell does that argue for the entire interface being �complex
and hard to understand�?

> Personally, I would simplify to an api with either 2 calls or 1 call:
> 1. (Both) void operator()(const std::string& message)
> 2. void operator()(const std::string& key, const std::string& value)
> 
> The transformations from (double*, int) -> std::string would be handled by
> utility functions outside the individual writers. Same for (double*, int,
> int) -> std::string. The goal would be to reduce the chance of writer
> error.
> 
> Of course, I see what you're trying to do and am ok with it if we make it
> less error prone by providing a base implementation that is tested. That's
> what I tried to do to keep your api as-is because I believe you e touched
> on some use cases where this is easier. I can still imagine accomplishing
> the same thing with a simplified api, but I was trying to stick to your
> interface.

Clearly I have either completely lost it or nobody is listening.  You are
arguing to simplify the interface at the complete loss of useful information
to the interfaces.  If we can only write strings then the interfaces can�t
discriminate between error messages or info messages or vectors or
key/values or anything.  How is that at all useful?  

> Regarding looking at it on the list: I did and didn't have an objection at
> the time. Now that I see how it's implemented and how it's programmer error
> prone, I don't think it's the greatest idea, but can live with it with
> changes to the implementation.

What error proneness?  You have found one copy/paste bug that was
replicated a few times and would be caught by a simple test.  And the
first request I made was �what should we test?�. 

> In terms of easier to use: right now, it's exactly the same as your pull
> request. No easier, no harder. The API is the same.

No, it�s not.  The API should expose the abstract base class and 
implementations useful for internal testing.  By giving the base class
an implementation you�ve defined a default that the interfaces have
to override, causing potential confusion and conflicts. 

And for whatever reason the git diffs were a mess.  Instead of just
changing the base writer you renamed another writer the base writer
and vice versa.  

> P.S. I'm really trying to get your design into the code base with minimal
> changes to your code that will allow for maintenance other than yourself.
> Im taking it a step at a time so we can actually look over the code and the
> changes and make sure the little piece is right. The problem with larger
> pull requests is that there's too much to digest and that often leads to us
> making mistakes. I'm just proactively trying to protect against that.

Really?  Because from here all I�m seeing is an arbitrary decomposition
of the changes into irrelevant intermediaries with a constant push to change
the API.  No tests have been suggested, and still none of the actual implementation
changes have been added.  All while all the other code (especially RStan) gets 
hacked together with no consideration for a consistent, maintainable, and 
scalable API.  Remind me again why I�m wasting my time refactoring instead
of adding new features?  Oh, yeah, because no one else will do it.

In any case this is not sustainable.

bob-carpenter:
I don't understand the details of this whole discussion, but
I do want to point out that abstract base classes very often
have implementations.  But I agree that it's confusing if you
have a default implementation that must be overridden.  It's not
unheard of --- for instance, vari has a no-op implementation
of chain() that's typically overridden by specializations.  That
kind of no-op base implementation is common.  But usually you
don't want behvior that changes --- you just define helper functions
in terms of other abstract functions.

I find all of this code very confusing, both how it was to
start with and how it is now.  As I've said many times, I'd strongly
prefer standalone functions rather than this hierarchical config
because I find it easier to understand.  

- Bob

PR #1604, opened 2015-09-03
Feature/issue 1157 inv phi

bgoodri:
I can't get any of the unit tests to link ATM, so hopefully this is correct. It is just the doc / expose / test component to the inv_Phi PR in math a few days ago.


bob-carpenter:
Nice --- this looks perfect to me right down to the fonts in the doc and the sorting of the function index!

bob-carpenter:
The current test failure is because it's not finding `inv_Phi` to compile the C++ for a Stan model using `inv_Phi`.  If the math lib isn't updated yet, then we just need to rerun these tests when it has been.

PR #1605, opened 2015-09-05
Feature/issue 1565 sparse multiply

bob-carpenter:
#### Summary:

Add sparse matrix by vector multiplication function into Stan language with tests nd doc.

#### Intended Effect:

Allow faster sparse multiplies with less memory.

#### How to Verify:

Unit tests for model compilation.  Rea doc.

Unit tests and doc already in stan-dev/math to support this.

#### Side Effects:

None.

#### Documentation:

Yes, new chapter to manual.

#### Reviewer Suggestions:

anyone.


bob-carpenter:
New unit tests I added for function signatures are parsing but failing to compile in the integration tests.

bgoodri:
I pushed some tweaks to the doc but it is still not going to compile. The `csr_extract_v.hpp` in math has the signature
```
csr_extract_v(const SparseMatrix<T, RowMajor>& A)
```
and there is no `SparseMatrix` in the Stan language. In other words, if you pass it a dense matrix in the Stan language, it won't compile.

sakrejda:
I haven't had a chance to look at Bob's version of the code but the
original motivation for having a sparse and dense signature for those was
to make it easier to use from interfaces without exposing the sparse ones
to Stan (yet). It makes it easier to have a consistent constructor in R via
RcppEigen...

bgoodri:
That makes sense, but the .stan files are not finding the dense matrix
version.

sakrejda:
That's weird, before the changes I had a complete set of Stan files that
ran fine. I'll check but it won't be till Tuesday. K

bob-carpenter:
This should pass now if lib/stan_math gets updated to stan-dev/math/develop --- it passes for me locally.

bob-carpenter:
The issue was just behind on the math lib.

bob-carpenter:
?  This one's failing now because of `inv_Phi` in the integration test --- it can't find it.  Probably just a model header include issue.

bgoodri:
How did it pass when the inv_Phi branch was merged?

syclik:
I don't know... I just updated the submodule. Hopefully this will get everything to pass. Sorry you were struggling with it today.

bgoodri:
Allright, let's merge this before it breaks again. @sakrejda Do we need something for rstan? I was thinking that a `csr_extract()` function that took a matrix and returned a list of w, u, and v would be nice.

sakrejda:
Yeah. Maybe just that. For consistency I had one that used the Stan math
library extractors (pasted below).  You can take an R sparse matrix (Matrix
package) in row major format, pass it through Rcpp as an Eigen sparse
matrix without making any copies, and then use the extractors  with sparse
Eigen matrix signatures from the Stan math library.

The alternative is to make an R function that plays with the Matrix package
object slots directly (they're close to what we extract) but then we're
depending on Matrix keeping it's implementation the same. It'll probably
stay, so your choice, but I'd rather take a path that will be similar
enough in the different interfaces
Just for reference I had a function like this in rstan/R/sparse.R

extract_sparse_parts <- function(A, type='CSC') {
  if (class(A) == 'matrix') {
    A <- Matrix(A, sparse=TRUE)
  }
  if (type == 'CSR') {
    A <- t(A)
  }
  if (class(A) == 'dgCMatrix') {
    o <- .Call("extract_sparse_components",A)
  } else {
    stop("Argument 'x' must either be of class 'matrix' or 'dgCMatrix'.")
  }
  o[['type']] <- type
  return(o)
}

Then in rstan/src/aaa_sparse_extractors.cpp I added:

#include <stan/math/prim/mat/fun/sparse_extractors.hpp>

#include <Rcpp.h>
#include <RcppEigen.h>

RcppExport SEXP extract_sparse_components(SEXP A) {
  Eigen::SparseMatrix<double> AA(Rcpp::as<Eigen::SparseMatrix<double> >(A));
  Eigen::Matrix<double, Eigen::Dynamic,1> w = stan::math::extract_w(AA);
  std::vector<double> ww(w.size());
  for (unsigned int i=0; i < w.size(); ++i)
    ww[i] = w(i);

  std::vector<int> v = stan::math::extract_v(AA);
  std::vector<int> u = stan::math::extract_u(AA);
  std::vector<int> z = stan::math::extract_z(AA);

  return Rcpp::List::create(
      Rcpp::Named("w") = ww,
      Rcpp::Named("v") = v,
      Rcpp::Named("u") = u,
      Rcpp::Named("z") = z
  );

}

and exposed that through init.cpp so it could be called from R.  That's
it.  Let me know if you want me to figure out what this would look like for
the current rstan (the above was vs. 2.6.2, I think).

Krzysztof

syclik:
Just read the last two comments. Mind continuing the discussion on rstan?

sakrejda:
Yep, good reminder that I should do issues for all the interfaces and see
if there's enough interest to expose this stuff everywhere. K


Krzysztof Sakrejda

Department of Biostatistics and Epidemiology
University of Massachusetts, Amherst
319 Morrill Science Center South
611 N. Pleasant Street
Amherst, MA 01003

work #: 413-325-6555
email: sakrejda@umass.edu

2015-09-08 8:02 GMT-04:00 Daniel Lee <notifications@github.com>:

PR #1620, opened 2015-09-30
set BIT variable on windows conditionally depending on 32 or 64 bit

wds15:
#### Summary:

set BIT variable on windows conditionally on running on a 32bit or 64bit machine; this triggers the -m$BIT variable to be set correctly for g++ based (Rtools) windows builds

#### Intended Effect:

fix the makefile on windows for 32bit machines. Currently Stan generates 64bit binaries on 32bit machines. Not good.

#### How to Verify:

Run some test for example on a win 32bit machine and a win 64bit machine.

#### Side Effects:

None for Stan, but this change should be propagated to stan-math & cmdstan.

#### Documentation:

None.

#### Reviewer Suggestions:
@syclik

### DIFF
--- a/make/os_win
+++ b/make/os_win
@@ -17,7 +17,9 @@ EXE = .exe
 O_STANC = 3
 CFLAGS_GTEST += -DGTEST_HAS_PTHREAD=0
 PATH_SEPARATOR = '\'
-ifeq ($(PROCESSOR_ARCHITECTURE)$(PROCESSOR_ARCHITECTURE6432),x86)
+## Detecting Process Bitness:
+## http://blogs.msdn.com/b/david.wang/archive/2006/03/26/howto-detect-process-bitness.aspx
+ifeq ($(PROCESSOR_ARCHITECTURE)$(PROCESSOR_ARCHITEW6432),x86)
   BIT=32
 else
   BIT=64
--- a/make/os_win
+++ b/make/os_win
@@ -17,7 +17,11 @@ EXE = .exe
 O_STANC = 3
 CFLAGS_GTEST += -DGTEST_HAS_PTHREAD=0
 PATH_SEPARATOR = '\'
-BIT = 64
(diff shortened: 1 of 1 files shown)


syclik:
Thanks. I was able to verify this on a 64-bit windows machine.
I have a question that I'll ask inline.

make/os_win
```diff
@@ -17,7 +17,11 @@ EXE = .exe
 O_STANC = 3
 CFLAGS_GTEST += -DGTEST_HAS_PTHREAD=0
 PATH_SEPARATOR = '\'
-BIT = 64
+ifeq ($(PROCESSOR_ARCHITECTURE)$(PROCESSOR_ARCHITECTURE6432),x86)
```
syclik: I'm confused with the two variables. Are both needed? It looks like the second one is misspelled (`PROCESSOR_ARCHITECTURE6432`).
I'm looking at this as a reference:
http://blogs.msdn.com/b/david.wang/archive/2006/03/26/howto-detect-process-bitness.aspx

Will this set the `BIT` variable for an AMD processor?

wds15:
Hi!
Thanks for catching - yes the second variable was misspelled. It should be

PROCESSOR_ARCHITEW6432

Yes, the patch should work for amd64. I mean the logic is to detect a "true" 32bit only machine which has PROCESSOR_ARCHITECTURE equal to x86 and the other variable PROCESSOR_ARCHITEW6432 would not be defined. In any other situation the logic decides for the 64bit branch.

And yes, both are needed, because you can run a 32bit process on a 64bit machine, for this combination PROCESSOR_ARCHITECTURE is x86 while the other one PROCESSOR_ARCHITEW6432 is set to a value which indicates the specific 64bit architecture, AMD64. In this situation the code decides for 64bit which is correct as 64bit code will run on such a machine.

In case you are on a 64bit machine and the make runs in a 64bit process, then PROCESSOR_ARCHITECTURE is set to AMD64 and again the 64bit branch is set. Hope that makes sense.

syclik:
@wds15, updated the changes. I'm ok with this once it passes tests.

PR #1630, opened 2015-10-08
Implementing writers

syclik:
#### Summary:

Implement writers.

#### Intended Effect:

Update writers to match feature/issue-1361-writer_callback.
None of the behavior changes.

#### How to Verify:

Take a look at the implementation.

#### Side Effects:

All interfaces may need minor changes.

#### Documentation:

We may need to fill in more doc.

#### Reviewer Suggestions: 

@bob-carpenter


syclik:
We may want to add more tests and doc before this goes in.

src/stan/interface_callbacks/writer/base_writer.hpp
```diff
@@ -8,15 +9,55 @@ namespace stan {
   namespace interface_callbacks {
     namespace writer {
 
+      // base_writer is an absract base class defining the
```
syclik: needs doxygen doc

src/stan/interface_callbacks/writer/noop_writer.hpp
```diff
@@ -0,0 +1,33 @@
...
+
+#include <stan/interface_callbacks/writer/base_writer.hpp>
+#include <string>
+#include <vector>
+
+namespace stan {
+  namespace interface_callbacks {
+    namespace writer {
+
+      class noop_writer: public base_writer {
```
syclik: doc

src/stan/interface_callbacks/writer/stream_writer.hpp
```diff
@@ -0,0 +1,101 @@
...
+#include <stan/interface_callbacks/writer/base_writer.hpp>
+#include <iostream>
+#include <vector>
+#include <string>
+
+namespace stan {
+  namespace interface_callbacks {
+    namespace writer {
+
+      template <class Stream>
```
syclik: doc

src/stan/interface_callbacks/writer/stream_writer.hpp
```diff
@@ -0,0 +1,101 @@
...
+          --last;
+
+          for (std::vector<double>::const_iterator it = state.begin();
+               it != last; ++it)
+            output__ << *it << ",";
+          output__ << state.back() << std::endl;
+        }
+
+        void operator()() {
+          output__ << std::endl;
```
syclik: I know this matches the old behavior of inserting a blank line into the csv file, but I think we can change this behavior to insert the `key_value_prefix__` into a completely blank line.

Any objections?

src/stan/interface_callbacks/writer/stream_writer_typedefs.hpp
```diff
@@ -0,0 +1,18 @@
...
+#include <stan/interface_callbacks/writer/stream_writer.hpp>
+#include <sstream>
+#include <iostream>
+
+namespace stan {
+  namespace interface_callbacks {
+    namespace writer {
+      // An implementation of stream_writer
+      // using stringstream for unit tests
+      typedef stream_writer<std::stringstream> sstream_writer;
```
syclik: Any reason we don't put this typedef inside `stream_writer.hpp`?

src/test/unit/interface_callbacks/writer/sstream_writer_test.cpp
```diff
@@ -0,0 +1,120 @@
...
+    for (int n = 0; n < N; ++n)
+      x.push_back(boost::lexical_cast<std::string>(n));
+
+  EXPECT_NO_THROW(writer(x));
+  EXPECT_EQ("0,1,2,3,4\n", ss.str());
+}
+
+TEST_F(StanInterfaceCallbacksSStreamWriter, null) {
+  EXPECT_NO_THROW(writer());
+  EXPECT_EQ("\n", ss.str());
```
syclik: if we change the behavior, this should have the prefix: `#\n`

syclik: I guess we should be using `writer_prefix()` also.

src/stan/interface_callbacks/writer/base_writer.hpp
```diff
@@ -8,15 +9,55 @@ namespace stan {
...
       public:
+        // Virtual destructor to avoid compiler warnings
         virtual ~base_writer() {}
-        // Can't enforce this method with a pure virtual function
-        // template <class T>
-        // virtual void operator()(const std::vector<T>& x) = 0;
-        virtual void operator()(const std::string x) = 0;
+
+        // Key-scalar input
+        virtual void operator()(const std::string& key,
```
syclik: @bob-carpenter, think we should name these methods? The idea that this is a functor is nice, but I think the use here is a bit confusing. I don't have a suggestion for naming, though.

src/stan/interface_callbacks/writer/stream_writer.hpp
stan-buildbot: @betanalpha, any reason we have this templated instead of taking in an `std::ostream`?

betanalpha: For many of the same reasons templates are used to replace class
derivations.  Here it was just to avoid relying on a specific ostream
interface (and looking up how sstream and fstream all relate to each other).

src/stan/interface_callbacks/writer/stream_writer_typedefs.hpp
betanalpha: Just to separate out the base interface from the implementations
used internally in the API.

src/stan/interface_callbacks/writer/stream_writer.hpp
betanalpha: Why?

src/stan/interface_callbacks/writer/stream_writer.hpp
bob-carpenter: 1.  OO: The streams are object oriented and designed to
be used with std::ostream arguments.  The class template
doesn't give you any more flexibility because you need
to instantiate it to a class implementing some concept,
which is going to be defined by ostream itself.

2. References:  If we use std::ostream& arguments,
we don't have to do the == NULL and branch logic
everywhere for safety and we don't have to dereference
the objects in the code.  This assumes anything 
supplying one of these stream-based implementations will 
supply a stream, not NULL (0).

- Bob

> ...

> > +      template <class Stream>
> 
> For many of the same reasons templates are used to replace class derivations. Here it was just to avoid relying on a specific ostream interface (and looking up how sstream and fstream all relate to each other).

>  @betanalpha, any reason we have this templated instead of taking in an std::ostream? — Reply to this email directly or view it on GitHub.

src/stan/interface_callbacks/writer/base_writer.hpp
bob-carpenter: I don't see any reason it needs to be an operator from
a use point of view, but I'm OK with it.  I think of
operators like anonymous functions --- they apply
without the ".fun_name" bit.  How will they look in use?

Here's the comparison with an overloaded write function.
You construct a writer:

```
  writer_t w(...);
```

and then it's either the functor:

```
  w("foo");
```

or named method:

```
  w.write("foo");
```

It looks a bit different with longer variable names:

```
  writer_t writer(...);
```

and then

```
  writer("foo");
```

vs.

```
  writer.write("foo")
```


Also, do we have a write const char* that will work with "foo" without
forcing it to be constructed as a string (not without overhead --- there's
a malloc and copy of contents).  It could always be added as an optimization
later.

src/stan/interface_callbacks/writer/stream_writer_typedefs.hpp
bob-carpenter: We shouldn't need this typedef when the templating
is replaced with the simpler OO pattern.

src/stan/interface_callbacks/writer/stream_writer.hpp
syclik: It seems more consistent that way. As it stands, all the methods for writing have comments except for:
1. writing the names (header of csv)
2. writing the state
3. writing a blank line.
The third seems out of place when the other methods like writing a message and writing key / value pairs include the `key_value_prefix__`. (wow... I'm just realizing what the variable name is indicating.)
If we were to go the other way, I would much rather not print the `key_value_prefix__` all method calls that aren't for key / value pairs. I think that might change behavior, though.

betanalpha: After spending so much time promoting not changing the behavior 
why do you want to change it now?  The asymmetry here is exactly
to match the current behavior.

syclik: Hopefully we'd all agree that the asymmetry is non-ideal and we'd rather have it be symmetric.

Although the behavior is changing, I know how the csv files are parsed, so I know this doesn't have an effect. It also doesn't affect RStan's ability to read csv files. A test needs to change to accommodate this behavior change, but not the code.

Of course, I haven't accounted for other consumers of this format: do you need the extra blank lines in the csvs for something in particular? Or can they be marked as comments safely?

bob-carpenter: Although this isn't the place to be having this discussion, I think we
need to use CSV format as intended.  That is, our CSV readers ignore
all the comments, read the header, and read the contents.  I think every
other kind of reader should go somewhere else, even if it reads the
comments of a CSV file.

syclik: I agree to all points made. We should make this a separate thread, but for the time being, I'm going to implement the change that shouldn't break anything for the sake of consistency and we can move forward with this.

syclik:
@bob-carpenter or @betanalpha, is this ok? I need to put together a pull request for CmdStan to deal with a missing header file.

bob-carpenter:
Is what OK?

syclik:
Could you take a look at the pull request? See if there's enough doc,
testing, anything else that needs to happen for just this piece getting in?

syclik:
I have a pull request on CmdStan for this branch: stan-dev/cmdstan#171.
https://github.com/stan-dev/cmdstan/pull/171

If that passes, once CmdStan is updated, it will be compatible with this pull request.

betanalpha:
How many reviewers are we going to need for this thing?
Was any behavior of the writer changed or just the writer
code propagated to weird corners of the tests and ADVI?

bob-carpenter:
One per pull request should be enough.  I'm happy to
review things if you want a second or third pair of eyes
on it.

betanalpha:
The problem is Dan is taking my pull request and changing
a bunch of things and then asking for more reviews.  This
whole process is falling apart which is making is near
impossible for me to contribute anything.

syclik:
@betanalpha, I take this to mean you've looked at the pull request and oked it? Just give me an explicit ok and this one's good to merge. The failed Jenkins test is due to the change to CmdStan living on a pull request.

betanalpha:
What about all the changes to stan/mcmc, stan/optimization, and stan/services
to propagate these through?

syclik:
That's next.

betanalpha:
Then merge this and get to those ASAP.  I really need the new writer
in the MCMC code base to facilitate some research experiments with
new HMC algorithms.

PR #1638, opened 2015-10-13
Change argument of eta_adagrad to eta

dustinvtran:
#### Summary:
Changes an argument name of variational inference, from `eta_adagrad` to `eta`. This follows the greek argument convention used in MCMC, and abstracts away from ADVI's tie to a particular learning rate. To be reviewed in combination with https://github.com/stan-dev/cmdstan/pull/166.

Note: I don't suggest we merge this request now, but the request will serve as discussion on whether 1. we should allow these sorts of changes; 2. implement it with backwards-compatibility; 3. wait until Stan 3.0.

#### Intended Effect:
None

#### How to Verify:
Run unit tests as typically.

#### Side Effects:
None

#### Documentation:
- [x] CmdStan (see https://github.com/stan-dev/cmdstan/pull/166)

#### Reviewer Suggestions:
None

### DIFF
--- a/src/stan/services/arguments/arg_variational.hpp
+++ b/src/stan/services/arguments/arg_variational.hpp
@@ -6,7 +6,7 @@
 #include <stan/services/arguments/arg_variational_algo.hpp>
 #include <stan/services/arguments/arg_variational_iter.hpp>
 #include <stan/services/arguments/arg_variational_num_samples.hpp>
-#include <stan/services/arguments/arg_variational_eta_adagrad.hpp>
+#include <stan/services/arguments/arg_variational_eta.hpp>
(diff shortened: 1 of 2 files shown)


bob-carpenter:
We've still got all of this variational stuff marked
as experimental, and it's only in CmdStan, so I think
we should be OK to monkey with the interface until we
remove the "experimental" label --- it's odd to have part
of the code in beta and part not, but that's what we
decided to do.

- Bob

syclik:
Jenkins, retest this please.

syclik:
@dustinvtran: I didn't realize why this was causing trouble downstream. The `eta` parameter was inadvertently changed to `string_argument`. We need to change this back to `real_argument`.

Issue #1588, opened 2015-08-17
Print warning in ADVI if max number of iterations reached or ELBO is higher at at previous iteration


dustinvtran:
Discussed in [stan-dev](https://groups.google.com/forum/#!topic/stan-dev/IB9BIGS1F_8). If the maximum number of iterations are reached, `MAX ITERATIONS REACHED` is currently printed. However, there should be more information, warning users that because the max number of iterations are reached, the optimization hasn't converged and thus one should be very careful about using samples from the output.

Same for the case when the ELBO is higher at a previous iteration that isn't the last. (can happen in SGD for nonconvex optimization).

akucukelbir:
@dustinvtran should i take a shot at this in https://github.com/stan-dev/stan/tree/feature/advi-adapt-eta since i'll need to check for decreasing ELBO there? or do you want to tackle these separately?

dustinvtran:
Since they’re minor fixes I think they can be in a different branch so that it is merged more quickly into develop.

akucukelbir:
in feature/advi-adapt-eta, i establish some scaffolding for tracking the ELBO. i just want to make sure we don't re-invent solutions for similar problems.

dustinvtran:
Ah, I see what you mean. Then I’ll add the changes in your branch then. :-)

Dustin

syclik:
Fixed in #1590.

PR #1640, opened 2015-10-14
Add more informational messages for ADVI

dustinvtran:
#### Summary:
This is a fix to issue #1588 and one additional new warning feature. In general, it implements three verbose message changes:
- Instead of outputting "MAX_ITERATIONS REACHED", the algorithm has a more user-friendly message.
- It adds a new message when drawing approximate posterior samples.
- It adds a new message if the ELBO was evaluated to be much larger at a previous set of iterations than what it in fact converged to. "Much larger" here is defined by greater than 5%; it's heuristic because the ELBO computation is a Monte Carlo estimate.

#### Intended Effect:
More verbose warning messages.

#### How to Verify:
Let me know if any of this/what needs unit testing.

#### Side Effects:
None

#### Documentation:
N/A

#### Reviewer Suggestions:
N/A


syclik:
Jenkins, retest this please.

syclik:
Jenkins, retest this please.

syclik:
Jenkins, retest this please.

syclik:
@dustinvtran, this needs tests.

Add some tests for `robbins_monro_adagrad` that trigger these messages. It would also be good to check that the core of the message is correctly triggered when it's happened.

We've been lenient so far with allowing more flexible changes, but with @akucukelbir's latest issue with regards to ctrl-c, we should be moving towards better code.

src/stan/variational/advi.hpp
```diff
@@ -334,6 +357,14 @@ namespace stan {
 
         // draw more samples from posterior and write on subsequent lines
         if (out_stream_) {
+          if (print_stream_) {
```
syclik: Why is this inside `if (out_stream_)`? Is that intentional?

dustinvtran: if `out_stream_` doesn't exist, then the algorithm doesn't need to draw any samples. (that's currently how it works)

bob-carpenter: Can print_stream_ exist if out_stream_ doesn't exist?

Also, you almost never want to do

```
if (cond1) { if (cond2) { ... }}
```

It's better in terms of efficiency and readability as:

```
if (cond1 && cond2) { ... }
```

- Bob

P.S.  I really do not like these 0 pointer patterns.
The are dangerous because they open up easy segfault
opportunities because <b>every single use</b> of `out_stream_`
with `if (out_stream_)`.

dustinvtran: `advi.hpp` makes no assumptions on any dependencies among `print_stream_`, `out_stream_`, and `diagnostic_stream_`.

The code doesn't do `if (cond 1 && cond2)` because it checks `cond2` multiple times within the control flow of the `if (cond1)` loop. Reimplementing it would change it to be `if (cond1 && cond 2) print stuff ... if (cond1) draw stuff ... if (cond1 && cond2) print stuff`. I don't know if that's more readable.

bob-carpenter: Sorry --- I should've looked at the larger context.  That
if (out_stream_) test isn't just for that I/O,
it's for a whole block of work, which is itself confusing,
but that seems to be the intention (out_stream_ is both a
flag indicating whether to do some kinds of work and the
stream to which draws are written).

- Bob

akucukelbir:
@syclik @bob-carpenter: cpplint is failing, seemingly not due to any file we're touching here. what should i do?

syclik:
We need to merge develop into the branch. I'll try to do that now. Give me a couple mins.

syclik:
I just pushed. cpplint shouldn't fail anymore.

akucukelbir:
awesome. thank you daniel.

i have another branch ready to be "pull requested". i'll do that as soon as this pull request gets reviewed and merged.

syclik:
I reviewed it. It does look like it'll only affect output. Thanks for adding the tests and showing that you actually trigger the two conditions.

syclik:
(this will get merged once tests pass. @dustinvtran, go ahead and create the new pull request.)

akucukelbir:
@syclik i'll wait for this to get merged, then merge develop into the new branch, make sure tests pass locally, and then create the new pull request.

syclik:
Cool. That works. And sorry -- I thought it was Dustin that responded (I
think it was cause he created this PR). Thanks.

akucukelbir:
no worries. tests on the new branch are passing locally (without the impending merge). assuming the also pass after the merge, i'll only have some unit tests to modify and we'll be ready for another pull request. awesome.

akucukelbir:
umm... again, not sure what's happening this time. :(

akucukelbir:
it says that 
> 'test/test-models/good/variational/univariate_no_constraint.hpp' file not found

that's odd.

akucukelbir:
ooooh. hold on. i need to edit that tests makefile thing... damn.

PR #1645, opened 2015-10-27
Fixes variational eta argument

syclik:
Fixes argument bug introduced with #1638.

### DIFF
--- a/src/stan/services/arguments/arg_variational_eta.hpp
+++ b/src/stan/services/arguments/arg_variational_eta.hpp
@@ -7,9 +7,9 @@ namespace stan {
 
   namespace services {
 
-    class arg_variational_eta: public string_argument {
+    class arg_variational_eta: public real_argument {
     public:
-      arg_variational_eta(): string_argument() {
+      arg_variational_eta(): real_argument() {
         _name = "eta";
         _description = "Stepsize scaling parameter for variational inference";
         _validity = "0 < eta <= 1.0";


dustinvtran:
I approve!

syclik:
(I don't know what we should do to test these things... it wasn't clear
that this would break CmdStan [seg fault], but it did.)

dustinvtran:
I'm writing unit tests in the `advi/internal-changes` branch, which will eventually be a pull request when I am finished. It's a complete overhaul of the ADVI code and current unit tests, as it was somewhat vague what each test was even testing for.

I'm not sure how to test the arguments themselves though.

syclik:
cool. off topic and no response required: can you create an issue and name
the branch "feature/issue-#-description" so we have a hope of tracking
changes for releases?

syclik:
This passed the Stan tests. We still don't have a good way to specify upstream branches.

betanalpha:
https://github.com/stan-dev/cmdstan/blob/develop/src/test/interface/services/argument_configuration_test.cpp

tries to run every combinatorial combination of arguments — all combinations
of good values are expected to pass, any bad value is expected to cause a
failure.  The good/bad values are defined in the arguments themselves.

syclik:
that test didn't fail on this branch.

in CmdStan, there's a cast to real_argument from string_argument that was
causing a seg fault.

betanalpha:
Create an issue — my best guess is that the seg fault is short circuiting
the test which causes it not to fail.

ariddell:
Is there anything I can do to help with this (i.e., this future pointer-less argument refactor mentioned by @syclik on the list)? Is there a branch working on this or is this on the backburner?

PR #1648, opened 2015-10-28
Docs standardized variable notation

migueldvb:
Add subindex for inverse transform to reverse stadardization of y.


bob-carpenter:
Thanks for sending the patch.  I sort of botched the traditional notation, which doesn't treat $z$ as a function, but as the transformed random variable from $y$.  But I'll merge it because I think the patch is a bit clearer than what I originally wrote.

Issue #1255, opened 2015-01-28
reverse deprecation on abs() and fabs()


bob-carpenter:
I think we should deprecate
-  `real fabs(real)`

and go with just
- `real abs(real)`
- `int abs(int)`

The only drawback is that we won't have a function that applies to an integer and returns a real value, but I don't think we need that.  Returning an integer won't lose information, and Stan will just promote it to a real value if necessary to use it elsewhere.  And we'll still get integer absolute values for integers.

I propose getting rid of everything we're deprecating in Stan 3.
- [ ] undeprecate `abs(real)`
- [ ] deprecate `fabs()`
- [ ] update manual explaing what's going on

rtrangucci:
@syclik we fixed with #1649 which was merged, ok to close?

bob-carpenter:
yes.

bob-carpenter:
I think we can solve this one and it'll be less confusing to users not familiar with the C++ function fabs.

rok-cesnovar:
Currently in the language we have the following abs() functions: 

```
int abs(int)
real abs(real)
```

and the following for fabs(): 

```
real fabs(int)
real fabs(real)
vector fabs(vector)
row_vector fabs(row_vector)
matrix fabs(matrix)
real[] fabs(int[])
real[] fabs(real[])
vector[] fabs(vector[])
row_vector[] fabs(row_vector[])
matrix[] fabs(matrix[])
real[,] fabs(int[,])
real[,] fabs(real[,])
vector[,] fabs(vector[,])
row_vector[,] fabs(row_vector[,])
matrix[,] fabs(matrix[,])
real[,,] fabs(int[,,])
real[,,] fabs(real[,,])
vector[,,] fabs(vector[,,])
row_vector[,,] fabs(row_vector[,,])
matrix[,,] fabs(matrix[,,])
real[,,,] fabs(int[,,,])
real[,,,] fabs(real[,,,])
vector[,,,] fabs(vector[,,,])
row_vector[,,,] fabs(row_vector[,,,])
matrix[,,,] fabs(matrix[,,,])
real[,,,,] fabs(int[,,,,])
real[,,,,] fabs(real[,,,,])
vector[,,,,] fabs(vector[,,,,])
row_vector[,,,,] fabs(row_vector[,,,,])
matrix[,,,,] fabs(matrix[,,,,])
real[,,,,,] fabs(int[,,,,,])
real[,,,,,] fabs(real[,,,,,])
[...]

PR #1649, opened 2015-10-28
replace deprecated function abs by fabs

migueldvb:
Replace `abs` by `fabs` in lasso example.

### DIFF
--- a/src/docs/stan-reference/advanced.tex
+++ b/src/docs/stan-reference/advanced.tex
@@ -275,7 +275,7 @@ data {
 model {
   // ...
   for (k in 1:K)
-    increment_log_prob(- lambda * abs(beta[k]));
+    increment_log_prob(- lambda * fabs(beta[k]));
 }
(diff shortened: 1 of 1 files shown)


andrewgelman:
“fabs”???

bob-carpenter:
See the manual index for definitions and signatures
for abs() and fabs() and the corresponding min/max and fmin/fmax.

See old mail threads for discussion on why it's like that.

But before you tell us again that it doesn't have to be that way
and that it's "tacky", see:

  https://github.com/stan-dev/stan/issues/1255

It's scheduled for Stan 3 because it's going to break backward
compatibility (unless we keep fabs and deprecate it).

This low-level maintenance coding never gets done because
nobody wants to do it.

- Bob

andrewgelman:
That’s fine.  If it’s in the manual, I can deal with it.
A

syclik:
This looks fine.

Issue #1361, opened 2015-03-11
Introduce Writer Callback


betanalpha:
Refactor internal Stan code to use a writer callback class left for the interfaces to implement.  Using the writer callback everywhere will remove the last traces of cout/cerr from the code (addressing #699) while giving the interfaces complete control over output (addressing #453).
- [ ] #1601: remove unused code
- [ ] #1602: implement writers

ariddell:
I appreciate what you're doing here, but it seems a bit hasty (esp. probe_args). It feels like one is inventing a protocol from scratch to communicate key value pairs via strings. At minimum, I think I'd prefer we pick something recognizable (like JSON) -- or do a little bit more work to articulate what are valid types of values and spec it in Protobuf.

I suppose I'd also like to see what the big C++ object approach looks like. I don't have that much time this week, but I'd be willing to give a shot at articulating my ideas in a cmdstan branch like this one.

betanalpha:
To be clear, the argument stuff is not new.
probe_args is for automated testing.

ariddell:
Hmm.. it would be cool if the branch was referenced here somehow (maybe add "Closes #1361" in the commit?). Here's a commit I was looking at: c68ac51d2edb3fb8e59cbeccdfffa31c1879338d

For Python and R I think it's fair to say that it's much easier to create a (pointer to) a struct (or C++ class) and pass that around than to write a C++ callback.

betanalpha:
The callbacks here are just classes that implement
certain functions — they should be straightforward
to implement (see the cout/cerr and stream
implementations that will eventually move to CmdStan).

ariddell:
I suppose the argument against having interfaces have to write C++ functions starts like this: it's easier to for interfaces to pass around a pointer or a chunk of bytes without knowing too much about it. In fact, with JSON or protobuf they don't even need to touch C++ code potentially. Having them implement a C++ class with specific properties is a significantly bigger hurdle.

It might be helpful to check out how RStan (and PyStan, more or less) currently deal with getting config and defaults setup:

https://github.com/stan-dev/rstan/blob/master/rstan/rstan/inst/include/rstan/stan_args.hpp

bob-carpenter:
If you really don't want to work at the C++ level, it
should be possible to completely rewrite PyStan to follow
Stan.jl and MatlabStan (and my original code for RStan), which
would have the user specify a command-line argument as a string,
write the data out to a file, and then execute a CmdStan process
and then collect up the CSV data.

[...]

PR #1652, opened 2015-10-29
Added explicit keyword to constructors with one argument

syclik:
#### Summary:

Adds the `explicit` keyword to constructors with one argument.

#### Intended Effect:

Code cleanup. Matches closer to #1361.

#### How to Verify:

Visually inspect the code for correctness.

#### Side Effects:

I don't foresee any problems, but may trickle upstream to CmdStan.

#### Documentation:

None required.

#### Reviewer Suggestions: 

@betanalpha

### DIFF
--- a/src/stan/services/arguments/argument.hpp
+++ b/src/stan/services/arguments/argument.hpp
@@ -15,7 +15,7 @@ namespace stan {
         : indent_width(2),
           help_width(20) { }
 
-      explicit argument(const std::string name)
+      explicit argument(const std::string& name)
         : _name(name),
           indent_width(2),
           help_width(20) { }
@@ -31,7 +31,7 @@ namespace stan {
       }
 
       virtual void print(std::ostream* s, const int depth,
-                         const std::string prefix) = 0;
+                         const std::string& prefix) = 0;
       virtual void print_help(std::ostream* s, const int depth,
                               const bool recurse) = 0;
 
@@ -44,8 +44,8 @@ namespace stan {
 
       virtual void probe_args(argument* base_arg, std::stringstream& s) {}
 
-      virtual void find_arg(std::string name,
-                            std::string prefix,
+      virtual void find_arg(const std::string& name,
+                            const std::string& prefix,
                             std::vector<std::string>& valid_paths) {
         if (name == _name) {
           valid_paths.push_back(prefix + _name);
@@ -65,7 +65,7 @@ namespace stan {
         }
       }
 
-      virtual argument* arg(const std::string name) {
+      virtual argument* arg(const std::string& name) {
         return 0;
       }
 
--- a/src/stan/services/arguments/argument.hpp
+++ b/src/stan/services/arguments/argument.hpp
@@ -15,7 +15,7 @@ namespace stan {
         : indent_width(2),
           help_width(20) { }
 
-      argument(const std::string name)
+      explicit argument(const std::string name)
         : _name(name),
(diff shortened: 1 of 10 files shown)


syclik:
Jenkins, retest this please.

syclik:
Jenkins, retest this please.

src/stan/services/arguments/singleton_argument.hpp
```diff
@@ -44,7 +44,7 @@ namespace stan {
     template<typename T>
     class singleton_argument: public valued_argument {
     public:
-      singleton_argument(): _validity("All") {
+      explicit singleton_argument(): _validity("All") {
```
bob-carpenter: Why are you making this unary constructor explicit?  Will this still work:

```
singleton_argument a;
```

?

src/stan/services/arguments/argument.hpp
```diff
@@ -15,7 +15,7 @@ namespace stan {
         : indent_width(2),
           help_width(20) { }
 
-      argument(const std::string name)
+      explicit argument(const std::string name)
```
bob-carpenter: Shouldn't this be taking a constant reference, not copying the string?

syclik:
Bob, thanks for reviewing the code! My fault: I did the laziest thing possible. I copied code over from the other branch so squash another 4 file diffs.

I think it's clear that I shouldn't have done this. Here are some options now:
1. close this pull request and fix Michael's branch to match `develop`
2. walk through all of the arguments and fix this permanently on the `develop` branch
3. walk through just these changes and fix it for this pull request

I would feel most comfortable with 2, but 1 is the easiest thing to do. Thoughts?

bob-carpenter:
I think (2) with as much of (3) as you have time for is good.
No reason to put off changes that move in the right direction just
because there are more waiting in the wings.

This is a direct consequence of the cpplint test because the inclination
every time I see them spit out warnings is to fix them.

- Bob

betanalpha:
I don’t remember touching the arguments in the originating
pull request so I’m not sure where this came up, or why the
warnings wouldn’t be in development.

syclik:
It was part of 136a1a2 "Fix a swarm of cpplint errors". No action required.
If you want to help with this particular thing, just make the right fixes
to develop in a branch and we'll get it merged in. It was just cluttering
up the file diff for that pull request, so I'm trying to knock it down to
something manageable. It's down to 119 now! I'm still really hopeful.

src/stan/services/arguments/singleton_argument.hpp
syclik: My bad. It should have gone on the next constructor. I'm fixing it.

syclik:
I updated the branch. @betanalpha, want to take a look?

syclik:
Jenkins, retest this please.

betanalpha:
Passing the strings by reference isn't going to help much here as the calls almost always are done with a string that gets constructed on the fly anyways (pretty sure the compiler can optimize this to a single constructor), and they don't get called enough for it really matter.  But it's largely irrelevant.

:+1:

syclik:
Agreed. I was reading the stack overflow threads about passing string by
value being optimized for c++11, but it looks like the right thing to do in
most cases is pass by reference.

Either way, this won't affect runtime noticeably, but I'm doing this as a
way to clean up our code base and to reduce the unnecessary diffs on that
large pull request.

I'll merge when tests pass.

Issue #1617, opened 2015-09-28
next manual, 2.8.0++


bob-carpenter:
This is where updates for the manual for the next release go if they are not related to a pull request (new features, bug fixes, etc.)

bob-carpenter:
* [ ] Add Krzysztof Sakrejda to list of developers (functions, Stan Math, Stan, C++)

* [ ] make it clear which repos/modules each developer is contributing to

bob-carpenter:
Julian King points out:
```
p(y | lambda) = Poisson(y | lambda) / (1 - PoissonCDF(0 | lambda))
so log p = log_Poisson(y | lambda) - log(1 - PoissonCDF(0 | lambda))
 = log_Poisson(y | lambda) - log(1 - exp(-lambda)
 = log_Poisson(y | lambda) - log1m_exp(-lambda)
```
* [ ] Update zero-inflated/hurdle section (p. 109 now) to note a further reduction
* [ ] cite Julian King for pointing this out (and add to acks)

bob-carpenter:
* [ ] thank Ashley Ford for reporting a bug on test-headers

bob-carpenter:
* [ ] update description of how Wiener distribution is defined according to following.

From Joachim Vandekerckhove on stan-users:

Sorry I'm a little late to this party, I had a long summer of travel.  The reason we originally implemented the lower-bound hits as negative reaction times was to have a cheap way of encoding what is essentially a bivariate distribution (strictly positive RT and binary response).  Because both the RT and the outcome are random variables, the current implementation (with only one bound) isn't entirely satisfactory if at any point we want to generate random numbers from the distribution.

I always liked the negative-RT solution even though it's a little counter-intuitive.  I'd probably prefer it in Stan since it's how it works in JAGS... but the more conventional solution would be to make it actually bivariate, and let it operate on a [RT response] pair, where the first is a strictly positive real and the second is a boolean.

bob-carpenter:
* [ ] fix broken URL links in manual
    * [ ] link to  http://mc-stan.org/examples.html

Howard Zail points out on stan-users:
<blockquote>
Most of the hyperlinks in the Stan Manual seem to be broken.  In particular, I am looking for the http://mc-stan.org/examples.html page.  This link is also broken on the mc-stan website.
</blockquote>

bob-carpenter:
* [ ] organize examples from manual on stan-dev/example-models better

[...]

PR #1665, opened 2015-11-03
Fix advi typo in manual

akucukelbir:
#### Summary:
This fixes a typo in the variational inference section of the manual. (Thanks to Bruno Jacobs for catching this one!)

#### Side Effects:
None

#### Documentation:
This only affects the documentation.

#### Reviewer Suggestions:
Anyone.


syclik:
@akucukelbir, please undo the change to `lib/stan_math`. You're pointed at a commit that doesn't exist.

akucukelbir:
how do i do that? (i don't understand submodules [yet?])

syclik:
I think the easiest thing is for me to undo it... it's just harder to
explain than to do it.

For future reference, git tracks submodules as its commit hash. If you
change the stan math library, you'll see it in your git status as
lib/stan_math having changes. Don't git add unless you want to intend to
change the math library.

akucukelbir:
roger that. thanks!

bob-carpenter:
Thanks for fixing these things --- the details are important.

Whenever someone suggests a fix we implement, please thank
them in the acknowledgements chapter of the manual.  There's
a list for code and a list for doc.

- Bob

bob-carpenter:
Also, you may find it easier to just comment on the ongoing
issue or open a branch matching whatever the next manual issue
currently is and fix it there.  Here's the current one for 2.8.0++
(they are always named with "next manual <current-release>++"):

  https://github.com/stan-dev/stan/issues/1617

I added an issue to thank Bruno Jacobs in the acknowledgements.

- Bob

akucukelbir:
thanks bob!

syclik:
Jenkins, retest this please.

syclik:
@bob-carpenter, if you think this is ok, can you merge? I don't know if you migrated this to the manual issue.

bob-carpenter:
Two issues with that, so I don't think it should go in as is:

* [ ] it uses \mathcal{N} for normals rather than \distro{Normal} we use elsewhere;  we should be consistent

* [ ] Andrew insists on no semicolons in conditional distriutions

If I fix the whole thing, will it conflict with other requests?

akucukelbir:
fine on my end!

syclik:
Bob, I don't understand your question. You can fix it here and it shouldn't
conflict. If it does, we'll figure it out.

bob-carpenter:
Even easier --- I'll just update notation when I deal with the manual.

PR #1669, opened 2015-11-04
Silencing warning from new clang++

syclik:
#### Summary:

Squashes warning from updated compiler on Mac.

#### Intended Effect:

Silences a new warning due to updated toolchain. To silence the cause is going to be more difficult than I'm willing to undertake now.

#### How to Verify:

Run 
```
> ./runTests.py src/test/unit/lang/reject/reject_func_call_model_test.cpp
```

#### Side Effects:

We'll need to see if this breaks on an older clang++. I'm on it.

#### Documentation:

None.

#### Reviewer Suggestions:

None.


syclik:
Jenkins, retest this please.

PR #1671, opened 2015-11-07
Updated initialize_state to use the writer

syclik:
#### Summary:

Change initialize_state to use the writer callbacks.

#### Intended Effect:

Refactor code.

#### How to Verify:

Run.

#### Side Effects:

CmdStan, RStan, and PyStan will need the writer. Action is on me.

#### Documentation:

Included in the doc.

#### Reviewer Suggestions: 

@betanalpha

### DIFF
--- a/src/stan/services/init/initialize_state.hpp
+++ b/src/stan/services/init/initialize_state.hpp
@@ -25,7 +25,6 @@
 namespace stan {
   namespace services {
     namespace init {
-      using stan::interface_callbacks::writer::base_writer;
       namespace {
 
         /**
@@ -119,7 +118,8 @@ namespace stan {
       template <class Model>
       bool initialize_state_values(Eigen::VectorXd& cont_params,
                                    Model& model,
-                                   base_writer& writer) {
+                                   interface_callbacks::writer::base_writer&
+                                   writer) {
         try {
           validate_unconstrained_initialization(cont_params, model);
         } catch (const std::exception& e) {
@@ -171,7 +171,7 @@ namespace stan {
       template <class Model>
       bool initialize_state_zero(Eigen::VectorXd& cont_params,
                                  Model& model,
-                                 base_writer& writer) {
+                                 interface_callbacks::writer::base_writer& writer) {
         cont_params.setZero();
         return initialize_state_values(cont_params, model, writer);
       }
@@ -195,7 +195,7 @@ namespace stan {
                                    Eigen::VectorXd& cont_params,
                                    Model& model,
                                    RNG& base_rng,
-                                   base_writer& writer) {
+                                   interface_callbacks::writer::base_writer& writer) {
         int num_init_tries = -1;
 
         boost::random::uniform_real_distribution<double>
@@ -259,7 +259,7 @@ namespace stan {
                                               Eigen::VectorXd& cont_params,
                                               Model& model,
                                               RNG& base_rng,
-                                              base_writer& writer,
+                                              interface_callbacks::writer::base_writer& writer,
                                               ContextFactory& context_factory) {
         try {
           boost::random::uniform_real_distribution<double>
@@ -347,7 +347,8 @@ namespace stan {
                                    Eigen::VectorXd& cont_params,
                                    Model& model,
                                    RNG& base_rng,
-                                   base_writer& writer,
+                                   interface_callbacks::writer::base_writer&
+                                   writer,
                                    ContextFactory& context_factory,
                                    bool enable_random_init = false,
                                    double R = 2) {
@@ -417,7 +418,7 @@ namespace stan {
                             Eigen::VectorXd& cont_params,
                             Model& model,
                             RNG& base_rng,
-                            base_writer& writer,
+                            interface_callbacks::writer::base_writer& writer,
                             ContextFactory& context_factory,
                             bool enable_random_init = false,
                             double init_r = 2) {
--- a/src/stan/services/init/initialize_state.hpp
+++ b/src/stan/services/init/initialize_state.hpp
@@ -171,7 +171,7 @@ namespace stan {
       template <class Model>
       bool initialize_state_zero(Eigen::VectorXd& cont_params,
                                  Model& model,
-                                 interface_callbacks::writer::base_writer& writer) {
+                             interface_callbacks::writer::base_writer& writer) {
         cont_params.setZero();
         return initialize_state_values(cont_params, model, writer);
       }
@@ -195,7 +195,7 @@ namespace stan {
                                    Eigen::VectorXd& cont_params,
                                    Model& model,
                                    RNG& base_rng,
-                                   interface_callbacks::writer::base_writer& writer) {
+                             interface_callbacks::writer::base_writer& writer) {
         int num_init_tries = -1;
 
         boost::random::uniform_real_distribution<double>
@@ -259,7 +259,7 @@ namespace stan {
                                               Eigen::VectorXd& cont_params,
                                               Model& model,
                                               RNG& base_rng,
-                                              interface_callbacks::writer::base_writer& writer,
+                               interface_callbacks::writer::base_writer& writer,
                                               ContextFactory& context_factory) {
         try {
           boost::random::uniform_real_distribution<double>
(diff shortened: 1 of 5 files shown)


src/stan/services/init/initialize_state.hpp
```diff
@@ -24,7 +25,7 @@
 namespace stan {
   namespace services {
     namespace init {
-
+      using stan::interface_callbacks::writer::base_writer;
```
syclik: I did this to cut down on the line length. If anyone objects, I can get rid of this line and fully specify the writer in the function declarations.

src/stan/services/init/initialize_state.hpp
```diff
@@ -113,52 +114,44 @@ namespace stan {
        *                            right size and set to 0.
        * @param[in,out] model       the model. Side effects on model? I'm not
        *                            quite sure
-       * @param[in,out] output      output stream for messages
+       * @param[in,out] writer      writer callback for messages
```
syclik: @betanalpha, I got rid of the template. That ok with you?

src/stan/services/init/initialize_state.hpp
```diff
@@ -113,52 +114,44 @@ namespace stan {
...
       bool initialize_state_values(Eigen::VectorXd& cont_params,
                                    Model& model,
-                                   std::ostream* output) {
+                                   base_writer& writer) {
         try {
           validate_unconstrained_initialization(cont_params, model);
         } catch (const std::exception& e) {
-          if (output)
-            *output << e.what() << std::endl;
+          writer(e.what());
```
syclik: @syclik: fix indentiation.

src/stan/services/init/initialize_state.hpp
```diff
@@ -113,52 +114,44 @@ namespace stan {
...
           return false;
         }
         double init_log_prob;
         Eigen::VectorXd init_grad = Eigen::VectorXd::Zero(model.num_params_r());
         try {
           stan::model::gradient(model, cont_params, init_log_prob,
-                                init_grad, output);
+                                init_grad);
```
syclik: @betanalpha, I was taking your lead here in getting rid of the output. Is that the right behavior?

src/stan/services/init/initialize_state.hpp
```diff
@@ -219,20 +212,23 @@ namespace stan {
...
-          if (output)
-            *output << std::endl << std::endl
-            << "Initialization between (" << -R << ", " << R
-            << ") failed after "
-            << MAX_INIT_TRIES << " attempts. " << std::endl
-            << " Try specifying initial values,"
-            << " reducing ranges of constrained values,"
-            << " or reparameterizing the model."
-            << std::endl;
+          std::stringstream R_ss, MAX_INIT_TRIES_ss;
```
syclik: @bob-carpenter or @bgoodri, is there a better way to output `double` and `int` as strings?

src/stan/services/init/initialize_state.hpp
```diff
@@ -290,34 +286,36 @@ namespace stan {
...
+                              cont_vecs_constrained, false, false, 0);
             std::vector<std::vector<size_t> > dims;
             model.get_dims(dims);
             stan::io::array_var_context random_context(cont_names,
                                                        cont_vecs_constrained,
                                                        dims);
             stan::io::chained_var_context cvc(context, random_context);
-            model.transform_inits(cvc, cont_params, output);
-            if (initialize_state_values(cont_params, model, output))
+            model.transform_inits(cvc, cont_params, 0);
```
syclik: @betanalpha, we're removing output from `model.*`. That's the right behavior, right? (I took this from the 1361 branch.)

syclik:
I'm ok with this. I think we're ok with passing 0 into the model's functions requiring `std::ostream*`.

src/stan/services/init/initialize_state.hpp
bgoodri: Not without C++11 AFAIK

syclik: Thanks. We'll leave it then.

src/stan/services/init/initialize_state.hpp
betanalpha: My only concern is that this will potentially affect the entire namespace and it won't be obvious where the typedef is defined if base_writer is used somewhere else.  Is it too pedantic to add the using to every file?

src/stan/services/init/initialize_state.hpp
betanalpha: It's this, using the old C conversion functions, or Boost's lexical cast.

src/stan/services/init/initialize_state.hpp
betanalpha: This stuff will all change with the var_context refactor so it may not matter for long.  We can remove the output or use local string streams to capture any error output and pass them to the writer.

src/stan/services/init/initialize_state.hpp
betanalpha: What template?  I don't see any template arguments changing in the diff.

src/stan/services/init/initialize_state.hpp
betanalpha: Gradient passes errors through an exception so no output should be needed.  May want to double check.

betanalpha:
:+1:

src/stan/services/init/initialize_state.hpp
syclik: Ok. Great.

src/stan/services/init/initialize_state.hpp
syclik: There's no typedef. Just a using.

src/stan/services/init/initialize_state.hpp
syclik: In the original implementation, there is a template parameter.

src/stan/services/init/initialize_state.hpp
betanalpha: Sorry, meant the using.  That carries over to the entire namespace, right?
Which means that another file in init can use “base_writer” without a scope
and there won’t be any local indication of what that scope might be.

src/stan/services/init/initialize_state.hpp
betanalpha: I guess given how little this function will be called the template isn't necessary and we can go with virtual functions.

src/stan/services/init/initialize_state.hpp
syclik: Yes, but I think we'll be alright. We don't have too much in the stan::interface_callbacks::writer namespace.

src/stan/services/init/initialize_state.hpp
bob-carpenter: I don't know a better way to do it.  

If we're finding ourselves doing this a lot, we can write
our own functions to do the conversions that lean on the
streams.  The streams have the property of always using
whatever the stream's precision is for the output.

By the way, is that something we're going to control in
the writer?  The digits thing is a big deal.  I think 
C++ defaults are higher than R defaults for print.  Always
a problem with ASCII formats.  The higher precision means
more storage requirements and slower I/O.

- Bob

src/stan/services/init/initialize_state.hpp
bob-carpenter: -1 on namespace-level includes.  Just spell it all out of you have to.  You can drop the `stan::` by the way, too.

syclik: Removed.

src/stan/services/init/initialize_state.hpp
syclik: Double checked. It doesn't output through streams.

src/stan/services/init/initialize_state.hpp
syclik: Agreed. I think we can do a better job of printing NaN and Inf on Windows. And control the output so it's more uniform across platforms. I think we should open up a new issues for this topic.

syclik:
Thank, all. We're ready to go once it passes tests.

syclik:
Jenkins, retest this please.

PR #1677, opened 2015-11-09
Updating ps_point and points to use writer callback

syclik:
#### Summary:

Use the base_writer in ps_point and *_point classes.

#### Side Effects:

None.

#### Documentation:

None.

#### Reviewer Suggestions: 

This is @betanalpha's code. I reviewed it and it's ok.


src/test/unit/mcmc/hmc/hamiltonians/ps_point_test.cpp
```diff
@@ -62,10 +63,9 @@ namespace stan {
       stan::test::capture_std_streams();
 
       ps_point point(2);
-      EXPECT_NO_THROW(point.write_metric(0));
-
       std::stringstream out;
-      EXPECT_NO_THROW(point.write_metric(&out));
+      stan::interface_callbacks::writer::stream_writer writer(out);
```
syclik: Note: this construction is here while we get the writer worked through the more of the code.

PR #1681, opened 2015-11-11
Feature/issue 1057 generalize indexing

bob-carpenter:
#### Summary:

Allow multi-indexing in expressions and assignment statements.

#### Intended Effect:

Adds indexing functions, types, and extends language to allow multiple indexes (lower bounds, upper bounds, lower and upper bounds, array of int, all) to both expressions and statements.

The API code is in `stan/model/indexing` contains the following functionality:

* `index.hpp`: structures for single and multiple indexing
* `index_list.hpp`: sequence of indexes
* `deep_copy.hpp`:  deep copy arbitrary Stan data types (avoids aliasing)
* `rvalue_index_size.hpp`: metaprogram for size of result container given index and size of underlying container
* `rvalue_at.hpp`: map index object plus result position to underlying position
* `rvalue_return.hpp`: metaprogram for calculating return type of indexing
* `lvalue.hpp`: assignment value to variable given an index list

The structures in index lists are mirrored in the abstract syntax tree (`stan/lang/ast.hpp`) and it also contains code for type inference.  There's also a new program to compute whether a variable shows up in an expression to detect the possibility of aliasing.

The C++ code generator changes are surprisingly straightforward and in the usual place (`stan/lang/generator.hpp`) --- the only hitch is checking if the left-hand variable occurs in the right-hand side expression, which causes the right side to be wrapped in a deep copy.  The harder part is generating neat code for user-facing and code-generation purposes --- there is separate logic for expressions printed for users and for C++.

The parser code is in `stan/lang/grammars/':

* `indexes_grammar...`: grammar for sequence of single and multiple indexes
* `statement_grammar...`: adds multiple assignment case with semantic actions
* `term_grammar...`: adds case for multiple indexed expression with semantic actions

In the statement and term grammar, the original logic remains, but now no loger has expectation operators `>` but just concatenation `>>` so that failure to match can backtrack to the multiple indexing.  This means multiple indexing has to deal with informative error messages.

#### How to Verify:

There are full unit tests for the boundary conditions in the API and template metaprograms.  There are also exhaustive unit tests for up to three dimensions of indexes both at the API level and the grammar/parser level that verify values and sizes are all correctly assigned.  (There's a repeated tested idiom with helper functions for expressions and statements and a whole lot of cut and paste and emacs macros in the grammar examples.)

#### Side Effects:

Error messages are slightly different for indexing errors because they are handled in the multi-indexing blocks.  Tests have been updated to reflect this change.

Expressions and assignment statements with no multiple indexing (that is, no indexing or just single integer indexes) will be parsed and code-generated as before.

#### Documentation:

New manual chapter for user's guide and new chapter for reference manual;  updated BNF description and added some cross-references.

API functions all documented.

#### Reviewer Suggestions:

Daniel (he'll be pickiest and this one should be looked at pretty closely)


syclik:
@bob-carpenter, this failure isn't the fault of this pull request. It looks like we snuck in some compiler warnings prior to this and Jenkins wasn't smart enough to pick it out. I think the warnings have come from the new clang++ compiler on Mac.

syclik:
I take it back... Bob, does this run on your machine:
`make src/test/test-models/good/indexing/good-ints.hpp-test`

It first compiles `src/test/test-models/good/indexing/good-ints.stan` to `src/test/test-models/good/indexing/good-ints.hpp`. Then it tries to compile the header file.

Mine doesn't compile.

bob-carpenter:
Nope, broken here, too.  I'll fix and push a patch.

Is there a make target for the integration tests?  I don't see
it with "make help" --- shouldn't it be under "Tests:"?

- Bob

syclik:
It's just:
`./runTests.py src/test/integration`
(or `./runTests.py src/test/integration/compile_models_test.cpp`)

I did some make magic to get all the good models to be run through stanc
then compiled with the C++ compiler.

bob-carpenter:
Cool, 

  > ./runTests.py -j4 src/test

seems to kick everything off.  May I update the make doc while
I'm at it to make this more visible?

- Bob

  ./runtests.py

syclik:
Yes, please.

syclik:
@bob-carpenter, this all looks great! Thanks for adding the doc and the tests.

PR #1684, opened 2015-11-12
Update ADVI's stepsize sequence for more reliable convergence

akucukelbir:
#### Summary:
This implements three changes all related to convergence of ADVI.
- The learning rate is now multiplied by `1/t^{1/2}`. This forces the optimization procedure to converge to a point and satisfy Robbins-Monro assumptions for various stochastic approximation theory. RMSProp, which is what is currently implemented, does not satisfy this and currently works based on only heuristics from the deep learning literature. Technically, the rate should be `1/t^{1/2 + epsilon}` for some `epsilon>0`, but due to the efficiency of square roots, we're doing `1/t^{1/2}` instead.
- The buffer size when looking at changes in the ELBO is forced to at least be of size 2. Previously it was only of size 1.
- This signals the "may be diverging" message after `2*eval_elbo_` (default is `2*100=200`), rather than after 100. This is based on two reasons: 1. It's obvious that the message should occur after a constant multiple of `eval_elbo_`, because that is when a new line of convergence information is given, and `eval_elbo_` is not necessarily 100; 2. There should be more leeway than after 1 line to spit out if the procedure might be diverging. Naturally then, why 2 instead of some other multiple? It's a heuristic we base on what we see in practice; we won't be able to solve this by finding an "optimal" number any time soon.

#### Intended Effect:
Optimization is forced to converge to a point and satisfy Robbins-Monro assumptions for stochastic approximations.

#### How to Verify:
Run unit tests.

#### Side Effects:
None

#### Documentation:
N/A

#### Reviewer Suggestions:
N/A


akucukelbir:
@syclik @bob-carpenter: ping.

this one should (hopefully) be a straightforward merge too. the upcoming pull requests will come with new unit and improved tests, for sure.

bob-carpenter:
This isn't part of the diff, but it looks like there's an initialization issue starting on line 217 of `advi.hpp`:

```
        int iter_counter = 1;
        while (do_more_iterations) {
          // Compute gradient using Monte Carlo integration
          calc_ELBO_grad(variational, elbo_grad);

          // RMSprop moving average weighting
          if (iter_counter == 0) {
            params_adagrad += elbo_grad.square();
          } else {

          ...

          ++iter_counter;
        }

```

The counter is initialized to 1 then is tested against 0.  The only code in the loop increments it by one.  

* [ ] figure out what should be happening with the above for initialization (either remove the conditional behavior because it's unnecessary, init at 0, or change the test to 1).

* [ ] For clarity, rewrite the while loop as a for loop (it helps clarify scope and idiomatically the counter tends not to get changed in the block):

```
for (int iter_counter = 0; do_more_iterations; ++iter_counter) {
  ...
}
```

There's no other increment of `iter_counter` anywhere.  

* close curly brace 2 spaces too far in on line 81.

bob-carpenter:
On lines 103--119 or so, you're using a try/catch construction which as far as I can tell should be a conditional that deals with the exceptional case.  So rather than this:

```
          try {
            double energy_i = model_.template log_prob<false, true>(zeta,
              print_stream_);
            stan::math::check_finite(function, "energy_i", energy_i);
            elbo += energy_i;
            i += 1;
          } catch (std::exception& e) {
            this->write_error_msg_(print_stream_, e);
            n_monte_carlo_drop += 1;
            if (n_monte_carlo_drop >= n_monte_carlo_elbo_) {
              const char* name = "The number of dropped evaluations";
              const char* msg1 = "has reached its maximum amount (";
              int y = n_monte_carlo_elbo_;
              const char* msg2 = "). Your model may be either severely "
                "ill-conditioned or misspecified.";
              stan::math::domain_error(function, name, y, msg1, msg2);
            }
```

I'm pretty sure you want to do this (but not so sure I was going to just change it myself):

```
double energy_i = model_.template log_prob<false, true>(zeta, print_stream_);
if (!math::is_finite(energy_i)) {
  this->write_error_msg_(print_stream_, "energy_i is infinite");
  n_monte_carlo_drop += 1;
  if (n_monte_carlo_drop >= n_monte_carlo_elbo_) {
    const char* name = "The number of dropped evaluations";
    const char* msg1 = "has reached its maximum amount (";
    int y = n_monte_carlo_elbo_;
    const char* msg2 = "). Your model may be either severely "
        "ill-conditioned or misspecified.";
    stan::math::domain_error(function, name, y, msg1, msg2);
}
elbo += energy_i;
++i;
```

I also replaced the final `i += 1` with `++i`.

akucukelbir:
thank you @bob-carpenter. great catch!

akucukelbir:
hmm. yes, i think those should be equivalent.

bob-carpenter:
I updated the comment --- turns out we go straight to Boost for finiteness checks and I added a couple more things.

akucukelbir:
shall i make your suggested change and push again?

bob-carpenter:
Ack --- they're not quite going to be the same.  As you have it, if the `log_prob` function throws an exception, you catch it in this loop.  If that was your intent, you really have to do it your way.  Sorry for the misdirection (though the general comments on preferable style stand!).

dustinvtran:
The problem with this change is the line
```{C++}
double energy_i = model_.template log_prob<false, true>(zeta, print_stream_);
```
outside the try-catch. When I first implemented this I recall that running `log_prob` can spit out an error and halt the program. (I could be wrong, in which otherwise the change looks good to me)

akucukelbir:
ok. let's see if the current commit passes the tests. if so, let's leave this as is. we can re-think bob's suggestion down the road, perhaps.

bob-carpenter:
I don't think my suggestion is workable for the reasons
Dustin (and I, after thinking about it) laid out.  I also think
I may have suggested this before, so doubly sorry about bringing it
up again.  If I bring it up a third time, I'll comment it :-)

akucukelbir:
no worries.

and it passes! :)

akucukelbir:
@bob-carpenter @syclik friendly ping :)

src/stan/variational/advi.hpp
```diff
@@ -280,7 +281,7 @@ namespace stan {
               do_more_iterations = false;
             }
 
-            if (iter_counter > 100) {
+            if (iter_counter > 2*eval_elbo_) {
```
syclik: @akucukelbir, was this intentional?

akucukelbir: yes. see the third bullet point in the summary that @dustinvtran provided

PR #1686, opened 2015-11-13
add support for FreeBSD

maxc01:
Just make a small change in Makefile and everything works without flaws, under g++ 4.8.5 and clang++ 3.4.1. 

There is no guarantee that further version will work so well, to separate system difference, a new file `os_freebsd` is created, which is just a copy of `os_linux`.


syclik:
@maxc01, do you see the configuration drifting? If not, could I suggest we only change `make/detect_os`:

```
ifeq ($(uname_S), FreeBSD)
  OS_TYPE := linux
endif
```

I'd prefer not to have a duplicate of the linux configuration if possible. If you think these will have to be configured separately, then it makes sense to have a separate configuration file.

maxc01:
@syclik FreeBSD and Linux has different directory structure, under FreeBSD, the header file is in /usr/local/include and lib file is in /usr/local/lib

1. Now `stan` puts its dependencies in `stan/lib/stan_math/lib`, but there is no guarantee that further version of these dependencies will still build well under FreeBSD. If this bad situation happens, `stan` won't build. In this case, a FreeBSD user can change `CFLAGS` and `LDFLAGS` to the system libraries(boost, eigen, etc.), FreeBSD development teams will make effects to make these libraries to build and perform well and this can be done by just changing from `/usr/include` to `/usr/local/include` and from `/usr/lib` to `/usr/local/lib`.
2. If further version of `stan` depends on some system libraries, `stan` will build fail for the same resaon.

But if these situations won't happen, `OS_TYPE := linux` is sufficient. But I think it is a safer choice to have a separate configuration file.

make/os_freebsd
```diff
@@ -0,0 +1,36 @@
...
+##
+
+ifeq (g++,$(CC_TYPE))
+  LDFLAGS += -lpthread
+  CFLAGS_GTEST += -DGTEST_HAS_PTHREAD=0
+  ifeq (4.8,$(CC_MAJOR).$(CC_MINOR))
+    CFLAGS += -Wno-unused-local-typedefs
+  endif
+endif
+ifeq (mingw32-g++,$(CC_TYPE))
```
ariddell: Is there mingw32 on FreeBSD?

maxc01: Yes, FreeBSD has mingw32.

ariddell: I see. It's for cross-compilation to Windows.

syclik:
@maxc01, I'm approaching this from the maintainability of the project and I think it's safer with less files. That is, until it's necessary to have different options on FreeBSD.

The names are a bit misleading and maybe we should be splitting configuration based on compilers. I think the reason we've split these according to OS was due to Mac OSX having different compiler options for the same version number of the g++ compiler.

bob-carpenter:
@maxc01 Can you verify you're OK with our releasing your contribution under the new BSD license (what we use for the core Stan code)?  You'll retain copyright for your work.  We're trying to make this process more formal, and we're starting here.  Sorry you're the guinea pig!

maxc01:
@syclik If further version of `stan` needs user to install some dependencies by themselves, for example, `stan` needs users to install `openblas` or something else, in this case, Linux and FreeBSD need different configuration. But if this is not a problem, then I prefer the single one configuration, it is cleaner and  easier to maintain and thus a safer choice.

maxc01:
@bob-carpenter Sure:)

PR #1693, opened 2015-11-16
Implement ADVI variational family zero constructor and set to zero functionality 

akucukelbir:
#### Summary:
Implements two changes to the `meanfield` and `fullrank` variational families in ADVI:

1. Implements a new constructor that only takes in a `size_t dimension` and initializes everything to a zero vector/matrix, as appropriate. 
2. Implements a `set_to_zero()` function that sets everything to a zero vector/matrix, as appropriate.

These are both important because we use instantiations of these objects to track the norm of our gradients across iterations. Initializating them to zero is useful. Being able to set them to zero will become important when we merge the automatic tuning branch for `eta`.

#### How to Verify:
I've updated the unit tests. Should be straightforward to see the changes.

#### Side Effects:
None

#### Documentation:
None.

#### Reviewer Suggestions:
@dustinvtran for variational accuracy, @syclik @bob-carpenter for overall accuracy and mergability.


src/stan/variational/families/normal_fullrank.hpp
```diff
@@ -49,12 +49,15 @@ namespace stan {
 
     public:
       // Constructors
+
+      // Constructs a zero-filled variational distribution
       explicit normal_fullrank(size_t dimension) :
         dimension_(dimension) {
         mu_     = Eigen::VectorXd::Zero(dimension_);
-        L_chol_  = Eigen::MatrixXd::Identity(dimension_, dimension_);
+        L_chol_ = Eigen::MatrixXd::Zero(dimension_, dimension_);
```
bob-carpenter: These shold all be initializers, as in `mu_(Eigen::VectorXd::Zero(dimension))`.

You should prefer to use local variables like `dimension` instead of member variables like `dimension_`.  Just make sure to put them in the same order as the declaration, where dimension is last.

I'm like a broken record in this regard, but wherever possible, it's better to have immutable objects (const member variables in C++ terms).  The main reason is that it's easier to make sure the class is in a sound state when it's constructed.  If you have setters, you need to be extra careful where you validate that a class is finally ready to go.  Here, the class isn't ready to go presumably until L_chol_ is set to something more useful.

Why set L_chol_ to the identity?

We would very strongly prefer doxygen doc on the classes and functions, not line comments.  No need to add comments like "Constructors" --- follow the Google style guide and assume the people reading your code know C++ and can tell that's a constructor by its name.

bob-carpenter: The constructor with a Cholesky factor should make sure the diagonal is positive so that it's the Cholesky factor of a positive definite matrix.

bob-carpenter: As far as I can tell, nothing ever changes the dimension so it should be declared `const`.

bob-carpenter: You should pull out the Cholesky factor validation --- as is, it's cut and pasted into the mutator (setter) and constructor.  Yet another reason why mutators are evil.

bob-carpenter: I'm just piling things on here that aren't local.  The entropy() function should declare the `tmp` variable where it's used.  There's nothing to be gained from declaring a primitive higher up, whereas declaring it further down makes the code more readable because its scope matches where it's used.

bob-carpenter: Arguments should be declared `const` wherever possible.  For instance, the model m in `calc_grad` (at least I hope the functions are const enough in the model --- if not, let me know and I'll fix that --- you won't be able to fix it in this pull request). 

It would help if you provided `@param[in,out]` with in for when parameters must be defined coming in and out when they're modified.  Obviously, the in only ones should be constant (other than primitives, which don't need const declarations).

bob-carpenter: I think it'd be cleaner with a k index on the outside and i and j on the inside rather than i, ii, and jj in calc_grad.  No big deal if you're already fed up with these comments.  We tend to use ii and jj for random effects indexing following Gelman and Hill's book.

bob-carpenter: I would prefer

```
for (int i; i < monte_carlo_grad; ) 
  ...
```

to

```
int i = 0;
while (i < monte_carlo_grad)
  ...
```

bob-carpenter: In the entropy calculation, why the loop to avoid zeroes.  Those can be super costly computationally in branch-point misprediction.  Shouldn't the diagonals be positive by construction and constraint?  And if they are not positive (the code only checks identity to zero, but there are also negative and not-a-number values), shouldn't the entropy explode?

Also, operators should start lines, not end lines as in the variable `result` for `entropy()`.

src/stan/variational/families/normal_fullrank.hpp
```diff
@@ -280,12 +288,11 @@ namespace stan {
             }
             i += 1;
           } catch (std::exception& e) {
-            this->write_error_msg_(print_stream, e);
             n_monte_carlo_drop += 1;
-            if (n_monte_carlo_drop >= 5*n_monte_carlo_grad) {
+            if (n_monte_carlo_drop >= 10*n_monte_carlo_grad) {
```
bob-carpenter: You need space between operators --- isn't cpplint catching this?

src/stan/variational/families/normal_fullrank.hpp
```diff
@@ -280,12 +288,11 @@ namespace stan {
...
               const char* name = "The number of dropped evaluations";
               const char* msg1 = "has reached its maximum amount (";
-              int y = 5*n_monte_carlo_grad;
+              int y = 10*n_monte_carlo_grad;
```
bob-carpenter: Rather than cut-and-paste the magic number 10 in two places, it's better to call it out as a named static constant.

src/stan/variational/families/normal_meanfield.hpp
```diff
@@ -262,12 +269,11 @@ namespace stan {
...
             i += 1;
           } catch (std::exception& e) {
-            this->write_error_msg_(print_stream, e);
             n_monte_carlo_drop += 1;
-            if (n_monte_carlo_drop >= 5*n_monte_carlo_grad) {
+            if (n_monte_carlo_drop >= 10*n_monte_carlo_grad) {
               const char* name = "The number of dropped evaluations";
               const char* msg1 = "has reached its maximum amount (";
-              int y = 5*n_monte_carlo_grad;
+              int y = 10*n_monte_carlo_grad;
```
bob-carpenter: Especially since the pesky `10` keeps showing up.

bob-carpenter:
Overall this looks great.  I like the way it's being done with operators.  

Please take that stream of comments constructively!  Most of them are related to code and comment style more than substance.  Not to say the style isn't important --- it is for readability and maintainability.

dustinvtran:
Thanks Bob. Many of your suggestions you made at some point earlier, which I implemented in another branch. :) Not sure whether/how to tease out the various changes from there to here.

akucukelbir:
hi @bob-carpenter,

i'll make the changes to `10` and push.

we'll make the other internal changes afterwards. let's just get the functionality changes in for this pull request.

cheers
alp

p.s. doing `mu_(Eigen::VectorXd::Zero(dimension))` gives
> error: no matching function for call to
      object of type 'Eigen::VectorXd' (aka 'Matrix<double, Dynamic, 1>')
lib/stan_math/lib/eigen_3.2.4/Eigen/src/Core/DenseCoeffsBase.h:392:5: note: candidate function not
      viable: no known conversion from 'const Eigen::VectorXd'
      (aka 'const Matrix<double, Dynamic, 1>') to 'Index' (aka 'long') for 1st argument
    operator()(Index index)

syclik:
Jenkins, retest this please.

bob-carpenter:
@syclik Are you OK with merging this without doxygen doc and without the fixes I suggested and then coming back and making more fixes later?  I don't know what this is holding up in terms of branches.  

I should've just made the fixes myself rather than leaving a gazillin comments in the first place.

bob-carpenter:
I'll clean this up a bit tonight and push.

akucukelbir:
many thanks bob. i'm working hard on the next pull request, which implements the actual "autotuning" of eta. that pull request merits close scrutiny. so i want to make sure we all have enough time to evaluate it.

bob-carpenter:
I'm almost done.  There's a lot to doc in there!

- Bob

bob-carpenter:
In `normal_fullrank()`, the definition of `entropy()` looks wrong.   The doc says this:

```
	// 0.5 * dim * (1+log2pi) + 0.5 * log det (L^T L) =
	// 0.5 * dim * (1+log2pi) + sum(log(abs(diag(L))))
```

and implements it as the second line.  But we don't necessarily have a diagonal Cholesky factor, so I don't think you can reduce it as specified.

bob-carpenter:
I finished the doc, but now I can't find the test for `variational/families/normal_fullrank.hpp`.  It should be in `src/test/unit/variational/families/normal_fullrank.hpp` with the same path.

bob-carpenter:
Ah --- I found it in `test/unit/variational/advi_params_fullrank_test.cpp`. 

1. I'll rename it and move it to where it should be.

2.  Someone needs to write the rest of the tests.  Very few of the functions in the file are tested in the unit test.   We want to test not only the calls, but that the right error messages are thrown when an erroneous input is included.

bob-carpenter:
I think this code was being written with the intention of maintaining the invariant that entries are not NaN.  There are several places I just documented where that can happen, such as in multiplying by a scalar.  That scalar input's not checked.  This is OK if the algorithms using this know that it's their job to guarantee the invariant.  I made sure to doc that the invariant is not maintained and where it can be violated.

bob-carpenter:
Never mind w.r.t. https://github.com/stan-dev/stan/pull/1693#issuecomment-157909806;  I see what's going on now.  It's using `det(L * L') = det(L) * det(L')`, and `L` is guaranteed to be diagonal.  Then you add both on the log scale and the 0.5 goes away.  Very neat.  No wonder everyone likes Cholesky factors!

bob-carpenter:
I'll fix the doc and location of test for `normal_meanfiled.hpp`, too.  It'll probably take the rest of the night, but hopefully all the tests will have passed by morning.

Then we'll need another issue, I think, to add tests for the functions that are missing tests unless those are already in another branch somewhere.

bob-carpenter:
I pushed all the changes to the doc and moved and renamed the two tests that were in the wrong place.

We can leave future improvements to future pull requests.  In particular, more tests unless Daniel wants to tag in and add some before merging this.

bob-carpenter:
Oh, and I also changed the return types of some of the operators to the more conventional reference types.

bob-carpenter:
Alp and Dustin should review.  I didn't change any functionality.  And I didn't clean up the testing into its own function in mean field.

akucukelbir:
thank you so much @bob-carpenter. i'll take a look shortly.

bob-carpenter:
@syclik I don't know what's up with the Travis CI build --- it looks like it's hung.  And I don't know why the default says "No test results found." but has a check next to it.

syclik:
I'm looking into it.

PR #1694, opened 2015-11-17
Fix typo in initialize_state.hpp

ariddell:
Visual Studio choked on this. Not sure why the other compilers didn't. 

Looks like @syclik has been working on this most recently.


syclik:
@ariddell, thanks! That's clearly the wrong thing.

I might add a few tests before merging to prevent this mistake from happening again and hopefully I haven't made the same mistake elsewhere.

PR #1699, opened 2015-11-19
ADVI stepsize sequence parameter (eta) adaptation

akucukelbir:
#### Summary:

(Note: goes with https://github.com/stan-dev/cmdstan/pull/223)

Implements a heuristic grid-search for selecting a decent value for `eta`.

- We do a grid search over `eta = (100, 10, 1, 0.1, 0.01)`
- The grid search is robust; if `eta=100` is so big that ADVI diverges, it does not halt Stan.
- We implement an adaptation argument structure that mimics NUTS. The new syntax is
```
./my_stanprogram variational eta=[1.0] adapt engaged=[0,1] iter=[50]
```
If `adapt engaged=1`, then the value for `eta` (default or manual) is ignored. If `adapt engaged=0`, then the value for `eta` is used without any adaptation. `adapt iter` is the number of iterations to spend on each configuration of `eta` during the grid search.

##### TODO
- [x] Convert `queue` logic to something else.

#### Intended Effect:

Makes ADVI more "black box". It should converge faster and more reliably for more models than before.

#### How to Verify:

Run your favorite model. Run the tests. 

```
$ ./runTests.py src/test/unit/variational
```

It needs more tests. Here's a list:

- [x] Need to test `adapt_eta()` in `advi.hpp`
- [x] Need to test `stochastic_gradient_ascent()` in `advi.hpp`
- [x] Need to test `src/stan/services/variational/print_progress.hpp`

#### Side Effects:

We write the "adapted" value of eta into `output.csv`. This breaks the current `stanprint` summary functionality. So I had to adjust `src/stan/io/stan_csv_reader.hpp`. We could drop that change and take care of it elsewhere.

We no longer evaluate `lp` for all the draws from the posterior. I don't know when/how/why we decided to do this, but it's not useful and it slows down the final stage of ADVI significantly.

#### Documentation:

- write documentation in cmdstan manual (separate PR)
- write documentation in stan manual (separate PR)

#### Reviewer Suggestions:

Everyone. I'd like everyone to be happy about these changes so that PyStan and RStan can interface to ADVI.


akucukelbir:
@bob-carpenter @syclik this is gonna be fun :)

src/stan/services/arguments/arg_variational.hpp
```diff
@@ -31,6 +31,7 @@ namespace stan {
                                  "of ELBO (objective function)",
                                  100));
         _subarguments.push_back(new arg_variational_eta());
```
bob-carpenter: The convention for member variables is to use final underscores, not initial ones.  Initial ones will run into trouble with the C++ spec if they get capitalized.

src/stan/services/arguments/arg_variational_adapt.hpp
```diff
@@ -0,0 +1,25 @@
...
+#include <stan/services/arguments/categorical_argument.hpp>
+#include <stan/services/arguments/arg_variational_adapt_engaged.hpp>
+#include <stan/services/arguments/arg_variational_adapt_iter.hpp>
+
+namespace stan {
+  namespace services {
+
+    class arg_variational_adapt: public categorical_argument {
+    public:
+      arg_variational_adapt() {
```
bob-carpenter: OK, I see this is widespread.  Is this required by the way the arguments are set up now?  If so, it all needs to be changed, but not as part of this pull request.

src/stan/services/arguments/arg_variational_adapt_engaged.hpp
```diff
@@ -0,0 +1,26 @@
...
+  namespace services {
+
+    class arg_variational_adapt_engaged: public bool_argument {
+    public:
+      arg_variational_adapt_engaged(): bool_argument() {
+        _name = "engaged";
+        _description = "Adaptation engaged?";
+        _validity = "[0, 1]";
+        _default = "1";
+        _default_value = true;
```
bob-carpenter: I think if we say the default is "1", then the default value should be 1, not true.  Or we should change the way the default's written.  This is again probably copied from other examples which should also be changed.

src/stan/services/variational/print_progress.hpp
```diff
@@ -0,0 +1,69 @@
...
+                                   refresh);
+
+        int it_print_width = std::ceil(std::log10(static_cast<double>(finish)));
+        if (io::do_print(m - 1, (start + m == finish), refresh)) {
+          o << prefix;
+          o << "Iteration: ";
+          o << std::setw(it_print_width) << m + start
+            << " / " << finish;
+          o << " [" << std::setw(3)
+            << static_cast<int>( (100.0 * (start + m)) / finish )
```
bob-carpenter: This should be "100" rather than "100.0" and then you don't need the cast.

src/stan/variational/advi.hpp
```diff
@@ -96,32 +104,29 @@ namespace stan {
...
+        int n_dropped_evaluations = 0;
+        for (int i = 0; i < n_monte_carlo_elbo_;) {
           variational.sample(rng_, zeta);
           try {
-            double energy_i = model_.template log_prob<false, true>(zeta,
+            double log_prob = model_.template log_prob<false, true>(zeta,
               print_stream_);
-            stan::math::check_finite(function, "energy_i", energy_i);
-            elbo += energy_i;
-            i += 1;
```
bob-carpenter: ++i

src/stan/variational/advi.hpp
```diff
@@ -150,18 +157,170 @@ namespace stan {
...
+                                   "Number of adaptation iterations",
+                                   adapt_iterations);
+
+
+        if (print_stream_) {
+          *print_stream_ << "Begin eta adaptation." << std::endl;
+        }
+
+        // Sequence of eta values to try during adaptation
+        std::queue<double> eta_sequence;
```
bob-carpenter: I don't think we need to build a queue at runtime for this --- it's just iterating through a fixed sequence.

src/stan/variational/advi.hpp
```diff
@@ -281,7 +438,7 @@ namespace stan {
               do_more_iterations = false;
             }
 
-            if (iter_counter > 2*eval_elbo_) {
+            if (iter_counter > 10*eval_elbo_) {
```
bob-carpenter: space between operators --- didn't cpplint catch this one?

src/stan/variational/advi.hpp
```diff
@@ -398,6 +562,7 @@ namespace stan {
       }
 
       // Helper function: compute relative decrease between two doubles
+      // TODO(akucukelbir): move to stan math and test there
       double rel_decrease(double prev, double curr) const {
         return std::abs(curr - prev) / std::abs(prev);
```
bob-carpenter: Two problems.

1.  This should only involve a single abs, and it should be fabs, as in `std::fabs((curr - prev) / prev)`.

2.  Because of the absolute value, it's not computing a relative decrease, so it should be renamed. Maybe relative diff?  

And a question.  This is only relative to `prev` rather than to both scales. 

And a bonus:

3.  Use doxygen doc, not line doc.

bob-carpenter:
LIke the other files, this one needs proper doc.  And some idiom tidying up, but we can do that later.  I don't see anything wrong with the logic.

syclik:
The tests are failing because `src/stan/services/variational/print_progress.hpp` doesn't include enough headers.

src/stan/services/arguments/arg_variational_adapt.hpp
syclik: yeah... this is all part of the arguments. I'll create a new issue for this.

syclik:
@akucukelbir, this needs tests for `print_progress`.

src/stan/variational/advi.hpp
syclik: I just looked at the cpplint python script. It looks like it's searching for an equal number of spaces around operators. I'm not sure how to configure it to enforce spaces.

syclik:
@akucukelbir, I fixed some of what bob suggested. Tests are still going to fail due to not having enough includes in print_progress.hpp. Please add tests for that function.

src/stan/services/arguments/arg_variational_adapt.hpp
akucukelbir: ok. ignoring comments about arguments in this PR.

src/stan/variational/advi.hpp
akucukelbir: daniel took care of this.

src/stan/variational/advi.hpp
akucukelbir: we've been playing around with the values and the number of `eta` values to test. the queue made sense. we may still want to play around with `eta` values down the road. so i think the queue still makes sense.

src/stan/variational/advi.hpp
akucukelbir: done.

akucukelbir:
headers for `print_progress.hpp` fixed.

akucukelbir:
@syclik the upstream `cmdStan` tests will fail. how can we point it to https://github.com/stan-dev/cmdstan/pull/223 ?

src/stan/variational/advi.hpp
bob-carpenter: You mean adjust the queue dynamically?  If not, it should be a static
constant array or std::vector with an index into the current position.

akucukelbir: i'll try rewriting it with a `std::vector` soon.

bob-carpenter: The main problem in my mind is allocating any container
structure which you then destructively modify to add and subtract
elements.  It's no big deal here because it's not in an embedded
loop.  It's just not the way this kind of thing is usually done
because std::vector and std::queue are expensive to manipulate
dynamically and because objects with varying state are harder to
reason about compared to indexes into an array.

akucukelbir: done.

akucukelbir:
@syclik: the tests will never pass because we need stan-dev/cmdstan#223. since that pull request only affects the ADVI portion of `cmdStan`, could you merge it? 

@bob-carpenter: looking okay to you, i hope?

@dustinvtran: could you take a quick look too?

all: i will submit separate pull requests for the manual changes. so i'm removing them from the checklist.

syclik:
Merging the cmdstan pull request first isn't the right way to do it. That
will point the Stan submodule to something that isn't valid.

I'll set up the test to be tested with that branch tomorrow.

akucukelbir:
@syclik ah i see. makes sense. thanks!

bob-carpenter:
OK with me.

syclik:
Jenkins, retest this please.

PR #1732, opened 2015-12-08
Bugfix/issue 1731 argument configuration advi

akucukelbir:
#### Summary:

This gives reasonable values for variational input arguments. 

#### Intended Effect:

It should make sure that the arguments tests don't fail.

#### How to Verify:

Run tests.

#### Documentation:

None.

#### Reviewer Suggestions: 

@syclik

### DIFF
--- a/src/stan/services/arguments/arg_variational_adapt_iter.hpp
+++ b/src/stan/services/arguments/arg_variational_adapt_iter.hpp
@@ -15,7 +15,7 @@ namespace stan {
         _default = "50";
         _default_value = 50;
         _constrained = true;
-        _good_value = 2.0;
+        _good_value = 50.0;
         _bad_value = -1.0;
         _value = _default_value;
       }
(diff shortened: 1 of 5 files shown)


syclik:
it's not quite right... I think you touched the submodule in some weird way. Mind backing out the change to the math submodule?

akucukelbir:
@syclik i did as you said, but i'm not sure it helped :(

bob-carpenter:
Currently this pull request has merge conflicts, so you're going to need to resolve those before anything's going to pass.

akucukelbir:
that's odd. it was fine when i submitted it last week?

daniel said that this was blocking you guys. this pull request is all
standard stan/argument stuff; nothing specific to ADVI. so feel free to
hack at it, if you prefer.

syclik:
Unfortunately, it was never fine. It might have merged in the past, but it
definitely doesn't now. It never passed tests. (you can see the red x's for
when it was run and failed)

This is blocking Stan. Tests like should never fail and I've already seen
it fail a handful of times.

If you look back at the merge problem, I think it's due to the stan_math
submodule being in a conflicting state. Can you merge develop into the
branch locally and fix the conflict?

akucukelbir:
@syclik looking okay now, i hope?

PR #1760, opened 2016-02-03
Updating Stan with service methods

syclik:
#### Summary:

Adding service methods to be called by interfaces (CmdStan, RStan, PyStan).

#### Intended Effect:

Refactor complete. This has all the required callbacks for writer and interrupt.

#### How to Verify:

Visual inspection. CmdStan will have a pull request soon that will take advantage of the new service methods.

#### Side Effects:

This will affect the interfaces. I can and will handle most of the changes on my own.

#### Documentation:

Doc updated.

#### Reviewer Suggestions: 

@betanalpha


betanalpha:
I can be talked into each sampler variant be in its own file for testing,
but I don’t like how the interfaces will call each such function individually.
It gives the interfaces too much freedom to screw around with the UI.

I am strongly in favor of having argument containers that are built and
configured by the interfaces, passed into the services, and then unpacked
as needed.  We still need to design these containers but this PR moves
us away from that approach to having giant monolithic argument lists.

ariddell:
an enthusiastic :+1: This looks great.

bob-carpenter:
I'm not sure what you mean by "screw around with the UI" here.

By argument containers, you mean a map from arguments to values?
Nested?  Error checked at container construction time?

What I imagine is that the interfaces will have functions that call
the inference routines (optimization with L-BFGS, optimization with BFGS,
sampling with NUTS, etc.).  These won't all be one big "stan()" function
like it is for CmdStan.  So I imagine each of these functions will marshal
its arguments and pass them to the UI.  What can go wrong with the direct
call to the service that can't go wrong with the intermediate step of packing
into a container and then calling the service?

betanalpha:
> I'm not sure what you mean by "screw around with the UI" here.

Have each interface specify the arguments in different ways
(for example how RStan does tier and warmup instead of
n_warmup and n_samples).

> By argument containers, you mean a map from arguments to values?
> Nested? Error checked at container construction time?

We haven�t designed anything yet, but in general some
collection of name/value pairs with some ability to validate.

> What I imagine is that the interfaces will have functions that call
> the inference routines (optimization with L-BFGS, optimization with BFGS,
> sampling with NUTS, etc.). These won't all be one big "stan()" function
> like it is for CmdStan. So I imagine each of these functions will marshal
> its arguments and pass them to the UI. What can go wrong with the direct
> call to the service that can't go wrong with the intermediate step of packing
> into a container and then calling the service?

One issue is that the service routes may not map to interface calls.
For example, in our initial draft of routes unit/diag/dense metrics were
all under one ehmc function call, but they would require calling different
functions the way Daniel has done it.

But mostly it�s code/logic duplication.  In the original refactor the interfaces
called one �sample� function and passed in an argument configuration
object, and the API took care of unpacking it and figuring out which 
function to call.  This restricts the interfaces to have similar UI while
also avoiding their having to implement such switching logic on their
own.

Basically in the original refactor the �command.hpp� file in CmdStan
dropped to a few hundred lines of code, where as here it will still be
pretty hefty due to all of the logic it has to implement.

ariddell:
re: UI; I think the function signatures in PyStan should be as close as possible to the C++ function signatures. The current deviations are unfortunate, in my opinion.

betanalpha:
Agreed — the problem is that I don’t think we’ll end up with interface
functions with this level of granularity (again, the structure of the metric 
is most naturally an argument, but also there’s the question of adapting
or not).

bob-carpenter:
I think it would be best to lay out what the functions
are going to be in RStan and PyStan and CmdStan and then
design Stan to support that.

When I wrote the CmdStan doc, I found I had to break
the commands apart into more bite-size units.  

I also don't want to set up a monolithic config object
that rigidly assigns groupings so that if an argument to
VI changes, then arguments to every other function have to
change.

- Bob

ariddell:
I think everything is spelled out here:

https://github.com/stan-dev/stan/wiki/User-Interface-Guidelines-for-Developers

For example, look at the sections: Typical R session, Typical PyStan
session.

Nothing's set in stone at this point, of course.

bob-carpenter:
I like that high level sketch.  It lays out what
actually has to happen.  But we need to dig down a layer
into the actual arguments and figure out just how many
things get run through ehmc (diag and dense mass? static
and NUTS?) and optimize (BFGS and L-BFGS?) and vi (mean field
and dense).

I just think writing it all out would help to define
the scope of what we're trying to do with the interfaces.

Michael wrote up the commands for CmdStan in chapter 8 of this:

https://github.com/stan-dev/cmdstan/releases/download/v2.9.0/cmdstan-guide-2.9.0.pdf

and then starting in fig 8.4 on p. 59, I break them down
into what look like separate calls but are really just
ways to specify arguments in the bigger structure defined
earlier by Michael.

The tradeoff is between 

 * a monolithic function approach,
   which has lots of interdependencies between arguments and only
   a sparse set of the arguments applicable to any call, and

 * a separate function approach, 
   which has lots of redundancy/cut-and-paste.

I believe the separate function approach, while redundant, 
is easier to understand and maintain.  And I believe it would
be even easier if we further took each of those functions
and flattened their arguments.

A kind of halfway house is like Git, where there's a subfunction,
as in "git add ...".  I think of that as just another syntax for
the separate function approach which we'd essentially have to use
in CmdStan because of the way the compilation works (at least now).

- Bob

betanalpha:
Allen,

We had a sketch that we did on a shared text editor hosted on your website last year.  Do you still have that?

ariddell:
That's what we entered into the Wiki:

https://github.com/stan-dev/stan/wiki/User-Interface-Guidelines-for-Developers

Let me know if there's something missing that you expected to find.

syclik:
[sorry I've been awol since submitting the pull request. I got sick and it
left my head pretty hazy for the past week. I'm finally getting over it.]

Michael, I think this is the right way to do it. For hmc, there are only 12
different entry points. We only need to get this right in CmdStan, PyStan,
and RStan. I've already gotten it right for CmdStan. I'll take care of it
for PyStan and RStan. I understand your concern for the code drifting and
it being hard. This is a step in the right direction -- perhaps not where
we'll end up, but it's still a step in the right direction. You're right in
that it makes the CmdStan code a little ugly because it deals with
unpacking the arguments, but this will make it a whole lot easier for the
other interfaces.

Allen, I'll update PyStan accordingly. I'm still a little fuzzy from my
cold so I was having trouble following the wiki page.


Michael, what would make this good enough to get in? The major difference
was splitting apart the "sample()" method by removing references to the
argument pointers so that all the other interfaces could call the functions
easily.

betanalpha:
> Michael, I think this is the right way to do it. For hmc, there are only 12
> different entry points.

Here�s the problem: I have a bunch of new samplers that I�m waiting to
add.  This includes Riemannian, Exhaustive HMC, and some more useful
variants of static HMC.  Each of these has adaptation or not, and all but
Riemannian have the three metric choices.  So all of a sudden we go
from 12 to, what, 48 or so?

Having intermediate argument containers would facilitate adding new
samplers/optimizers/etc because the API routes would not change.
As Bob noted we have two extremes � complete unpacking on the
outside or on the inside � but in the end I think we need to decide on
some compromise in the middle (split out functionality into O(10)
major routes and then have containers for the specifics).  This requires
difficult design choices but I think it�s the only way to scale everything
up.

Ultimately I was hoping to use the current argument containers as a
stop gap until we build better ones, at which point we could plum them
through all of the API routes and interfaces.  Because you took one of
the extremes, this PR will make it much harder to add new samplers
in the intermediate and make any changes in the end.  I�m sure you�ll
say that you�ll handle all of that, and I have no doubt that you will, but
then it�ll take forever to get anything done.

> Michael, what would make this good enough to get in? The major difference
> was splitting apart the "sample()" method by removing references to the
> argument pointers so that all the other interfaces could call the functions
> easily.

See above.

ariddell:
If we figured out a way to move in the direction @betanalpha is suggesting I think it still would be easier to start with having all this functionality broken up into separate functions.

The C++ code in PyStan/RStan is borderline unmaintainable and I think this change will improve things there a great deal.

@syclik One way to go forward with updating PyStan would be to just update RStan and then point me to that PR. I can do the PyStan updating.

syclik:
I think going from 12 to 48 or so is the right move. There's a build
limitation that we need to deal with; we can't include forward-mode for
everyone because last we checked, it doesn't build across all platforms
within a reasonable amount of time. I'm not sure what we need to do yet,
but my guess is that we'll need to create different executables (from
CmdStan) that includes different portions. (Setting up the code this way is
also the way to go for CloudStan -- not sure if you saw those emails.)

I don't see the work being harder this way. The total amount of work
necessary is the same. If you have the stuff on a branch, maybe I can take
a look and see if I agree with you then.

Here's what I'd suggest: move forward this way and if I'm wrong, I'll fix
it. It's easier to go from something flat to something with routes. The
reason this took a while was that I had to be careful as I removed things.
If you were hoping it'd be a stopgap, then I'd suggest letting this be the
way we do it for now because it's a lot easier routing things through from
this point rather than the one with a structure that has a bit too much
abstraction behind it.

On Mon, Feb 8, 2016 at 6:44 AM, Michael Betancourt <notifications@github.com

betanalpha:
> I think going from 12 to 48 or so is the right move. There's a build
> limitation that we need to deal with; we can't include forward-mode for
> everyone because last we checked, it doesn't build across all platforms
> within a reasonable amount of time. I'm not sure what we need to do yet,
> but my guess is that we'll need to create different executables (from
> CmdStan) that includes different portions. (Setting up the code this way is
> also the way to go for CloudStan -- not sure if you saw those emails.)

Build issues aside, this is a nightmare for doc and UX.  We absolutely
cannot expect our users to learn O(100) functions, most of which are
very simply related.

> I don't see the work being harder this way. The total amount of work
> necessary is the same. If you have the stuff on a branch, maybe I can take
> a look and see if I agree with you then.

No, it is absolutely not.  Flat means replicating code that could be
reduced into one function or a loop/etc if we had a better abstraction.

> Here's what I'd suggest: move forward this way and if I'm wrong, I'll fix
> it. It's easier to go from something flat to something with routes. The
> reason this took a while was that I had to be careful as I removed things.
> If you were hoping it'd be a stopgap, then I'd suggest letting this be the
> way we do it for now because it's a lot easier routing things through from
> this point rather than the one with a structure that has a bit too much
> abstraction behind it.

Sorry, this kind of approach is no longer maintainable.  You keep making
going this way, and then big changes end up being entirely dependent on
you and we have to wait for months on end for anything to happen.  The
�do it my way and I�ll fix it later� is not an option from my side of things.

syclik:
On Mon, Feb 8, 2016 at 11:53 AM, Michael Betancourt <
notifications@github.com> wrote:

>
> > I think going from 12 to 48 or so is the right move. There's a build
> > limitation that we need to deal with; we can't include forward-mode for
> > everyone because last we checked, it doesn't build across all platforms
> > within a reasonable amount of time. I'm not sure what we need to do yet,
> > but my guess is that we'll need to create different executables (from
> > CmdStan) that includes different portions. (Setting up the code this way
> is
> > also the way to go for CloudStan -- not sure if you saw those emails.)
>
> Build issues aside, this is a nightmare for doc and UX. We absolutely
> cannot expect our users to learn O(100) functions, most of which are
> very simply related.
>

This doesn't affect the user. CmdStan stays the same in terms of interface
between the user and CmdStan. The same is true of RStan and PyStan.


> > I don't see the work being harder this way. The total amount of work
> > necessary is the same. If you have the stuff on a branch, maybe I can
> take
> > a look and see if I agree with you then.
>
> No, it is absolutely not. Flat means replicating code that could be
> reduced into one function or a loop/etc if we had a better abstraction.
>

No. I don't think using a god object is going to help us here (
https://en.wikipedia.org/wiki/God_object). I've had enough difficulty with
trying to tease apart the argument structure within Stan. It's one of the
reasons things have moved so slow on my end -- it's tricky. It's much
trickier than I would prefer due to the dynamic casts, the pointers, having
to know what the structure is to make sense of it. The reason I moved away
from one sample() function that took an argument pointer is that it would
not be easy to populate this object and call sample() from within RStan and
PyStan. I'm looking at this from the point of view of maintaining the whole
family of interfaces. I still don't know how to create an argument
substructure in the correct way without knowing how it should be
constructed -- that's a bad pattern and not the right way to do it. It's
error prone and opens up chances for seg faults.

In short: no. If you look at this code vs your pull request, you'll see
that the logic isn't more complicated or that there's more. In fact, the
logic is much simpler and easier to follow.


> > Here's what I'd suggest: move forward this way and if I'm wrong, I'll fix
> > it. It's easier to go from something flat to something with routes. The
> > reason this took a while was that I had to be careful as I removed
> things.
> > If you were hoping it'd be a stopgap, then I'd suggest letting this be
> the
> > way we do it for now because it's a lot easier routing things through
> from
> > this point rather than the one with a structure that has a bit too much
> > abstraction behind it.
>
> Sorry, this kind of approach is no longer maintainable. You keep making
> going this way, and then big changes end up being entirely dependent on
> you and we have to wait for months on end for anything to happen. The
> �do it my way and I�ll fix it later� is not an option from my side of
> things.
>

The alternative was for me to have said no to your PR months ago and let
you continue to fix things until they were right. I'm trying to be helpful
-- it took a lot longer because there were many more underlying code issues
that weren't addressed. I haven't been forcing my way on the branch, btw.
I've gone along with a lot of your underlying implementation, even if that
wasn't what I would have done, in order to move your branch along.

Let's get back to being productive. What's wrong with having 12 routes
here? I've already got the branch ready for CmdStan. This change is going
to make RStan and PyStan a whole lot better. The original branch won't;
that'll require having to construct the argument structure by allocating
memory and having to know the structure within RStan and PyStan. It's the
same amount of work, this one is just easier to maintain (across the
interfaces, not just for Stan / CmdStan).

Are we at a standstill? If you used your original sample() function, it
won't get through a code review due to the use of a god object. If you're
rejecting this on the basis of the design of having modular functions, then
I'm not sure what our next steps are. Do you want me to implement the RStan
and PyStan updates just to show how much it cleans up code?

betanalpha:
> > Build issues aside, this is a nightmare for doc and UX. We absolutely
> > cannot expect our users to learn O(100) functions, most of which are
> > very simply related.
> >
> 
> This doesn't affect the user. CmdStan stays the same in terms of interface
> between the user and CmdStan. The same is true of RStan and PyStan.

No, that is not guaranteed.  The only way to avoid drift amongst the
interfaces is to have the exposed functions mirror the UX.  So either
the UX will be at this very fine level or the onus is on the interfaces 
and there no way to enforce a consistent UX.

And this ignores many of the other UX frustrations with functions
that have so many arguments � they _all_ have to be specified
with no potential for defaults.  It�ll make the C++ API a giant pain
in the ass to use.

> No. I don't think using a god object is going to help us here (
> https://en.wikipedia.org/wiki/God_object). I've had enough difficulty with
> trying to tease apart the argument structure within Stan. It's one of the
> reasons things have moved so slow on my end -- it's tricky. It's much
> trickier than I would prefer due to the dynamic casts, the pointers, having
> to know what the structure is to make sense of it. The reason I moved away
> from one sample() function that took an argument pointer is that it would
> not be easy to populate this object and call sample() from within RStan and
> PyStan. I'm looking at this from the point of view of maintaining the whole
> family of interfaces. I still don't know how to create an argument
> substructure in the correct way without knowing how it should be
> constructed -- that's a bad pattern and not the right way to do it. It's
> error prone and opens up chances for seg faults.

The pointer explosion is an implementation issue �  now that we know
more about C++ we can ensure compile time checks and all the niceties.
Additionally, the config is nontrivial to construct because it was never
designed as such � I have said multiple times that the final step in the
refactor is redesigning the argument config objects so they are easy
to set and validate.  

> In short: no. If you look at this code vs your pull request, you'll see
> that the logic isn't more complicated or that there's more. In fact, the
> logic is much simpler and easier to follow.

I fundamentally disagree.  The logic is just flattened, requiring far more
hand-written code.  Just look at the tests � each function has to be
separate tested in a bespoke fashion, where as the argument config
allowed for automated testing (currently in CmdStan, but easily moved
to Stan once the refactor is complete).

> The alternative was for me to have said no to your PR months ago and let
> you continue to fix things until they were right. I'm trying to be helpful
> -- it took a lot longer because there were many more underlying code issues
> that weren't addressed. I haven't been forcing my way on the branch, btw.
> I've gone along with a lot of your underlying implementation, even if that
> wasn't what I would have done, in order to move your branch along.

What did I say when I first submitted the PR?  �Everyone take a look and
let me know what you think of the architecture, what should be tested, etc.�
Did anyone do that?  Of course not, most people didn�t look and you took
it as full PR and started micromanaging.  This is incredibly frustrating because 
everyone says that they can�t wrap their heads around the refactor and need 
examples, but then there is little to no constructive feedback and we end up 
in quagmires like this.

> Let's get back to being productive. What's wrong with having 12 routes
> here? I've already got the branch ready for CmdStan. This change is going
> to make RStan and PyStan a whole lot better. The original branch won't;
> that'll require having to construct the argument structure by allocating
> memory and having to know the structure within RStan and PyStan. It's the
> same amount of work, this one is just easier to maintain (across the
> interfaces, not just for Stan / CmdStan).

> Are we at a standstill? If you used your original sample() function, it
> won't get through a code review due to the use of a god object. If you're
> rejecting this on the basis of the design of having modular functions, then
> I'm not sure what our next steps are. Do you want me to implement the RStan
> and PyStan updates just to show how much it cleans up code?

I�m not sure how many times I need to say this.  The problem is the immediate
future � this approach makes adding samplers a pain in the ass and, what do
you know, there are a bunch of samplers I�m waiting to put in.  Then there�s
the complications with the refactor.

I don�t care what happens to the interfaces _now_.  I care about how much
we can clean them up with the refactor is done.  All of these incremental
updates are taking forever and making the further steps in the refactor more
challenging.  And since no one else is proposing any useful designs for the
refactor that frustration all ends up on me.  Awesome.

syclik:
From the technical perspective:
- splitting out the calls is easier for implementing the interfaces
- the maintenance is going to be easier; you may not believe me, but it is
- I agree with you regarding defaults, but how often have we changed defaults? Overengineering this is not the way to go.

I'm really adamant that this is the way we move forward in order to make things easier on the interfaces. With RHMC, we'll expose more routes, but it's easier integrating them than you may think. And it doesn't affect the user interface as much as you might think.

So... will you give this the ok? If you're really insistent on building the structure, I still think it should end at these routes, so I don't see the harm.


Aside:
Btw, I have just as much frustration. For the pull request, yes, it was a sketch, but that sketch had so much buried in it. I didn't even see that this was how you split apart the calls until I started working on this last pull request. This was just one of the many things that you did under the guise of a sketch. If you actually want feedback, you have to put things in a form where you can get feedback. I'm mostly frustrated that instead of building working pull requests that could be reviewed and merged, I had to work through this one just to uncover the different things you did. I couldn't have given you the appropriate feedback without doing just that. Out of respect, I did exactly that -- I split things out so I could see what needed to be done.

bob-carpenter:
I like them split out as much as possible so that
all arguments are relevant and of a single type and
there's a single return type.  

I primarily like it this way because it makes things
easier to understand for both the developers and users.

I think it makes the API/doc/testing for everything much
easier because of fewer interactions.

Most users won't use that many of the functions, but even
if they do, it will be easier to learn one simple(r) function
first, then move on than to try to grasp the whole argument
structure for everything.

For the shared functionality required to implement these top-level
calls (like, say, reading in a diagonal mass matrix), the behavior
including error checking can be encapsulated in a function.  

- Bob

syclik:
I'm closing this pull request for now. I'm going to make sure more changes and reopen the pull request with updates.

PR #1767, opened 2016-02-15
init int variables to min int, fixes #1766

### LINKED ISSUE #1766: remove init to NaN for int values
The generated code assigns `NaN` to integer values on init.

* [ ] make it assign the smallest number (negative number of greatest magnitude) so it's easy to spot if users don't further assign

This is the root cause of the math library issue https://github.com/stan-dev/math/issues/232

bob-carpenter:
#### Summary:

Change behavior in generated code for Stan programs to assign minimum integer value to integers rather than a double not-a-number behavior.

#### Intended Effect:

Remove undefined behavior arising from assigning not-a-number to integers.

#### How to Verify:

New unit test on generated code.

#### Side Effects:

n/a

#### Documentation:

n/a

#### Reviewer Suggestions:

anyone


syclik:
Looks good. Thanks for adding a test.

Issue #1768, opened 2016-02-16
function that translates to C++ but doesn't compile


bob-carpenter:
Matthew Zeigenfuse reported on stan-users that a program involving a forward definition of a `_log` function translates to C++ but doesn't compile.  Here's a simplified example that reproduces the problem:

```
functions {
  real n_log(real y);

  real n_log(real y) {
    return -0.5 * square(y);
  }
}
parameters {
  real mu;
}
model {
  mu ~ n();
}
```

which produces the error

```
In file included from <built-in>:332:
In file included from <command line>:5:
/Users/carp/temp2/rep.hpp:53:1: error: redefinition of 'n_log'
n_log(const T0__& y, std::ostream* pstream__) {
^
/Users/carp/temp2/rep.hpp:28:1: note: previous definition is here
n_log(const T0__& y, std::ostream* pstream__) {
^
1 error generated.
```

* [ ] fix the generator
* [ ] thank Matthew Zeigenfuse for the bug report

PR #1769, opened 2016-02-17
remove dup def for _log funs, fixing #1768

bob-carpenter:
#### Summary:

Remove duplicate definition of `propto=false` version of `_log` user-defined functions when there is a forward declaration.

#### Intended Effect:

Allow recursive density definitions.

#### How to Verify:

There is a new model test that passes now and didn't pass before.

#### Side Effects:

None.

#### Documentation:

Behavior now matches existing doc.

#### Reviewer Suggestions:

Anyone.


syclik:
Thanks!

PR #1779, opened 2016-02-26
 reparameterize tau in example for  hierarchical regression model

gbiele:


jgabry:
I would remove the comment that says `// prior scale` where you declare `tau_unif` because it's `tau` and not `tau_unif` that's the prior scale.

gbiele:
I've removed the "// prior scale" comment from 
tau_unif and added it to tau in the transformed
parameters block

PR #1788, opened 2016-03-08
Feature/issue 1787 compound assign

bob-carpenter:
#### Summary:

Add `=` as alternative to `<-` for assignment; deprecate `<-`.

#### Intended Effect:

Make Stan look more like other programming languages and pave way for new `+=` syntax.

#### How to Verify:

Unit tests.

#### Side Effects:

Deprecation warnings now come up for `<-` with message indicating it should be replaced with `=`

#### Documentation:

The manual was updated to reflect the new assignment operator `=` and a note was included in the language reference chapter indicating that `<-` was the old assignment syntax.  As usual, we're not maintaining doc in parallel on both the old way and the new way.

#### Reviewer Suggestions: 

Anyone.

### DIFF
--- a/src/docs/stan-reference/advanced.tex
+++ b/src/docs/stan-reference/advanced.tex
@@ -125,14 +125,14 @@ parameters {
 }
 transformed parameters {
   real<lower=0> squared_error;
-  squared_error <- dot_self(y - x * beta);
+  squared_error = dot_self(y - x * beta);
 }
 model {
   increment_log_prob(-squared_error);
 }
 generated quantities {
   real<lower=0> sigma_squared;
-  sigma_squared <- squared_error / N;
+  sigma_squared = squared_error / N;
 }
 \end{stancode}
 %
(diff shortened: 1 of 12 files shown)


bob-carpenter:
Actually, hang on.  I'll do the doc while I'm at it.

bob-carpenter:
OK, ready to be reviewed now.

syclik:
Cool! Thanks!

bob-carpenter:
Let's make sure at the meeting it's what everyone wants
before merging!

syclik:
Sure!

bgoodri:
@bob-carpenter The next time you do a PR for the language, can you add "in the Stan language" after deprecated to the deprecation message about `<-`. It is a bit jarring to see in R that
```
posterior <- stan("foo.stan")
Warning (non-fatal): assignment operator <- deprecated; use = instead.
```
because it could be interpreted (particularly if using a package that auto-generates Stan code) that R has deprecated `<-`.

syclik:
Hi Ben,

Can you open up a new issue? Can you add the exact wording you want for the message?


Daniel

Issue #1790, opened 2016-03-08
Update HMC Diagnostics


betanalpha:
#### Description:

Rename divergent iterations and add the value of the Hamiltonian to identify poorly chosen kinetic energies.

#### Current Version:
develop

bob-carpenter:
What is being renamed to what?  Is this user facing?

betanalpha:
Please look at the code — there are only a few file changes.
It’s user facing only if the interfaces change — of course that
has now been left to the interface designers so who the hell
knows?

bob-carpenter:
What code am I supposed to be looking at?  Is there a branch or a pull request?

betanalpha:
Huh?  You were commenting on it earlier. https://github.com/stan-dev/stan/pull/1791

PR #1791, opened 2016-03-08
Clean up HMC diagnostics, update corresponding tests

betanalpha:
#### Summary:

Addresses issue #1790.

#### Intended Effect:

Rename divergent iterations and add the value of the Hamiltonian to identify poorly chosen kinetic energies.

#### How to Verify:

Run unit tests.

#### Side Effects:

Changes sampler parameter output.  This will break the automated tests that will try to push this into CmdStan -- a super dev friendly problem that @syclik will have to fix by hand.

#### Documentation:

None -- the Stan manual does not discuss specific HMC diagnostics.

#### Reviewer Suggestions: 

Anyone.


jgabry:
For consistency, should we rename n_divergent in the diagnostics given by
RStan (and ShinyStan) when you add this? I like the name change but that
would break any user scripts that access the sampler parameter diagnostics
by name. Or should we wait until making other changes that break things?

>

bob-carpenter:
Either support both and deprecate the old one with a warning,
or wait to make breaking changes for a major version update.

Semantic versionining:  MAJOR.MINOR.PATCH

backward compatibility breaking:  MAJOR++
new features without backward compatibility breaks:  MINOR++
bug fixes without new features: PATCH++

- Bob

jgabry:
Right, and that's what we would do if we were talking about the name of a function argument or something like that, but for this kind of change I don't think we can allow both simultaneously with one deprecated. Here's a quick example illustrating that: 

If a user has an R script containing something like the following
```r
fit <- stan(...) 
sampler_pars <- get_sampler_params(fit)
sampler_pars_chain1 <- sampler_pars[[1]]
```
then `sampler_pars_chain1 ` will be a matrix with column names corresponding to the names of the various sampler parameters (`"accept_stat__"`, `"n_divergent__"`, etc).  So if we change the name from `n_divergent__` to `divergent__` and a user has a script that uses column names rather than column numbers to access the columns of the matrix then they would _have to_ use `divergent__` and not `n_divergent__` and there's no way for us to allow both. Once they have the matrix then they need to play by R's rules.

So this seems like something we should wait to change, at least for anything user-facing.

ariddell:
There are a lot of variables that have been mentioned as being good candidates for renaming.

For example, in "Recommendations and style guidelines" in https://github.com/stan-dev/stan/wiki/User-Interface-Guidelines-for-Developers

I mention "consistent variable names (either num_warmup and num_iter or use warmup and iter, not a mix)"

It would be nice to track them some place centrally so we can do the renaming all at once. While RStan/PyStan/CmdStan may differ I don't think they should differ on variable names.

bob-carpenter:
Or you can introduce a new function name that returns
the new thing while still supporting the old function
that returns the old thing.

And while I have your attention about renaming, I'd recommend
removing "get_" prefixes;  "set_" is OK, but "get_" doesn't
really add any info, just four more characters.

- Bob

jgabry:
On Tuesday, March 8, 2016, Bob Carpenter <notifications@github.com> wrote:

> Or you can introduce a new function name that returns
> the new thing while still supporting the old function
> that returns the old thing.


Ok yeah, I wasn't thinking about adding a new function but that's probably
the best option.


>
> And while I have your attention about renaming, I'd recommend
> removing "get_" prefixes; "set_" is OK, but "get_" doesn't
> really add any info, just four more characters.
>
> - Bob
>

Fine by me. I'll add an issue at the RStan repo.

>

betanalpha:
Looks like the unit tests are failing with a compilation error -- both unit and performance tests pass locally on my machine.

syclik:
Jenkins, retest this please.

syclik:
It looks like some io stream (not related to the tests) died on Jenkins. I
just restarted it.

On Wed, Mar 9, 2016 at 9:47 AM, Michael Betancourt <notifications@github.com

betanalpha:
Now it's just the performance test which I changed to get running locally on my machine.  Why is it yielding different results on Jenkins?  This test is a mess.

syclik:
Yes, the test is a mess. Thanks for reminding me. I'll fix it with the services refactor. Having a direct route to call the algorithm will clean up the test.

betanalpha:
"ld: not enough disk space for writing" Jenkins out of space?

betanalpha:
Jenkins, retest this please.

betanalpha:
Ready to merge.

src/stan/mcmc/hmc/nuts/base_nuts.hpp
```diff
@@ -152,14 +152,16 @@ namespace stan {
         names.push_back("stepsize__");
         names.push_back("treedepth__");
         names.push_back("n_leapfrog__");
-        names.push_back("n_divergent__");
+        names.push_back("divergent__");
+        names.push_back("E__");
```
syclik: @jgabry, @bgoodri: what impact does this have on RStan and ShinyStan? Is this sort of change to the output easy to handle?

syclik: Summary: 
- `n_divergent__` moves to `divergent__`
- `E__` is inserted

I just need to know what sort of impact this has. It's easy for CmdStan.

bgoodri: I don't think it will be a big deal.

betanalpha: CmdStan already passes and there’s a separate PR for updated doc.
So are we merging or what?

syclik: @betanalpha, I got confirmation from @jgabry. It'll break existing things, but not hard to change.

What's `E__`? Could we rename that to something descriptive?

betanalpha: It’s the value of the Hamiltonian at the sample (included the momentum
which gets thrown away), or energy.  I would prefer we stick with E as it
is concise and matches the notation being used in my papers.

syclik: would you be ok with something slightly more verbose like `energy__`? I'm thinking about what users will see and ask. This is only going in the header of the output csv file, correct?

syclik:
@betanalpha, let me know if we can use `energy__` or something a little more descriptive than `E__`.

@jgabry, 
- the current version of ShinyStan will not work with this change if it gets into RStan. Is this correct?
- is there a fix you can apply to have ShinyStan handle both existing behavior and this change?
- is there a way to tell users that upgrade rstan to also upgrade ShinyStan?
- what happens if this change hits the current ShinyStan?

src/stan/mcmc/hmc/nuts/base_nuts.hpp
bob-carpenter: It's easy for a user who has read Michael's papers
to map energy__ to E__, but the reverse step is a little
more confusing.  On the other hand, just seeing energy__ isn't
going to help most of our users without going to Michael's
papers.  So I'd prefer energy__ to match the other params and
be a little less confusing, but I don't think it's a huge deal
either way.  Narrower column names do have advantages.

- Bob

betanalpha: I don’t know why we’re getting into such pedantic name arguments here,
especially given that I’m the one expert on these concepts and the sole
author of the corresponding papers/notes.  Moreover, the interfaces can 
do whatever they want with the names — as established consistency is 
no longer a concern.  

I want to get working on a bunch of other changes that are waiting on this 
so I made the change.  The Windows tests are failing due to what looks
like a bad configuration — given how terribly this has gone I hope this
failure will be addressed promptly.

Having a PR wait this long for such pedantic concerns is absolutely not 
sustainable.

syclik:
@betanalpha: thanks for changing it. I guess you did something I didn't expect, though -- I wasn't suggesting changing the code... just the output name so it's friendlier on the users. I trust you're ok with the addition of the member variable?

It looks like github.com went down when the Windows test kicked off. I'll start the tests again.

syclik:
Jenkins, retest this please.

betanalpha:
This was going to be done in the next PR anyways, I just
took advantage of having to change things.

syclik:
Sweet.

jgabry:
On Sun, Mar 13, 2016 at 7:59 PM, Daniel Lee <notifications@github.com>
wrote:

> @jgabry <https://github.com/jgabry>,
>
>    - the current version of ShinyStan will not work with this change if
>    it gets into RStan. Is this correct?
>
> Correct, but once we get the change into a branch on the RStan repo it
should be easy for me to make the necessary tweaks to ShinyStan. @bgoodri
and/or @maverickg would know best where any changes to RStan need to be
made.

>
>    - is there a fix you can apply to have ShinyStan handle both existing
>    behavior and this change?
>    - is there a way to tell users that upgrade rstan to also upgrade
>    ShinyStan?
>
> These are related and it depends on how RStan implements the change. For
example, Bob suggested that we add a new function to RStan that has the
same behavior as get_sampler_params but uses the new names instead of
changing get_sampler_params (which would break user's scripts). If that's
what RStan ends up doing then ShinyStan could look for the new function and
use that if the user has it, and if not it could use the old function.

As for the names of the sampler_params, I'm fine with whatever Michael
wants.

syclik:
Thanks! Merge will happen when tests pass!

Issue #1846, opened 2016-04-07
Multinomial sampling of NUTS trajectories.


betanalpha:
#### Summary:
Change how NUTS samples from each dynamically generated trajectory.

#### Description:
Right now NUTS uses a slice sampler to select each point which is a bit ungainly.  Sampling points multinomially using their Metropolis probabilities will streamline the code and make it easier to experiment with algorithmic generalizations.

#### Current Version:
v2.9.0

bob-carpenter:
Along the entire trajectory including the original point?  And I take it this is the one that's easy to Rao-Blackwellize rather than multinomial sample.  Any a priori theory on how this will compare to the slice sampling?  I don't imagine the normalization for the multinomial is going to be too compute intensive compared to everything else going on.

betanalpha:
> Along the entire trajectory including the original point?
> 
Yes, including the original point is important for avoiding periodicity.
> And I take it this is the one that's easy to Rao-Blackwellize rather than multinomial sample.
> 
That is part of the motivation, although we are very far away from anything close to an implementation.
> Any a priori theory on how this will compare to the slice sampling?
> 
Marginally (over the slice level or multinomial sample) they are the same so they shouldn’t yield
appreciable difference.  Having all of the probabilities at hand is technically more information
but we’re not using that yet.
> I don't imagine the normalization for the multinomial is going to be too compute intensive compared to everything else going on.
> 

No, it’s negligible.

bob-carpenter:
OK, I think you should be in charge of what we do for sampling.
So I'm OK with this change.  I still need to review the code, though.

PR #1848, opened 2016-04-07
Feature/issue 1846 multinomial nuts

betanalpha:
#### Submisison Checklist

#### Summary

Change the defaults NUTS sampler to sample from trajectories multinomially instead of with the original slice sampler.  The original implementation has been preserved as "nuts_classic".  Addresses #1846.

#### Intended Effect

Minimal user-facing affects.  Allows for cleaner code and facilitates research.

#### How to Verify

Run tests, try recovering known expectations from analytic models.

#### Side Effects

Small changes to exact sampler output but expectations should be the same with the MCMC standard error.

#### Documentation

None.

#### Reviewer Suggestions

Anyone.

#### Copyright and Licensing

Copyright: Michael Betancourt

### DIFF
--- a/src/stan/mcmc/hmc/nuts/base_nuts.hpp
+++ b/src/stan/mcmc/hmc/nuts/base_nuts.hpp
@@ -59,7 +59,7 @@ namespace stan {
 
         double H0 = this->hamiltonian_.H(this->z_);
         int n_leapfrog = 0;
-        double sum_metro_prob = 1; // exp(H0 - H0)
+        double sum_metro_prob = 1;  // exp(H0 - H0)
 
         // Build a trajectory until the NUTS criterion is no longer satisfied
         this->depth_ = 0;
--- a/src/stan/mcmc/hmc/nuts/base_nuts.hpp
+++ b/src/stan/mcmc/hmc/nuts/base_nuts.hpp
@@ -13,28 +13,14 @@
 
 namespace stan {
   namespace mcmc {
-
-    struct nuts_util {
-      // Constants through each recursion
-      double log_u;
-      double H0;
-      int sign;
-
-      // Aggregators through each recursion
-      int n_tree;
-      double sum_prob;
-      bool criterion;
-    };
-
-    // The No-U-Turn Sampler (NUTS).
-
+    // The No-U-Turn Sampler (NUTS) with multinomial sampling
     template <class Model, template<class, class> class Hamiltonian,
               template<class> class Integrator, class BaseRNG>
     class base_nuts : public base_hmc<Model, Hamiltonian, Integrator, BaseRNG> {
     public:
       base_nuts(Model &model, BaseRNG& rng)
         : base_hmc<Model, Hamiltonian, Integrator, BaseRNG>(model, rng),
-        depth_(0), max_depth_(5), max_delta_(1000),
+        depth_(0), max_depth_(5), max_deltaH_(1000),
         n_leapfrog_(0), divergent_(0), energy_(0) {
       }
 
@@ -46,18 +32,16 @@ namespace stan {
       }
 
       void set_max_delta(double d) {
-        max_delta_ = d;
+        max_deltaH_ = d;
       }
 
       int get_max_depth() { return this->max_depth_; }
-      double get_max_delta() { return this->max_delta_; }
+      double get_max_delta() { return this->max_deltaH_; }
 
       sample transition(sample& init_sample,
                         interface_callbacks::writer::base_writer& writer) {
         // Initialize the algorithm
         this->sample_stepsize();
-
-        nuts_util util;
 
         this->seed(init_sample.cont_params());
 
@@ -70,83 +54,69 @@ namespace stan {
         ps_point z_sample(z_plus);
         ps_point z_propose(z_plus);
 
-        int n_cont = init_sample.cont_params().size();
+        Eigen::VectorXd rho = this->z_.p;
+        double sum_weight = 1;
 
-        Eigen::VectorXd rho_init = this->z_.p;
-        Eigen::VectorXd rho_plus(n_cont); rho_plus.setZero();
-        Eigen::VectorXd rho_minus(n_cont); rho_minus.setZero();
+        double H0 = this->hamiltonian_.H(this->z_);
+        int n_leapfrog = 0;
+        double sum_metro_prob = 1; // exp(H0 - H0)
 
-        util.H0 = this->hamiltonian_.H(this->z_);
-
-        // Sample the slice variable
-        util.log_u = std::log(this->rand_uniform_());
-
-        // Build a balanced binary tree until the NUTS criterion fails
-        util.criterion = true;
-        int n_valid = 0;
-
+        // Build a trajectory until the NUTS criterion is no longer satisfied
         this->depth_ = 0;
         this->divergent_ = 0;
 
-        util.n_tree = 0;
-        util.sum_prob = 0;
+        while (this->depth_ < this->max_depth_) {
+          // Build a new subtree in a random direction
+          Eigen::VectorXd rho_subtree(rho.size());
+          rho_subtree.setZero();
 
-        while (util.criterion && (this->depth_ <= this->max_depth_)) {
-          // Randomly sample a direction in time
-          ps_point* z = 0;
-          Eigen::VectorXd* rho = 0;
+          bool valid_subtree = false;
+          double sum_weight_subtree = 0;
 
           if (this->rand_uniform_() > 0.5) {
-            z = &z_plus;
-            rho = &rho_plus;
-            util.sign = 1;
+            this->z_.ps_point::operator=(z_plus);
+            valid_subtree
+              = build_tree(this->depth_, rho_subtree, z_propose,
+                           H0, 1, n_leapfrog,
+                           sum_weight_subtree, sum_metro_prob, writer);
+            z_plus.ps_point::operator=(this->z_);
           } else {
-            z = &z_minus;
-            rho = &rho_minus;
-            util.sign = -1;
+            this->z_.ps_point::operator=(z_minus);
+            valid_subtree
+              = build_tree(this->depth_, rho_subtree, z_propose,
+                           H0, -1, n_leapfrog,
+                           sum_weight_subtree, sum_metro_prob, writer);
+            z_minus.ps_point::operator=(this->z_);
           }
 
-          // And build a new subtree in that direction
-          this->z_.ps_point::operator=(*z);
+          sum_weight += sum_weight_subtree;
+          if (!valid_subtree) break;
 
-          int n_valid_subtree = build_tree(depth_, *rho, 0, z_propose, util,
-                                           writer);
+          // Sample from an accepted subtree
           ++(this->depth_);
 
-          *z = this->z_;
-
-          // Metropolis-Hastings sample the fresh subtree
-          if (!util.criterion)
-            break;
-
-          double subtree_prob = 0;
-
-          if (n_valid) {
-            subtree_prob = static_cast<double>(n_valid_subtree) /
-              static_cast<double>(n_valid);
-          } else {
-            subtree_prob = n_valid_subtree ? 1 : 0;
-          }
-
-          if (this->rand_uniform_() < subtree_prob)
+          double accept_prob = sum_weight_subtree / sum_weight;
+          if (this->rand_uniform_() < accept_prob)
             z_sample = z_propose;
 
-          n_valid += n_valid_subtree;
-
-          // Check validity of completed tree
+          // Break when NUTS criterion is not longer satisfied
           this->z_.ps_point::operator=(z_plus);
-          Eigen::VectorXd delta_rho = rho_minus + rho_init + rho_plus;
 
-          util.criterion = compute_criterion(z_minus, this->z_, delta_rho);
+          rho += rho_subtree;
+          if (!compute_criterion(z_minus, this->z_, rho))
+            break;
         }
 
-        this->n_leapfrog_ = util.n_tree;
+        this->n_leapfrog_ = n_leapfrog;
 
-        double accept_prob = util.sum_prob / static_cast<double>(util.n_tree);
+        // Compute average acceptance probabilty across entire trajectory,
+        // even over subtrees that may have been rejected
+        double accept_prob
+          = sum_metro_prob / static_cast<double>(n_leapfrog + 1);
 
         this->z_.ps_point::operator=(z_sample);
         this->energy_ = this->hamiltonian_.H(this->z_);
-        return sample(this->z_.q, - this->z_.V, accept_prob);
+        return sample(this->z_.q, -this->z_.V, accept_prob);
       }
 
       void get_sampler_param_names(std::vector<std::string>& names) {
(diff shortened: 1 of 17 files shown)


betanalpha:
Where are the Windows tests failing?

syclik:
It's a compiler warning. Windows never ceases to amaze me. 

d1m1s1b1.stat.columbia.edu:8080/job/Stan%20Pull%20Request%20-%20Windows%20-%20Tests%20-%20Unit/639/warnings18Result/new/

Change EXPECT_EQ() to EXPECT_FALSE()

src/stan/mcmc/hmc/nuts/base_nuts.hpp
```diff
@@ -70,83 +54,69 @@ namespace stan {
...
-          ps_point* z = 0;
-          Eigen::VectorXd* rho = 0;
+          bool valid_subtree = false;
+          double sum_weight_subtree = 0;
 
           if (this->rand_uniform_() > 0.5) {
-            z = &z_plus;
-            rho = &rho_plus;
-            util.sign = 1;
+            this->z_.ps_point::operator=(z_plus);
```
bob-carpenter: Why not use the operator syntax?

```
this->z_ = z_plus;
```

src/stan/mcmc/hmc/nuts/base_nuts.hpp
```diff
@@ -70,83 +54,69 @@ namespace stan {
...
+              = build_tree(this->depth_, rho_subtree, z_propose,
+                           H0, -1, n_leapfrog,
+                           sum_weight_subtree, sum_metro_prob, writer);
+            z_minus.ps_point::operator=(this->z_);
           }
 
-          // And build a new subtree in that direction
-          this->z_.ps_point::operator=(*z);
+          sum_weight += sum_weight_subtree;
+          if (!valid_subtree) break;
```
bob-carpenter: Should the break come before adding the weight?

src/stan/mcmc/hmc/nuts/base_nuts.hpp
```diff
@@ -70,83 +54,69 @@ namespace stan {
...
         }
 
-        this->n_leapfrog_ = util.n_tree;
+        this->n_leapfrog_ = n_leapfrog;
 
-        double accept_prob = util.sum_prob / static_cast<double>(util.n_tree);
+        // Compute average acceptance probabilty across entire trajectory,
+        // even over subtrees that may have been rejected
+        double accept_prob
+          = sum_metro_prob / static_cast<double>(n_leapfrog + 1);
```
bob-carpenter: This doesn't need the static cast to double --- whenever you have a mixed operation, the lower type is promoted automatically.

src/stan/mcmc/hmc/nuts/base_nuts.hpp
```diff
@@ -46,19 +32,17 @@ namespace stan {
...
+      double get_max_delta() { return this->max_deltaH_; }
 
       sample transition(sample& init_sample,
                         interface_callbacks::writer::base_writer& writer) {
         // Initialize the algorithm
         this->sample_stepsize();
 
-        nuts_util util;
-
         this->seed(init_sample.cont_params());
```
bob-carpenter: I know this isn't new code, but why all the `this -> `?  Is there an issue with just calling `seed(...)` and other functions directly?  (I seem to recall something like that happening that we couldn't figure out).

src/stan/mcmc/hmc/nuts/base_nuts.hpp
betanalpha: This copies only the position and momenta, not the mass matrix.  It avoids unnecessary copies which was a significant burden.

src/stan/mcmc/hmc/nuts/base_nuts.hpp
```diff
@@ -70,83 +54,69 @@ namespace stan {
...
+          // Build a new subtree in a random direction
+          Eigen::VectorXd rho_subtree(rho.size());
+          rho_subtree.setZero();
 
-        while (util.criterion && (this->depth_ <= this->max_depth_)) {
-          // Randomly sample a direction in time
-          ps_point* z = 0;
-          Eigen::VectorXd* rho = 0;
+          bool valid_subtree = false;
+          double sum_weight_subtree = 0;
```
bob-carpenter: It looks like you're doing the sum and normalization on the linear scale, which seems like it might be prone to underflow.  Is that going to be robust enough or should we be using `log_sum_exp()`?

src/stan/mcmc/hmc/nuts/base_nuts.hpp
```diff
@@ -70,83 +54,69 @@ namespace stan {
...
-          double subtree_prob = 0;
-
-          if (n_valid) {
-            subtree_prob = static_cast<double>(n_valid_subtree) /
-              static_cast<double>(n_valid);
-          } else {
-            subtree_prob = n_valid_subtree ? 1 : 0;
-          }
-
-          if (this->rand_uniform_() < subtree_prob)
```
bob-carpenter: If I'm following this, you're using an implicit binary search with a random prob in (0,1) at each branching point.  It should only require a single RNG that can then be used to guide the search down the subtrees.  I can't quite follow this closely enough to see if what you're doing then allows a subtree that isn't selected to be freed, which would be great.  I'd love to see the pseudo-code for this in the NUTS III paper.

src/stan/mcmc/hmc/nuts/base_nuts.hpp
betanalpha: It's a template issue -- lookup occurs only after template instantiation so the internal class members and methods aren't in scope.  See http://stackoverflow.com/questions/4643074/why-do-i-have-to-access-template-base-class-members-through-the-this-pointer.

src/stan/mcmc/hmc/nuts/base_nuts.hpp
```diff
@@ -171,72 +141,75 @@ namespace stan {
...
-            static_cast<double>(n1 + n2);
+          bool valid_right
+            = build_tree(depth - 1, rho_subtree, z_propose_right,
+                         H0, sign, n_leapfrog,
+                         sum_weight_right, sum_metro_prob, writer);
 
-          if ( util.criterion && (this->rand_uniform_() < accept_prob) )
-            z_propose = z_propose_right;
+          sum_weight += sum_weight_right;
+          if (!valid_right) return false;
```
bob-carpenter: Same question as above w.r.t. order of sum and rejection.

src/stan/mcmc/hmc/nuts/base_nuts.hpp
betanalpha: No, we should be good because each weight is normalized by exp(H0).  If the Hamiltonian gets smaller then the algebra is fine and if it gets too large then we'll break with a divergence anyways.  I can change to log_sum_exp() if people would prefer.

src/stan/mcmc/hmc/nuts/base_nuts.hpp
```diff
@@ -171,72 +141,75 @@ namespace stan {
...
-          Eigen::VectorXd& subtree_rho = left_subtree_rho;
-          subtree_rho += right_subtree_rho;
-
-          rho += subtree_rho;
-
-          util.criterion &= compute_criterion(z_init, this->z_, subtree_rho);
+          // Multinomial sample from right subtree
+          double accept_prob = sum_weight_right / sum_weight;
+          if (this->rand_uniform_() < accept_prob)
+            z_propose = z_propose_right;
```
bob-carpenter: I take it `z_propose` is equal to the `z_propose_left` at this point so everything happens online?  This is very slick.

src/stan/mcmc/hmc/nuts/base_nuts.hpp
betanalpha: Doesn't matter as the weight isn't used after the break.

src/stan/mcmc/hmc/nuts_classic/adapt_dense_e_nuts_classic.hpp
```diff
@@ -0,0 +1,55 @@
+#ifndef STAN_MCMC_HMC_NUTS_ADAPT_DENSE_E_NUTS_CLASSIC_HPP
```
bob-carpenter: Is the point of `nuts_classic` to keep the old implementation?  And if so, do we really need both?  I don't care about having backward compatibility at the underlying algorithm level.

src/stan/mcmc/hmc/nuts_classic/adapt_diag_e_nuts_classic.hpp
```diff
@@ -0,0 +1,56 @@
...
+namespace stan {
+  namespace mcmc {
+
+    // The No-U-Turn Sampler (NUTS) on a
+    // Euclidean manifold with diagonal metric
+    // and adaptive stepsize
+
+    template <class Model, class BaseRNG>
+    class adapt_diag_e_nuts_classic:
+      public diag_e_nuts_classic<Model, BaseRNG>,
```
bob-carpenter: Multiple inheritance is a really dangerous pattern.  I'm not saying you have to change this here, but you really want to think about including a reference to a class rather than using multiple inheritance.  It's much easier to reason about and usually the relationship you have is not an is-a relationship.  Here it'd seem to make sense to have the `stepsize_var_adaptor` be held as a member variable.

src/stan/mcmc/hmc/nuts/base_nuts.hpp
betanalpha: Yes, it's a bit wasteful on RNGs but we also don't know the depth of three a priori so there would have to be at least some dynamic RNG generation.  Not sure if Boost has a binary RNG that uses up less of the generator or not.

betanalpha: What do you mean freed?  The entire tree is never held in memory.

src/stan/mcmc/hmc/nuts/base_nuts.hpp
betanalpha: Same response -- once there's an invalid subtree the weights aren't used.

src/stan/mcmc/hmc/nuts_classic/base_nuts_classic.hpp
```diff
@@ -0,0 +1,249 @@
...
+      base_nuts_classic(Model &model, BaseRNG& rng):
+        base_hmc<Model, Hamiltonian, Integrator, BaseRNG>(model, rng),
+        depth_(0), max_depth_(5), max_delta_(1000),
+        n_leapfrog_(0), divergent_(0), energy_(0) {
+      }
+
+      ~base_nuts_classic() {}
+
+      void set_max_depth(int d) {
+        if (d > 0)
```
bob-carpenter: I'd prefer an exception if `d <= 0`, as it indicates a programmer error on the outside.  If it's only called internally, you don't need the check as you can make sure it's called properly.

src/stan/mcmc/hmc/nuts/base_nuts.hpp
betanalpha: Yes.  See http://arxiv.org/abs/hep-lat/9208011 and http://arxiv.org/abs/1601.00225.

src/stan/mcmc/hmc/nuts_classic/base_nuts_classic.hpp
```diff
@@ -0,0 +1,249 @@
...
+        // Build a balanced binary tree until the NUTS criterion fails
+        util.criterion = true;
+        int n_valid = 0;
+
+        this->depth_ = 0;
+        this->divergent_ = 0;
+
+        util.n_tree = 0;
+        util.sum_prob = 0;
+
```
bob-carpenter: I'd prefer to see util constructed with all its pieces in one line, but I can see how you'd like the names here.

src/stan/mcmc/hmc/nuts_classic/adapt_dense_e_nuts_classic.hpp
betanalpha: Yes, just to keep the slice sampler around in case someone wants it.

betanalpha:
@bob-carpenter NUTS classic is a copy of the existing implementation which has been in production for some time.  Many of the points being raised were cleaned up in the new implementation -- I'd rather keep NUTS classic as it is.

src/stan/mcmc/hmc/nuts_classic/base_nuts_classic.hpp
```diff
@@ -0,0 +1,249 @@
...
+          if (this->rand_uniform_() > 0.5) {
+            z = &z_plus;
+            rho = &rho_plus;
+            util.sign = 1;
+          } else {
+            z = &z_minus;
+            rho = &rho_minus;
+            util.sign = -1;
+          }
+
```
bob-carpenter: I think using references is a lot cleaner here.

```
bool is_plus = this->rand_uniform_() > 0.5;
ps_point& z = is_plus ? z_plus : z_minus;
VectorXd& rho = is_plus ? rho_plus : rho_minus;
int util.sign = is_plus ? 1 : -1;  // or more efficiently, is_plus * 2 - 1
```
because then you can use them like ordinary variables instead of pointers and you're guaranteed by the compiler to know they aren't going to segfault.

src/stan/mcmc/hmc/nuts_classic/base_nuts_classic.hpp
```diff
@@ -0,0 +1,249 @@
...
+          double subtree_prob = 0;
+
+          if (n_valid) {
+            subtree_prob = static_cast<double>(n_valid_subtree) /
+              static_cast<double>(n_valid);
+          } else {
+            subtree_prob = n_valid_subtree ? 1 : 0;
+          }
+
```
bob-carpenter: You can get away from redefining with

```
double subtree_prob
  = n_valid 
  ? static_cast<double>(n_valid_subtree) / n_valid
  : n_valid_subtree ? 1 : 0;
```

src/stan/mcmc/hmc/nuts_classic/base_nuts_classic.hpp
```diff
@@ -0,0 +1,249 @@
...
+          // Check validity of completed tree
+          this->z_.ps_point::operator=(z_plus);
+          Eigen::VectorXd delta_rho = rho_minus + rho_init + rho_plus;
+
+          util.criterion = compute_criterion(z_minus, this->z_, delta_rho);
+        }
+
+        this->n_leapfrog_ = util.n_tree;
+
+        double accept_prob = util.sum_prob / static_cast<double>(util.n_tree);
```
bob-carpenter: doesn't need the cast

src/stan/mcmc/hmc/nuts_classic/base_nuts_classic.hpp
```diff
@@ -0,0 +1,249 @@
...
+
+            if (z_init_parent) *z_init_parent = this->z_;
+            z_propose = this->z_;
+
+            double h = this->hamiltonian_.H(this->z_);
+            if (boost::math::isnan(h))
+              h = std::numeric_limits<double>::infinity();
+
+            util.criterion = util.log_u + (h - util.H0) < this->max_delta_;
+            if (!util.criterion) ++(this->divergent_);
```
bob-carpenter: Shouldn't `divergent_` be `bool` rather than `int` now?  In which case, the line above can just set the value to `true`.

src/stan/mcmc/hmc/nuts_classic/base_nuts_classic.hpp
```diff
@@ -0,0 +1,249 @@
...
+
+            util.sum_prob += std::min(1.0, std::exp(util.H0 - h));
+            util.n_tree += 1;
```
bob-carpenter: `++util.n_tree;`

src/stan/mcmc/hmc/nuts_classic/base_nuts_classic.hpp
```diff
@@ -0,0 +1,249 @@
...
+          } else {
+          // General recursion
+          Eigen::VectorXd left_subtree_rho(rho.size());
+          left_subtree_rho.setZero();
+          ps_point z_init(this->z_);
+
+          int n1 = build_tree(depth - 1, left_subtree_rho, &z_init,
+                              z_propose, util, writer);
+
+          if (z_init_parent) *z_init_parent = z_init;
```
bob-carpenter: I'd really like to lean less heavily on pointers, but this one's going to be too difficult to disentangle!

src/stan/mcmc/hmc/nuts_classic/base_nuts_classic.hpp
```diff
@@ -0,0 +1,249 @@
...
+
+          if ( util.criterion && (this->rand_uniform_() < accept_prob) )
+            z_propose = z_propose_right;
+
+          Eigen::VectorXd& subtree_rho = left_subtree_rho;
+          subtree_rho += right_subtree_rho;
+
+          rho += subtree_rho;
+
+          util.criterion &= compute_criterion(z_init, this->z_, subtree_rho);
```
bob-carpenter: Be careful with this, because `&=` is bitwise, not logical, so it'd be clearner with an explicit `util.criterion && ...` on the RHS.

src/stan/mcmc/hmc/nuts_classic/base_nuts_classic.hpp
```diff
@@ -0,0 +1,249 @@
...
+          return n1 + n2;
+        }
+      }
+
+      int depth_;
+      int max_depth_;
+      double max_delta_;
+
+      int n_leapfrog_;
+      int divergent_;
```
bob-carpenter: Should `divergent_` be `bool`?

src/stan/mcmc/hmc/nuts/base_nuts.hpp
bob-carpenter: No, that's fine if it's going to be stable.  I just thought it'd
be easier to ask you than work through the logic myself!

src/stan/mcmc/hmc/nuts/base_nuts.hpp
bob-carpenter: In which case it'd save an operation to put it after the break, not
that a single add and assign is going to break the operation bank.  I
just think it's clearer.

src/stan/mcmc/hmc/nuts/base_nuts.hpp
bob-carpenter: Thanks for the pointer.  I thought it was some awful C++ thing like that.

src/stan/mcmc/hmc/nuts/base_nuts.hpp
bob-carpenter: Great --- I think it'll just fall off the stack.

src/stan/mcmc/hmc/nuts_classic/adapt_dense_e_nuts_classic.hpp
bob-carpenter: I'd rather not keep it around if you think the new one's going
to be better (or at least as good).  I don't want the long term
support hassle and I don't want to give people too many knobs to
twist that aren't going to have much effect.  Your call, though.

bob-carpenter:
Got it.  See previous comment.  This inline commenting thing is tough
to follow!

bob-carpenter:
This is good to go from the code perspective.  All of my comments were superficial points about coding style and idioms.  I like this new algorithm approach, too.

@syclik Is this going to be OK upstream with the interfaces?  If so, go ahead and merge.

syclik:
Yup! No issues upstream.

Issue #1849, opened 2016-04-07
Add uniform static HMC sampler


betanalpha:
#### Summary:
Add a static HMC variant.

#### Description:
Add a static HMC sampler that samples trajectories of fixed sized uniformly around the initial point instead of deterministically forward in time.

PR #1850, opened 2016-04-07
Add uniform static HMC sampler

betanalpha:
#### Submisison Checklist

#### Summary

Adds a static HMC sampler that samples trajectories of fixed sized uniformly around the initial point instead of deterministically forward in time.  Addresses #1849.

#### Intended Effect

Adds a new HMC sampler.

#### How to Verify

Run tests.

#### Side Effects

None.

#### Documentation

None.

#### Reviewer Suggestions

Anyone.

#### Copyright and Licensing

Copyright: Michael Betancourt

### DIFF
--- a/src/stan/mcmc/hmc/static_uniform/adapt_dense_e_static_uniform.hpp
+++ b/src/stan/mcmc/hmc/static_uniform/adapt_dense_e_static_uniform.hpp
@@ -1,0 +1,53 @@
+#ifndef STAN_MCMC_HMC_STATIC_UNIFORM_ADAPT_DENSE_E_STATIC_UNIFORM_HPP
+#define STAN_MCMC_HMC_STATIC_UNIFORM_ADAPT_DENSE_E_STATIC_UNIFORM_HPP
+
+#include <stan/mcmc/stepsize_covar_adapter.hpp>
+#include <stan/mcmc/hmc/static_uniform/dense_e_static_uniform.hpp>
+
+namespace stan {
+  namespace mcmc {
+    // Hamiltonian Monte Carlo on a
+    // Euclidean manifold with dense metric,
+    // static integration time,
+    // and adaptive stepsize
+    template <typename M, class BaseRNG>
+    class adapt_dense_e_static_uniform:
+      public dense_e_static_uniform<M, BaseRNG>,
+      public stepsize_covar_adapter {
+    public:
+      adapt_dense_e_static_uniform(M &m, BaseRNG& rng):
+        dense_e_static_uniform<M, BaseRNG>(m, rng),
+        stepsize_covar_adapter(m.num_params_r()) { }
+
+      ~adapt_dense_e_static_uniform() { }
+
+      sample transition(sample& init_sample,
+                        interface_callbacks::writer::base_writer& writer) {
+        sample s = dense_e_static_uniform<M, BaseRNG>::transition(init_sample,
+                                                                  writer);
+
+        if (this->adapt_flag_) {
+          this->stepsize_adaptation_.learn_stepsize(this->nom_epsilon_,
+                                                    s.accept_stat());
+
+          bool update = this->covar_adaptation_.learn_covariance
+            (this->z_.mInv, this->z_.q);
+
+          if (update) {
+            this->init_stepsize(writer);
+            this->stepsize_adaptation_.set_mu(log(10 * this->nom_epsilon_));
+            this->stepsize_adaptation_.restart();
+          }
+        }
+        return s;
+      }
+
+      void disengage_adaptation() {
+        base_adapter::disengage_adaptation();
+        this->stepsize_adaptation_.complete_adaptation(this->nom_epsilon_);
+      }
+    };
+  }  // mcmc
+}  // stan
+
+#endif
--- a/src/stan/mcmc/hmc/static_uniform/adapt_dense_e_static_uniform.hpp
+++ b/src/stan/mcmc/hmc/static_uniform/adapt_dense_e_static_uniform.hpp
@@ -6,10 +6,12 @@
 
 namespace stan {
   namespace mcmc {
-    // Hamiltonian Monte Carlo on a
-    // Euclidean manifold with dense metric,
-    // static integration time,
-    // and adaptive stepsize
+    /**
+      * Hamiltonian Monte Carlo implemetnation that uniformly samples
+      * from trajectories with a static integration time with a
+      * Gaussian-Euclidean disintegration and dense metric and
+      * adaptive step size
+    */
     template <typename M, class BaseRNG>
     class adapt_dense_e_static_uniform:
       public dense_e_static_uniform<M, BaseRNG>,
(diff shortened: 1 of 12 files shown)


src/stan/mcmc/hmc/static_uniform/adapt_dense_e_static_uniform.hpp
```diff
@@ -0,0 +1,53 @@
+#ifndef STAN_MCMC_HMC_STATIC_UNIFORM_ADAPT_DENSE_E_STATIC_UNIFORM_HPP
+#define STAN_MCMC_HMC_STATIC_UNIFORM_ADAPT_DENSE_E_STATIC_UNIFORM_HPP
+
+#include <stan/mcmc/stepsize_covar_adapter.hpp>
+#include <stan/mcmc/hmc/static_uniform/dense_e_static_uniform.hpp>
+
+namespace stan {
+  namespace mcmc {
+    // Hamiltonian Monte Carlo on a
```
bob-carpenter: This should be a class comment rather than `//` for the doc.  No big deal given our doc is so poor for this stuff, but we ought to at least move in the right direction here.

bob-carpenter:
Is there a description of the algorithm somewhere?  By uniformly, do you mean if the user specifies L steps, L1 < L go backward and L - L1 go forward?  Knowing this will make it much easier to review the code.

Why do this rather than going one way?  Naively, it'd seem going one way is likely to get you further.

Should the manual be updated at the same time?

Is this replacing the old static HMC?  (I assume not, as it says "adds" rather than "replaces".)  If it's not replacing the old one, why do we need two?  I don't remember discussing this and didn't see any motivation on the issue.

src/stan/mcmc/hmc/static_uniform/adapt_dense_e_static_uniform.hpp
betanalpha: Example, please.

bob-carpenter:
```
/**
 * Hamiltonian Monte Carlo on a...
 */
```

That will get picked up as the class doc from doxygen.

bob-carpenter:
Looks like this also has conflicts that need to be resolved if you didn't notice.

It's possible to branch from other branches, and Git figures out to merge everything back into develop.

betanalpha:
http://arxiv.org/abs/1601.00225, Section 2.  There is a reason I write these papers....

The nominal implementation of HMC uses only the final point whereas ideally we'd sample uniformly from the entire trajectory based on the Metropolis probabilities.  But to do that we have to randomize where the initial point lies in the overall trajectory, hence integrating both forwards and backwards in time.

The user specifies L (or really time and L gets computed from the step size) then L1 is sampled uniformly from 0 to L.  Then the trajectory is built up from L1 steps backwards in time and (L - L1) steps forward in time.  A point is sampled from the trajectory dynamically.

This is not user facing for the moment as it is aimed for upcoming research, so no doc is needed.

betanalpha:
Doc updated and merge conflict resolved (apparently git is not very good at merging parallel changes to makefiles).

bob-carpenter:
Thanks.  That makes sense.

I find your papers very challenging, whereas these two-sentence
just-in-time summaries are much easier to digest :-)

If it's only for resaerch, is there a reason to put it into Stan
itself rather than keeping it on a branch?

- Bob

betanalpha:
Yes, it helps stage various research studies and potential
changes to the main Stan features without having to maintain
a complete fork.  It also facilitates future decisions about
deprecating/removing older (naive) implementations.

bob-carpenter:
OK.  It's not going to be hard to maintain, so I'm OK with
this.

betanalpha:
So are we good to merge?

src/stan/mcmc/hmc/static_uniform/base_static_uniform.hpp
```diff
@@ -0,0 +1,159 @@
...
+
+        this->z_.ps_point::operator=(z_init);
+
+        for (int l = 0; l < L_ - 1 - Lp; ++l) {
+          this->integrator_.evolve(this->z_, this->hamiltonian_,
+                                   this->epsilon_, writer);
+
+          double h = this->hamiltonian_.H(this->z_);
+          if (boost::math::isnan(h))
+            h = std::numeric_limits<double>::infinity();
```
bob-carpenter: Can we predict more of what's going to happen downstream once `h` becomes infinite?   This is more a question than a request for a code change.

src/stan/mcmc/hmc/static_uniform/base_static_uniform.hpp
```diff
@@ -0,0 +1,159 @@
...
+
+      void set_nominal_stepsize_and_L(const double e, const int l) {
+        if (e > 0 && l > 0) {
+          this->nom_epsilon_ = e;
+          L_ = l;
+          T_ = this->nom_epsilon_ * L_; }
+      }
+
+      void set_T(const double t) {
+        if (t > 0) {
```
bob-carpenter: If this is controlled through the API and you know `t > 0` is satisifed by call calls, you don't need this test.  Usually you want these tests if something external can mess things up, and then you usually want to raise an exception, not just ignore the call where `t <= 0`.  Or is there some t = 0 case that matters here?

src/stan/mcmc/hmc/static_uniform/unit_e_static_uniform.hpp
```diff
@@ -0,0 +1,27 @@
...
+    /**
+      * Hamiltonian Monte Carlo implemetnation that uniformly samples
+      * from trajectories with a static integration time with a
+      * Gaussian-Euclidean disintegration and unit metric
+    */
+    template <typename M, class BaseRNG>
+    class unit_e_static_uniform
+      : public base_static_uniform<M, unit_e_metric, expl_leapfrog, BaseRNG> {
+    public:
+      unit_e_static_uniform(M &m, BaseRNG& rng):
```
bob-carpenter: If M is the model class here, shouldn't it be const?  I don't think a model has anything mutable in it.

bob-carpenter:
I had a couple more questions, but yes, this is basically ready to merge whenever you think it's ready.

src/stan/mcmc/hmc/static_uniform/base_static_uniform.hpp
betanalpha: Perhaps, but nothing I’d feel comfortable with at the moment.

src/stan/mcmc/hmc/static_uniform/unit_e_static_uniform.hpp
betanalpha: It would make sense, but I don’t think we’ve followed that
convention anywhere else.  Also, if there aren’t any mutable
members then it wouldn’t matter.  Perhaps worth a note for
later code cleanup.

Issue #1852, opened 2016-04-10
Streamline Generalized NUTS Criterion


betanalpha:
The current implementation of the (generalized) No-U-Turn criterion needs to be optimized to use less memory and become suitable for Riemannian HMC.

PR #1854, opened 2016-04-10
Feature/issue 1852 streamline nuts criterion

betanalpha:
#### Submisison Checklist

#### Summary

Tweaks the (generalized) No-U-Turn criterion.  Addresses #1852.

#### Intended Effect

The (generalized) No-U-Turn criterion is slightly tweaked which not only allows for an implementation that uses less memory but also makes adding Riemannian NUTS much simpler.  The change is essentially an edge case in the trajectory discretization (include the last point or not?) that is ambiguous in the theoretical definition, so it shouldn't make any appreciable difference in sampler performance.

#### How to Verify

Run unit tests or try running some multivariate models with known expectations.  Everything should fit just as well as before.

#### Side Effects

The tweak in the criterion definition will slightly change the exact sampler output but should not affect the accuracy of any expectations.  For small step sizes any such effects should be negligible.

#### Documentation

No user-facing changes.

#### Reviewer Suggestions

Anyone.

#### Copyright and Licensing

Copyright: Michael Betancourt

### DIFF
--- a/src/stan/mcmc/hmc/nuts/adapt_dense_e_nuts.hpp
+++ b/src/stan/mcmc/hmc/nuts/adapt_dense_e_nuts.hpp
@@ -7,10 +7,11 @@
 
 namespace stan {
   namespace mcmc {
-
-    // The No-U-Turn Sampler (NUTS) on a
-    // Euclidean manifold with dense metric
-    // and adaptive stepsize
+    /**
+      * The No-U-Turn sampler (NUTS) with multinomial sampling
+      * with a Gaussian-Euclidean disintegration and adaptive
+      * dense metric and adaptive step size
+    */
     template <class Model, class BaseRNG>
     class adapt_dense_e_nuts : public dense_e_nuts<Model, BaseRNG>,
                                public stepsize_covar_adapter {
--- a/src/stan/mcmc/hmc/nuts/adapt_diag_e_nuts.hpp
+++ b/src/stan/mcmc/hmc/nuts/adapt_diag_e_nuts.hpp
@@ -7,11 +7,11 @@
 
 namespace stan {
   namespace mcmc {
-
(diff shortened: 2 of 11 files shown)


bob-carpenter:
I really like that this removes a lot of code duplication.  But I can't tell what the specific tweak is.  Could you give me a hint so that I can try to check if it's doing what it's supposed to be doing?  Are `p_sharp_plus` and `p_sharp_minus` the current forward-in-time and backward-in-time endpoints?  If so, what's `p_sharp_left`?

Also, does there need to be a change to the manual algorithms chapters to match this change?

betanalpha:
The theoretical definition is

<p_{N}, rho>

where

rho = \int dt p(t).

The old implementation approximated rho without the final point,

rho_old = \sum_{n = 0}^{n - 1} p_{n},

while the new implementation does use the final point, just to
avoid having to keep the momenta values around for the subtraction,

rho_new = \sum_{n = 0}^{N} p_{n},

This way instead of needing to save the momentum and the
metric to compute M^{-1} * p_{N} in the inner product _and_
subtract p_{N} from rho we can just store p_sharp = M^{-1} * p.
Abstracting to p_sharp, which is computed by each Hamiltonian
implementation, is what admits the clean up of the criterion
implementations.

minus/plus refer to the forward/backward endpoints where as
left/right refers to the endpoint in the current subtree which may 
be flipped with respect to the global definition of minus/plus.  It’s
all reversible so the actual order doesn’t matter, but the naming
should be self-consistent.

The doc doesn’t touch on any of these details so no changes
are necessary.

bob-carpenter:
Thanks.  Just what I needed.

bob-carpenter:
This one looks good to go, too.  I love the way you have all this code organized; it makes it very easy to review.

PR #1859, opened 2016-04-12
Feature/issue 1855 add xhmc

betanalpha:
#### Submisison Checklist

#### Summary

Introduces Exhaustive Hamiltonian Monte Carlo (XHMC), an algorithm with a similar dynamic integration time scheme to NUTS but with a different termination criterion.

#### Intended Effect

New sampler implementation in the MCMC code with some service support in case we expose it later.

#### How to Verify

Run unit tests.

#### Side Effects

MCMC tests were reorganized and cleaned up.

#### Documentation

None as we haven't exposed it yet.

#### Reviewer Suggestions

Anyone.

#### Copyright and Licensing

Copyright: Michael Betancourt


src/stan/mcmc/hmc/hamiltonians/base_hamiltonian.hpp
```diff
@@ -36,6 +36,8 @@ namespace stan {
         return T(z) + V(z);
       }
 
+      virtual double dG_dt(Point& z) = 0;
```
bob-carpenter: Can this be `const Point&`?  It's much easier to get the const correctness in from the get go.  And same question for `dtau_dq()` below.

bob-carpenter: Can you add some doc at least in the base class explaining what dG_dt is?  Just a note and maybe a pointer to a paper.

src/stan/mcmc/hmc/xhmc/adapt_dense_e_xhmc.hpp
```diff
@@ -0,0 +1,55 @@
...
+#define STAN_MCMC_HMC_xhmc_ADAPT_DENSE_E_XHMC_HPP
+
+#include <stan/interface_callbacks/writer/base_writer.hpp>
+#include <stan/mcmc/stepsize_covar_adapter.hpp>
+#include <stan/mcmc/hmc/xhmc/dense_e_xhmc.hpp>
+
+namespace stan {
+  namespace mcmc {
+    /**
+      * The No-U-Turn sampler (XHMC) with multinomial sampling
```
bob-carpenter: The following * go under the first, not the second one.  Emacs just does this automatically.

src/stan/mcmc/hmc/xhmc/adapt_dense_e_xhmc.hpp
```diff
@@ -0,0 +1,55 @@
+#ifndef STAN_MCMC_HMC_xhmc_ADAPT_DENSE_E_XHMC_HPP
+#define STAN_MCMC_HMC_xhmc_ADAPT_DENSE_E_XHMC_HPP
+
```
bob-carpenter: The xhmc should be XHMC to be consistent with the rest of our header guards.

src/stan/mcmc/hmc/xhmc/adapt_dense_e_xhmc.hpp
bob-carpenter: Is there a top-level note about the different flavors of NUTS in play?  I'm not sure what XHMC is (I was guessing it was exhaustive from your papers, but here it's just listed as being a name for NUTS.

src/stan/mcmc/hmc/xhmc/adapt_dense_e_xhmc.hpp
```diff
@@ -0,0 +1,55 @@
...
+    public:
+        adapt_dense_e_xhmc(Model &model, BaseRNG& rng)
+          : dense_e_xhmc<Model, BaseRNG>(model, rng),
+          stepsize_covar_adapter(model.num_params_r()) {}
+
+      ~adapt_dense_e_xhmc() {}
+
+      sample transition(sample& init_sample,
+                        interface_callbacks::writer::base_writer& writer) {
+        sample s = dense_e_xhmc<Model, BaseRNG>::transition(init_sample,
```
bob-carpenter: I like this arrangement of calling (one of) the superclass(es) first and then doing adaptation.

src/stan/mcmc/hmc/xhmc/adapt_diag_e_xhmc.hpp
```diff
@@ -0,0 +1,55 @@
...
+          stepsize_var_adapter(model.num_params_r()) {}
+
+      ~adapt_diag_e_xhmc() {}
+
+      sample transition(sample& init_sample,
+                        interface_callbacks::writer::base_writer& writer) {
+        sample s = diag_e_xhmc<Model, BaseRNG>::transition(init_sample,
+                                                           writer);
+
+        if (this->adapt_flag_) {
```
bob-carpenter: This is so close to being duplicated code, only with variance instead of covariance plugged in.  But I don't see a good way to factor it without making it too hard to follow.

src/stan/mcmc/hmc/xhmc/adapt_unit_e_xhmc.hpp
```diff
@@ -0,0 +1,43 @@
+#ifndef STAN_MCMC_HMC_xhmc_ADAPT_UNIT_E_XHMC_HPP
```
bob-carpenter: This lower-case xhmc again.  But what I really wanted to ask is whether it would be possible to factor the mass matrix adaptation from the step size adaptation so that unit_e is really just a specified mass matrix with no adaptation of the mass matrix.  Or will that make the arithmetic slower to actually have the mass matrix in there doing nothing?

src/stan/mcmc/hmc/xhmc/base_xhmc.hpp
```diff
@@ -0,0 +1,223 @@
...
+      base_xhmc(Model &model, BaseRNG& rng)
+        : base_hmc<Model, Hamiltonian, Integrator, BaseRNG>(model, rng),
+          depth_(0), max_depth_(5), max_deltaH_(1000), x_delta_(0.1),
+          n_leapfrog_(0), divergent_(0), energy_(0) {
+      }
+
+      ~base_xhmc() {}
+
+      void set_max_depth(int d) {
+        if (d > 0)
```
bob-carpenter: I really think these should either throw exceptions or be eliminated and the checking left up to the caller.

src/stan/mcmc/hmc/xhmc/base_xhmc.hpp
```diff
@@ -0,0 +1,223 @@
...
+        // Compute average acceptance probabilty across entire trajectory,
+        // even over subtrees that may have been rejected
+        double accept_prob
+          = sum_metro_prob / static_cast<double>(n_leapfrog + 1);
+
+        this->z_.ps_point::operator=(z_sample);
+        this->energy_ = this->hamiltonian_.H(this->z_);
+        return sample(this->z_.q, -this->z_.V, accept_prob);
+      }
+
```
bob-carpenter: This is really awesome.  The algorithm's as clean (if not cleaner) in C++ than in the paper's pseudocode!

src/stan/mcmc/hmc/xhmc/base_xhmc.hpp
```diff
@@ -0,0 +1,223 @@
...
+            double pi = exp(H0 - h);
+            sum_numer += pi * this->hamiltonian_.dG_dt(this->z_);
+            sum_weight += pi;
+            sum_metro_prob += pi > 1 ? 1 : pi;
+
+            z_propose = this->z_;
+
+            return !this->divergent_;
+
+        } else {
```
bob-carpenter: Don't need the `else` block here given that you returned from the `if` branch.  This is one of the standard refactors that seemed wrong to me at first.   But once you get used to always putting the "special" cases first, it becomes much easier to read the "normal" control flow.

bob-carpenter:
I really like the overall organization of this code.  It would really help to add doc to the base class function declarations.  The xhmc in the header guards should be caps.  And there should be some pointer back to the papers or manual or wherever XHMC is defined.  Otherwise, the comments above are all just suggestions and don't even call for responses.  (I should mark that as I go somehow, since I trail a lot of suggestions in these reviews.)

The const correctness would be nice if it can be done relatively locally --- otherwise I agree it needs a later pass.

src/stan/mcmc/hmc/hamiltonians/base_hamiltonian.hpp
betanalpha: They actually can't be made const as the Point objects also hold auxiliary state to avoid recomputation, such as the gradients and, for Riemannian HMC, metric and other intermediate variables.  Many of these classes will modify those internal states.

src/stan/mcmc/hmc/xhmc/adapt_unit_e_xhmc.hpp
betanalpha: Technically it would be slower due to the additional multiplication by 1's everywhere.

betanalpha:
Since I reorganized all of the tests in this PR I went ahead and cleaned up all of the samplers.  This includes doc, the bad header capitalization in XHMC, removing the extraneous else from the build_tree methods, and changing all of the model references to const references.  Once the tests pass I'll merge, then it's on to Riemannian metrics...

src/stan/mcmc/hmc/xhmc/adapt_unit_e_xhmc.hpp
bob-carpenter: Tha's what I suspected.  Oh well.

PR #1861, opened 2016-04-14
Feature/issue 304 add rhmc

betanalpha:
#### Submisison Checklist

#### Summary

Introduce Riemannian SoftAbs metrics to the active samplers.

#### Intended Effect

Allows each sampler to use the SoftAbs metric in addition to
the existing Euclidean metric variants.

#### How to Verify

Run unit tests, inspect code.  There are lots of potential design
and performance improvements so this is to be considered
a preliminary implementation.

#### Side Effects

None.

#### Documentation

None.

#### Reviewer Suggestions

Anyone.

#### Copyright and Licensing

University of Warwick


betanalpha:
Anyone understand the Windows error trying to instantiate the higher-order autodiff types?

syclik:
Bob and I have seen this before. Something isn't in the right order. Our specialization of Eigen's traits needs to be included before Eigen instantiates it on its own or you get that message. Might be a problem with the header include, but maybe there's something included before that? I haven't dug into it yet. 

Does it not happen on Mac?

betanalpha:
Everything passed locally on my machine.

bob-carpenter:
See

http://stackoverflow.com/questions/21112148/specialization-of-member-function-template-after-instantiation-error-and-order

http://stackoverflow.com/questions/7774188/explicit-specialization-after-instantiation

It's about member functions, but the point's the same.  You need
to have all the definitions included before any of the instantiations.
For some reason, some compilers are pickier than others about this.

Oh how I miss Java, where the order didn't matter!

betanalpha:
So I need prototypes of the specializations of the autodiff functions
with higher-order autodiff types and the softabs functional defined
in softabs_metric?

bob-carpenter:
You just need to make sure the declarations (not necessarily the
definitions) come before the instantiations.

The bad case is this:

====================================
template <typename A>
struct foo {
  A a_;
  foo(const A& a) : a_(a) { }
};

int bar() {
  return foo<int>(3).a_;
}

template <>
struct foo<int> {
  int x_;
  foo(int x) : x_(2 * x) { }
}

int main() {
  return bar();
}
=====================================

./bad.cpp:12:8: error: explicit specialization of 'foo<int>' after instantiation
struct foo<int> {
       ^~~~~~~~

If you drop in the forward declaration before the definition of bar():


=======================================
template <> struct foo<int>;
===================================

you get

./bad.cpp:10:10: error: 'foo<int>' is an incomplete type
  return foo<int>(3).a_;
         ^
./bad.cpp:7:20: note: forward declaration of 'foo<int>'
template <> struct foo<int>;


So the point is that 

1)  you can't instantiate a general template class then specialize
later (first case)

2)  you must define before you instantiate (second case)

- Bob

betanalpha:
Would this be due to the fact that I call the explicit double instantiations of
hessian and grad_hessian_times_matrix?  Sorry, I’m just having trouble
wrapping my head around how all of this interacts with the higher-order
autodiff.

bob-carpenter:
Right --- you can't instantiate with double until you've
declared all the specializations that are coming.   

BAD:
define base template version T
instantiate base template with double
define template specialization for var, fvar

GOOD:
define base template version T
declare specializations
instantiate base template with double
define specializations
instantiate specializations with var, fvar, etc.

- Bob

bob-carpenter:
So, if we add the `-fpermissive` flag, everything works just fine.  Turns out the problem is that g++ is being particularly strict.  It's probably better to fix the code, though, so I'll try to dive a little deeper and see if I can uncover the root cause.  

From the g++ doc:

```
-fpermissive
Downgrade some diagnostics about nonconformant code from errors to warnings. Thus, using -fpermissive will allow some nonconforming code to compile. 
```

betanalpha:
Okay, we're good to go.  Now who wants to code review this mess?

bob-carpenter:
I can code review it.  Anything I should be paying attention to among the "mess" of code?

betanalpha:
The Hamiltonians and integrators were refactored to admit both
explicit and implicit leapfrog integrators within the same framework.
The softabs_metric might be doing a lot of unnecessary computation
that could be cached — it’s certainly creating lots of big temporary
matrices.  Not sure if we want to worry about such optimizations yet.

Everything should be fully tested.

bob-carpenter:
Thanks --- those complications are to be expected.
I won't worry about optimization unless something
jumps out at me.

- Bob

syclik:
Looks good to me.

syclik:
@betanalpha, before merging, I just noticed that some of the tests need to be added to the `make/tests` file to automatically generate the header files. I'm going to add a comment to the file. Mind updating it?

make/tests
```diff
@@ -133,6 +133,8 @@ src/test/unit/mcmc/hmc/hamiltonians/diag_e_metric_test.cpp: src/test/test-models
 src/test/unit/mcmc/hmc/hamiltonians/unit_e_metric_test.cpp: src/test/test-models/good/mcmc/hmc/hamiltonians/funnel.hpp
 src/test/unit/mcmc/hmc/integrators/expl_leapfrog_test.cpp: src/test/test-models/good/mcmc/hmc/integrators/command.hpp
 src/test/unit/mcmc/hmc/integrators/expl_leapfrog2_test.cpp: src/test/test-models/good/mcmc/hmc/integrators/gauss.hpp
+src/test/unit/mcmc/hmc/integrators/impl_leapfrog_test.cpp: src/test/test-models/good/mcmc/hmc/integrators/command.hpp
+src/test/unit/mcmc/hmc/integrators/impl_leapfrog2_test.cpp: src/test/test-models/good/mcmc/hmc/integrators/gauss.hpp
```
syclik: Need to add a few more test dependencies:
- `src/test/unit/mcmc/hmc/hamiltonians/softabs_metric_test.cpp`
- `src/test/unit/mcmc/hmc/nuts/softabs_nuts_test.cpp`
- ` src/test/unit/mcmc/hmc/xhmc/softabs_xhmc_test.cpp`

betanalpha:
Updated.

src/stan/mcmc/hmc/hamiltonians/base_hamiltonian.hpp
```diff
@@ -37,29 +37,50 @@ namespace stan {
       }
 
       // The time derivative of the virial, G = \sum_{d = 1}^{D} q^{d} p_{d}.
-      virtual double dG_dt(Point& z) = 0;
+      virtual double dG_dt(
+        Point& z,
```
bob-carpenter: I'd strongly prefer arguments start on the same line as the function when they'll fit to reduce vertical space.

src/stan/mcmc/hmc/hamiltonians/base_hamiltonian.hpp
```diff
@@ -37,29 +37,50 @@ namespace stan {
...
+      virtual const Eigen::VectorXd dtau_dq(
+        Point& z,
+        interface_callbacks::writer::base_writer& info_writer,
+        interface_callbacks::writer::base_writer& error_writer) = 0;
 
       virtual const Eigen::VectorXd dtau_dp(Point& z) = 0;
 
       // phi = 0.5 * log | Lambda (q) | + V(q)
-      virtual const Eigen::VectorXd dphi_dq(Point& z) = 0;
+      virtual const Eigen::VectorXd dphi_dq(
```
bob-carpenter: If the virtual overhead's too high here, we could probably go with a CRTP implementation for the Hamiltonians.  I only bring this up because this is a pretty inner-loop kind of code.

src/stan/mcmc/hmc/hamiltonians/base_hamiltonian.hpp
```diff
@@ -37,29 +37,50 @@ namespace stan {
...
 
-      virtual void
-      update(Point& z,
-             interface_callbacks::writer::base_writer& info_writer,
-             interface_callbacks::writer::base_writer& error_writer) {
+      void update_potential(
+        Point& z,
+        interface_callbacks::writer::base_writer& info_writer,
+        interface_callbacks::writer::base_writer& error_writer) {
+        try {
```
bob-carpenter: (fix now) Negation is faster (much much faster) and I think it's easier to read.

src/stan/mcmc/hmc/hamiltonians/base_hamiltonian.hpp
```diff
@@ -37,29 +37,50 @@ namespace stan {
...
+        } catch (const std::exception& e) {
+          this->write_error_msg_(e, error_writer);
+          z.V = std::numeric_limits<double>::infinity();
+        }
+      }
+
+      void update_potential_gradient(
+        Point& z,
+        interface_callbacks::writer::base_writer& info_writer,
+        interface_callbacks::writer::base_writer& error_writer) {
```
bob-carpenter: (question) Should the exception case use NaN instead of a randomly signed infinity?  Is it ever recoverable once it goes to infinity()?

src/stan/mcmc/hmc/hamiltonians/diag_e_metric.hpp
```diff
@@ -30,19 +30,27 @@ namespace stan {
...
 
-      const Eigen::VectorXd dtau_dq(diag_e_point& z) {
+      const Eigen::VectorXd dtau_dq(
+        diag_e_point& z,
+        interface_callbacks::writer::base_writer& info_writer,
+        interface_callbacks::writer::base_writer& error_writer) {
         return Eigen::VectorXd::Zero(this->model_.num_params_r());
       }
 
       const Eigen::VectorXd dtau_dp(diag_e_point& z) {
```
bob-carpenter: (questions) 

1.  argument diag_e_point& should be const as it's not modified (though you may have problems with `z.p`;  if we cared more about const correctness, you'd use a const method to return `z.p`.

2.  I don't think the const on the return matters because it's a copy.

3. If this gets used to assign, it might be best to pass that in so that there's not a redundant copy on the `cwiseProduct()` return.

src/stan/mcmc/hmc/hamiltonians/ps_point.hpp
```diff
@@ -71,9 +71,7 @@ namespace stan {
        * @param writer writer callback
        */
       virtual void
-      write_metric(stan::interface_callbacks::writer::base_writer& writer) {
-        writer("No free parameters for unit metric");
-      }
+      write_metric(stan::interface_callbacks::writer::base_writer& writer) {}
```
bob-carpenter: (comment) I like this change.

src/stan/mcmc/hmc/hamiltonians/base_hamiltonian.hpp
betanalpha: The use of the Hamiltonian is always templated out, so we should probably remove these visual calls altogether!

src/stan/mcmc/hmc/hamiltonians/softabs_metric.hpp
```diff
@@ -0,0 +1,213 @@
...
+    template <class Model, class BaseRNG>
+    class softabs_metric
+      : public base_hamiltonian<Model, softabs_point, BaseRNG> {
+    private:
+      typedef typename stan::math::index_type<Eigen::VectorXd>::type idx_t;
+    public:
+      explicit softabs_metric(const Model& model)
+        : base_hamiltonian<Model, softabs_point, BaseRNG>(model) {}
+
+      ~softabs_metric() {}
```
bob-carpenter: (fix now) The default destructor is a no-op, so you don't need this.

src/stan/mcmc/hmc/hamiltonians/softabs_metric.hpp
```diff
@@ -0,0 +1,213 @@
...
+            double delta =   z.eigen_deco.eigenvalues()(i)
+                           - z.eigen_deco.eigenvalues()(j);
+
+            if (std::fabs(delta) < 1e-10) {
+              double lambda = z.eigen_deco.eigenvalues()(i);
+              double alpha_lambda = z.alpha * lambda;
+
+              // Thresholds defined such that the approximation
+              // error is on the same order of double precision
+              if (std::fabs(alpha_lambda) < 1e-4) {
```
bob-carpenter: (fix now)  This 1e-4 threshold appears all over, so should be defined as a static const somewhere.

src/stan/mcmc/hmc/hamiltonians/softabs_metric.hpp
```diff
@@ -0,0 +1,213 @@
...
+                z.pseudo_j(i, j) =   (2.0 / 3.0) * alpha_lambda
+                                   * (1.0 -   (2.0 / 15.0)
+                                            * alpha_lambda * alpha_lambda);
+              } else if (std::fabs(alpha_lambda) > 18) {
```
bob-carpenter: (fix now) Give `18` a name as a const or at least document it.

src/stan/mcmc/hmc/hamiltonians/softabs_point.hpp
```diff
@@ -0,0 +1,51 @@
+#ifndef STAN_MCMC_HMC_HAMILTONIANS_SOFTABS_POINT_HPP
+#define STAN_MCMC_HMC_HAMILTONIANS_SOFTABS_POINT_HPP
+
+#include <stan/interface_callbacks/writer/base_writer.hpp>
+#include <stan/mcmc/hmc/hamiltonians/ps_point.hpp>
+
+namespace stan {
+  namespace mcmc {
+
+    // Point in a phase space with a base
```
bob-carpenter: (nice to have) doxygen class and function --- it's the one form of doc I fully support

src/stan/mcmc/hmc/hamiltonians/softabs_point.hpp
```diff
@@ -0,0 +1,51 @@
...
+        alpha(1.0),
+        hessian(Eigen::MatrixXd::Identity(n, n)),
+        eigen_deco(n),
+        log_det_metric(0),
+        softabs_lambda(Eigen::VectorXd::Zero(n)),
+        softabs_lambda_inv(Eigen::VectorXd::Zero(n)),
+        pseudo_j(Eigen::MatrixXd::Identity(n, n)) {}
+
+        // SoftAbs regularization parameter
+        double alpha;
```
bob-carpenter: (fix now)  

1. The indenting is wrong for these member variables.

2.  The ones that are const (if any) should be marked as such.  It's as much for the reader of the code as the compiler and future writer of more code.

src/stan/mcmc/hmc/integrators/impl_leapfrog.hpp
```diff
@@ -0,0 +1,113 @@
...
+        Eigen::VectorXd q_init = z.q + 0.5 * epsilon * hamiltonian.dtau_dp(z);
+        Eigen::VectorXd delta_q(z.q.size());
+
+        for (int n = 0; n < this->max_num_fixed_point_; ++n) {
+          delta_q = z.q;
+          z.q.noalias() = q_init + 0.5 * epsilon * hamiltonian.dtau_dp(z);
+          hamiltonian.update_metric(z, info_writer, error_writer);
+
+          delta_q -= z.q;
+          if (delta_q.cwiseAbs().maxCoeff() < this->fixed_point_threshold_)
```
bob-carpenter: (comment) This could be a bit more efficient as a loop that breaks when it finds a coeff greater than the threshold.

src/stan/services/arguments/arg_softabs_alpha.hpp
```diff
@@ -0,0 +1,29 @@
...
+
+#include <stan/services/arguments/singleton_argument.hpp>
+
+namespace stan {
+  namespace services {
+
+    class arg_softabs_alpha: public real_argument {
+    public:
+      arg_softabs_alpha(): real_argument() {
+        _name = "alpha";
```
bob-carpenter: (fix now)

1. no initial underscores

2.  these should be initializers, not assignments in the body

src/stan/services/sample/init_softabs.hpp
```diff
@@ -0,0 +1,33 @@
+#ifndef STAN_SERVICES_SAMPLE_INIT_SOFTABS_HPP
```
bob-carpenter: Does this mean we'll be able to run RHMC in CmdStan?

src/test/unit/mcmc/hmc/base_hmc_test.cpp
```diff
@@ -124,7 +124,7 @@ TEST(McmcBaseHMC, streams) {
   stan::interface_callbacks::writer::stream_writer writer(output);
 
   EXPECT_NO_THROW(sampler.write_sampler_state(writer));
-  EXPECT_EQ("Step size = 0.1\nNo free parameters for unit metric\n",
+  EXPECT_EQ("Step size = 0.1\n",
```
bob-carpenter: (fix now) We don't want to hard code line returns.  If we use `std::endl` then the newline will be platform specific.  So I'd just test without the newline `\n` symbol.  

(future) Avoid newlines in exception messages.

src/stan/services/sample/init_softabs.hpp
betanalpha: We can — but it hasn’t been exposed to CmdStan yet.
I don’t want to further slow down compiles.

src/test/unit/mcmc/hmc/hamiltonians/dense_e_metric_test.cpp
```diff
@@ -29,114 +29,110 @@ TEST(McmcDenseEMetric, sample_p) {
...
-    
   }
-  
+
   Eigen::VectorXd g2 = metric.dtau_dp(z);
-  
+
   for (int i = 0; i < z.q.size(); ++i) {
-    
+
```
bob-carpenter: (just a test, so no need to fix)  No newlines at beginnings of blocks --- cpplint will flag it.

src/test/unit/mcmc/hmc/hamiltonians/diag_e_metric_test.cpp
```diff
@@ -12,48 +12,48 @@
...
+
   for (int i = 0; i < n_samples; ++i) {
     metric.sample_p(z, base_rng);
-    double T = metric.T(z);
-    
-    double delta = T - m;
+    double tau = metric.tau(z);
+
+    double delta = tau - m;
     m += delta / static_cast<double>(i + 1);
```
bob-carpenter: (fix whenever + Future) You don't need to static cast one of two arguments --- it's built into all the arithmetic operators.  So this can just be `delta / (i + 1)`.

bob-carpenter:
This all looks great. 

I didn't check that the tests covered all the cases, or really even look at the test content other than for form.

My one comment overall is that when you have `a OP b` and one of the operands is `double` and the other is `int`, then the `int` gets cast up to double automatically.  You should almost never need `static_cast<double>(...)`

src/test/unit/mcmc/hmc/base_hmc_test.cpp
betanalpha: ```std::endl``` is not a platform-independent newline, it just adds ```\n``` and then flushes the iostream buffer.

src/stan/mcmc/hmc/hamiltonians/base_hamiltonian.hpp
betanalpha: I thought ```std::numeric_limits<double>::infinity()``` was defined to be _larger_ than all doubles, hence representing positive infinity.  By using positive infinity the samplers will interpret these failures as divergences and each HMC transition will fail gracefully.

betanalpha:
Many of the comments have been addressed.  The only things that I didn't change were

* I will get violent about having arguments aligned, even at the expensive of an extra vertical line.  This will all go away once we have a decent logger, anyways.

* I left the input ```ps_points``` as non-const as there are lots of tricky interactions between the functions that in some implementations, especially ```softabs_metric```, require modifying the members of the ```ps_points```.

* I did not change ```arg_softabs``` as _all_ of the arguments have the same problem and should all be addressed at once.

* You can take my ```static_cast<double>```s from my cold, dead hands.  😁 I can never remember which languages promote in which ways, so I tend to err on the side of being too conservative.  I'll keep this in mind if they pop up in the source code.

In the future we should think about the ```virtual``` modifiers in ```base_hamiltonian.hpp```.  I had them there to define pure virtual methods to enforce an interface, but it might be fruitful to remove them to sacrifice enforcement for performance.

Issue #1907, opened 2016-06-09
stan_csv_reader is sensitive (crashes) to whitespace in rhs of metadata.


sakrejda:
#### Summary:
Stansummary (and I assume other binaries that use stan_read_csv) throw an exception
when there is no whitespace after the "=" in metadata.  

#### Description:
Taking a .csv file from rstan or pystan and trying to run it through stansummary throws and error, if the file is edited to add whitespace after "=" in metadata the file is read fine.  

#### Reproducible Steps:
1. Run any stan model in Pystan or rstan and write out a .csv file. In R, with the standard-normal.stan file:
~~~
parameters {
  real x;
}

model {
  x ~ normal(0,1);
}
~~~

~~~
library(rstan)
m1 <- stan('standard-normal.stan', sample_file='/tmp/test.csv', chains=1);
~~~
2. Crash stansummary: In the terminal run, from the root of cmdstan repo:
3. Edit the file to add a space after each "=" in metadata.
4. Re-run stansummary and it reads the file fine.

#### Current Output:

~~~
SPH-KSakrejdaL:cmdstan krzysztof$ bin/stansummary test.csv
libc++abi.dylib: terminating with uncaught exception of type boost::exception_detail::clone_impl<boost::exception_detail::error_info_injector<boost::bad_lexical_cast> >: bad lexical cast: source type value could not be interpreted as target
Abort trap: 6
~~~

#### Expected Output:
~~~
krzysztof@hagrid:~/packages/cmdstan$ bin/stansummary /tmp/test.csv
Warning: non-fatal error reading adapation data
Inference for Stan model: 
1 chains: each with iter=(2000); warmup=(0); thin=(1); 2000 iterations saved.

Warmup took (0.013) seconds, 0.013 seconds total
Sampling took (0.011) seconds, 0.011 seconds total

                    Mean     MCSE  StdDev    5%    50%       95%  N_Eff  N_Eff/s  R_hat
lp__            -5.0e-01  2.2e-02    0.70  -2.0  -0.24  -1.1e-03    999  9.0e+04    1.0
accept_stat__    8.5e-01  5.6e-03    0.25  0.14   0.96   1.0e+00   2000  1.8e+05    1.1
stepsize__       1.3e+00  3.2e-02     1.2  0.60    1.2   2.2e+00   1404  1.3e+05    1.0
treedepth__      1.7e+00  1.6e-02    0.62   1.0    2.0   3.0e+00   1494  1.3e+05   1.00
[...]

PR #1908, opened 2016-06-09
Make stan_csv_reader insensitive to whitespace around rhs of .csv fil…

sakrejda:
#### Submisison Checklist

#### Summary
There is a two-line change to stan_csv_reader that trims whitespace before using rhs.

#### Intended Effect
Read .csv files produced by rstan/pystan which do no follow the cmdstan convention of including whitespace around the "=" in metadata.

#### How to Verify
There is a new test file and test, it fails with the old stan_csv_reader and passes with the new one.  It has three lines modified with different whitespace patterns. It probably is possible to make the test smaller but the test should pass regardless of any stray whitespace so I'd rather keep the large file.

#### Side Effects
None.

#### Copyright and Licensing

Krzysztof Sakrejda

…e metadata.

Previously running stansummary on a .csv file produced by either rstan or pystan
would fail on a metadata line with no space after the "=" sign.  Editing the
files to say "= " would fix the problem.  With this change any amount of spacing
before or after the value works as it gets trimmed before the lexical_cast.

### DIFF
--- a/src/stan/io/stan_csv_reader.hpp
+++ b/src/stan/io/stan_csv_reader.hpp
@@ -101,7 +101,8 @@ namespace stan {
           if (equal != std::string::npos) {
             name = lhs.substr(0, equal);
             boost::trim(name);
-            value = lhs.substr(equal + 2, lhs.size());
+            value = lhs.substr(equal + 1, lhs.size());
+            boost::trim(value);
             boost::replace_first(value, " (Default)", "");
           } else {
             if (lhs.compare(" data") == 0) {
--- a/src/test/unit/io/stan_csv_reader_test.cpp
+++ b/src/test/unit/io/stan_csv_reader_test.cpp
@@ -10,6 +10,7 @@ public:
   void SetUp () {
     blocker0_stream.open("src/test/unit/io/test_csv_files/blocker.0.csv");
     metadata1_stream.open("src/test/unit/io/test_csv_files/metadata1.csv");
+    metadata3_stream.open("src/test/unit/io/test_csv_files/metadata3.csv");
     header1_stream.open("src/test/unit/io/test_csv_files/header1.csv");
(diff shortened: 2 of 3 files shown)


sakrejda:
I updated the branch moments after clicking "create pull request" if the CI checks fail maybe that's why?

bob-carpenter:
You need to make sure this works with the CmdStan output, too. The files changed make it look like you changed the behavior rather than supporting both. Don't we want to just skip whitespace before and after `=`?

sakrejda:
This version passes the original cmdstan output test too
(test_csv_files/metadata1.csv)

sakrejda:
@bob-carpenter To be specific about the tests: [metadata1](https://github.com/stan-dev/stan/blob/develop/src/test/unit/io/test_csv_files/metadata1.csv) is used to test that standard cmdstan output is correctly interpreted and [metadata3](https://github.com/stan-dev/stan/blob/bugfix/csv_reader_insensitive_whitespace/src/test/unit/io/test_csv_files/metadata3.csv) is used to test that a file with extra whitespace or missing whitespace is also correctly read.  You might have looked at it before I added the test/test file.

syclik:
I think this is good enough for the time being. I wrote the first version and really poorly. I get what's going on. (I'd prefer if it were just rewritten, but not necessary for this pull request.)

This fixes #1907. 

@sakrejda, mind following the naming convention for future pull requests? `bugfix/issue-#-description`? It helps when closing these pull requests so the issues are closed with the pull request.

sakrejda:
@syclik Will do about pull request naming.  I agree a re-write is best and I have more extensive changes on another branch that should be more robust.  I just didn't want to try to move a lot at once without understanding the current thing.  More long-term I'm working on finishing the spec for output and doing something simpler but that will really require understanding output and coordinating more with interface maintainers so I'm not ready for it.

PR #1916, opened 2016-06-13
Issue 1125 conditional operator

mitzimorris:
#### Submisison Checklist

#### Summary
Add C++ style ternary operator to the Stan language:   
`condition ?  true_val : false_val;`

#### Intended Effect

Provides an alternative to if/then/else construct that can be used in assignment statements
and which only requires evaluation of one branch of the conditional.

#### How to Verify

The test model src/test/test-models/good/validate_conditional_op_good.stan
is used for unit tests for the expression grammar.
To test the generator, run:
`make -j2 src/test/test-models/good/validate_conditional_op_good.hpp-test`

#### Documentation

Stan manual documentation written by Bob

#### Copyright and Licensing


runTests.py
```diff
@@ -14,7 +14,7 @@
 
 winsfx = ".exe"
 testsfx = "_test.cpp"
-debug = False
+debug = True
```
syclik: @mitzimorris, mind changing this back?

src/docs/stan-reference/distributions.tex
```diff
@@ -1911,13 +1911,12 @@ \subsubsection{Probability Density Function}
 \distro{Wiener}(y|\alpha, \tau, \beta, \delta)
 =
 \frac{\alpha}{(y-\tau)^3/2}
-\,
+%
```
syclik: what's with the changes to the distributions? I haven't seen it built yet, but I wanted to make sure these were intentional.

src/docs/stan-reference/functions.tex
```diff
@@ -494,7 +494,13 @@ \subsection{Logical Functions}
       x & \mbox{if } cond \neq 0 \\
       y & \mbox{otherwise}
     \end{cases}
-    \]}
+    \]
+    {\bfseries Warning:} This function should not be used.  Rather,
```
syclik: I think we should deprecate `if_else` with this pull request. I don't see why we should warn users in the doc without warning them at compile time.

src/docs/stan-reference/language.tex
```diff
@@ -1807,6 +1809,123 @@ \subsection{Operator Precedence and Associativity}
 is interpreted as \mbox{\code{-(n \textasciicircum\ 3)}}.
 
 
+\section{Conditional Operator}\label{conditional-operator.section}
+
+\subsection{Conditional Operator Syntax}
+
+The ternary conditional operator is unique in that it takes three
+arguments and uses a mixed syntax.  If \code{a} is an expression of
+type \code{int} and \code{b} and \code{c} are expressions that can be
```
syclik: I don't think this is true. I think `\code{a}` can be of type `real`, but evaluate to non-zero and that'll be treated as `true`, right? (that's what you say below)

src/stan/lang/generator.hpp
```diff
@@ -346,6 +357,57 @@ namespace stan {
         }
         o_ << ')';
       }
+
+
+
+      //    struct conditional_op has members:
```
syclik: remove this block of comments.

syclik:
@mitzimorris, the code looks good. There were just a couple of things I noticed that should be changed before it gets in.

src/docs/stan-reference/distributions.tex
bob-carpenter: No change---just reformatting so that there weren't overfull hboxes.  Also a bunch of other changes to add spaces to the code when I took a pass replacing if_else.

runTests.py
mitzimorris: reverted to False.

src/docs/stan-reference/language.tex
mitzimorris: Daniel is correct - the code as written allows \code{a} (the condition) to be either int or real.
do we want to restrict it to int?

src/stan/lang/generator.hpp
mitzimorris: removed.

src/docs/stan-reference/language.tex
bob-carpenter: It'll match the rest of our conditionals and C++ better to
allow real there.  I don't mind either way.

- Bob

syclik:
Thanks. We're good.

PR #2004, opened 2016-08-09
Stan refactor

syclik:
#### Submisison Checklist
#### Summary

This pull request accomplishes two major things:
1. In the `src/stan/services` folder (the `stan::services` namespace), we now have functions that should be called by the interfaces.
2. Removes a lot of the old code.

It does this while maintaining all current behavior.
#### Intended Effect

The intended effect is to make the downstream interfaces easier to implement and maintain.

The code is much cleaner. Each of the functions in `src/stan/services` excluding the `src/stan/services/util` folder are meant to be called by RStan, PyStan, and CmdStan. This will also make it much easier to deal with the upcoming template build issue we will have with the introduction of forward-mode and RHMC.
#### How to Verify

(I'm sorry -- the diff is large.)
1. All tests should pass. (The upstream tests will fail due to CmdStan `develop` branch not having changes necessary, but I'll submit a pull request on the CmdStan branch.)
2. Take a look at all the new functions in `src/stan/services`. I've covered all the use cases of CmdStan, PyStan, and RStan. See side effects below.
3. There were a couple of additional utilities outside of `src/stan/services` that were added. Take a look.

If there are some oddly implemented things in there, most weird implementations are due to keeping the existing behavior intact. This way, we can roll these changes out with a minor release and keep moving forward.
#### Side Effects

CmdStan, RStan, and PyStan will break under the current branch, but will maintain the same behavior.

I have a branch of CmdStan that works. Allen has a branch of PyStan that's almost working. I have a branch of RStan that's almost working.
#### Documentation

The C++ code may need some more doc.

If anyone wants to help, there's that and adding more tests.
#### Reviewer Suggestions

This one's big, so it'd be helpful to have a few eyes on it.

@ariddell, @bgoodri: since this will effect both of you directly.
@sakrejda: since you've been looking at this for a while.
@bob-carpenter, @betanalpha, @rtrangucci: if you've got the time.

Although I'm confident that this behaves identically, this one shouldn't be taken lightly.
#### Copyright and Licensing


bob-carpenter:
Very exciting.  I'm happy to help out.  Should we
wait until we have working RStan and PyStan versions
before reviewing?  I don't want to commit something to
Stan's develop branch that might break the interfaces beyond
repair.

- Bob

syclik:
Thanks for offering!

No, I don't think we need to wait, but I don't feel comfortable merging until we have RStan and PyStan versions ready to go.

I've implemented branches almost all the way through... enough to get samples out and into the Rcpp. I hadn't figured out how to configure the writers correctly, but that should be do-able.

syclik:
One thing this doesn't is a good way to get default arguments.

Two reasonable options (and I'm sure there are more) include:
1. Having the default arguments in C++ as part of the function signature. This isn't enforceable, but it can be treated like doc.
2. Exposing functions for default arguments.

src/stan/callbacks/writer/chained_writer.hpp
```diff
@@ -0,0 +1,88 @@
...
+#include <ostream>
+#include <vector>
+#include <string>
+
+namespace stan {
+  namespace callbacks {
+    namespace writer {
+
+      /**
+       * stream_writer writes to an std::ostream.
```
sakrejda: should get chained_writer,

```
/**
 * chained_writer writes output to a pair of writers.
 */
```

src/stan/io/random_var_context.hpp
```diff
@@ -0,0 +1,221 @@
...
+            unconstrained_params_[n] = 0.0;
+        } else {
+          boost::random::uniform_real_distribution<double>
+            unif(-init_radius, init_radius);
+          for (size_t n = 0; n < num_unconstrained_; n++)
+            unconstrained_params_[n] = unif(rng);
+        }
+
+        std::vector<double> constrained_params;
+        std::vector<int> int_params;
```
sakrejda: I guess this is "params" in some non-Stan sense since we don't have integer parameters? Just confusing every time I see it.

src/stan/io/random_var_context.hpp
```diff
@@ -0,0 +1,221 @@
...
+        model.write_array(rng,
```
sakrejda: Maybe this is general confusion about the design on my part rather than a real problem but why does constructing a var_context modify the model? Maybe the issue is that var_context is poorly documented at the top level. I thought var_context was supposed to hold data and allow access to it.

src/stan/io/random_var_context.hpp
syclik: When we first started Stan, we had placeholders for both real and integer parameters. It hasn't been fully removed and I actually don't mind having them around for the generality. Maybe one day we can figure out how to sample integer parameters generally.

src/stan/io/random_var_context.hpp
syclik: Where is the model being modified? It shouldn't. The intent of this class is to generate a random initialization given a model, so it needs to know the model. It need to know types of parameters the model is looking for and the sizes.

syclik:
i didn't change all the tests; I'll work on that today.

src/stan/io/random_var_context.hpp
sakrejda: Sorry, I thought the line comments were more specific.  Line 87, what does that model.write_array do?

src/stan/io/random_var_context.hpp
sakrejda: Thanks, that makes a lot more sense now!

src/stan/io/random_var_context.hpp
syclik: (the inline code comments are attached to lines.)
`model.write_array()` is poorly named. As input, it takes a random number generator, the unconstrained parameters, unconstrained integer parameters. For output, it fills in the constrained parameters by applying the transforms to the unconstrained parameters. There are a couple of additional boolean arguments, one of which is an indicator for generating things in the generated quantities block -- if so, it uses the random number generator provided.
This method doesn't change the model. It's essentially going from unconstrained to constrained parameters.

sakrejda: That makes so much more sense!

syclik:
Just FYI, the Jenkins tests pass; the failed test is from the upstream CmdStan branch trying to build with `develop`.

syclik:
There are a couple more (minor) things that I need to do on this branch:
- change `stan::callbacks::base_writer` to `stan::callbacks::writer`
- remove `src/stan/callbacks/var_context_factory`. If any of the things need to move to RStan or CmdStan, then we should move them. (Or if they need to be in both, then we should leave it, but I don't think it's necessary anymore.)

sakrejda:
@syclik I agree that getting the services well-defined in one directory makes this much easier to turn into interfaces, really nice to see this get worked out!  I looked through (all) the diffs and I didn't really find anything problematic.  I did see an opportunity to add doc in a bunch of places and I will have some time later this week to work on tests.  Can you suggest start and end points for the services tests?  Would it make sense to generate code for a specific simple model and test from the services entry points all the way to log-density and gradient calculations or should the end point be earlier?

src/stan/io/random_var_context.hpp
bob-carpenter: The model code still has a vector of integer parameters
awaiting some implementation.  I should've never coded that
in the first place ahead of any plan to implement it.
But they are params in the usual Stan sense (as opposed to
arguments to probabilty functions, which are always called
"parameters" even if they're data).

src/stan/io/random_var_context.hpp
bob-carpenter: A var_context is a *reader* for data, it doesn't
store data.  For example, in R and Python, it can
sit on top of their memory and read variables out of
their environment and not hold anything itself.  The
random var_context also doesn't need to store any data.
The var_context doesn't modify a model directly, but
the constructor requires a var_context for the data.

bob-carpenter:
I think we should have 

* functions for the defaults
* functions to validate

I don't see how to do anything useful in C++ argument structure
because the defaults have to be some final subsequence, not an
arbitrary subset.

src/stan/io/random_var_context.hpp
bob-carpenter: It would be so awesome to include more const correctness
throughout.  We've never succeeded in that refactor before,
but we're better at C++ now.

syclik:
Any suggestions on how to deal with functions to defaults and validation?

One way I could see it happening is to move all the functions in `src/stan/services/` to static member functions of a struct with the defaults for the arguments also being member functions within that struct. That seems less than ideal too, but at least we'd have the functions for default be tied to the function it belongs with.

For validation, do you think we should have validation outside of the function itself? Or should we just pass inputs into the function and have that validate inputs? (My fault... I should have taken care of that with this implementation.)

bob-carpenter:
As a specific proposal, how about:

```
struct stepsize {
  static std::string description();
  static void validate(double x);
  static double default();
};
```

Maybe a short and long description if we need both.

The class is just a wrapper to keep things together, so this
could also be broken into functions `stepsize_description()`, etc.

General rule on validation is to always validate arguments to
functions called externally and never validate arguments from
functions called internally (caller's responsibility).

The only way to not have the caller have to specify everything
is with a builder pattern that involves an object per function
(this is a problem with having so many functions), and gets called
as

  arg_builder().set_stepsize(0.01).set_int_time(1.0)....build();

Typically that build() call will do all the validation and return
a structure with defaults filled in.  I think this is overkill
with the functions expanded out as you have them, but would've
been critical with functions with complex interactions among
arguments.

syclik:
I'm perfectly happy using `struct`. And for validating outside the calls.

One use case, which is possibly hypothetical, is the overloading of variable names and their meanings. We would have difficulty distinguishing the different uses and defaults if we had global defaults.

I just looked and that's currently a hypothetical use case. I think instead of global defaults, we can have the defaults live inside each subfolder / namespace. We can enforce that argument names are overloaded within one of those subfolders / namespaces. E.g. within sampling, there will only be one meaning for alpha.

bob-carpenter:
I can see that we might have stepsize in both optimization
and sampling (or even in EHMC and RHMC) with different defaults.

I absolutely do not want to use includes to sort this out.
We can use namespaces or clutter up the names for defaults
with suffixes.

I was thinking struct per argument, but maybe it's better with
a struct per "route".  But there's so damn many of them that
it's offputting.  We could organize with inheritance, thereby
rebuilding the stuff Michael had in CmdStan into a C++
class hierarchy.  At some point, we'll just be back to having
the builder pattern.

- Bob

syclik:
Yup. No need for include magic. I'm suggesting namespaces. I'll put together some code on the branch; struct per argument within a namespace.

sakrejda:
+1 to struct per route.  When I was trying to organize the arguments in the wiki it became clear just how different the meanings of the names are dependent on route (especially when people start calling things "alpha" rather than descriptive names like "step_size").  Maybe we should require descriptive names at this level.

bob-carpenter:
I still don't understand.  Are you suggesting a namespace per route
then overloading all the structures?  Because we'll only guarantee
that a name has a constant meaning on a per-route level.

Could you sketch one out before coding dozens of them so I can
understand what you're proposing?

- Bob

P.S.  Can anyone think of a better term than "route"?

bob-carpenter:
Yes, adapt_delta absolutely has to be replaced with something
more meaningful.  I think that means supporting the old name
*and* a new name until Stan 3 arrives.

- Bob

bob-carpenter:
Or just waiting for Stan 3 to make these changes.  That's
probably the better plan.

src/stan/services/optimize/lbfgs.hpp
```diff
@@ -0,0 +1,204 @@
...
+                double init_alpha,
+                double tol_obj,
+                double tol_rel_obj,
+                double tol_grad,
+                double tol_rel_grad,
+                double tol_param,
+                int num_iterations,
+                bool save_iterations,
+                int refresh,
+                Interrupt& interrupt,
```
sakrejda: @syclik The type of the interrupt callback here is different from the interrupt callback type for sample/fixed_param_sampler.hpp (or similar file name).  I started writing tests for the sample/* services functions but I can't tell if these are supposed to be the same or not.

syclik: They should be the same! Thanks! I'll fix that.

sakrejda: BTW: if I write tests, do you want me to do a pull request to this branch or just push them here?

syclik: Just push to the branch.

sakrejda:
@syclik I tested out pushing to the branch with just one of the files I added but since it was originally cloned from my fork of this branch a second commit appeared that I didn't see.  Sorry if I messed something up... :/  It just looks like a merge of the changes you already made but I'm not sure why it's showing up at all....

syclik:
That's fine. No problem. Doesn't look messed up. We tend not to worry at the commit level. (Even though we've been advised that we should.) just way too much effort.

sakrejda:
ok, think I've got it figured out now :+1:

src/stan/services/optimize/lbfgs.hpp
bob-carpenter: Is Interrupt a typedef or a template parameter?  

If the interrupts are only to interrupt, then I don't
see why they couldn't have the same signature for all
routes and be regular arguments rather than template parameters.

- Bob

syclik: Yup. They're not a template parameter any more. They all have the same
signature.

sakrejda:
unit test failed from 1 new warning, I'll fix that with the next tests.

sakrejda:
@syclik As I'm writing tests.... I think "init" is a weird argument name for the context on the C++ side.  Could just call it "context" or spell out "initial_values" or at least to match things like "num_warmup" call it "init_values".... except that it's not initial values, it's actually the data, so maybe just "data"...  pretty much anything other than "init", what do you think?

syclik:
I'm fine with renaming it, but the type is a var_context and it's for initialization. We could call it init_context, but I'd prefer init since the context part is obvious from the type. 

It's definitely not "the data." That's already part of the instantiated model. 

Any other suggestions?

sakrejda:
Never mind then, init is a fine name but the model in the tests is just instantiated as `stan_model model;`... so I assumed "init" was for data.  Is there  a test somewhere that instantiates the model with data so I can figure out what the types are here?

ariddell:
The sense that some of the parameter names might be improved is one that
I share. I'll trust @syclik's judgment on whether or not such a feeling
is warranted.

syclik:
I accept it with open arms! 

But I think the init var_context was misinterpreted as being data when it is indeed initialization. That said, I support it.

sakrejda:
What I really need to do is make a flowchart for how a Stan model is put together! I think I get all the pieces now.

src/stan/services/optimize/lbfgs.hpp
bob-carpenter: If the Interrupt template parameters are all instantiated
the same way, can we make them a regular type?  Then if
the model can be made into a regular type (I don't see why
it couldn't be if the base class is enriched), then we could
get rid of the templating altogether.

bob-carpenter:
The main motivation to work at the commit level is to support
git bisect.  Given that we're much more careful on testing now,
it hasn't been such a big deal not to have every commit working.
But we like to have commits at least build, even if they don't pass
all the unit tests.  I've always been afraid to rebase because
I tend to push my commits to the origin.

bob-carpenter:
Allen --- do you have suggestions for new names?  If so, now's
the time to bring them up!

bob-carpenter:
And what the project really needs is doc.  The best we have
is this:

https://github.com/stan-dev/stan/wiki/Model-Concept

I think we want to move to where there's a base class and we
pass models by ordinary type rather than template parameter.
That makes everything easier to understand because then there's
an object-level spec in C++ as opposed to meta or comment level doc like
on a C++ "concept".

syclik:
I've just pushed changes that take away the template parameter for the interrupt callback. I've also updated the finite diff code in diagnose mode so it takes an interrupt. (This will help with RStan; right now, there's no way to stop RStan when running finite diffs.)

Left on my plate is a prototype for defaults.

syclik:
@sakrejda, if you want to see how this is all done in CmdStan, see [feature/issue-287-stan-services](https://github.com/stan-dev/cmdstan/tree/feature/issue-287-stan-services). In particular, [src/cmdstan/command.hpp](https://github.com/stan-dev/cmdstan/blob/feature/issue-287-stan-services/src/cmdstan/command.hpp). Lines 53 -- 118 set up most of the things needed to run any command.

ariddell:
@bob-carpenter I'll give the names some thought  I'll also double check for places where doc changes/additions would be helpful.. I'm more comfortable with the doc and var names for the writers than var_context but that could just be me.

sakrejda:
@syclik Thanks---I had gotten confused with the fact that you need a var_context with data first for the model, and then another var_context to initialize sampling but now it's really clear.

@bob-carpenter now that I understand I think the "model concept" wiki is pretty good. The problem is that you totally loose perspective on what was hard once you get it.

syclik:
@bob-carpenter: how about this for defaults?
https://github.com/stan-dev/stan/pull/2004/commits/f614952e2e0e615d8048fe80eda0d873edccdb3f

We can have one `defaults.hpp` per folder / namespace. I couldn't use the function name `default()` cause it's a keyword, so I made it more verbose. I'm happy to change it.

bob-carpenter:
Exactly what I was imagining.

But don't you want to use our validation functions
rather than throwing exceptions directly?

- Bob

syclik:
I do, but they aren't templated for exception type. I'm adding is_* checks on a branch. Once those are in, I'll update these. 

I'll put together the rest of these. One per namespace.

bob-carpenter:
Oh, you want "std::invalid_argument"?  I suppose that makes
sense.

I don't think you need different namespaces for all these,
do you?  I'd go lighter on this and just put everything in
services.  Eigen puts almost everything in Eigen:: and Boost
is very light on namespaces.  Would there be conflicts otherwise?

- Bob

sakrejda:
@syclik How about this for classes for testing writer/interrupt use (737520b27b9102ffd7a20317833853d8dc03b257). Classes are in src/test/unit/instrumented_callbacks.hpp and there's a usage example in src/test/unit/services/util/generate_transitions_test.cpp

sakrejda:
there was a still one more missing include, fixed it and util.hpp now passes locally without warnings.  Hopefully that was it.

syclik:
I added all the defaults with tests. They're in:
- `src/stan/services/diagnose/defaults.hpp`
- `src/stan/services/experimental/advi/defaults.hpp`
- `src/stan/services/optimize/defaults.hpp`
- `src/stan/services/sample/defaults.hpp`

@sakrejda, I'll need to look at your tests. I think once the tests are there, Stan will be code complete and we'll only need to worry about the interfaces.

syclik:
These are the files that still don't have tests. We'll need to knock them off one by one. @sakrejda: do you have any tests written for any of this? If so, could you push soon?

- [ ] src/stan/callbacks/chained_writer.hpp
- [ ] src/stan/callbacks/interrupt.hpp
- [ ] src/stan/callbacks/noop_interrupt.hpp
- [ ] src/stan/services/experimental/advi/fullrank.hpp
- [ ] src/stan/services/experimental/advi/meanfield.hpp
- [ ] src/stan/services/sample/hmc_nuts_dense_e.hpp
- [ ] src/stan/services/sample/hmc_nuts_dense_e_adapt.hpp
- [ ] src/stan/services/sample/hmc_nuts_diag_e.hpp
- [ ] src/stan/services/sample/hmc_nuts_diag_e_adapt.hpp
- [ ] src/stan/services/sample/hmc_nuts_unit_e.hpp
- [ ] src/stan/services/sample/hmc_nuts_unit_e_adapt.hpp
- [ ] src/stan/services/sample/hmc_static_dense_e.hpp
- [ ] src/stan/services/sample/hmc_static_dense_e_adapt.hpp
- [ ] src/stan/services/sample/hmc_static_diag_e.hpp
- [ ] src/stan/services/sample/hmc_static_diag_e_adapt.hpp
- [ ] src/stan/services/sample/hmc_static_unit_e.hpp
- [ ] src/stan/services/sample/hmc_static_unit_e_adapt.hpp
- [ ] src/stan/services/util/experimental_message.hpp
- [ ] src/stan/services/util/generate_transitions.hpp
- [ ] src/stan/services/util/initialize.hpp
- [ ] src/stan/services/util/mcmc_writer.hpp
- [ ] src/stan/services/util/rng.hpp
- [ ] src/stan/services/util/run_adaptive_sampler.hpp
- [ ] src/stan/services/util/run_sampler.hpp

sakrejda:
@syclik I wrote classes I wanted to use for testing but they seem like the sort of complicated thing you guys sometimes dislike so I asked for your opinion (above) before I wrote a bunch of test code.  I also got busy with other stuff so it was kind of convenient that you didn't have time! Could you take a look at the classes in src/test/unit/instrumented_callbacks.hpp and the usage example in src/test/unit/services/util/generate_transitions_test.cpp?  I know they're not pretty but I think using these will make the test code shorter and easier to write.

syclik:
Ok. Will do today.

sakrejda:
Thanks, I'd rather get your opinion than push something you'll see as a nuissance in the future.

syclik:
@sakrejda, looks fine to me. I would move it down to the test/unit/services folder since that's the only place it'll be used for now. We can leave it in the stan::test namespace. We don't need to be super-careful about the test namespace.

It definitely beats reimplementing that time and time again.

I don't see tests as a nuisance. I don't see most things as a nuisance if it has function now. I'll always annoyed by my past self, though. Never enough tests.

sakrejda:
Ok, thanks! I have a couple of tests I need to update to use the class, I'll fix them up and push.

syclik:
Thanks. Wherever you leave off, I can pick up.

src/stan/services/util/experimental_message.hpp
```diff
@@ -0,0 +1,27 @@
+#ifndef STAN_SERVICES_UTIL_EXPERIMENTAL_MESSAGE_HPP
+#define STAN_SERVICES_UTIL_EXPERIMENTAL_MESSAGE_HPP
+
+#include <stan/callbacks/writer.hpp>
+
+namespace stan {
+  namespace services {
+    namespace util {
+
+      void experimental_message(stan::callbacks::writer&
```
sakrejda: @syclik I'm seeing this function and thinking that it would be easier to create a namespace, say stan::message, with only static strings so that the calling function could just do writer(stan::message::experimental_procedure) rather than experimental_message(writer).  I think it would be clearer, less code, fewer calls to the writer, less formatting implied by calls to writer (such as when we call the message_writer() here to add newlines (I think that's what that's for).  It would also make it easier to check whether the string came through correctly on tests.  What do you think? I could create the strings as I write tests---there are a lot of places (e.g.in stan::services::util::initialize) where these multi-line strings are used.

sakrejda:
@syclik That's all the time I had today, I'll pick up again another day and check things off the list when I do.  hmc_nuts_dense_e_adapt.hpp has at least some tests implemented (counting that things get called, that the right columns show up in output, etc..) but I don't know when it's supposed to throw or other more detailed behavior.  I don't have snapshot output to do the exact regression tests against either so there's only a placeholder for that.

src/stan/services/util/experimental_message.hpp
syclik: I don't think so. Newlines are handled by the interfaces themselves.

syclik:
@sakrejda, I added the remaining tests! If you want to add any more doc, please do. Or we can at least mark what needs doc. For now, I think this is pretty code complete and needs a thorough review.

CmdStan is working with this branch.

I'm working on RStan and PyStan.

sakrejda:
@syclik Congrats on getting it done! I'll take another look at code/doc this weekend.

syclik:
I've marked this for v2.13.1++. This needs to get in sooner rather than later.

@bob-carpenter, mind reviewing?

@bgoodri, @jgabry: can you verify that the branch I put up for RStan works?
@ariddell: we'll talk in person next week, but let's make sure it works with PyStan too.

I'm confirming this works with CmdStan. I'll run tests with the right upstream branch and the tests will all be good.

bob-carpenter:
I'll review.  I just need to find an old-fashioned turntable with a record changer to queue up enough albums.  And it looks like they made their diffs more robust by letting me pull them up file by file.  Yay!

syclik:
Jenkins, retest this please.

syclik:
Jenkins, retest this please.

syclik:
Bump. @bob-carpenter, @thelseraphim, and anyone else: can you start reviewing this code?

We have branches working for RStan, PyStan, and CmdStan. Maybe not 100% yet, but definitely showing that it works.

src/stan/io/random_var_context.hpp
```diff
@@ -0,0 +1,221 @@
...
+       * @param[in] init_radius the unconstrained variables are uniform draws
+       *   from -init_radius to init_radius. If this is 0, the unconstrained
+       *   values are set to 0. If the init_radius is non-zero, it is expected
+       *   to be greater than 0.
+       */
+      template <class Model, class RNG>
+      random_var_context(Model& model,
+                         RNG& rng,
+                         double init_radius,
+                         bool init_zero)
```
sakrejda: doc doesn't say what init_zero is

sakrejda: looking below, isn't this redundant with setting init_radius = 0.0?

src/stan/model/finite_diff_grad.hpp
```diff
@@ -35,6 +38,7 @@ namespace stan {
       std::vector<double> perturbed(params_r);
       grad.resize(params_r.size());
       for (size_t k = 0; k < params_r.size(); k++) {
+        interrupt();
```
sakrejda: How often is the interrupt supposed to be called? This is once per log density evaluation but I don't think that's how ti works with HMC (?)

src/stan/services/optimize/bfgs.hpp
```diff
@@ -0,0 +1,200 @@
...
+       * @param chain chain id to advance the pseudo random number generator
+       * @param init_radius radius to initialize
+       * @param[in] init_alpha
+       * @param[in] tol_obj
+       * @param[in] tol_rel_obj
+       * @param[in] tol_grad
+       * @param[in] tol_rel_grad
+       * @param[in] tol_param
+       * @param[in] num_iterations maximum number of iterations
+       * @param[in] save_iterations indicates whether all the interations should
```
sakrejda: Should say something like "if true, values at each iteration are saved in addition to the ifnal result, otherwise only the final result is saved." Currently unclear what happens when false (does it save anything?)

src/stan/services/optimize/defaults.hpp
```diff
@@ -0,0 +1,326 @@
...
+       * Amount of history to keep for L-BFGS.
+       */
+      struct history_size {
+        /**
+         * Description of history_size.
+         *
+         * @returns description
+         */
+        static std::string description() {
+          return "Amount of history to keep for L-BFGS.";
```
sakrejda: This is related to memory use so it should have units instead of just "amount"?

src/stan/services/sample/defaults.hpp
```diff
@@ -0,0 +1,516 @@
...
+       * Number of warmup iterations.
+       */
+      struct num_warmup {
+        /**
+         * Description of num_warmup.
+         *
+         * @returns description
+         */
+        static std::string description() {
+          return "Number of warmup iterations.";
```
sakrejda: "Number of samples (from 'num_samples') to use for warmup."  Otherwise it's not clear that this is a subset of num_samples.

src/stan/services/sample/defaults.hpp
```diff
@@ -0,0 +1,516 @@
...
+        /**
+         * Validates num_warmup; num_warmup must be greater than or
+         * equal to 0.
+         *
+         * @param[in] num_warmup argument to validate
+         * @throw std::invalid argument if num_warmup < 0
+         */
+        static void validate(int num_warmup) {
+          if (!(num_warmup >= 0))
+            throw std::invalid_argument("num_warmup must be greater "
```
sakrejda: It also must be less than 'num_samples'.

src/stan/services/sample/defaults.hpp
```diff
@@ -0,0 +1,516 @@
...
+
+        /**
+         * Validates thin; thin must be greater than 0.
+         *
+         * @param[in] thin argument to validate
+         * @throw std::invalid argument if thin <= 0
+         */
+        static void validate(int thin) {
+          if (!(thin > 0))
+            throw std::invalid_argument("thin must be greater than 0.");
```
sakrejda: Also must be at most num_samples

src/stan/services/sample/hmc_nuts_dense_e_adapt.hpp
```diff
@@ -0,0 +1,126 @@
...
+       * @param init_buffer
+       * @param term_buffer
+       * @param window
+       * @param interrupt Callback for interrupts
+       * @param message_writer Writer for messages
+       * @param error_writer Writer for errors
+       * @param init_writer Writer callback for unconstrained inits
+       * @param sample_writer Writer for draws
+       * @param diagnostic_writer Writer for diagnostic information
+       * @return error code; 0 if no error
```
sakrejda: I think with optimization you were referring to stan::services::error_codes::OK for teh error codes.

betanalpha:
NONONONONONONONONONONONONONONONO
to using RStan’s convention for num_samples as all
iterations and num_warmup as some subset.  It’s 
confusing, awkward, and in general the worst.  I
wrote the sampling code to explicitly avoid this
(and CmdStan doesn’t use it).

sakrejda:
@betanalpha that too but whatever you choose it needs to be clear in the doc because at the moment I do something stupid whenever I switch from CmdStan to rstan.

bob-carpenter:
NOOOOO!  There is *one* sample consisting of a bunch of
draws.  The warmup iterations aren't draws from the posterior
and hence aren't kept as part of the sample.

Let's keep our language crisp at least in the doc and descriptions.  

There should *not* be a "num_samples" parameter anywhere!

bob-carpenter:
Yes, what is going on should be clear in the doc.

We don't want to change names in the interfaces for 2.x.  But
it'd be great to make everything consistent across interfaces for 3.0.

In the future, we don't want to conflate a draw with a sample or
indicate that what we get in the posterior iterations are independent
draws.

betanalpha:
We’re going to have to sit down and talk through this,
but from the math perspective “sample” does not mean
independent.  “Correlated samples” are exactly what
a Markov chain generates.  There are rigorous definitions
for all of these terms.

andrewgelman:
Hi, now I'm confused.  rstan uses "iter," not "num_samples," for number of iterations.

jgabry:
@andrewgelman rstan uses the names `iter` and `warmup`, but I think the convention Michael was referring to doesn't have to do with the names but rather the fact that `iter` is used for the total number rather than `total = iter + warmup`.

src/stan/io/random_var_context.hpp
syclik: I'll add better doc.

src/stan/model/finite_diff_grad.hpp
syclik: Right now, the test gradient option doesn't call the interrupt. This makes it very difficult to use from within RStan now. I intentionally had it call the interrupt once per parameter. I'm updating the doc, but this is intentional.

src/stan/services/optimize/bfgs.hpp
```diff
@@ -0,0 +1,200 @@
...
+        bfgs._conv_opts.tolAbsGrad = tol_grad;
+        bfgs._conv_opts.tolRelGrad = tol_rel_grad;
+        bfgs._conv_opts.tolAbsX = tol_param;
+        bfgs._conv_opts.maxIts = num_iterations;
+
+        double lp = bfgs.logp();
+
+        std::stringstream msg;
+        msg << "Initial log joint probability = " << lp;
+        message_writer(msg.str());
```
syclik: (This all needs to change, but we can't change it until we get the refactor in. The I/O is something that needs to be consolidated first through the refactor.)

src/stan/services/sample/defaults.hpp
syclik: Outside the scope of this pull request. Please open a new issue and see this file for reference: src/stan/services/arguments/arg_num_warmup.hpp

src/stan/services/sample/defaults.hpp
syclik: No, num_warmup is independent of num_samples.

src/stan/services/sample/defaults.hpp
syclik: also, no.

src/stan/services/sample/fixed_param.hpp
```diff
@@ -0,0 +1,72 @@
...
+#include <stan/services/util/rng.hpp>
+#include <stan/services/util/initialize.hpp>
+#include <vector>
+
+namespace stan {
+  namespace services {
+    namespace sample {
+
+      template <class Model>
+      int fixed_param(Model& model,
```
syclik: Yes, I'll add doc.

syclik:
Hi all, let's not derail this pull request due to naming disagreements. There is now a discussion on discourse, so please refrain from having that discussion on this pull request.

This pull request is a refactor. I've painstakingly maintained behavior from the `develop` branch. Even though I wanted to fix a lot of things along the way, I did not. This is not the right place for changes to old behavior. (I realize it's hard to tell what's what; if it's come along due to old behavior, I'll let you know and you can create a new issue for it.)

Please, let's keep this moving. The 4 things it needs:
- code approval on Stan. This is what I'm asking people to do now.
- confirmation that CmdStan works. I'm confirming.
- confirmation that RStan works. I'm waiting for @bgoodri to confirm.
- confirmation that PyStan works. I'm waiting for @ariddell to confirm.

I really need those looking at the C++ to approve the Stan code independently of the interfaces.

src/stan/services/experimental/advi/meanfield.hpp
```diff
@@ -0,0 +1,84 @@
...
+#include <stan/variational/advi.hpp>
+#include <boost/random/additive_combine.hpp>
+#include <string>
+#include <vector>
+
+namespace stan {
+  namespace services {
+    namespace experimental {
+      namespace advi {
+
```
sakrejda: Missing doc for arguments... don't know the code well enough to write doc here...

src/stan/services/optimize/bfgs.hpp
sakrejda: Good point, sorry about the off-topic stuff.

src/stan/services/experimental/advi/meanfield.hpp
syclik: I just added doc! This is exhausting.

src/test/unit/callbacks/chained_writer_test.cpp
```diff
@@ -0,0 +1,113 @@
+#include <stan/callbacks/chained_writer.hpp>
+#include <gtest/gtest.h>
+
+namespace test {
+  class mock_writer : public stan::callbacks::writer {
```
sakrejda: I did write a generic class that could get used instead of making one of these every time but... this is fine too. The tests just need to go, they don't need to be pretty.

src/stan/services/experimental/advi/meanfield.hpp
sakrejda: Yeah, I'm really not sure where you get the energy for this (I mean, I get it's part of your job but still...)

sakrejda:
@syclik So, a) this is amazing; b) I did look over all the code and most of the changes to tests.  I don't see any specific problems or cut-and-paste problems, having the interface doc consolidated to these interface functions makes it easier to understand, I read through the logic of the interface functions last time and we wrote tests for most of it so I didn't go through it again in detail.  Last time the argument defaults/doc static functions weren't there but I think they're good too--it's clear and makes all that stuff programatically accessible.... I checked that doc matched functions and that the defaults were reasonable for most of them.  Are there specific other things I should look at? I'm sure (hope!) somebody else looks over this as well but I think it's good to go.

ariddell:
OK. I'm on notice for checking that everything works. I'm going to push
this off until after the PyStan release if that's alright. (There are a
variety of new tests to add.)

src/stan/services/optimize/bfgs.hpp
```diff
@@ -0,0 +1,202 @@
...
+            msg << " " << std::setw(10) << std::setprecision(4)
+                << bfgs.alpha0() << " ";
+            msg << " " << std::setw(7)
+                << bfgs.grad_evals() << " ";
+            msg << " " << bfgs.note() << " ";
+            message_writer(msg.str());
+            msg.str("");
+          }
+
+          if (bfgs_ss.str().length() > 0) {
```
bob-carpenter: This is another one where I don't want to see this logic repeated everywhere.  Can we write an auxiliary function that is something like `write_then_clear(message_writer, bfgs_ss)`.

src/stan/services/optimize/bfgs.hpp
```diff
@@ -0,0 +1,202 @@
...
+            message_writer(bfgs_ss.str());
+            bfgs_ss.str("");
+          }
+
+          if (save_iterations) {
+            std::vector<double> values;
+            msg.str("");
```
bob-carpenter: This `msg.str("")` seems either dangerous (if there's something in `msg`) or redundant (if there's nothing in it).

src/stan/services/optimize/bfgs.hpp
```diff
@@ -0,0 +1,202 @@
...
+            model.write_array(rng, cont_vector, disc_vector, values,
+                              true, true, &msg);
+            if (msg.str().length() > 0)
```
bob-carpenter: functions --- this is all over the place.  I think it should be built into the message writer if it's going to be this prevalent.

src/stan/services/optimize/bfgs.hpp
```diff
@@ -0,0 +1,202 @@
...
+            parameter_writer(values);
+          }
+        }
+
+        if (!save_iterations) {
+          std::vector<double> values;
+          msg.str("");
+          model.write_array(rng, cont_vector, disc_vector, values,
+                            true, true, &msg);
+          if (msg.str().length() > 0)
```
bob-carpenter: Will there be an easy way to find all these to fix?  I won't comment on any more, but the pattern is ubiquitous.

src/stan/services/optimize/bfgs.hpp
```diff
@@ -0,0 +1,202 @@
...
+            message_writer(msg.str());
+          values.insert(values.begin(), lp);
+          parameter_writer(values);
+        }
+
+        int return_code;
```
bob-carpenter: Why is the message writer writing something based on `ret` twice, once in the message writer inside the conditionals and once out?  I think it's much cleaner to not declare undefined variables then define them in conditionals then use them again outside if it can be avoided.  As in:

```
if (ret >= 0) {
  message_writer("Opt...
  return error_codes::OK;
} else {
  ...
}
```

If you must write the `bfgs.get_code_string( )` thing, do you really need to write anyting else?

src/stan/services/optimize/defaults.hpp
bob-carpenter: I take it then you'll say the same thing about all the rewording I was suggesting.  This is going to be a long series of pull requests.  I think it may be easier just to review everything again because nobody's creating these issues when you say things are out of scope.  I don't really want to create an issue to change a single word in an output.  We haven't really nailed down the API at that level.  Is it problematic because of all the brittle exact-wording tests would require major rewriting of tests?

src/stan/services/optimize/defaults.hpp
```diff
@@ -0,0 +1,326 @@
...
+         * Description of save_iterations.
+         *
+         * @returns description
+         */
+        static std::string description() {
+          return "Save optimization interations to output.";
+        }
+
+        /**
+         * Validates save_iterations. This is a noop.
```
bob-carpenter: "no-op"!  Unless you can get a diaresis into the code and write "nöop".

src/stan/services/optimize/defaults.hpp
```diff
@@ -0,0 +1,326 @@
...
+         *
+         * @param[in] save_iterations argument to validate
+         */
+        static void validate(bool save_iterations) {
```
bob-carpenter: Why is there a no-op validation for booleans?  The validation is going to be up at the level of some string getting converted to a `bool` --- once it's a `bool` there's nothing to validate.  Given that these aren't extending some abstract base class, can't we just get rid of useless validation code?

src/stan/services/optimize/lbfgs.hpp
```diff
@@ -0,0 +1,206 @@
...
+          = stan::services::util::initialize(model, init, rng, init_radius,
+                                             false,
+                                             message_writer, init_writer);
+
+        std::stringstream lbfgs_ss;
+        typedef stan::optimization::BFGSLineSearch
+          <Model, stan::optimization::LBFGSUpdate<> > Optimizer;
+        Optimizer lbfgs(model, cont_vector, disc_vector, &lbfgs_ss);
+        lbfgs.get_qnupdate().set_history_size(history_size);
+        lbfgs._ls_opts.alpha0 = init_alpha;
```
bob-carpenter: I"m confused now.  Where does the validation hpapne to make sure we a are passing reasonale values to lbfgs?

src/stan/services/optimize/lbfgs.hpp
```diff
@@ -0,0 +1,206 @@
...
+                            true, true, &msg);
+          if (msg.str().length() > 0)
+            message_writer(msg.str());
+
+          values.insert(values.begin(), lp);
+          parameter_writer(values);
+        }
+        int ret = 0;
+
+        while (ret == 0) {
```
bob-carpenter: There's a lot of repeated code pattern, but I don't want to duplicate my comments.  Can you just carry the earlier ones down, like making this a for loop?

syclik:
Thanks, @bob-carpenter! I took a quick read through the comments. I'll try to address as many as I can today.

A couple of bigger points:
- naming. We should definitely fix that now. I'm happy going with `no_op` instead of `noop`. But I think I'd rather have a base implementation that acts as a `no_op`. Thoughts?
- I'll try to generalize your comments to other places in the code.
- Places where I can't change, I'll leave comments in place.

src/stan/callbacks/chained_writer.hpp
```diff
@@ -0,0 +1,89 @@
...
+  namespace callbacks {
+
+    /**
+     * <code>chained_writer</code> is an implementation that writes to
+     * multiple writers.
+     *
+     * For any given call to this writer, it will delegate to both writers
+     * provided in the constructor.
+     */
+    class chained_writer : public writer {
```
syclik: You're absolutely right. Any suggestions for names? `tee_writer` or `teed_writer`? `forked_writer`?

src/stan/callbacks/chained_writer.hpp
```diff
@@ -0,0 +1,89 @@
...
+        writer2_(key, value);
+      }
+
+      void operator()(const std::string& key, const std::string& value) {
+        writer1_(key, value);
+        writer2_(key, value);
+      }
+
+      void operator()(const std::string& key,
+                      const double* values,
```
syclik: The doc is inherited in doxygen.
The reason for the pointer was to keep in line with the old pull request. I really think IO should be handled better, so I'm going to scale back the writers to what we use now:
- `writer()`
- `writer(std::string)`
- `writer(std::vector<std::string>)`
- `writer(std::vector<double>)`

It'll cut down on our overhead for now. I think it's the right way to go and we won't have pointers floating around.

src/stan/callbacks/interrupt.hpp
```diff
@@ -0,0 +1,22 @@
...
+     * <code>interrupt</code> is an abstract base class defining the interface
+     * for Stan interrupt callbacks.
+     *
+     * The interrupt is called from within Stan algorithms to allow
+     * for the interfaces to handle interrupt signals (ctrl-c).
+     */
+    class interrupt {
+    public:
+      interrupt() {}
+      virtual void operator()() = 0;
```
syclik: +1 and thanks. Given that doxygen inherits doc from superclasses, should we just doc the superclass and not the subclass if it's trivial?

src/stan/callbacks/interrupt.hpp
```diff
@@ -0,0 +1,22 @@
...
+    /**
+     * <code>interrupt</code> is an abstract base class defining the interface
+     * for Stan interrupt callbacks.
+     *
+     * The interrupt is called from within Stan algorithms to allow
+     * for the interfaces to handle interrupt signals (ctrl-c).
+     */
+    class interrupt {
+    public:
+      interrupt() {}
```
syclik: Absolutely. I'll add virtual destructors.

src/stan/io/random_var_context.hpp
```diff
@@ -0,0 +1,222 @@
...
+       * @param model instantiated model to generate variables for
+       * @param[in,out] rng pseudo-random number generator
+       * @param[in] init_radius the unconstrained variables are uniform draws
+       *   from -init_radius to init_radius.
+       * @param[in] init_zero indicates whether all unconstrained variables
+       *   should be initialized at 0. When init_zero is false, init_radius
+       *   must be greater than 0.
+       */
+      template <class Model, class RNG>
+      random_var_context(Model& model,
```
syclik: yeah... I tried to make add the `const` keyword at one point; it doesn't work. We're not there yet. But this is making it easier.

src/stan/mcmc/base_mcmc.hpp
```diff
@@ -18,15 +18,15 @@ namespace stan {
...
+                 callbacks::writer& info_writer,
+                 callbacks::writer& error_writer) = 0;
 
       virtual void get_sampler_param_names(std::vector<std::string>& names) {}
 
       virtual void get_sampler_params(std::vector<double>& values) {}
 
       virtual void
-      write_sampler_state(interface_callbacks::writer::base_writer& writer) {}
+      write_sampler_state(callbacks::writer& writer) {}
```
syclik: @bob-carpenter, that'll come later. I completely agree that we need sampler-state specific writers rather than overloading a generic writer. We'll get there, but after this pull request. I'm thinking about it.

src/stan/mcmc/hmc/base_hmc.hpp
```diff
@@ -34,7 +34,7 @@ namespace stan {
           epsilon_jitter_(0.0) {}
```
syclik: Yes, but it shouldn't hurt too much here. I think git will handle the merge fine.

bob-carpenter:
> On Dec 27, 2016, at 9:51 AM, Daniel Lee <notifications@github.com> wrote:
> 
> You're absolutely right. Any suggestions for names? tee_writer or teed_writer? forked_writer?

I like tee_writer from the unix:

See:  https://en.wikipedia.org/wiki/Tee_(command)

> The doc is inherited in doxygen.

Fantastic.  Then if the doc doesn't change in the subclass, I'd rather
just inherit.  I don't think you need to write "teed version of ..." in
front of all those.  Maybe with all the time and editorial staff in the
world, but not worth it for us.

> The reason for the pointer was to keep in line with the old pull request. I really think IO should be handled better, so I'm going to scale back the writers to what we use now:
> 
> 	• writer()
> 	• writer(std::string)
> 	• writer(std::vector<std::string>)
> 	• writer(std::vector<double>)
> It'll cut down on our overhead for now. I think it's the right way to go and we won't have pointers floating around.

Nice!

> +1 and thanks. Given that doxygen inherits doc from superclasses, should we just doc the superclass and not the subclass if it's trivial?

Yes, see above.

> Absolutely. I'll add virtual destructors.

Thanks.  That'll cover us if anyone creates a subclass
that frees heap memory in the destructor.
> 
> yeah... I tried to make add the const keyword at one point; it doesn't work. We're not there yet. But this is making it easier.

I know.  Can't wait.  I've felt bad about not getting
const right before and I spent days in the past on
more than one occasion trying to fix it.  I think I could
do it now.  Skills training from compiling C++ night
and day for 5 years! 

>  I completely agree that we need sampler-state specific writers rather than overloading a generic writer. We'll get there, but after this pull request.

Understood.
> 
> 
> In src/stan/mcmc/hmc/base_hmc.hpp:
> 
> > @@ -34,7 +34,7 @@ namespace stan {
>            epsilon_jitter_(0.0) {}
> 
> Yes, but it shouldn't hurt too much here. I think git will handle the merge fine.

OK.  And I think we should still support it, just present a
deprecation warning until Stan 3.

- Bob

bob-carpenter:
> On Dec 27, 2016, at 9:19 AM, Daniel Lee <notifications@github.com> wrote:
> 
> Thanks, @bob-carpenter! I took a quick read through the comments. I'll try to address as many as I can today.
> 
> A couple of bigger points:
> 
> 	• naming. We should definitely fix that now. I'm happy going with no_op instead of noop. But I think I'd rather have a base implementation that acts as a no_op. Thoughts?

I commented at length on that in one of the comments.  Short version:
absolutely if you can delegate calls to one method to another.  Makes
it easier and less error prone to implement subclasses.

> 	• I'll try to generalize your comments to other places in the code.

Thanks.

> 	• Places where I can't change, I'll leave comments in place.

Sounds good.

- Bob

src/stan/io/empty_var_context.hpp
```diff
@@ -0,0 +1,113 @@
...
+     * This is an implementation of a var_context that doesn't contain
+     * any variables.
+     */
+    class empty_var_context : public var_context {
+    public:
+      /**
+       * Destructor
+       *
+       */
+      ~empty_var_context() {}
```
syclik: I'll make sure the base class, `var_context`, has a virtual destructor. As far as I can tell, the subclass's destructor doesn't have to be virtual. I don't think there will be a measurable hit if there's a performance difference: we only create this once in CmdStan for the lifetime of the program.

src/stan/io/empty_var_context.hpp
```diff
@@ -0,0 +1,113 @@
...
+
+      /**
+       * Return a list of the names of the floating point variables in
+       * the context. This contet has no variables.
+       *
+       * @param names Vector to store the list of names in.
+       */
+      void names_r(std::vector<std::string>& names) const {
+        names.clear();
+      }
```
syclik: This follows the design of the super class, `var_context`, which has been around for a long time.

src/stan/io/random_var_context.hpp
```diff
@@ -0,0 +1,222 @@
...
+        model.get_param_names(names_);
+        model.get_dims(dims_);
+
+        // cutting names_ and dims_ down to just the constrained parameters
+        std::vector<std::string> constrained_params_names;
+        model.constrained_param_names(constrained_params_names, false, false);
+        size_t keep = constrained_params_names.size();
+        size_t index = 0;
+        size_t num = 0;
+        for (index = 0; index < dims_.size(); index++) {
```
syclik: yes, it's needed past the loop. I bet there's a cleaner way to write the function.

bob-carpenter:
> On Dec 28, 2016, at 9:30 AM, Daniel Lee <notifications@github.com> wrote:
> 
> @syclik commented on this pull request.
> 
> In src/stan/io/empty_var_context.hpp:
> 
> > +#include <vector>
> +
> +namespace stan {
> +  namespace io {
> +
> +    /**
> +     * This is an implementation of a var_context that doesn't contain
> +     * any variables.
> +     */
> +    class empty_var_context : public var_context {
> +    public:
> +      /**
> +       * Destructor
> +       *
> +       */
> +      ~empty_var_context() {}
> 
> I'll make sure the base class, var_context, has a virtual destructor. As far as I can tell, the subclass's destructor doesn't have to be virtual. I don't think there will be a measurable hit if there's a performance difference: we only create this once in CmdStan for the lifetime of the program.

The subclass's destructor will be virtual if the base class's
is declared as virtual.

Performance hit or no, it's necessary for a properly formed
base class if you want to extend it generically.  Otherwise,
the destructor and class should be carefully documented to 
ensure nobody extends it with a class that manages its own
memory.

syclik:
I just pushed a lot of changes. I need to walk through the src/stan/services folder for doc next. This is going to affect RStan, PyStan, and CmdStan, so the upstream tests won't pass, but I'll take care of that next.

src/stan/services/experimental/advi/fullrank.hpp
```diff
@@ -0,0 +1,112 @@
...
+         * @param eval_elbo evaluate ELBO every Nth iteration
+         * @param output_samples number of posterior samples to draw and save
+         * @param[out] interrupt interrupt callback to be called every iteration
+         * @param[out] message_writer output for messages
+         * @param[out] init_writer Writer callback for unconstrained inits
+         * @param[out] parameter_writer output for parameter values
+         * @param[out] diagnostic_writer output for diagnostic values
+         * @return stan::services::error_codes::OK if successful
+         */
+        template <class Model>
```
syclik: Yes, I believe so. I don't think `prob_grad` has enough of the methods -- we've just tacked on additional methods through the generator instead of actually updating the appropriate base class. So yes, let's create an issue for that.

src/stan/services/optimize/bfgs.hpp
```diff
@@ -0,0 +1,202 @@
...
+
+        std::vector<std::string> names;
+        names.push_back("lp__");
+        model.constrained_param_names(names, true, true);
+        parameter_writer(names);
+
+        if (save_iterations) {
+          std::vector<double> values;
+          msg.str("");
+          model.write_array(rng, cont_vector, disc_vector, values,
```
syclik: yes.

src/stan/services/optimize/bfgs.hpp
```diff
@@ -0,0 +1,202 @@
...
+                            true, true, &msg);
+          if (msg.str().length() > 0)
```
syclik: This looks odd, but this is to maintain the old printing behavior. I'll add a comment.

src/stan/services/optimize/bfgs.hpp
```diff
@@ -0,0 +1,202 @@
...
+            message_writer(msg.str());
+
+          values.insert(values.begin(), lp);
+          parameter_writer(values);
+        }
+        int ret = 0;
```
syclik: Removed space. No, it can't be just a for loop -- `ret` is used outside the while loop.

src/stan/services/optimize/bfgs.hpp
syclik: Can't change it to that: there is no method for accepting double values and it'll change the spacing to something that's different. I left the output identical to the old output to be safe.

The repeated code isn't new code. This is code that's moved that GitHub isn't tracking so well as moved code.

src/stan/services/optimize/bfgs.hpp
syclik: We'll walk through all instances where writers are used and we'll fix it.

src/stan/services/optimize/bfgs.hpp
```diff
@@ -0,0 +1,202 @@
...
+
+        while (ret == 0) {
+          interrupt();
+          if (refresh > 0
+              && (bfgs.iter_num() == 0
+                  || (bfgs.iter_num() + 1 % refresh == 0)))
```
syclik: Fixed

src/stan/services/optimize/bfgs.hpp
```diff
@@ -0,0 +1,202 @@
...
+            message_writer("    Iter "
```
syclik: No. Of course not. But I'm trying hard not to fix everything that's wrong with our whole code base or we'll never get this in. This has been in since it was written -- it bothers me too, but I'm really trying to fix the things that need to be fixed. We can walk through the code base later.

src/stan/services/optimize/bfgs.hpp
```diff
@@ -0,0 +1,202 @@
...
+
+          ret = bfgs.step();
+          lp = bfgs.logp();
+          bfgs.params_r(cont_vector);
+
+          if (refresh > 0
+              && (ret != 0
+                  || !bfgs.note().empty()
+                  || bfgs.iter_num() == 0
+                  || (bfgs.iter_num() + 1 % refresh == 0))) {
```
syclik: found the rest of the instances

src/stan/services/optimize/bfgs.hpp
syclik: I want to push this off until later -- this only happens in a few places.

src/stan/services/optimize/bfgs.hpp
syclik: I don't know. This was old behavior that I'm just trying to maintain with this refactor.

src/stan/services/optimize/bfgs.hpp
bob-carpenter: Is the problem that you get a free newline when you send `msg.str()` or `msg.str("")`.  I don't like this pattern---I'd rather see the newlines be either explicit if you're printing just strings or part of a higher-level interface like a sample writer if you put ends of lines after them automatically (I think `write_array` does that now).

bob-carpenter:
> On Dec 28, 2016, at 11:21 PM, Daniel Lee <notifications@github.com> wrote:
> 
> @syclik commented on this pull request.
> 
> In src/stan/services/optimize/bfgs.hpp:
> 
> > +          model.write_array(rng, cont_vector, disc_vector, values,
> +                            true, true, &msg);
> +          if (msg.str().length() > 0)
> +            message_writer(msg.str());
> +
> +          values.insert(values.begin(), lp);
> +          parameter_writer(values);
> +        }
> +        int ret = 0;
> +
> +        while (ret == 0) {
> +          interrupt();
> +          if (refresh > 0
> +              && (bfgs.iter_num() == 0
> +                  || (bfgs.iter_num() + 1 % refresh == 0)))
> +            message_writer("    Iter "
> 
> No. Of course not. But I'm trying hard not to fix everything that's wrong with our whole code base or we'll never get this in. This has been in since it was written -- it bothers me too, but I'm really trying to fix the things that need to be fixed. We can walk through the code base later.

Isn't the behavior going to be wrong here given that % binds
more tightly than +.  If so, could you please create an issue
for this?  Thanks.  I can't actually understand the logic or I'd
do it myself.

bob-carpenter:
> This one hurts. It's in ADVI and it's propagated through to CmdStan. I don't think all our developers are on board with the terminology.

Michael's OK with it, and I think we can just enforce it.
But I'm OK taking a pass through this later.  We can't actually
remove the existing naming until Stan 3, but we could replace
it and deprecate the old behavior.

Sorry if the comments are overwhelming.  I'm just noting
everything that pops out at me reviewing the code.

> 
> I don't think it's hopeless, but I think fixing this should come with fixing all of ADVI. I'm going to create a new issue for all the things you've written up.

Thanks!

> +           * @param[in] tol_rel_obj argument to validate
> 
> We haven't been too consistent with this. I can either add them everywhere or remove them. Or just let it be haphazard until we know what we want.

OK.  It's very inconsistent in the code base.  I often
forget because it wasn't in Javadoc when I used it (or I
didn't know about it).   

As to what we want, we can have the `[in]` everywhere if
you think it's helpful to appear more consistent.  It's 
exactly like adding `const` to a primitive---it doesn't do
anything to the external API because C++ pass by value.
That is, it has to be `[in]` by construction if it's a primitive
passed by value, the same way it's `const` by construction.

> In src/stan/services/experimental/advi/fullrank.hpp:
> 
> > +
> +          stan::variational::advi<Model,
> +                                  stan::variational::normal_fullrank,
> +                                  boost::ecuyer1988>
> +            cmd_advi(model,
> +                     cont_params,
> +                     rng,
> +                     grad_samples,
> +                     elbo_samples,
> +                     eval_elbo,
> +                     output_samples);
> +          cmd_advi.run(eta, adapt_engaged, adapt_iterations,
> +                       tol_rel_obj, max_iterations,
> +                       message_writer, parameter_writer, diagnostic_writer);
> +
> +          return 0;
> 
> It can throw. ADVI is a mess, so I can't really track through all the different types of things it can or can't do, so I'm going to stick it on the ADVI issue I'm creating.

Thanks.  I'm not too worried about ADVI until it's a bit more
robust and we can recommend it to users in some situations.

syclik:
Jenkins, retest this please.

src/stan/services/optimize/bfgs.hpp
```diff
@@ -0,0 +1,202 @@
...
+       * @param[in] tol_rel_grad convergence tolerance on the relative norm of
+       *   the gradient
+       * @param[in] tol_param convergence tolerance on changes in parameter value
+       * @param[in] num_iterations maximum number of iterations
+       * @param[in] save_iterations indicates whether all the interations should
+       *   be saved to the parameter_writer
+       * @param[in] refresh how often to write output to message_writer
+       * @param[out] interrupt interrupt callback to be called every iteration
+       * @param[out] message_writer output for messages
+       * @param[out] init_writer Writer callback for unconstrained inits
```
syclik: Ok. I can doc them all as `[in, out]`.

src/stan/services/optimize/bfgs.hpp
syclik: Yes. We'll tackle the output issue after this is in. There are hacks in RStan now that parse output: I don't know how much it tries to do through regex.

After this pull request, I should be able to get into the different places where RStan does that and remove make it not rely on regex.

src/stan/services/optimize/bfgs.hpp
syclik: sorry -- I misread the original comment. It does happen in a lot of places. Let me see how many spots and get this going if it's not too disruptive.

src/stan/services/optimize/bfgs.hpp
syclik: We can fix this with using local streams. I'll see how hard it is to switch things over.

src/stan/services/optimize/defaults.hpp
syclik: The validation happens from within the interfaces. Although these don't extend from a shared base class, the intent was to share a common concept that could be accessed programmatically from the interfaces. I think it's better to just leave the `validate()` functions there even if they are no-ops.

src/stan/services/optimize/lbfgs.hpp
syclik: From CmdStan, RStan, PyStan. I'm pretty sure that's what we decided. Personally, I would have wanted to have (redundant) validation inside the service functions, but I think at some point I was convinced that it was better practice to validate arguments outside.

src/stan/services/optimize/lbfgs.hpp
syclik: I wish. Bad coding practice that's in lbfgs that I'm not going to tackle here.

bob-carpenter:
> 
> sorry -- I misread the original comment. It does happen in a lot of places. Let me see how many spots and get this going if it's not too disruptive.

Or just create a new issue so we don't forget.  I really don't want to drag this one out---I was just commenting on all the stuff that I saw.

bob-carpenter:
> From CmdStan, RStan, PyStan. I'm pretty sure that's what we decided. Personally, I would have wanted to have (redundant) validation inside the service functions, but I think at some point I was convinced that it was better practice to validate arguments outside.

We wanted to do this outside so that the interfaces could produce error messages in their own language.

syclik:
Yup. That's right. I think the current implementation isn't perfect, but
it'll work for now.

ariddell:
> We wanted to do this outside so that the interfaces could produce error messages in their own language.

I think one source of error messages (via throwing exceptions) would be fine. I thought there were other reasons pushing validation outside the function.

syclik:
Bob, action is back on you to review the pull request.

I made most of the changes you suggested. If I missed anything major, please remind me. The ADVI issues, I've created a new issue: #2195.

bob-carpenter:
@syclik  Thanks!  That's awesome.  I'll be able to take another pass tomorrow (1 January 2017!).

src/stan/services/optimize/newton.hpp
```diff
@@ -0,0 +1,130 @@
...
+               << std::setw(2) << (m + 1) << "."
+               << " Log joint probability = " << std::setw(10) << lp
+               << ". Improved by " << (lp - lastlp) << ".";
+          message_writer(msg2.str());
+
+          if (std::fabs(lp - lastlp) <= 1e-8)
+            break;
+        }
+
+        {
```
bob-carpenter: these braces don't do anything

src/stan/services/util/experimental_message.hpp
```diff
@@ -0,0 +1,33 @@
...
+       *
+       * @param[in,out] message_writer writer for experimental algorithm message
+       */
+      void experimental_message(stan::callbacks::writer&
+                                message_writer) {
+        message_writer("------------------------------"
+                       "------------------------------");
+        message_writer("EXPERIMENTAL ALGORITHM:");
+        message_writer("  - please expect frequent updates to the procedure");
+        message_writer("  - please expect unexpected inference results");
```
bob-carpenter: expect the unexpected?  how about instead of both lines:  

This procedure has not been thoroughly tested and may be unstable or buggy.  The interface is subject to change.

bob-carpenter:
This pull request was too large---it overwhelmed my mac and ipad trying to review.  I lost about ten comments, mostly about all the repeated code snippets that should be oulled into functions.

Never mind.  I just approved it because as much as I hate to see all this code duplication, our code doesnt need to be perfectly factored going forward and I want to be pragmatic and get this in ASAP

src/stan/services/experimental/advi/fullrank.hpp
```diff
@@ -0,0 +1,112 @@
...
+                                               message_writer, init_writer);
+
+
+          std::vector<std::string> names;
+          names.push_back("lp__");
+          model.constrained_param_names(names, true, true);
+          parameter_writer(names);
+
+          Eigen::VectorXd cont_params;
+          cont_params.resize(cont_vector.size());
```
syclik: I'm changing both instances to construct with the right size. We can't use `Eigen::Map` -- it doesn't compile.

bob-carpenter:
> On Jan 3, 2017, at 3:48 PM, Daniel Lee <notifications@github.com> wrote:
> 
> @syclik commented on this pull request.
> ...
> +          Eigen::VectorXd cont_params;
> +          cont_params.resize(cont_vector.size());
> 
> I'm changing both instances to construct with the right size. We can't use Eigen::Map -- it doesn't compile.

It compiles and passes tests on my machine with both assignment:

```
          Eigen::VectorXd cont_params
            = Eigen::Map<Eigen::VectorXd>(&cont_vector[0], cont_vector.size(), 1);
```

and construction:

```
          Eigen::VectorXd cont_params(Eigen::Map<Eigen::VectorXd>(&cont_vector[0], cont_vector.size(), 1));
```


Not sure it's faster or clearer, though.  Here's the lowdown on Map:

  https://eigen.tuxfamily.org/dox/group__TutorialMapClass.html

- Bob

syclik:
Just communicated with @ariddell and he gave it a thumbs up.

This is now ready to merge as soon as the final tests pass. We have:
- a review of the Stan branch by @bob-carpenter 
- sign-off by a CmdStan maintainer, me
- sign-off by an RStan maintainer, @bgoodri 
- sign-off by the PyStan maintainer, @ariddell 

As soon as the tests pass, I'm merging.

bgoodri:
I think it is fine.

bob-carpenter:
W00T!

Issue #1459, opened 2015-05-22
conflict with function `pi()` and user variable `pi`


bob-carpenter:
Bee pointed out on the stan-users list that a user-defined `pi` and the built-in function `pi()` will conflict in the C++ compiler.  I'm not quite sure why this is, but get the same issue with this simplified example.

```
parameters {
  real pi;
}
model {
}
generated quantities {
  real mu;
  mu = 2 * pi();
}
```

Here's the error from clang:

```
In file included from <command line>:7:
/Users/carp/temp2/pime.hpp:247:43: error: called object type 'double' is not a function or function pointer
            stan::math::assign(mu, (2 * pi()));
```

* [ ] qualify `pi()` as `stan::math::pi()` and it will compile

bob-carpenter:
With a `using stan::math::pi;` in the try-catch block with the assignment, there's no problem.  With a top-level `using stan::math::pi` the same error as above arises.  But then using the full `stan::math::pi()` fails with a different error.  I'm not sure what's going on with the scope and member variable vs. function namespace.  But I'll track it down.

PR #2006, opened 2016-08-10
qualify nullary fun names to avoid clash, finishes #1459

bob-carpenter:
#### Submisison Checklist

#### Summary

Qualify namespace with `stan::math::` for built-in nullary functions.

#### Intended Effect

Avoid name conflicts that currently arise if a variable in Stan is declared with the same name as a built-in and both are used.

#### How to Verify

New model test exercising everything.

#### Side Effects

No.

#### Documentation

No, now the code matches the doc.

#### Reviewer Suggestions

Anyone.

#### Copyright and Licensing

Columbia University

### DIFF
--- a/src/stan/lang/grammars/semantic_actions_def.cpp
+++ b/src/stan/lang/grammars/semantic_actions_def.cpp
@@ -24,6 +24,24 @@ namespace stan {
 
   namespace lang {
 
+    /**
+     * Add qualifier "stan::math::" to nullary functions defined in the
+     * Stan language.  The original name is set to the name here and
+     * the name is converted to have the prefix.
+     *
+     * @param f Function to qualify.
+     */
+    void qualify_builtins(fun& f) {
+      if (f.args_.size() > 0) return;
+      if (f.name_ == "e" || f.name_ == "pi" || f.name_ == "log2"
+          || f.name_ == "log10" || f.name_ == "sqrt2"
+          || f.name_ == "not_a_number" || f.name_ == "positive_infinity"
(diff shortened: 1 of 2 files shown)


sakrejda:
Code looks good to me as long as the new test passes but do we really want to allow `negative_infinity`, `positive_infinity`, and `machine_precision` as parameter names? I think it would make sense to keep those reserved unless somebody is trying to run a Stan code obfuscation contest.

syclik:
What do you mean "parameter names"? Those are existing functions / keywords in the current version of Stan (and have been for a while).

sakrejda:
@syclik Look at the new test---doing this allows users to use these as parameter names (in the Stan sense).  It makes sense for things like "pi" that have established usages as parameter names but positive_infinity as a parameter name only belongs in an obfuscation contest unless there's some use case that I'm missing :confused:

syclik:
@bob-carpenter: cool. I think this is awesome.

@sakrejda: I don't see any problem with it. Thanks for pointing me to the test. We allow users to do with the language what they want, even if it's bad for them.

sakrejda:
@syclik That's fine, I didn't know that it was a design decision.

bob-carpenter:
It was more that I thought we should be consistent and either rule out all the nullary (zero arguments) functions as variable names or rule out none of them.

sakrejda:
@bob-carpenter I think it's a fine decision, just wanted to raise the issue so it would get thought through once.  It's not like we rule out other stupid names like "james_bond" or "nice_rabbit" as a parameter name.

PR #2010, opened 2016-08-12
Manual: Added indexing-based simplification to Multivariate Regression Example

mike-lawrence:
#### Summary
The Multivariate Regression Example included some optimizations but didn't include the, in my opinion, useful simplification recommended at the beginning of the Multiple Indexing and Range Indexing chapter. My substitution should probably be checked, as the order of variables in the "clunky assignment to a local variable" of the MRE and MIRI sections were different.

#### Copyright and Licensing
I assert no rights with regards to copyright & licensing.

### DIFF
--- a/src/docs/stan-reference/programming.tex
+++ b/src/docs/stan-reference/programming.tex
@@ -2533,6 +2533,13 @@ non-vectorized alternative:
     y[n] ~ normal(x[n] * beta[jj[n]], sigma);
 \end{stancode}
 %
+For equivalent performance to the vectorized version, but less typing, the entire block can be replaced with:
+%
+\begin{stancode}
+  y ~ normal(beta[jj] .* x,sigma);
+\end{stancode}
+%
+which takes advantage of indexing introduced in Stan version 2.9 (see \refchapter{multi-indexing}).
 
 The code in the Stan program above also builds up an array of vectors
 for the outcomes and for the multivariate normal, which provides a
@@ -2635,14 +2642,10 @@ transformed parameters {
   beta = u * gamma + (diag_pre_multiply(tau,L_Omega) * z)';
 }
 model {
-  vector[N] x_beta_jj;
-  for (n in 1:N)
-    x_beta_jj[n] = x[n] * beta[jj[n]]';
-  y ~ normal(x_beta_jj, sigma);
-
   to_vector(z) ~ normal(0, 1); 
   L_Omega ~ lkj_corr_cholesky(2);
   to_vector(gamma) ~ normal(0, 5);
+  y ~ normal(beta[jj] .* x, sigma);
 }
 \end{stancode}
 
--- a/src/docs/stan-reference/programming.tex
+++ b/src/docs/stan-reference/programming.tex
@@ -2475,14 +2475,14 @@ data {
   int<lower=1> J;              // num groups
   int<lower=1> L;              // num group predictors
   int<lower=1,upper=J> jj[N];  // group for individual
-  matrix[N, k] x;               // individual predictors
+  matrix[N, K] x;               // individual predictors
   row_vector[L] u[J];          // group predictors
   vector[N] y;                 // outcomes
 }
 parameters {
   corr_matrix[K] Omega;        // prior correlation
   vector<lower=0>[K] tau;      // prior scale
-  matrix[L, k] gamma;           // group coeffs
+  matrix[L, K] gamma;           // group coeffs
   vector[K] beta[J];           // indiv coeffs by group
   real<lower=0> sigma;         // prediction error scale
 }
--- a/src/docs/stan-reference/programming.tex
+++ b/src/docs/stan-reference/programming.tex
@@ -2496,12 +2496,8 @@ model {
       u_gamma[j] = u[j] * gamma;
     beta ~ multi_normal(u_gamma, quad_form_diag(Omega, tau));
   }
-  {
-    vector[N] x_beta_jj;
-    for (n in 1:N)
-      x_beta_jj[n] = x[n] * beta[jj[n]];
-    y ~ normal(x_beta_jj, sigma);
-  }
+  for (n in 1:N)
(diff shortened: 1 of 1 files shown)


bob-carpenter:
Thanks!  That looks great.  Mitzi and I are going to take a run through all of our models and try to update them as soon as we get the compound declare/define and unary vectorization features in.  

If someone lets me know how I can check out this change and make sure the manual builds and the link works, I can merge it.

syclik:
Jenkins, ok to test.

syclik:
@bob-carpenter, Jenkins will build the manual for us; I'll put up a link once it's done.

It won't pass all the upstream tests since it the commit isn't on stan-dev/stan, but we can merge once you're satisfied that is passes the manual build.

mike-lawrence:
FYI, found some typos while checking if the model compiles so I fixed these in a couple new commits.

mike-lawrence:
The model won't compile, so something is wrong with how I attempted to apply the solution shown at the beginning of the "Multiple Indexing and Range Indexing" chapter. As it is now, it's complaining about the line "y ~ normal(beta[jj] .* x, sigma);" , because the "beta[jj] .* x" part yields a matrix when it needs to be a real, array of reals, vector or row vector.

bob-carpenter:
If you'd like our help debugging, please include the whole
Stan program that doesn't compile.

mike-lawrence:
@bob-carpenter Oh, I was referring to the model in the manual that my pull affects. Here's the full model:
`
data {
  int<lower=0> N;              // num individuals
  int<lower=1> K;              // num ind predictors
  int<lower=1> J;              // num groups
  int<lower=1> L;              // num group predictors
  int<lower=1,upper=J> jj[N];  // group for individual
  matrix[N, K] x;               // individual predictors
  row_vector[L] u[J];          // group predictors
  vector[N] y;                 // outcomes
}
parameters {
  matrix[K, J] z;
  cholesky_factor_corr[K] L_Omega;  
  vector<lower=0,upper=pi()/2>[K] tau_unif;  
  matrix[L, K] gamma;                         // group coeffs
  real<lower=0> sigma;                       // prediction error scale
}
transformed parameters {
  matrix[J, K] beta;
  vector<lower=0>[K] tau;     // prior scale
  for (k in 1:K) tau[k] = 2.5 * tan(tau_unif[k]);
  beta = u * gamma + (diag_pre_multiply(tau,L_Omega) * z)';
}
model {
  to_vector(z) ~ normal(0, 1); 
  L_Omega ~ lkj_corr_cholesky(2);
  to_vector(gamma) ~ normal(0, 5);
  y ~ normal(beta[jj] .* x, sigma);
}
`

mike-lawrence:
I believe I got it to work. It seems that rows_dot_product() is necessary.

mike-lawrence:
Once I got it working, the version using rows_dot_product() was quite a bit faster than the version that created x_beta_jj, so I posted another commit that eliminates use/reference to the latter altogether.

mike-lawrence:
Not sure it's of use to anyone, but I ended up creating an example (r code to generate data, then various stan models to test different approaches) while fixing this: 
https://gist.github.com/mike-lawrence/bce5f79852e01c904eaf67405585feb7

bob-carpenter:
Thanks.  

I'd suggest changing this declaration:


  matrix[nS,nX] Smu ; 

to

  vector[nX] Smu[nS];

in order to vectorize this

    Smu[ns] ~ multi_normal(mu,Z) ;

to

   Smu ~ multi_normal(mu, Z);

That'll ensure that Z only has to be factored once.

- Bob

mike-lawrence:
@bob-carpenter unfortunately rows_dot_product() needs Smu to be a matrix. Your suggestion also breaks the hmod_c0 version that creates a variable q instead of using rows_dot_product().

bob-carpenter:
If you need those matrix operations, it'd still be
better to copy the final matrix Smu into a local variable that is
an array of vectors before calling multi_normal.  Way
faster than factoring the matrix in each loop iteration.

We should really add the matrix call.  It wouldn't be hard,
but we have quite a few of these convenience additions stacked
up, so I don't know when we'll get to it.

- Bob

mike-lawrence:
For your suggestion of creating a copy into a local variable that is an array of vectors, would this be what you're thinking?
  {
    vector[nX] temp[nS] ;
    for(ns in 1:nS){
      temp[ns] = Smu[ns]' ;
    }
    temp ~ multi_normal(mu,Z) ;
  }

bob-carpenter:
Exactly.

Though if efficiency is a concern, it might be
better to do the transpose, then copy the columns,
because Eigen presumably does the transposes more
efficiently using blocking rather than going one row/column
at a time, which is a big memory locality hit (memory locality
and branch misprediction are the main causes of inefficienty
for code on modern CPUs.  Stan instantiates Eigen in the default
way so as to store the matrices in column-major order, so
that Smu.

But I'm not sure if it'll matter for a simple transpose.

{
  matrix[nX, nS] Smu_tr;
  vector[nX] Smu_tr_arr[nS];

  Smu_tr = Smu';
  for (n in 1:nS) Smu_tr_arr[n] = Smu_tr[ , n];
  Smu_tr_arr ~ multi_normal(mu, Z);
}

Probably not worth the trouble and lack of clarity, even if
it is a bit faster. 

- Bob

P.S.  Never space before semicolons, always space after commas.
      I'm finding our users aren't used to the C/C++/Java conventions,
      and their code is therefore very hard to scan for me because
      the syntax is such a distraction.

      Also notice that I tried to use meaningful variable names rather
      than "temp".  I figure now that I'm heading down this road, I
      might as well go all the way in syntax/formatting instructions.

      Also, always indent (we recommend two spaces as it's the convention
      in C and C++, but some users like four, which is the convention for Java).

syclik:
@bob-carpenter, the test failure is fine. Feel free to merge when you think it's ready. 

The built manual isn't persistent on Jenkins, so I think you'll have to build it locally to see if the manual looks right.

PR #2026, opened 2016-08-23
update error msg for , instead of | in prob funs, fixes #1940

### LINKED ISSUE #1940: Parser errors/warnings for new syntax are confusing
#### Summary:
When playing around with the new syntax I found that the parser errors and warnings were not very informative. Well, they were informative enough for me because I already know the correct syntax and was intentionally trying to trigger the error messages, but I imagine they would pretty opaque to users. Example in **Reproducible Steps**, below.

#### Description:
See below in **Reproducible Steps**

* [x] fix error message for missing vertical bar

The duplication problem is a duplicate isseu with #1946, where there are more cases to fix.  

#### Reproducible Steps:
First put the following in `logistic.stan`
```stan,
data {
  int<lower=0> N;
  int<lower=0,upper=1> y[N];
}
parameters {
  real theta;
}
model {
  theta ~ normal(0, 1);
  y ~ bernoulli_logit(theta);
}
generated quantities {
  vector[N] log_lik;
  for (n in 1:N)
    log_lik[n] = bernoulli_logit_log(y[n], theta);
}
```
then run
```r,
library("rstan")
stanc(file = "logistic.stan")
```
This gives the warning: 
```
DIAGNOSTIC(S) FROM PARSER:
Warning: Deprecated function 'bernoulli_logit_log'; please replace suffix '_log' with '_lpdf' for density functions or '_lpmf' for mass functions
Warning: Deprecated function 'bernoulli_logit_log'; please replace suffix '_log' with '_lpdf' for density functions or '_lpmf' for mass functions
```
which is good (not one of the ones I found confusing), but why does it appear twice if 
`bernoulli_logit_log` is only used once in the Stan program? Is it because the sampling
statement `y ~ bernoulli_logit(theta)` is also used? Either way it's kind of weird.

The stuff I found more confusing comes next, when trying to follow the suggestion in the warning.  First I modify the `bernoulli_logit_log` line in generated quantities:
```stan,
log_lik[n] = bernoulli_logit_lpdf(y[n], theta); \\ I know this is incorrect
```
and the error is 
```
SYNTAX ERROR, MESSAGE(S) FROM PARSER:
ERROR at line 15

 13:      vector[N] log_lik;
 14:      for (n in 1:N)
 15:        log_lik[n] = bernoulli_logit_lpdf(y[n], theta);
                                             ^
 16:    }

PARSER EXPECTED: <probability function arguments>
Error in stanc(model_code = paste(program, collapse = "\n"), model_name = model_cppname,  : 
  failed to parse Stan model 'logistic' due to the above error.
```
which just says that it expected `<probability function arguments>`. I assume that's trying to tell me that I forgot the vertical bar but that's not what the message says. Would it be possible to mention the vertical bar in the message? In addition, I used a function that doesn't exist (`bernoulli_logit_lpdf` even though it's a mass function) but didn't get any error about that. So it seems to be giving me an error about using the wrong arguments to a function that doesn't exist.

So next I use the vertical bar:

```stan,
log_lik[n] = bernoulli_logit_lpdf(y[n] | theta);
```
and that gets rid of the mysterious `<probability function arguments>` and now I get the error about `bernoulli_logit_lpdf` not existing:
```
SYNTAX ERROR, MESSAGE(S) FROM PARSER:

No matches for: 

  bernoulli_logit_lpdf(int, real)

Function bernoulli_logit_lpdf not found.

ERROR at line 15

 13:      vector[N] log_lik;
 14:      for (n in 1:N)
 15:        log_lik[n] = bernoulli_logit_lpdf(y[n] | theta);
                                                           ^
 16:    }

Error in stanc(model_code = paste(program, collapse = "\n"), model_name = model_cppname,  : 
  failed to parse Stan model 'logistic' due to the above error.
```
except it doesn't tell me that I just need to use `lpmf` instead of `lpdf`. Would it be possible to detect that that someone is using a function that just has an incorrect suffix?

#### Current Output:
The current output is given above in **Reproducible Steps**

#### Expected Output:
I think it would be helpful if:

* Errors about functions not existing were thrown before errors about incorrect arguments
* Errors about missing vertical bars said something about missing vertical bars instead of 
just `<probability function arguments>`
* Only one warning is issued if deprecated function is only once in a Stan program
* Using `lpdf` instead of `lpmf` or vice-versa resulted in a more informative message
indicating that the correct function just has a different suffix 

Are there reasons I'm overlook why any of these would be difficult?
 
#### Current Version:
v2.10.0

bob-carpenter:
#### Submisison Checklist

- [x] Run (parser) unit tests: `./runTests.py src/test/unit`

#### Summary

Clarify error messages for probability functions.

#### Intended Effect

Make it easier for users to understand error messages.

#### How to Verify

New model unit tests for new message.

#### Side Effects

No.

#### Documentation

Not necessary.

#### Reviewer Suggestions

Anyone.

#### Copyright and Licensing

Columbia University

### DIFF
--- a/src/stan/lang/grammars/term_grammar_def.hpp
+++ b/src/stan/lang/grammars/term_grammar_def.hpp
@@ -231,7 +231,7 @@ namespace stan {
         %= lexeme[char_("a-zA-Z")
                   >> *char_("a-zA-Z0-9_.")];
 
-      prob_args_r.name("probability function arguments");
+      prob_args_r.name("probability function argument");
       prob_args_r
         %= (lit('(') >> lit(')'))
         | hold[lit('(')
@@ -239,7 +239,9 @@ namespace stan {
                >> lit(')')]
         | (lit('(')
            >> expression_g(_r1)
-           >> lit('|')
+           >> (lit(',')
+               [require_vbar_f(_pass, boost::phoenix::ref(error_msgs_))]
+               | (eps > lit('|')))
            >> (expression_g(_r1) % ',')
            >> lit(')'));
 
--- a/src/stan/lang/grammars/semantic_actions.hpp
+++ b/src/stan/lang/grammars/semantic_actions.hpp
@@ -659,6 +659,11 @@ namespace stan {
     };
     extern boost::phoenix::function<set_var_type> set_var_type_f;
 
+    struct require_vbar : public phoenix_functor_binary {
+      void operator()(bool& pass, std::ostream& error_msgs) const;
+    };
+    extern boost::phoenix::function<require_vbar> require_vbar_f;
+
     struct validate_no_constraints_vis : public boost::static_visitor<bool> {
       std::stringstream& error_msgs_;
       explicit validate_no_constraints_vis(std::stringstream& error_msgs);
--- a/src/stan/lang/grammars/semantic_actions_def.cpp
+++ b/src/stan/lang/grammars/semantic_actions_def.cpp
@@ -1912,6 +1912,17 @@ namespace stan {
     }
     boost::phoenix::function<set_var_type> set_var_type_f;
 
+    void require_vbar::operator()(bool& pass, std::ostream& error_msgs) const {
+      pass = false;
+      error_msgs << "Probabilty functions with suffixes _lpdf, _lpmf,"
+                 << " _lcdf, and _lccdf," << std::endl
+                 << "require a vertical bar (|) between the first two"
+                 << " arguments." << std::endl;
+    }
+    boost::phoenix::function<require_vbar> require_vbar_f;
+
+
+
     validate_no_constraints_vis::validate_no_constraints_vis(
                                                std::stringstream& error_msgs)
       : error_msgs_(error_msgs) { }
--- a/src/test/test-models/bad/prob-fun-args-no-bar.stan
+++ b/src/test/test-models/bad/prob-fun-args-no-bar.stan
@@ -1,0 +1,6 @@
+parameters {
+  real y;
+}
+model {
+  target += normal_lpdf(y, 0, 1);
+}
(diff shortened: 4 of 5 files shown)


jgabry:
I didn't review the C++ code (not my specialty) but I took a look at the
new error messages in the code and they look good to me!

bob-carpenter:
Could someone review the actual code?  It's a minor change.

src/stan/lang/grammars/term_grammar_def.hpp
```diff
@@ -231,15 +231,17 @@ namespace stan {
...
+      prob_args_r.name("probability function argument");
       prob_args_r
         %= (lit('(') >> lit(')'))
         | hold[lit('(')
                >> expression_g(_r1)
                >> lit(')')]
         | (lit('(')
            >> expression_g(_r1)
-           >> lit('|')
+           >> (lit(',')
```
sakrejda: @bob-carpenter I'm reading this as triggering the new error message function if you run into a literal ',' after `expression_g` consumes the first function argument or continuing if it gets a literal '|'?

sakrejda:
I looked at the code, there's one line comment I left where the code is hard to understand b/c I'm not familiar with the language processing other than having skimmed the doc a few times but the test confirms what I think is going on so if it passes I think it's good.

src/stan/lang/grammars/term_grammar_def.hpp
bob-carpenter: Correct.  The `>>` is for sequencing and allows backtracking through it.  The `>` is for sequencing, but cuts (in the Prolog sense) backtracking.  The `require_vbar_f` just prints a message that you want the `|` instead of the `,` and then sets `_pass = false` so that the rule match fails.  The Qi non-terminal `eps` matches the empty sequence, so `eps > lit('|')` says that you just require the literal `|` and can't backtrack if you don't get it.

PR #2048, opened 2016-09-01
Fixes #2045. Making includes to the math library consistent

### LINKED ISSUE #2045: Clean up includes to the math library
#### Summary:
The includes to the math library should be limited to `stan/math/{prim, rev, fwd, mix}/{scal, arr, mat}.hpp` from the Stan library. 

#### Description:
Including other headers may cause problems with instantiating traits in the correct order. 

#### Additional Information:
stan-dev/math#375 depends on this issue. 

#### Current Version:
v2.11.0

syclik:
#### Submisison Checklist

#### Summary
I replaced the individual includes to the math library with a single include.

#### Intended Effect
This will ensure the traits get instantiated correctly.

#### How to Verify
Look at code.

#### Side Effects
This might include a few more files, but from each of our interfaces, this shouldn't have an effect.

#### Documentation
None necessary.

#### Reviewer Suggestions
Anyone.

#### Copyright and Licensing

Columbia University

### DIFF
--- a/src/stan/io/dump.hpp
+++ b/src/stan/io/dump.hpp
@@ -1,19 +1,15 @@
 #ifndef STAN_IO_DUMP_HPP
 #define STAN_IO_DUMP_HPP
 
+#include <stan/io/validate_zero_buf.hpp>
+#include <stan/io/var_context.hpp>
+#include <stan/math/prim/mat.hpp>
 #include <boost/lexical_cast.hpp>
 #include <boost/throw_exception.hpp>
 #include <boost/type_traits/is_floating_point.hpp>
 #include <boost/type_traits/is_integral.hpp>
 #include <boost/type_traits/is_arithmetic.hpp>
 #include <boost/utility/enable_if.hpp>
-
-#include <stan/io/validate_zero_buf.hpp>
-#include <stan/io/var_context.hpp>
-#include <stan/math/prim/scal/meta/index_type.hpp>
-#include <stan/math/prim/arr/meta/index_type.hpp>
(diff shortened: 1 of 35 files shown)


syclik:
Jenkins, retest this please.

bob-carpenter:
Don't we want to go in exactly the opposite direction to have files include only what they use?  The problem I'm finding with these include-the-world things is that they cause the unit tests to all need to be recompiled.  And they're going to slow down compilation times (though I don't know by how much).  

The other thing we really want to do is avoid circular includes.  But it looks like you ordered things here so that you only included top-level math headers outside of math (obviously, as this is in stan-dev/stan, not in stan-dev/math).

bob-carpenter:
Also, by the same logic as the `is_inf` refactor, shouldn't we be including one of our headers rather than `<Eigen/Dense>` so we can have better control of what gets included (particularly adding more if we need to, like our own traits files).

bob-carpenter:
Otherwise, though, this looks OK.

bob-carpenter:
So merge it if you think this is the right way to go.

syclik:
For includes to the math library, we should include the most restrictive thing we can include, which I think I do for these changes. We should be including those in order to keep the traits instantiated in the correct order.

Regarding the changes to include Eigen directly: I did that for those files because they don't use the math library at all. They just use Eigen. Nothing from the `stan::math` namespace is used, so I just included the Eigen header directly.

bob-carpenter:
OK, we can do that for now and if it's only double instantiations, that's all we should need going forward.

PR #2050, opened 2016-09-01
Compile models are fixed. They will now try to instantiate models with the types we need.

syclik:
#### Submisison Checklist

#### Summary
The integration tests were broken before. We had thought that adding `-fsyntax-only` was sufficient for code to compile, but we found out that it didn't instantiate enough.

This pull request will now instantiate the Stan program with the correct template instantiations.

This currently passes on develop now, but should fail:
```
make src/test/test-models/good/validate_conditional_op_good.hpp-test
```
or
```
./runTests.py src/test/integration/compile_models_test.cpp 
```

With this pull request, it fails now. We'll need to make sure everything else passes.

#### Intended Effect
Makes sure our Stan programs are compilable in C++.

#### How to Verify
Run this for one file:
```
make src/test/test-models/good/validate_conditional_op_good.hpp-test
```

or to run all files:
```
./runTests.py src/test/integration/compile_models_test.cpp 
```

Here's what's used as a test file just to make sure we're instantiating enough (generated in `test/test-model-main.cpp` by make):
```c++
nt main() {
  stan::io::var_context *data = NULL;
  stan_model model(*data);

  std::vector<stan::math::var> params;
  std::vector<double> x_r;
  std::vector<int> x_i;

  model.log_prob<false, false>(params, x_i);
  model.log_prob<false, true>(params, x_i);
  model.log_prob<true, false>(params, x_i);
  model.log_prob<true, true>(params, x_i);
  model.log_prob<false, false>(x_r, x_i);
  model.log_prob<false, true>(x_r, x_i);
  model.log_prob<true, false>(x_r, x_i);
  model.log_prob<true, true>(x_r, x_i);

  return 0;
}
```

#### Side Effects
This might break some things we thought were ok. We'll need to fix them.

#### Documentation
I added a few notes to the makefile.

#### Reviewer Suggestions
Anyone.

#### Copyright and Licensing

Columbia University

### DIFF
--- a/make/tests
+++ b/make/tests
@@ -81,10 +81,40 @@ endif
 ##
 # Target to verify code generated Stan programs
 # (.stan -> .hpp) are compilable
+#
+# Running:
+# > make test/integration/compile_models
+# will make sure everything in src/test/test-models/good/
+# compiles
 ##
 .PHONY: %.hpp-test
-src/test/test-models/good/%.hpp-test : src/test/test-models/good/%.hpp
-	$(COMPILE.c) -fsyntax-only -O0 $<
+src/test/test-models/good/%.hpp-test : src/test/test-models/good/%.hpp test/test-model-main.cpp
+	$(COMPILE.c) -O0 -include $< -o $(DEV_NULL) test/test-model-main.cpp
+
+test/test-model-main.cpp:
+	@mkdir -p test
+	@touch $@
+	@echo "int main() {" >> $@
+	@echo "  stan::io::var_context *data = NULL;" >> $@
+	@echo "  stan_model model(*data);" >> $@
+	@echo >> $@
+	@echo "  std::vector<stan::math::var> params;" >> $@
+	@echo "  std::vector<double> x_r;" >> $@
+	@echo "  std::vector<int> x_i;" >> $@
+	@echo >> $@
+	@echo "  model.log_prob<false, false>(params, x_i);" >> $@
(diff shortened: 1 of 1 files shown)


syclik:
@bob-carpenter, the failure is right: it fails on the tenary operator.

mitzimorris:
Code changes from branch bugfix/issue-2044-ternary-op-result-for-mismatched-types
will pick up version of stan translater that generate a model.hpp file that compiles.
The console log errors report that build failed on the following line of c++ code:

```
 stan::math::assign(tpy, (logical_lt(y,100) ? double(x) : double(py) ));
```

for the following statement in the transformed parameters block of the model:

```
  tpy = y < 100 ? y : py;   // t,f : real,var
```

where tpy is a transformed parameter, y is type real, and py is type var.

In branch bugfix/issue-2044-ternary-op-result-for-mismatched-types, the above statement
is translated as:

```
stan::math::assign(tpy, (logical_lt(y,100) ? stan::math::promote_scalar<T__>(y) : stan::math::promote_scalar<T__>(py) ));
```

bob-carpenter:
I don't really understand the makefile, but this is going to fail the integration tests at least until the ternary op gets in.  That's the correct behavior for this particular commit, though, so you can just merge it as is if you want.

syclik:
Woo hoo! The other models we have don't fail to compile! That's a relief.

bob-carpenter:
Indeed.  Thanks for getting this working.  I was really
relying on it to catch errors.

- Bob

PR #2052, opened 2016-09-03
Feature/issue 2015 truncation

syclik:
#### Submisison Checklist

#### Summary
This will prevent built in distributions without the appropriate cdf functions to be used in truncation.

#### Intended Effect
I took a pass and checked that the function to be used for truncation isn't the same as the distribution function.

#### How to Verify
Run the unit tests. There's an additional unit test. To run that one:
```
./runTests.py src/test/unit/lang/parser/trunc_test.cpp
```

#### Side Effects
This will stop some existing models from being compiled. They shouldn't have been compiled in the first place.

#### Documentation
None.

#### Reviewer Suggestions
@bob-carpenter. There might be a better fix for this. I took the shortest path I could see.

#### Copyright and Licensing

Columbia University.

### DIFF
--- a/src/stan/lang/grammars/semantic_actions_def.cpp
+++ b/src/stan/lang/grammars/semantic_actions_def.cpp
@@ -1106,7 +1106,8 @@ namespace stan {
         std::vector<expr_type> arg_types_trunc(arg_types);
         arg_types_trunc[0] = s.truncation_.low_.expression_type();
         std::string function_name_ccdf = get_ccdf(s.dist_.family_);
-        if (!is_double_return(function_name_ccdf, arg_types_trunc,
+        if (function_name_ccdf == s.dist_.family_
+            || !is_double_return(function_name_ccdf, arg_types_trunc,
                               error_msgs)) {
           error_msgs << "lower truncation not defined for specified"
                      << " arguments to "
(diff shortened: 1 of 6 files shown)


bob-carpenter:
Looks OK, but I think it's missing tests for some of the functionality branches.  And it hasn't addressed Ben's questions, which I haven't really sorted through.  So let me take a look before merging.  Of course, these tests will then fail when we add appropriate lcdfs and lccdfs, but that's OK.  Better safe than sorry here in testing land.

syclik:
Yes. You're right about the tests. I was lazy and only added one when I
should have added multiple tests.

bob-carpenter:
I can add the rest of the tests.  I'm going through Ben's dev list comments now to make sure we're not doing something off-by-one on the generation.

bob-carpenter:
I added the remaining tests.  As Ben pointed out in email, I think we have a further issue with our CDFs vs. CCDFs or how we're using them in generated code for truncation.  I'm opening a new issue for that.  But we shouldn't hold up the release for that, but we may need to warn the users.

syclik:
Thanks, Bob. 

I agree regarding the other issue, but I think this "fixes" the original issue. The original model shouldn't compile. The truncation needs to be done by hand until we implement the cdf functions.

PR #2069, opened 2016-09-12
Vectorizing exp in Stan

rayleigh:
#### Submisison Checklist

#### Summary
This pull request is to vectorize exp in Stan by using the vectorization done in stan-math.

#### Intended Effect
Users can use the function exp on a Stan data type up to 8 dimensions.

#### How to Verify
Run make src/test/test-models/good/function-signatures/math/matrix/exp.hpp-test.

#### Side Effects
An add_unary_vectorized function is added to ast.hpp and ast_def.cpp. This function can be used to push other unary vectorized functions in stan-math up to Stan.
 
#### Documentation
Will need to be done to inform users.

#### Copyright and Licensing

Per University of Michigan's copy right policy (http://spg.umich.edu/policy/601.28 Section C), Rayleigh Lei is the copyright holder.


syclik:
@rayleigh, can you check with U Mich that you own copyright or that they're willing to license it open-source? And instead of "This is my work.", would you mind putting down your name instead?

It's missing the doc in the manual. Anyone can fill it in if they have time. This PR won't get in until the doc is in place. (This PR is user facing and the doc should come with the PR.)

src/stan/lang/ast_def.cpp
```diff
@@ -284,6 +284,16 @@ namespace stan {
     void function_signatures::add_unary(const::std::string& name) {
       add(name, DOUBLE_T, DOUBLE_T);
     }
+    void function_signatures::add_unary_vectorized(const::std::string&
+    name) {
+      for (size_t i = 0; i < 8; ++i) {
+        add(name, expr_type(DOUBLE_T, i), expr_type(INT_T, i));
```
syclik: this will promote all the function outputs for `int` inputs. Are there any cases where the output is `int` with an `int` input? It seems like there would be.

syclik:
Btw, the code looks great! It's ready to go once the doc gets in and we clear up the IP.

syclik:
@rayleigh, looks like you own copyright as long as you're a student. See: http://spg.umich.edu/policy/601.28 Section C.

Mind updating the original text?

bob-carpenter:
I guess I was too slow off the mark.  I haven't even started
on the stan-dev/stan side, so Rayleigh can go ahead and add
all the vectorizations that exist to the language and we'll catch
up on the other ones later when I get the math lib done.

Which signatures got declared for exp()?  I don't recall
settling on a depth that it'll work to.

Are we going to do these one function at a time?  I guess it's
good to start with one.

We haven't been grilling everyone on their various university's
willingness to open source.  So we should either do this for
everyone on the dev list, everyone who contributes anything, or
just trust people to say who owns the copyright.

- Bob

src/stan/lang/ast_def.cpp
bob-carpenter: No
> On Sep 12, 2016, at 10:10 AM, Daniel Lee <notifications@github.com> wrote:
> 
> In src/stan/lang/ast_def.cpp:
> 
> > @@ -284,6 +284,16 @@ namespace stan {
> >      void function_signatures::add_unary(const::std::string& name) {
> >        add(name, DOUBLE_T, DOUBLE_T);
> >      }
> > +    void function_signatures::add_unary_vectorized(const::std::string&
> > +    name) {
> > +      for (size_t i = 0; i < 8; ++i) {
> > +        add(name, expr_type(DOUBLE_T, i), expr_type(INT_T, i));
> 
> this will promote all the function outputs for int inputs. Are there any cases where the output is int with an int input? It seems like there would be.

No.  Those'll be separate functions.    Only the double output
versions were vectorized.

With depth up to 7, that'll be 40 signatures per function.  The
parser should be able to handle that.

- Bob

bob-carpenter:
This needs doc in the manual, which means inventing some kind of
new notation for unary vectorizations, like:

   T exp(T x);
   T[...] exp(T[...] x);

with an indication that T can be any basic type and the arrays
can be up to 8 elements.  I think it's worth having the basic and
the array version as otherwise users are going to get confused.
I don't think we can have all 40 signatures listed explicitly!

src/stan/lang/ast_def.cpp
rayleigh: In the unary vectorize code in `math`, the function to deal with the base case of int returns double: 
(https://github.com/stan-dev/math/blob/1f2794953e5199717afd87bf40dbbe1f7165ce20/stan/math/prim/mat/vectorize/apply_scalar_unary.hpp#L101-L120). So, if the function is using the unary vectorize code without any modification, the return type for any int types will be double.

rayleigh:
-Mind updating the original text?
I updated the original text. Does that work? I haven't written copyright information before.

-Which signatures got declared for exp()?  I don't recall settling on a depth that it'll work to.
In `stan/math`, we only test the base type and the base type nested in a std::vector. However, the `stan/math` code should work for any depth. I chose 8 because I was using `dims` and `size` in `function_signatures.h` as models and both have 8 as the max depth.

-Are we going to do these one function at a time?  I guess it's good to start with one.
I only did one in this pull request because of what happened with the `stan/math` vectorize-all function pull request. I was hoping that if everything looked good here for exp(), the next pull request would vectorize the functions that are currently vectorized in `stan/math` because I could use `add_unary_vectorization` to vectorize them in Stan and a bash script to copy the `exp` file to create the appropriate `test-models` files.

-  T exp(T x);
   T[...] exp(T[...] x);
Based on the comment above, should we make a note that if T is int, the return type will be double?

bob-carpenter:
Depth 8 should be OK.  

And yes, good point about noting the result type is double.  Don't know how to indicate that but probably not `T exp(T x)`.  Should we have a separate `real[...] exp(int[...])` case?  I just don't want the manual to get junked up with a ton of repetetitive boilerplate that'll distract users.

rayleigh:
I looked at the doc. Could we list `T exp(T x)` in the Index, but underneath, put `Real-Valued Basic Functions`, `Array Operations`, and `Matrix Operations` and page links to the descriptions in these sections? Then, we would explain in the section linked to under `Array Operations`. Also, where should we put the section for `Elementwise Functions` in the `Array Operations` chapter in the doc?

bob-carpenter:
Let me take a shot at writing up the doc.  It'll need some LaTeX macros if we don't want to be duplicating this for 50 functions.

bob-carpenter:
Looks good.  I added general doc macros and a section on how vectorization works.  OK to merge when tests pass (which they should, because latest was only doc).

Issue #1978, opened 2016-07-26
selective suppression increment_log_prob warning


sdewaele:
#### Summary:

Introduce a tag in the comments of a line to selectively suppress the parser warning to call increment_log_prob.
#### Description:

The Stan parser issues a the following warning if the left-hand side of a sampling statement may contain a non-linear transform of a parameter or local variable. After performing this check, the programmer should be able add a tag to that line to suppress the warning in the future. This reduces unnecessary parser output, allowing the programmer to pay more attention to remaining, vaild warnings. Also, in this way a perfect program compiles without errors, which is also desirable.

For example, the syntax for this tag can be similar to [Matlab's syntax for warning suppression](http://www.mathworks.com/help/matlab/matlab_prog/check-code-for-errors-and-warnings.html#brqxeeu-169).
#### Reproducible Steps:

The parser generates the warning during compilation of the following script:

```
data {
    real mu;
    real sigma;
}

parameters {
    real a;
    real b;
}

transformed parameters {
    real s;
    real bsq;
    s <- a + b;
    bsq <- b^2; 
}

model {
    s ~ normal(mu, sigma); # %ok<NOJACOBIAN>
    bsq ~ lognormal(3, 2);
}
```
#### Current Output:

```
DIAGNOSTIC(S) FROM PARSER:
Warning (non-fatal):
Left-hand side of sampling statement (~) may contain a non-linear transform of a parameter or local variable.
If so, you need to call increment_log_prob() with the log absolute determinant of the Jacobian of the transform.
Left-hand-side of sampling statement:
    s ~ normal(...)
Warning (non-fatal):
Left-hand side of sampling statement (~) may contain a non-linear transform of a parameter or local variable.
If so, you need to call increment_log_prob() with the log absolute determinant of the Jacobian of the transform.
Left-hand-side of sampling statement:
    bsq ~ lognormal(...)
```
#### Expected Output:

The first warning is suppressed because the user has checked the line and added the tag `%ok<NOJACOBIAN>` in the comments of the indicated line:

```
DIAGNOSTIC(S) FROM PARSER:
Warning (non-fatal):
Left-hand side of sampling statement (~) may contain a non-linear transform of a parameter or local variable.
If so, you need to call increment_log_prob() with the log absolute determinant of the Jacobian of the transform.
Left-hand-side of sampling statement:
    bsq ~ lognormal(...)
```
#### Current Version:

v2.9.0

syclik:
[...]

Issue #1672, opened 2015-11-07
Feature request: suppress specific change-of-variable warnings


ksvanhorn:
My habit as a developer has always been to address all compiler warnings, either by changing the code, if that is not an acceptable option, issuing a directive to suppress the warning. My reason is that when you get used to seeing warnings, you start to ignore them, and you run the risk of ignoring warnings that indicate an actual problem.

With Stan, it is currently impossible to avoid warning messages with any model that involves a change of variable (transformed parameter that is given a distribution), even if the model is entirely correct. From a software engineering perspective, it would be highly desirable to have some way of suppressing unnecessary warning messages. The obvious way to do this would be some sort of compiler directive that says "I have included the necessary Jacobian adjustment for this transformed parameter." It would be better if there were some way for the compiler to verify this, but that appears to be a rather difficult task, except for some simple cases.

bob-carpenter:
You make a very good point.  I have the same habit, and you'll
see we do the same thing with the Stan code itself (though
suppressing all the warnings from Doxygen, cpplint, and the
various g++ and clang++ compiler versions starts to be a real
drag on the code itself).

I do think it's a good idea to have a way to suppress warnings
at a statement level.  I think it's going to require creating some
way to annotate statements generally and then passing flags into
the parser that can be stored and accessed locally in semantic actions
on the parser rules.

It's not only difficult, but impossible in general given Stan's
Turing completeness to do a compile-time verification of whether
a Jacobian is needed or not.

- Bob

tpapp:
Could we have simple nostrum like 
```stan
nowarning { // code below works fine
  ...
}
```

bob-carpenter:
I think that's a good idea.  I just don't know how to implement it easily.  Although the syntax is simple, the way the streams are plumbed through, this would require a lot of conditional logic in the parser semantic actions. 

We'll be moving to a logger-type interface, where we'll be able to control the level of output, but not individual blocks.

tpapp:
I was looking at the source, and perhaps I don't understand most of it (not having programmed in C++ for more than 10 years), but I thought of the following: add a `warnings` or similar slot to the AST representation of expressions, with a default that shows all warnings, and have
```stan
nowarning {
  expression1
}
```
set it to a value that suppresses it when parsing `expression1`. Then `validate_sample::operator` could check whether it is set and emit the warning only when not suppressed.

[...]

Issue #512, opened 2014-01-23
break info/warning/error/debug messages into log levels 


bob-carpenter:
(Originally suggested by John Hall on stan-users;  my words in the issue.)

Given the number of users we've confused with the rejection warnings, it would be nice to have levels of reporting like a typical logger.  I'm most familiar with log4j, which has messages of increasing verbosity, which are roughly  error, warning, info, debug (see <a href="http://logging.apache.org/log4j/1.2/apidocs/org/apache/log4j/lf5/LogLevel.html">LogLevel javadoc</a>).  

This would allow us to turn off the warnings most of the time and only turn them on if a model's having problems.

syclik:
I'll start a wiki on what I think the design should be. I think @sakrejda and I already started one, so maybe I'll dig that up.

This depends on #1751, but once we're past that, we can start moving on this.

PR #2075, opened 2016-10-03
Feature/issue 1951 compound declaration definition

mitzimorris:
#### Submisison Checklist

#### Summary

Adds compound declaration/definition statements such as `real x = 1.0` to Stan language.

#### Intended Effect

For any variable that is assignable, allow assignment to be combined with the variable definition.
Assignable variable are local variables and variables declared in the transformed data, transformed parameters, and generated quantities blocks.

#### How to Verify

To verify that the parser correctly recognizes valid compound declaration/definition statements, run unit test `stan/src/test/unit/lang/parser/var_decls_grammar_test.cpp`.

To verify that the generated code compiles correctly, make the test models in `src/test/test-models/good/declare-define-var*.stan`,  e.g.:
```
for x in src/test/test-models/good/declare-define-*.stan
do
echo ${x%.stan}
make ${x%.stan}.hpp-test
done
```

#### Side Effects

The compiler cannot check that the dimension sizes match.  See example model: `src/test/test-models/good/declare-define-var-vec-runtime-error.stan`

#### Documentation

Basic syntax described in Stan manual.  More documentation is needed.  Future planned work is to update the example models to use this feature.

#### Reviewer Suggestions

Daniel Lee

#### Copyright and Licensing

Mitzi Morris


src/test/test-models/good/declare-define-var-double.stan
```diff
@@ -0,0 +1,67 @@
...
+  real tp4[5];
+  real tp5[5] = tp4;
+  tp3 = tp1 + tp2;
+
+  // local variables
+  {
+    real lp1 = 1;      // real_p <- int_d 
+    real lp2 = 2.0;    // real_p <- real_d
+    real lp3 = tp2;      // real_p <- real_p
+      real lp4[6];
```
syclik: fix indenting

bgoodri:
I still think the parser should warn if the user declares without defining on the same line in all blocks except data and parameters.

src/docs/stan-reference/language.tex
```diff
@@ -1267,6 +1267,84 @@ \subsection{Type Naming Notation}
...
+even though \code{0} is of type \code{int} and \code{sum} is of type
+\code{real}, because integer-typed scalar expressions can be assigned
+to real-valued scalar variables.  In all other cases, the type of the
+expression on the right-hand side of the assignment must be identical
+to the type of the variable being declared.
+
+Any type may be assigned.  For example,
+%
+\begin{stancode}
+real matrix[3, 2] a = b;
```
mitzimorris: changes made and pushed

src/docs/stan-reference/language.tex
```diff
@@ -1267,6 +1267,84 @@ \subsection{Type Naming Notation}
...
+their static typing and cannot be validated until run time.
+
+
+\subsection{Right-Hand Side Expressions}
+
+The right-hand side may be any expression which has a type which is
+assignable to the variable being declared.  For example,
+%
+\begin{stancode}
+real matrix[3, 2] a = 0.5 * (b + c);
```
mitzimorris: changes made and pushed

src/test/test-models/good/declare-define-var-double.stan
mitzimorris: changes made and pushed

mitzimorris:
would prefer to create new issue for adding parser warning for variables declarations with missing definitions in blocks which allow compound declaration/definition statements.

syclik:
+1.

bob-carpenter:
Because the only way to define a variable is through assignment
(e.g., no pass by reference), then it's relatively easy to
figure out if a variable's being used before it's defined, which
is really where we want the warning to go.  It'd be easy to put
one in a declaration without a definition, but it might be confusing
to users who've actually defined the variable.

bgoodri:
One of the more common problems with users' Stan programs is

model {
  real z;
  z ~ normal(0, 1);
}

so nothing actually gets assigned to z. I presumed that since we didn't
already have a parser warning for that case, it was difficult to determine
whether to throw it. Warning whenever z is declared but not defined would
fix that, although it would be somewhat overkill for

model {
  real z;
  z = 0;
}

but there is nothing particularly to be said in favor of writing a model
block like that, other than it has always been mandatory.

bob-carpenter:
> On Oct 4, 2016, at 10:47 PM, bgoodri <notifications@github.com> wrote:
> 
> One of the more common problems with users' Stan programs is
> 
> model {
> real z;
> z ~ normal(0, 1);
> }

Agreed that this should produce a warning.  And I think the
warning should point to where the variable is used before being
defined, that is in the ~ statement. 

Harder to detect case:

  real z[10];

  for (n in 1:9)
    z[n] = ... something the user couldn't vectorize ... 

Can't do this statically.

> so nothing actually gets assigned to z. I presumed that since we didn't
> already have a parser warning for that case, it was difficult to determine
> whether to throw it. Warning whenever z is declared but not defined would
> fix that, although it would be somewhat overkill for
> 
> model {
> real z;
> z = 0;
> }

This one should not produce a warning.

I'd be OK producing warnings in cases where you have a declaration without
a definition if we had a way to turn those warnings off.  That's #1978 and
#1672 and is related to #512 and is also related to the "pedantic" mode
proposal:  https://github.com/stan-dev/stan/wiki/Stan-Parser-Pedantic-Mode

- Bob

Issue #410, opened 2013-11-22
test remaining src/stan/prob/internal_math.hpp functions


syclik:
Issue #333 shows the `gradregIncGamma` function getting into an infinite loop. The additional functions need to be tested:
- [ ] `F32`
- [ ] `gradF32`
- [ ] `grad2F1`
- [ ] `gradRegIncBeta`

syclik:
This issue was moved to stan-dev/math#27

PR #2083, opened 2016-10-14
Vectorize in Stan vectorized functions in math

rayleigh:
#### Submisison Checklist

#### Summary
This is to vectorize in Stan the functions already vectorized in `stan/math/` (see [here](https://github.com/stan-dev/math/pull/240) and [here](https://github.com/stan-dev/math/pull/276)).

#### Intended Effect
For a vectorized function, users can now use that function element-wise on a vector or matrix without writing a for loop. For instance, if `a` is a matrix, users can write `log(a)` to apply `log` element-wise on `a`.

#### How to Verify
It can be verified by applying a vectorized function on a vector or matrix and checking that the function was applied to each element. This pull request also adds function signature tests for these functions in Stan and there are tests in `stan/math`. 

#### Side Effects
Before, `int abs` would have a return type of `int`. Now it'll have a return type of `real`.

#### Documentation
`functions.tex` has been updated to include the vectorized functions. The macro used was also updated to be more flexible.

#### Copyright and Licensing

Per Michigan's copyright policies, Rayleigh Lei is the copyright holder.

### DIFF
--- a/src/docs/stan-reference/functions.tex
+++ b/src/docs/stan-reference/functions.tex
@@ -147,8 +147,7 @@ This is a no-op.
 
 \begin{description}
 %
-\fitem{int}{abs}{int \farg{x}}{
-  Returns the absolute value of \farg{x}. 
+  \fitemUnaryVec{abs}{absolute value of \farg{x}}{
   \[\mbox{abs}(x) = |x|\]}
 %
 \fitemnobody{int}{int\_step}{int \farg{x}}
@@ -681,10 +680,7 @@ as \code{x~+~y} and \code{-x}, respectively.
 
 \begin{description}
 %
-  \fitem{real}{abs}{real \farg{x}}{     
-    Returns the absolute value of \farg{x}\,;
-    this function is deprecated and will be removed in the future;
-    please use \code{fabs} instead.
+  \fitemUnaryVec{abs}{absolute value of \farg{x}\,; this function is deprecated and will be removed in the future; please use \code{fabs} instead}{
     \[
     \mbox{\code{abs}}(x) = |x|
     \]}
@@ -771,25 +767,17 @@ of these in the future, but as of now, there is no good workaround.
 
 \begin{description}
   %
-  \fitem{real}{floor}{real \farg{x}}{
-    Returns the floor of \farg{x}, which is the largest integer less
-    than or equal to \farg{x}, converted to a real value;
-    see warning at start of \refsection{step-functions}.
+  \fitemUnaryVec{floor}{floor of \farg{x}, which is the largest integer less than or equal to \farg{x}, converted to a real value; see warning at start of \refsection{step-functions}}{
     \[
     \mbox{\code{floor}}(x) = \lfloor x \rfloor
     \]}
   %
-  \fitem{real}{ceil}{real \farg{x}}{
-    Returns the ceiling of \farg{x}, which is the smallest integer greater
-    than or equal to \farg{x}, converted to a real value;
-    see warning at start of \refsection{step-functions}.
+  \fitemUnaryVec{ceil}{ceiling of \farg{x}, which is the smallest integer greater than or equal to \farg{x}, converted to a real value; see warning at start of \refsection{step-functions}}{
     \[
     \mbox{\code{ceil}}(x) = \lceil x\rceil
     \]}
   %
-  \fitem{real}{round}{real \farg{x}}{
-    Returns the nearest integer to \farg{x}, converted to a real value;
-    see warning at start of \refsection{step-functions}.
+  \fitemUnaryVec{round}{nearest integer to \farg{x}, converted to a real value; see warning at start of \refsection{step-functions}}{
     \[
     \mbox{\code{round}}(x) = 
     \begin{cases}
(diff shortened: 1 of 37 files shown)


lib/stan_math
```diff
@@ -1 +1 @@
-Subproject commit dfbb65161e9ebf703101204f76538109836157f4
+Subproject commit 26b2d80136653467e685c8b54cabcf06b929a5c9
```
bob-carpenter: This needs to be reverted to the current develop version.  Feature pull requests shouldn't update the submodule.

bob-carpenter:
The manual needs to be updated as part of this pull request.

syclik:
I moved the submodule revision back to the state of develop.

I second @bob-carpenter's comment regarding requiring the doc with this pull request.

rayleigh:
I can work on the documentation. 

Also, I see that the tests are passing anymore. I think it's because the commit stan/math got reverted to doesn't have the changes that I made to `stan/math/prim/mat.hpp`. Instead, it's to an earlier commit from October 4th. I've been trying to Google another way, but I haven't been able to find one. I have seen [this](http://stackoverflow.com/questions/13844996/git-submodule-init-not-pulling-latest-commit) and [this](https://blog.sourcetreeapp.com/2012/02/01/using-submodules-and-subrepositories/). 

Finally, should I email the bash script that I used? It probably could save whoever vectorizes the other functions time.

syclik:
The pull request needs to get merged. For whatever reason, we're behind on
the math tests. It'll probably get merged this weekend as long as it's ok.
Then we can kick off the Stan tests. So... for now, don't worry about it.
Fix the doc if you've got it in you.

What does the bash script do? It'll probably be useful.


Daniel

bob-carpenter:
stan-dev/stan should be updated in a separate pull request so that the
submodule matches the current develop version of stan-dev/math.

I'm still not quite sure of the mechanics of those submodule updates,
but it looks like Jenkins is slammed and all the pull requests are
trickling through.  I think you may be waiting for this one:

https://github.com/stan-dev/stan/pull/2081

Also, I changed a bunch of underlying functions and will be updating
those mat.hpp files in stan-dev/math when #410 is merged (it vectorizes all the remaining
real-valued unary functions).  Then we'll need a further pull request to
update to that submodule (that happens automatically), then another 
pull request to add the remaining dozen or so functions and this will
be completely done.   Can't wait.

- Bob

rayleigh:
Got it. The bash script 
1) Updates `function_signatures.h` by replacing `add_unary(fct)` with `add_unary_vectorized(fct)` 
2) Copies the `exp` matrix function signature test and replaces `exp` calls with the function
3) Runs the created matrix signature test

bob-carpenter:
That'll be useful.  Where should we put scripts like this?

bob-carpenter:
Let me know if you want help debugging.  Probably best to do this after the math lib submodule gets updated to the latest pull requests---I removed all the `math.h` includes and the global using statements for MSVC and also vectorized the remaining functions.  The problem I see right now is in  the signature for `abs()` in the integration tests, which should be fixed when all the right submodules are in place.

syclik:
Regarding scripts, I've been collecting these things in my own repo: https://github.com/syclik/stan-scripts

@bob-carpenter, we talked about it briefly (< 10 mins) maybe a few years ago and at that point, we didn't have a lot of use of putting it under stan-dev. Perhaps I can move my set of scripts to the gelman org and collect more stuff there?

syclik:
Jenkins, retest this please.

rayleigh:
Got it. Are we trying to debug in `Stan` or `stan/math`?

Also, in the interim, I updated the manual using a Perl script that I wrote. Let me know if the formatting is correct for `functions.tex`. I also had to update the `fitemUnaryVec` macro created so I could be more flexible with the descriptions.

bob-carpenter:
The stan-dev/math submodule got merged, so Daniel kicked off tests again in stan-dev/stan for this pull request.

syclik:
@rayleigh, thanks for updating the manual. I don't think it's particularly clear now. Unfortunately, I don't have any clear guidance, but maybe we can come up with something?

Here's what the current text looks like in the rendered manual (rendered image followed by the actual text, which I can't format to look like the pdf using markdown):

screen shot 2016-10-18 at 11 29 44 am

> R abs(T x)
>     Returns the absolute value of x. It is defined by
>         abs(x) = |x|
>     Applies elementwise to any scalar, vector, matrix, or array argument type T; the return type is the same as the argument type with int promoted to real; for details, see Section 33.1.

What's confusing to me about this is the types: `R` and `T`. Perhaps we need a section describing the element-wise vectorization somewhere else in more detail rely on that.

I'm going to approve the pull request since it looks like the doc with `R` and `T` was there before, but before merging, I'd like to get @bob-carpenter's opinion on it. Or if @rayleigh has any more thoughts on making this more explicit.

rayleigh:
The only thing I can think of is to have the return type be T. Other than that, I think that section 33.1, which is referred to at the end of the description, does a good job explaining vectorization. The description itself is a good summary of the function and how it can be used.

Also, I just wanted to check before it gets merged, but is the formatting okay for `functions.tex`? I squished the description into one line. Should I switch it back to a multi-line description as it was before?

PR #2157, opened 2016-12-01
docs fix for multivariate regression example

mike-lawrence:


bob-carpenter:
Thanks.  We always have a "next manual" issue for little fixes like this.  It's easier for us to just fix them all at once in a single pull request than to kick off our entire process (which involves integration testing and then merging spawning upstream pull requests in the interfaces).  But we can merge this one when things pass.

Issue #2178, opened 2016-12-19
NUTS sampler broken


beckermr:
#### Summary:
It appears that the NUTS sampler is failing on a simple problem with a known solution.

#### Description:
I am sampling a 2D Gaussian with known parameters and not getting back the right variances or correlation coefficient.

#### Reproducible Steps:
I have put a test script in both python and R below. (Thank you to @wlattner for the R test!)
Here is the STAN model I am using
```
parameters {
  vector[2] z;
}

model {
  matrix[2,2] sigma;
  vector[2] mu;

  mu[1] <- 0.0;
  mu[2] <- 3.0;
  sigma[1][1] <- 1.0 * 1.0;
  sigma[1][2] <- 0.5 * 1.0 * 2.0;
  sigma[2][1] <- 0.5 * 1.0 * 2.0;
  sigma[2][2] <- 2.0 * 2.0;

  z ~ multi_normal(mu, sigma);
}
```

#### Current Output:
Here is an example output from the script. You can see the standard deviations and correlation coefficient are wrong. These are very long chains and so they should converge to the proper parameters with high precision.
```
pystan version: 2.12.0.0
algorithm: NUTS
Inference for Stan model: TWOD_Gaussian_c1141a5e1a103986068b426ecd9ef5d2.
4 chains, each with iter=100000; warmup=10000; thin=1;
post-warmup draws per chain=90000, total post-warmup draws=360000.

       mean se_mean     sd   2.5%    25%    50%    75%  97.5%  n_eff   Rhat
z[0]-5.1e-4  1.9e-3   0.95  -1.89  -0.61 7.7e-4   0.61   1.88 235552    1.0
z[1]    3.0  3.9e-3    1.9  -0.77   1.77    3.0   4.23   6.77 234959    1.0
lp__  -1.48  2.3e-3   0.96  -4.09  -1.84  -1.18   -0.8  -0.57 179274    1.0

Samples were drawn using NUTS at Fri Dec 16 21:23:49 2016.
For each parameter, n_eff is a crude measure of effective sample size,
and Rhat is the potential scale reduction factor on split chains (at
convergence, Rhat=1).
mean  : [ -5.08802129e-04   2.99765183e+00]
stddev: [ 0.94632959  1.89621228]
corr  : 0.44350873785
```

#### Expected Output:
I ran this example with standard HMC and it worked great.
```
pystan version: 2.12.0.0
algorithm: HMC
[...]

PR #2180, opened 2016-12-22
Bugfix/issue 2178 fix nuts criterion

betanalpha:
#### Submisison Checklist

#### Summary

Addresses #2178 by modifying the build_tree code to return proper tree boundaries and fixing the rho used for the right subtree.

#### Intended Effect

Restores exact validity of the NUTS sampler.  The NUTS tests were expanded to catch these issues.

#### How to Verify

Run unit tests in src/test/unit/mcmc/hmc/nuts.

Also, run the modified model I posted in the comments of #2178 to see that the two marginal variances and the correlation are both within the MCMC standard error, even with one million iterations.

#### Side Effects

Small changes to the sampler behavior which should decrease with the sampler step size.  May break the upstream integration test.

#### Documentation

None user facing.  Inline doc was updated.

#### Reviewer Suggestions

@syclik @bob-carpenter 

#### Copyright and Licensing

### DIFF
--- a/src/stan/mcmc/hmc/nuts/base_nuts.hpp
+++ b/src/stan/mcmc/hmc/nuts/base_nuts.hpp
@@ -76,7 +76,6 @@ namespace stan {
         while (this->depth_ < this->max_depth_) {
           // Build a new subtree in a random direction
           Eigen::VectorXd rho_subtree = Eigen::VectorXd::Zero(rho.size());
-
           bool valid_subtree = false;
           double log_sum_weight_subtree
             = -std::numeric_limits<double>::infinity();
--- a/src/stan/mcmc/hmc/nuts/base_nuts.hpp
+++ b/src/stan/mcmc/hmc/nuts/base_nuts.hpp
@@ -85,7 +85,7 @@ namespace stan {
             this->z_.ps_point::operator=(z_plus);
             valid_subtree
               = build_tree(this->depth_, z_propose,
-                           p_sharp_dummy, p_sharp_plus, rho_subtree, 
+                           p_sharp_dummy, p_sharp_plus, rho_subtree,
                            H0, 1, n_leapfrog,
                            log_sum_weight_subtree, sum_metro_prob,
                            info_writer, error_writer);
@@ -152,9 +152,9 @@ namespace stan {
         values.push_back(this->energy_);
       }
 
-      bool compute_criterion(Eigen::VectorXd& p_sharp_minus,
-                             Eigen::VectorXd& p_sharp_plus,
-                             Eigen::VectorXd& rho) {
+      virtual bool compute_criterion(Eigen::VectorXd& p_sharp_minus,
+                                     Eigen::VectorXd& p_sharp_plus,
+                                     Eigen::VectorXd& rho) {
         return    p_sharp_plus.dot(rho) > 0
                && p_sharp_minus.dot(rho) > 0;
       }
--- a/src/stan/mcmc/hmc/nuts/base_nuts.hpp
+++ b/src/stan/mcmc/hmc/nuts/base_nuts.hpp
@@ -206,16 +206,17 @@ namespace stan {
           return !this->divergent_;
         }
         // General recursion
-        Eigen::VectorXd p_sharp_left = this->hamiltonian_.dtau_dp(this->z_);
-
-        Eigen::VectorXd rho_subtree(rho.size());
-        rho_subtree.setZero();
 
         // Build the left subtree
         double log_sum_weight_left = -std::numeric_limits<double>::infinity();
 
+        Eigen::VectorXd rho_left(rho.size());
+        rho_left.setZero();
+
+        Eigen::VectorXd p_sharp_left = this->hamiltonian_.dtau_dp(this->z_);
+
         bool valid_left
-          = build_tree(depth - 1, rho_subtree, z_propose,
+          = build_tree(depth - 1, rho_left, z_propose,
                        H0, sign, n_leapfrog,
                        log_sum_weight_left, sum_metro_prob,
                        info_writer, error_writer);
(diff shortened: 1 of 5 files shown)


src/stan/mcmc/hmc/nuts/base_nuts.hpp
```diff
@@ -60,45 +60,44 @@ namespace stan {
...
 
+        double log_sum_weight = 0;  // log(exp(H0 - H0))
         double H0 = this->hamiltonian_.H(this->z_);
         int n_leapfrog = 0;
         double sum_metro_prob = 1;  // exp(H0 - H0)
 
         // Build a trajectory until the NUTS criterion is no longer satisfied
         this->depth_ = 0;
-        this->divergent_ = 0;
+        this->divergent_ = false;
```
bob-carpenter: Thanks for switching to booleans---makes it much easier to read the code.

src/stan/mcmc/hmc/nuts/base_nuts.hpp
```diff
@@ -152,9 +151,9 @@ namespace stan {
         values.push_back(this->energy_);
       }
 
-      bool compute_criterion(Eigen::VectorXd& p_sharp_minus,
-                             Eigen::VectorXd& p_sharp_plus,
-                             Eigen::VectorXd& rho) {
+      virtual bool compute_criterion(Eigen::VectorXd& p_sharp_minus,
```
bob-carpenter: This only needs to be virtual if you are (a) using OO and passing references to the base class, OR (b) using templating and don't implement `compute_criterion` in all subclasses.

src/stan/mcmc/hmc/nuts/base_nuts.hpp
```diff
@@ -60,45 +60,44 @@ namespace stan {
...
         this->depth_ = 0;
-        this->divergent_ = 0;
+        this->divergent_ = false;
 
         while (this->depth_ < this->max_depth_) {
           // Build a new subtree in a random direction
-          Eigen::VectorXd rho_subtree(rho.size());
-          rho_subtree.setZero();
-
+          Eigen::VectorXd rho_subtree = Eigen::VectorXd::Zero(rho.size());
```
bob-carpenter: Won't the original way you had it be more efficient?   I actually don't know if the compiler's clever enough to reuse the right-hand side object given its scope is only the assignment or if this causes an extra malloc and copy.  In which case, the way you were originally doing it would be more efficient.  Probably not worth worrying about.

src/stan/mcmc/hmc/nuts/base_nuts.hpp
```diff
@@ -249,8 +258,9 @@ namespace stan {
             z_propose = z_propose_right;
         }
 
+        Eigen::VectorXd rho_subtree = rho_left + rho_right;
         rho += rho_subtree;
```
bob-carpenter: This should be

```
rho += rho_left + rho_right;
```

to minimize copies.  The assignment to `rho_subtree` would otherwise alloc and assign, whereas the direct `+=` will avoid it.

src/stan/mcmc/hmc/nuts/base_nuts.hpp
betanalpha: Yes, for testing we need a derived class that overloads compute_criterion which is why I added it back in.

src/stan/mcmc/hmc/nuts/base_nuts.hpp
betanalpha: rho_subtree is used again in the following call to compute_criterion.  Because of this we'd have to dump it into a temporary or awkwardly call ```.eval()``` anyways.

betanalpha:
For the record, with 1,000,000 iterations I get 

```
                       Mean        MCSE      StdDev           5%          50%          95%       N_Eff     N_Eff/s       R_hat
z[1]            -3.6414e-04  1.2369e-03  1.0004e+00  -1.6443e+00  -7.6846e-04   1.6448e+00  6.5405e+05  2.5322e+04  1.0000e+00
```

One of the stats in the model I suggested to run is exactly the marginal variance for z[1] minus it's true value, and it is within the MCMC SE of zero.  This could not be true is the standard deviation was also somehow biased.

Remember that you have to compare deviations to the expected MCMC SE and this can be surprisingly large at times!

bob-carpenter:
I'm a knucklehead---forgot to check out the new branch before running!  It works much better on the new branch.  Everything looks good.  Didn't realize you reused that variable since I was only looking at the diff. 

```
                    Mean     MCSE   StdDev     5%       50%       95%    N_Eff  N_Eff/s    R_hat
z[1]            -3.5e-04  1.2e-03  1.0e+00   -1.6   1.4e-03   1.6e+00  6.5e+05    17636  1.0e+00
z[2]             3.0e+00  2.5e-03  2.0e+00  -0.29   3.0e+00   6.3e+00  6.4e+05    17490  1.0e+00
delta_var1      -9.6e-04  1.9e-03  1.4e+00  -1.00  -5.5e-01   2.8e+00  5.6e+05    15263  1.0e+00
delta_var2      -5.6e-03  7.5e-03  5.6e+00   -4.0  -2.2e+00   1.1e+01  5.6e+05    15354  1.0e+00
delta_corr      -1.7e-03  1.6e-03  1.1e+00   -1.2  -3.4e-01   2.2e+00  4.6e+05    12610  1.0e+00
```

src/stan/mcmc/hmc/nuts/base_nuts.hpp
bob-carpenter: This shouldn't be too much overhead, given that there's a dot-product getting called, but I don't like the idea in general of making functions virtual just for testing.  We shouldn't need to do that, but we can talk about that later.  Let's just get this in for now.

syclik:
Thanks, @betanalpha, for taking care of this. Thanks, @bob-carpenter, for reviewing. I looked over the changes too and they all looked good.

betanalpha:
It’s a holiday miracle!

Issue #2179, opened 2016-12-21
de/serialization issues


wds15:
#### Summary:
When using rstan 2.13 stan_rdump to serialize arrays with zero size , then these cannot be read with cmdstan 2.13

#### Description:
Serializing with stan_rdump and starting a cmdstan run does not work any more if the arrays has size zero. This used to work with rstan 2.12 and cmdstand 2.12.

This releated to the issue rstan#371 .

#### Reproducible Steps:
serialized data with rstan 2.13 which does not work with cmdstan 2.13:
```
N <- 0
y <- 
structure(integer(0), 
.Dim = c(0))
```

this data works with cmdstan 2.13 and was generated with rstan 2.12:
```
N <- 0
y <- 
structure(c(),
.Dim = c(0))
```

Here is a dummy Stan program:
```
data {
  int N;
  int y[N];
}
parameters {
  real mu;
}
model {
  target += normal_lpdf(mu|0, 1);
}
```

#### Current Output:
cmdstan fails to run

#### Expected Output:
Error from cmdstan:
```
[weberse2@node04:~/work/stan/misc] ./stan_int_bug sample data file=notok.R

Exception: data y value  beyond int range
Diagnostic information: 
stan_2.13.1/src/stan/io/dump.hpp(518): Throw in function bool stan::io::dump_reader::next()
Dynamic exception type: boost::exception_detail::clone_impl<boost::exception_detail::error_info_injector<std::invalid_argument> >
std::exception::what: data y value  beyond int range
```

#### cmdstan Stan Version:
2.13.1

#### Operating System:
RHEL 6.7

bob-carpenter:
This issue was moved to stan-dev/rstan#373

maverickg:
I will try to fix it in stan rather than rstan so all are happy.

bob-carpenter:
OK, if you want to fix it there, it should be a Stan issue.
But it should be opened as a Stan issue rather than as a comment
about RStan to CmdStan translation.

But why not just go back to having RStan generate the same output as before?

- Bob

maverickg:
see here for why rstan changed this function: https://github.com/stan-dev/rstan/issues/366

[...]

PR #2187, opened 2016-12-23
dump_reader reads integer(x) as zero arrays

maverickg:
fix issue #2179

#### Submisison Checklist

#### Summary

Allow dump_reader to read zero arrays of integer (or double) that could be generated from `R`'s dump.  This would fix issue #2179 . 

#### Intended Effect

Stan's dump interface can read data of the following format:

```
x <- integer(0)
y <- structure(integer(0), .Dim = c(2, 3))
z <- structure(double(0), .Dim = c(2, 3))
a <- double(1)
b <- structure(double(1), .Dim = c(1))
``` 

#### How to Verify

* see the added unit test
* build cmdstan with the new dump.hpp and run examples in #2179 

#### Side Effects

None.

#### Documentation

See the doxygen comments in the code. 

#### Reviewer Suggestions
@bob-carpenter @syclik  or anyone

#### Copyright and Licensing

Jiqiang Guo

### DIFF
--- a/src/stan/io/dump.hpp
+++ b/src/stan/io/dump.hpp
@@ -56,6 +56,15 @@ namespace stan {
      * a string such as "1." or "0.9e-5" represents a floating
      * point value.
      *
+     * <p>The dump reader treats "integer(x)" as an array of zeros
+     * (type being integer and length x), where x any non-negative
+     * integers and x can be omitted to indicate zero-length.
+     * So the following are all legitimate: * "x <- integer()",
+     * "x <- integer(0) ", and "x <- integer(3)". For array of zeros
+     * of type double, we can replace the above "integer" with "double".
+     * This is mainly for the purpose of supporting zero-size arrays
+     * such as "x <- structure(integer(0), .Dim = c(2, 0))".
+     *
      * <p>For dumping, arrays are indexed in last-index major fashion,
      * which corresponds to column-major order for matrices
      * represented as two-dimensional arrays.  As a result, the first
@@ -75,7 +84,7 @@ namespace stan {
      *
      * definitions ::= definition+
      *
-     * definition ::= name ("->" | '=') value optional_semicolon
+     * definition ::= name ("<-" | '=') value optional_semicolon
      *
      * name ::= char*
      *        | ''' char* '''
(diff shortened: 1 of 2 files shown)


bob-carpenter:
Thanks.  I'll review this ASAP, but I want to download it and run it, not just look at the code, so it'll take a bit of time.

maverickg:
Thanks.  By the way, is the manual of CmdStan the only place we documented this dump format for users?

bob-carpenter:
Yes, it is.  We should probably create a standalone document
somewhere.

- Bob

bob-carpenter:
I verified that the previous format works, but I don't know enough about R to understand what is supposed to happen with this:

```
y <- structure(integer(0), .Dim = c(2, 3))
```

What would the type be for `y` in a Stan program?  I tried

```
int mu[2, 3];
```

but it segfaulted when i used this data file:

```
mu <- structure(integer(0), .Dim = c(2, 3));
```

If I change the declaration to

```
int mu[0];
```

then I get this:

```
Exception: mismatch in number dimensions declared and found in context; processing stage=data initialization; variable name=mu; dims declared=(0); dims found=(2,3)
```

maverickg:
It the dump file is created from R correctly, it should not have 

```
y <- structure(integer(0), .Dim = c(2, 3))
```

Besides, it would error in R

```
> structure(integer(0), .Dim = c(2,3))
Error in structure(integer(0), .Dim = c(2, 3)) : 
  dims [product 6] do not match the length of object [0]
```

I will find out what it current does for this similar case:
```
y <- structure(c(0, 0), .Dim = c(3))
```

This is not correct in the sense of `R` either, but it's fine for stan right now.  Good thing is that it does not segfault.  

```
> y <- structure(c(0, 0), .Dim = c(3))
Error in structure(c(0, 0), .Dim = c(3)) : 
  dims [product 3] do not match the length of object [2]
```

At least, I would make `y <- structure(integer(0), .Dim = c(2))` not cause a segfault.

bob-carpenter:
> On Dec 23, 2016, at 6:40 PM, maverickg <notifications@github.com> wrote:
> 
> It the dump file is created from R correctly, it should not have
> 
> y <- structure(integer(0), .Dim = c(2, 3))
> 
> Besides, it would error in R

OK.  I was confused because I copied the above line
from the pull request itself (under "Intended Effect").

Not segfaulting would be nice, but we probably shouldn't
worry too much about the details of this format since I
think we're going to be replacing it with JSON or Protocol
Buffers soon (though we'll keep supporting the old format
at least until Stan 3).  Within CmdStan, segfaults are mainly
annoying because you don't get an error message;  they don't
bring down the environment like in R or Python.

bob-carpenter:
Let me know if you'd like me to approve and merge now or wait for more changes.

maverickg:
Thanks.  The segfault happens as well for the original case such as the following:

```
x <- structure(c(), .Dim =  c(2))
```

What is more, I think it would happen sometimes for the following (the length of the vector is less than what the `.Dim` says. 
```
x <- structure(c(1), .Dim = c(2))
```

But it did not segfault at least for me, perhaps for the same reason the following code does not segfault either.

```
mmac2:cmdstan jq$ more test.cpp

#include <iostream>
#include <vector>

int main(int argc, char **argv) {
  std::vector<int> a;
  a.push_back(3);
  std::cout << a[0] << std::endl;
  std::cout << a[1] << std::endl;
}
mmac2:cmdstan jq$ g++ test.cpp
mmac2:cmdstan jq$ clang++ test.cpp
mmac2:cmdstan jq$ ./a.out
3
268435456
```

But as these are not supposed to be produced by `stan_rdump` and using R's dump format won't last long, I am inclined to just merge this one.

PR #2210, opened 2017-01-23
Tweak average metro prob calc to not use initial state

betanalpha:
#### Submisison Checklist

#### Summary

Removes initial state from the calculation of the average Metropolis acceptance probability.

#### Intended Effect

Reduces the average Metropolis acceptance probability to be consistent with the average probability of accepting a jump to each new state in the HMC trajectory from the initial state.

#### How to Verify

Run unit tests or try it on your favorite model.

#### Side Effects

Should make adaptation a bit more robust by allowing a change in step size to more strongly affect the average acceptance probability.

#### Documentation

None user facing.

#### Reviewer Suggestions

Anyone.

#### Copyright and Licensing

Please list the copyright holder for the work you are submitting (this will be you or your assignee, such as a university or company):  Columbia University.

### DIFF
--- a/src/stan/mcmc/hmc/nuts/base_nuts.hpp
+++ b/src/stan/mcmc/hmc/nuts/base_nuts.hpp
@@ -67,7 +67,7 @@ namespace stan {
         double log_sum_weight = 0;  // log(exp(H0 - H0))
         double H0 = this->hamiltonian_.H(this->z_);
         int n_leapfrog = 0;
-        double sum_metro_prob = 1;  // exp(H0 - H0)
+        double sum_metro_prob = 0;
 
         // Build a trajectory until the NUTS criterion is no longer satisfied
         this->depth_ = 0;
@@ -128,7 +128,7 @@ namespace stan {
         // Compute average acceptance probabilty across entire trajectory,
         // even over subtrees that may have been rejected
         double accept_prob
-          = sum_metro_prob / static_cast<double>(n_leapfrog + 1);
+          = sum_metro_prob / static_cast<double>(n_leapfrog);
 
         this->z_.ps_point::operator=(z_sample);
         this->energy_ = this->hamiltonian_.H(this->z_);
--- a/src/test/performance/logistic_test.cpp
+++ b/src/test/performance/logistic_test.cpp
@@ -108,13 +108,13 @@ TEST_F(performance, values_from_tagged_version) {
     << "last tagged version, 2.9.0, had " << N_values << " elements";
 
   std::vector<double> first_run = last_draws_per_run[0];
-  EXPECT_FLOAT_EQ(-65.201302, first_run[0])
+  EXPECT_FLOAT_EQ(-65.319603, first_run[0])
     << "lp__: index 0";
 
-  EXPECT_FLOAT_EQ(0.93779498, first_run[1])
+  EXPECT_FLOAT_EQ(0.96712297, first_run[1])
     << "accept_stat__: index 1";
 
-  EXPECT_FLOAT_EQ(1.20701, first_run[2])
+  EXPECT_FLOAT_EQ(1.06545, first_run[2])
     << "stepsize__: index 2";
 
   EXPECT_FLOAT_EQ(2, first_run[3])
@@ -126,13 +126,13 @@ TEST_F(performance, values_from_tagged_version) {
   EXPECT_FLOAT_EQ(0, first_run[5])
     << "divergent__: index 5";
 
-  EXPECT_FLOAT_EQ(65.755402, first_run[6])
+  EXPECT_FLOAT_EQ(66.090698, first_run[6])
     << "energy__: index 6";
 
-  EXPECT_FLOAT_EQ(1.29672, first_run[7])
+  EXPECT_FLOAT_EQ(1.4068201, first_run[7])
     << "beta.1: index 7";
 
-  EXPECT_FLOAT_EQ(-0.47478199, first_run[8])
+  EXPECT_FLOAT_EQ(-0.59367198, first_run[8])
     << "beta.2: index 8";
 
   matches_tagged_version = !HasNonfatalFailure();
--- a/src/test/unit/mcmc/hmc/nuts/softabs_nuts_test.cpp
+++ b/src/test/unit/mcmc/hmc/nuts/softabs_nuts_test.cpp
@@ -326,7 +326,7 @@ TEST(McmcSoftAbsNuts, transition_test) {
   EXPECT_FLOAT_EQ(-0.86902142, s.cont_params()(1));
   EXPECT_FLOAT_EQ(1.6008, s.cont_params()(2));
   EXPECT_FLOAT_EQ(-3.5239484, s.log_prob());
-  EXPECT_FLOAT_EQ(0.99709648, s.accept_stat());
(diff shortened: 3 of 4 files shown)


betanalpha:
Tests passing -- anyone want to review?

jgabry:
I'm not yet familiar enough with the code to make useful comments on those
changes (my untrained eye doesn't see anything obviously problematic), but
I do have a question about the purpose.

Is the idea here that the initial state was counted as an acceptance and
removing it makes the mean more sensitive to each of the remaining values?

betanalpha:
Correct.

syclik:
I'm hoping I'll free up later today or tomorrow to look at it in depth. I
haven't peeked at it, but if it doesn't have some end-to-end tests that
show that it behaves well for some simpler problems, I will request those.
(you might have already done it... just haven't gotten to looking at it.)

betanalpha:
We do not have the infrastructure for supporting tests like that, nor
do we have the framework for even testing such values.

Seriously, everyone, this is a two lines change.

syclik:
We do have infrastructure for testing end-to-end. I'm trying to avoid us
releasing broken samplers.

sakrejda:
@betanalpha So I first looked at this and thought, hey, it's a two-line change.  Then I couldn't figure out how this was excluding the initial point.  It looks like it starts an accumulator at zero instead of one which would make sense if this zero/one was the relative acceptance probability for the initial point---is that it?

bob-carpenter:
I'm confused about how we supposedly have diagnostics for HMC
yet can't validate that HMC itself works.

Given that we've broken the sampler in two previous releases,
I'm very very reluctant to change things without a way to
measure what we gain from doing so.

As a minimum, can we do what Andrew was suggesting and simulate
data and check posterior coverage?  That leads directly to
a simple hypothesis test, right, where we expect binomial(N, p)
of the parameters to be in the posterior p intervals?

I'm not trying to be difficult, but I'm just very confused about
why you're saying it's impossible to test.  We don't need something
that's knife-edged balanced, just something that will find those
problems like the bivariate normal not estimating posterior variance
properly.  Or things slowing down by a factor of 50%.  I understand
that might not measure a 1% error in posterior variance, but I'd
feel a lot better doing that than just throwing our hands up and
saying it's impossible.  Or that it's too hard.

Can we combine this with a framework for testing approximate
inference, too?  

- Bob

betanalpha:
Yes, the accept probability for a transition from the initial state into itself is one.
Resetting the accumulator to zero and then dividing the accumulator by num_leapfrog
instead of num_leapfrog + 1 then ignores the initial point altogether.

betanalpha:
Because the output is stochastic.  Posterior coverage has no
frequentist guarantees, so a hypothesis test will have not have
a known coverage.  Even if we did know the coverage, a single
hypothesis test will be really noisy so to make something that
doesn’t false positive all the time we’d have to do an ensemble 
of tests and then hypothesis test the distribution of p-values.

What does have frequentist guarantees is the MCMC standard
error, which motivated the framework in the stat_valid_test 
branch (using the ensemble of tests approach mentioned above).

In any case, checking models by eye to see if they return 
reasonable values is fine because we can readily run again
with a different seed if we see something weird.  But once we
automate that in a test we have to confront the subtleties above
lest we end up with something that is completely insensitive to
problems or false positives too much.

What will absolutely not work is hard coding values and testing
that the samplers return those, which seems to be the favorite
test strategy of some people.  Any small change in the sampler
affects the stochastic evolution of the Markov chain which then
yields different return values, even for the same initial seed.
But there’s nothing pathological about this behavior — it’s the
nature of MCMC.  

The only reasonable way to introduce useful MCMC integration
tests is what’s in stat_valid_test right now, and that cannot be
implemented as a test until we can easily run the Stan from a
pure C++ interface.  Even with the refactor that is still awkward.

syclik:
Sorry, I think you misunderstand my intent. There are some tests in there to freeze behavior, which you keep alluding to. They're really there to trigger exactly that: make sure we don't accidentally change behavior with fixed seeds, etc.

I also wasn't suggesting doing some sort of hypothesis testing. We've tried that early on and it did help stop us from making lots of critical mistakes, but I think we're past that now.

What I'm suggesting is doing some due diligence and we try not to repeat shipping Stan with broken samplers. We do have a framework for running models either with fixed seeds or random seeds. Some models are analytic and quick enough where we can calculate some properties and know they should be within some range. If we did that, we wouldn't have made the 2.11 mistake. I'm just asking we don't change implementation details that affect everything without being a little more careful.

Hopefully that makes sense. I'm really not trying to be contentious. I'm trying not to ask you to write another test that's like the performance test. I'm trying to ask that you put in some effort to show that this isn't blatantly wrong.

sakrejda:
`sum_metro_prob` is local to the `transition` function and is modified by reference in `build_tree` using only the `+=` operator so I buy that this PR does what it says.  I think Bob and Dan are making a valid point that we've been burned by sampler bugs a few times and that it's easy for subtle bugs to sneak through.  Maybe rather than automated testing this PR come with examples of the manual tests past HMC bugs have caused failures in: 1) correctly sampling from a uniform (0,1); and 2) correct variances on MVN. (?) It wouldn't make any sense to get burned by those kinds of bugs again.

@betanalpha I wrote some of the tests for the services in the refactor branch and it seemed straightforward at the time. What else would need to be improved about the interface to be able to use your valid_stat_test branch?

betanalpha:
Does anyone really think that I submitted a PR without testing on a few models,
including the one that I used to validate the 2.14 fix?  I explicitly noted in the
PR details that reviewers should run tests and try on their favorite models to
ensure that inferences are still reasonable.  Telling reviewers which models to
test introduces all kinds of problems — much more robust for reviewers to
spot check on their own models without the bias of suggestion.

sakrejda:
The point is we want to document what did get checked so if a bug comes up
we know why we missed it.  Email sucks for communicating and I'd rather
keep what gets checked explicit.

bob-carpenter:
Let's discuss this in person at the Stan meeting tomorrow.

My goal is to understand how Michael did testing for this and
how we can codify that testing into regression tests that we can
use going forward.

- Bob

syclik:
+1.

I'm sorry I won't be at the meeting, but I trust we'll come up with
something that can be a repeatable test.

seantalts:
As one of the industrial programmers I feel obliged to chime in - I think even relatively fragile tests of very important code is vastly better than no tests.

To illustrate: an ex-Google company I used to work at had what they called "gold tests" on API boundaries that would test character-by-character all API responses. So whenever the API changed a little, even to add something new, the tests would change. We had a facility for updating the tests to comply with the new response automatically, so the burden on the programmer was fairly minimal, and then the diff would show exactly what changes happen with each commit, which was an extremely nice historical property to maintain.

I recognize that character-by-character obviously doesn't work for stochastic things, but you could imagine a similar system with heuristics / valid ranges. Maybe this is close to the stats testing framework you mentioned building 2 years ago?

bob-carpenter:
Feel free to speak up at the meetings!

It's exactly these "gold test" things that are at issue, with Michael on one side, Daniel on the other, and me somewhere in the middle.  

The issue here is whether we can do some tests that are better than nothing.  Sounds like Michael says we do have such things, but I'd like to make sure we at least have the last two things that were causing issues tested.  That's just in the spirit of software regression testing (as in making sure the software doesn't regress to former bad behaviors).

- Bob

PR #2216, opened 2017-01-27
Bugfix/2145 gq rng compound decl define

mitzimorris:
#### Submisison Checklist

#### Summary
Changed Stan parser to better track program block in which a variable is declared by changing `stan/lang/ast/var_origin.hpp` from an `int` to a `struct` consisting of the type of the enclosing program block and a bool indicating whether or not inside a local block.  This affects any grammar rule which sets or updates program block information and its associated semantic action.

#### Intended Effect
The specific bugfix allows compound declare define statements in local program blocks when the definition statement contains a function call.  The change to type `var_origin` is the first step in refactoring the parser logic to better keep track of what is/isn't allowed in a particular scope.

#### How to Verify
Test model `src/test/test-models/good/declare-define-gq-local-rng.stan` contains the reported bug.  The unit test `src/test/unit/parser/var_decls_grammar_test.cpp` tests that this model is parsable.

#### Side Effects
None

#### Reviewer Suggestions
Bob Carpenter or Daniel Lee

#### Copyright and Licensing

### DIFF
--- a/src/stan/lang/ast/var_origin.hpp
+++ b/src/stan/lang/ast/var_origin.hpp
@@ -35,7 +35,8 @@ namespace stan {
        *
        * @param program_block enclosing program block
        */
-      var_origin(const origin_block& program_block);  // NOLINT(runtime/explicit)
+      var_origin(const
+                 origin_block& program_block);   // NOLINT(runtime/explicit)
 
       /**
        * Construct an origin for a variable in specified outer program block,
--- a/src/stan/lang/ast/var_origin.hpp
+++ b/src/stan/lang/ast/var_origin.hpp
@@ -24,31 +24,33 @@ namespace stan {
       bool is_local_;
 
       /**
-       * Construct an empty var_origin
+       * No arg constructor, defaults:
+       * program_block_ : model_name_origin
+       * is_local_ : false
        */
       var_origin();
 
       /**
        * Construct an origin for variable in a specified block
        *
-       * @param program_block - enclosing program block
+       * @param program_block enclosing program block
        */
-      var_origin(const origin_block program_block);  // NOLINT(runtime/explicit)
+      var_origin(const origin_block& program_block);  // NOLINT(runtime/explicit)
 
       /**
        * Construct an origin for a variable in specified outer program block,
        * specify whether or not variable is in local program block 
        *
-       * @param program_block - enclosing program block
-       * @param is_local - flags whether or not in a local block
+       * @param program_block enclosing program block
+       * @param is_local flags whether or not in a local block
        */
-      var_origin(const origin_block program_block,
-                 const bool is_local);  // NOLINT(runtime/explicit)
+      var_origin(const origin_block& program_block,
+                 const bool& is_local);
 
       /**
        * Return true if declared in a local block.
        *
-       * @return bool if declared in a local block.
+       * @return value of is_local_
        */
       bool is_local() const;
     };
--- a/src/stan/lang/ast/var_origin.hpp
+++ b/src/stan/lang/ast/var_origin.hpp
@@ -1,87 +1,58 @@
 #ifndef STAN_LANG_AST_VAR_ORIGIN_HPP
 #define STAN_LANG_AST_VAR_ORIGIN_HPP
 
+
+#include <stan/lang/ast/origin_block.hpp>
+#include <cstddef>
+
 namespace stan {
   namespace lang {
 
     /**
-     * The type of a variable indicating where a variable was
-     * declared.   This is a typedef rather than an enum to get around
-     * forward declaration issues with enums in header files.
+     * Structure describes enclosing program block(s)
+     * in which variable is declared.
      */
-    typedef int var_origin;
+    struct var_origin {
+      /**
+       * Outermost enclosing program block.
+       */
+      origin_block program_block_;
 
-    /**
-     * Origin of variable is the name of the model.
-     */
-    const int model_name_origin = 0;
+      /**
+       * Flags whether in a nested (local) program block.
+       */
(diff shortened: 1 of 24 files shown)


src/stan/lang/ast/var_origin.hpp
```diff
@@ -1,86 +1,57 @@
...
+
+      /**
+       * Construct an origin for a variable in specified outer program block,
+       * specify whether or not variable is in local program block 
+       *
+       * @param program_block - enclosing program block
+       * @param is_local - flags whether or not in a local block
+       */
+      var_origin(const origin_block program_block,
+                 const bool is_local);  // NOLINT(runtime/explicit)
```
bob-carpenter: Don't need NOLINT on binary constuctor

src/stan/lang/ast/var_origin.hpp
```diff
@@ -1,86 +1,57 @@
...
+       * Construct an empty var_origin
+       */
+      var_origin();
+
+      /**
+       * Construct an origin for variable in a specified block
+       *
+       * @param program_block - enclosing program block
+       */
+      var_origin(const origin_block program_block);  // NOLINT(runtime/explicit)
```
bob-carpenter: Would this not work with an explicit declaration?  If it works as explicit, use that instead of NOLINT.

```
explicit var_origin(const origin_block& program_block);
```

Should also be constant reference or remove the `const`.

src/stan/lang/ast/var_origin.hpp
```diff
@@ -1,86 +1,57 @@
...
+       */
+      origin_block program_block_;
+
+      /**
+       * Flags whether in a nested (local) program block.
+       */
+      bool is_local_;
+
+      /**
+       * Construct an empty var_origin
```
bob-carpenter: This isn't empty---you should set the variables to something and doc what that is.

src/stan/lang/ast/var_origin_def.hpp
```diff
@@ -0,0 +1,22 @@
...
+#define STAN_LANG_AST_VAR_ORIGIN_DEF_HPP
+
+#include <stan/lang/ast/origin_block.hpp>
+#include <stan/lang/ast/var_origin.hpp>
+
+namespace stan {
+  namespace lang {
+
+    var_origin::var_origin()
+      : program_block_(model_name_origin), is_local_(false) { }
```
bob-carpenter: This is the values you use in the no-arg constructor.

src/stan/lang/ast/var_origin.hpp
```diff
@@ -1,86 +1,57 @@
...
+       */
+      var_origin();
+
+      /**
+       * Construct an origin for variable in a specified block
+       *
+       * @param program_block - enclosing program block
```
bob-carpenter: no hyphens after parameter names for doxygen

src/stan/lang/ast/var_origin.hpp
```diff
@@ -1,86 +1,57 @@
...
+       * @param is_local - flags whether or not in a local block
+       */
+      var_origin(const origin_block program_block,
+                 const bool is_local);  // NOLINT(runtime/explicit)
+
+      /**
+       * Return true if declared in a local block.
+       *
+       * @return bool if declared in a local block.
```
bob-carpenter: `return true if declared` not `return bool if declared`

src/stan/lang/grammars/semantic_actions_def.cpp
```diff
@@ -2534,7 +2544,7 @@ namespace stan {
     boost::phoenix::function<set_int_range_upper> set_int_range_upper_f;
 
     void validate_int_data_expr::operator()(const expression& expr,
-                                            int var_origin, bool& pass,
+                                            var_origin var_origin, bool& pass,
```
bob-carpenter: The variable `var_origin` should be passed by reference here.  It's also confusing to have the variable name and the type name be the same thing.  So how about `const var_origin& vo` instead?  Then all the `var_origin.` below get changed, which makes it less confusing as to value vs. data structure.

src/stan/lang/ast/var_origin.hpp
mitzimorris: updated doc

src/stan/lang/ast/var_origin.hpp
mitzimorris: fixed

src/stan/lang/ast/var_origin.hpp
mitzimorris: removed NOLINT, changed to constant reference

src/stan/lang/ast/var_origin.hpp
mitzimorris: fixed per above

bob-carpenter:
You will need to add "explicit" if you remove NOLINT(runtime/explicit) or Cpplint will balk.

src/stan/lang/ast/var_origin.hpp
mitzimorris: per comment, added NOLINT for 1-arg constructor, but not for binary constructor.

src/stan/lang/grammars/semantic_actions_def.cpp
mitzimorris: changed to const reference.  also changed all var_origin variables named "var_origin" to "vo".  other functions in the code have var_origin variables named "origin" - kept these unchanged.  in two functions, changed name "origin_allowed" to "origin" so that usage across this file is more consistent.

bob-carpenter:
Cpplint failed.  Please run cpplint before pushing.

PR #2260, opened 2017-04-03
Feature/1706 mass matrix

mitzimorris:
#### Submisison Checklist

#### Summary
Samplers for HMC static and HMC nuts allow specification of initial inverse mass matrix.  The functions which run these samplers are:

- `stan::services::sample::hmc_nuts_diag_e`
- `stan::services::sample::hmc_nuts_diag_e_adapt`
- `stan::services::sample::hmc_nuts_dense_e`
- `stan::services::sample::hmc_nuts_dense_e_adapt`
- `stan::services::sample::hmc_static_diag_e`
- `stan::services::sample::hmc_static_diag_e_adapt`
- `stan::services::sample::hmc_static_dense_e`
- `stan::services::sample::hmc_static_dense_e_adapt`

The initial mass matrix is passed in to these functions as a stan::io::var_context object which must o contain a variable named "mass_matrix".  Depending on the sampler, this is either a vector of the diagonal elements of the inverse mass matrix or a dense mass matrix.  The length diagonal vector should match the number of parameters in the model, likewise the number of rows and columns in the dense mass matrix should match the number of parameters in the model.  The input functions validate that the specified mass matrix has the correct dimensions and that it is positive definite and will return `error_codes::CONFIG` if the specified mass matrix is invalid.

If no initial mass matrix is specified, the sampler functions will use the identity matrix as the mass matrix.

This feature affects the following parts of the Stan API:

- `stan::mcmc::hmc::hamiltonians` constructors for the `ps_point` subclasses
- `stan::mcmc::hmc::nuts` constructors for the samplers
- `stan::mcmc::hmc::static` constructors for the samplers
- `stan::services::sample` functions which run the samplers

Utility functions to read, validate, and create identity diag and dense mass matrices have been added to `stan::services::util`

#### Intended Effect
Allows users to resume sampling with an evolved mass matrix.  This will only produce results when used with the same model and input data.

#### How to Verify
The following unit tests have been added:

- `src/test/unit/services/util/mass_matrix_test.hpp`
- `src/test/unit/services/sample/hmc_nuts_diag_mass_matrix_test.cpp`
- `src/test/unit/services/sample/hmc_nuts_dense_mass_matrix_test.cpp`
- `src/test/unit/services/sample/hmc_static_diag_mass_matrix_test.cpp`
- `src/test/unit/services/sample/hmc_static_dense_mass_matrix_test.cpp`

#### Side Effects
none

#### Documentation
doxygen comments added for all new functions

#### Reviewer Suggestions
Bob Carpenter, Michael Betancourt

#### Copyright and Licensing


src/stan/mcmc/hmc/nuts/base_nuts.hpp
```diff
@@ -27,6 +27,28 @@ namespace stan {
...
+       */
+      base_nuts(const Model& model, BaseRNG& rng,
+                Eigen::VectorXd& inv_mass_matrix)
+        : base_hmc<Model, Hamiltonian, Integrator, BaseRNG>(model, rng,
+                                                            inv_mass_matrix),
+          depth_(0), max_depth_(5), max_deltaH_(1000),
+          n_leapfrog_(0), divergent_(0), energy_(0) {
+      }
+
+      /**
```
betanalpha: Duplicate constructor.

src/stan/mcmc/hmc/base_hmc.hpp
```diff
@@ -33,14 +33,63 @@ namespace stan {
...
+          z_(model.num_params_r(), inv_mass_matrix),
+          integrator_(),
+          hamiltonian_(model),
+          rand_int_(rng),
+          rand_uniform_(rand_int_),
+          nom_epsilon_(0.1),
+          epsilon_(nom_epsilon_),
+          epsilon_jitter_(0.0) {}
+
+      /**
```
betanalpha: Duplicate constructor.

src/stan/mcmc/hmc/static/base_static_hmc.hpp
```diff
@@ -29,6 +29,28 @@ namespace stan {
...
+       */
+      base_static_hmc(const Model& model, BaseRNG& rng,
+                      Eigen::VectorXd& inv_mass_matrix)
+        : base_hmc<Model, Hamiltonian, Integrator, BaseRNG>(model, rng,
+                                                            inv_mass_matrix),
+        T_(1), energy_(0) {
+        update_L_();
+      }
+
+      /**
```
betanalpha: Duplicate constructor.

src/stan/services/util/read_dense_mass_matrix.hpp
```diff
@@ -0,0 +1,50 @@
...
+        try {
+          init_mass_matrix.validate_dims("read dense mass matrix",
+                           "mass_matrix", "matrix",
+                           init_mass_matrix.to_vec(num_params, num_params));
+          std::vector<double> dense_vals =
+            init_mass_matrix.vals_r("mass_matrix");
+          inv_mass_matrix =
+            stan::math::to_matrix(dense_vals, num_params, num_params);
+        } catch (const std::exception& e) {
+          error_writer("Cannot get mass matrix from input file");
```
betanalpha: Can we have two try/catch blocks here and be more verbose with the error messages?  In particular, one block and error for seeing if the right objects is in the given `var_context` and a second one to see if the object has the right shape.  We want the user to quickly be able to identify the problem.

src/stan/services/util/read_diag_mass_matrix.hpp
```diff
@@ -0,0 +1,50 @@
...
+          init_mass_matrix.validate_dims("read diag mass matrix", "mass_matrix",
+                                         "vector_d",
+                                         init_mass_matrix.to_vec(num_params));
+          std::vector<double> diag_vals =
+            init_mass_matrix.vals_r("mass_matrix");
+          for (size_t i=0; i < num_params; i++) {
+            inv_mass_matrix(i) = diag_vals[i];
+          }
+        } catch (const std::exception& e) {
+          error_writer("Cannot get mass matrix from input file");
```
betanalpha: Same comment about better identified error messages as for the dense case.

src/stan/services/util/run_sampler.hpp
```diff
@@ -62,6 +62,9 @@ namespace stan {
         clock_t end = clock();
         double warm_delta_t = static_cast<double>(end - start) / CLOCKS_PER_SEC;
 
+        writer.write_adapt_finish(sampler);
```
betanalpha: If `adapt_finish` is being called here to dump the non-adapted mass matrix then we should rename the function!

src/test/unit/services/check_adaptation.hpp
```diff
@@ -0,0 +1,118 @@
...
+
+      void check_adaptation(const size_t& num_rows, const size_t& num_cols,
+                            const std::vector<double>& param_vals,
+                            stan::test::unit::instrumented_writer& report,
+                            const double& err_margin) {
+        std::vector<std::string> param_strings = report.string_values();
+        size_t offset = 0;
+        for (size_t i = 0; i < param_strings.size(); i++) {
+          offset++;
+          if (param_strings[i].find("lements of inverse mass matrix:")
```
betanalpha: Typo or a very sad inverse mass matrix?

src/test/unit/services/sample/fixed_param_test.cpp
```diff
@@ -153,51 +158,5 @@ TEST_F(ServicesSamplesFixedParam, parameter_checks) {
   EXPECT_DOUBLE_EQ(0.0, parameter_values.back()[2])
     << "final memory_writer should be (0, 0)";
   EXPECT_EQ(return_code, 0);
-
 }
 
-TEST_F(ServicesSamplesFixedParam, output_regression) {
```
betanalpha: What's the motivation for concatenating all of these tests?

src/test/unit/services/sample/hmc_nuts_dense_mass_matrix_test.cpp
```diff
@@ -0,0 +1,264 @@
...
+                                             init,
+                                             parameter,
+                                             diagnostic);
+  EXPECT_EQ(0, return_code);
+
+  stan::io::dump dmp = 
+    stan::services::util::create_ident_dense_mass_matrix(3);
+  stan::io::var_context& inv_mass_matrix = dmp;
+  std::vector<double> dense_vals
+    = inv_mass_matrix.vals_r("mass_matrix");
```
betanalpha: Missing a `inv_` prefix of the name within the var context here and throughout the test?

src/test/unit/services/sample/hmc_nuts_diag_e_test.cpp
```diff
@@ -109,82 +112,7 @@ TEST_F(ServicesSampleHmcNutsDiagE, output_sizes) {
   // iteration.
   EXPECT_EQ("lp__", diagnostic_names[0][0]);
   EXPECT_EQ("accept_stat__", diagnostic_names[0][1]);
-}
```
betanalpha: More concatenation?

betanalpha:
Sorry, this is going to be annoying but I want to maintain a consistent naming scheme throughout the code.  Can you replace all instances of ``inv_mass_matrix`` with ``inv_metric`` or ``inv_e_metric``?  Similarly, instead of referring to a metric as ``ident`` can you use ``unit_e``?  For example, ``ident_dense`` would become ``unit_e_dense`` in line with the naming of the sampler variants.

betanalpha:
Similarly, we want the `var_context` to have an object called `inv_mass_matrix` and not `mass_matrix`, right?  The code doesn't do an inversion but takes the values as they are and assigns them to the inverse metric.  Under renaming this would become `inv_metric` or `inv_e_metric` is anyone prefers a slightly more specific naming.

src/test/unit/services/check_adaptation.hpp
mitzimorris: regex to identify both diagonal and dense mass matrix output messages, with differ.

src/test/unit/services/sample/fixed_param_test.cpp
mitzimorris: the invocation of the sampler is the same across all tests.  all of the outputs can be checked in one go.

src/test/unit/services/sample/hmc_nuts_dense_mass_matrix_test.cpp
mitzimorris: question as to whether or not the var_context object should be called "mass_matrix" or "inv_mass_matrix".   as implemented, expecting "mass_matrix".  if you think "inv_mass_matrix" or "inverse_mass_matrix" is preferable, I will change this.

src/stan/services/util/run_sampler.hpp
mitzimorris: yes, absolutely.  "write_sampler_state"?
could also omit message "Adaptation Finished".

src/test/unit/services/sample/hmc_nuts_dense_mass_matrix_test.cpp
betanalpha: Yes, see the general comment below.

src/test/unit/services/sample/fixed_param_test.cpp
betanalpha: True, but then it can be slightly ambiguous as to which particular failure caused the entire test to fail.  Typically we have avoided monolithic tests and kept them split up roughly by the functionality being tested even with the overhead of repeated instantiations.  It's not a deal-breaker, but I would prefer to maintain the current convention if possible.

src/test/unit/services/check_adaptation.hpp
betanalpha: But shouldn't it be "elements" instead of "lements"?  The latter reading as "laments", hence the sad object.

src/stan/services/util/run_sampler.hpp
betanalpha: Changing the message and the format of the output file will break a bunch of tests.  For now can you add a `write_sampler_state` function that doesn't output "Adaption Finished" and then call that function here?

bob-carpenter:
I think keeping the unit tests as units as Michael suggests is a good idea.

bob-carpenter:
It's OK to change the output if you also fix all the tests.  The key thing is making sure that the posterior analysis tools will still work with the CSV output if that changes.

src/test/unit/services/check_adaptation.hpp
mitzimorris: capitalization

src/stan/services/util/read_dense_mass_matrix.hpp
mitzimorris: var_context.validate_dims does all the work; changed code to include error message

src/stan/services/util/read_diag_mass_matrix.hpp
mitzimorris: same reply

src/stan/services/util/run_sampler.hpp
mitzimorris: function exists already - moved call from within `write_adapt_finish` into `run_sampler`

mitzimorris:
@seantalts any insight as to why header tests failed on Jenkins?   they pass on my machine.
locally I did a "make math-update" but didn't check in updated math lib for this branch.

seantalts:
That sounds like it could be the reason? Try merging develop in?

mitzimorris:
how to merge in develop?
command `git pull origin develop` says my branch is up-to-date with develop.

commands `make math-revert` followed by `make math-update` get rid of git message about changes in lib/stan_math, and so git says there's nothing to commit (or push).

do Jenkins scripts do `make math-revert` followed by `make math-update`?   I thought they used to - but maybe I'm misremembering.  most definitely confused.

seantalts:
It does do that stuff and you're up to date. I dug in and found the same problem we had before - there were artifacts from the 2.14 release(!) around on gelman-group-mac in the workspace for that job. I think I fixed it by:

```
ssh gelman-group-mac.stat.columbia.edu
cd /Users/Shared/Jenkins/gelman-group-mac/workspace/Stan\ Pull\ Request\ -\ Tests\ -\ Header/
rm -rf lib/stan_math_2.14.0
vim .git/modules/lib/stan_math/config 
```
And on the last step pointing git to the correct math directory. 

Jenkins, retest this please.

mitzimorris:
many thanks!  looks like header tests passed!

mitzimorris:
unit tests now failing, checked console log and still seeing math lib for 2.14.0

seantalts:
See if you can ssh in and fix it with the steps I outlined above for the
unit test job?

seantalts:
Jenkins, retest this please.

seantalts:
Jenkins, retest this please.

seantalts:
Okay I did something more robust - I changed Jenkins to delete this build's workspace before running. I would suggest trying that in the future instead of my steps above that require a login on the machines.

mitzimorris:
many thanks!

betanalpha:
The services code still uses `mass_matrix` instead of `metric`.  I strongly prefer having consistent naming everywhere and I have an ulterior motive to wean people off of the term "mass matrix" as it's not more intuitive than "metric" but is more limiting on the formal side.

betanalpha:
Everything else looks good except for one design element.  I strongly prefer having the `base_hmc` class ignorant of the metric and hence the values not specified through the constructor.  Instead I would rather have the `ps_point` derivations construct their metrics with default values and then have the service functions assign the initialized values after pulling them out of the `var_context`, similar to how the initial step size is set.  For example, 

```
Eigen::MatrixXd inv_metric;
        try {
          inv_metric =
            util::read_dense_metric(init_metric, model.num_params_r(), error_writer);
          util::validate_dense_metric(inv_metric error_writer);
        } catch (const std::domain_error& e) {
          return error_codes::CONFIG;
        }

stan::mcmc::adapt_dense_e_nuts<Model, boost::ecuyer1988>
          sampler(model, rng);
sampler.set_nominal_stepsize(stepsize);
sampler.z().inv_e_metric = inv_metric;
```

bob-carpenter:
I'm all for keeping the names consistent and "metric" is fine
by me.

I also like the idea of keeping the stepsize and metric on
the same footing in terms of where and how they're set.

That domain_error should get its message printed out somehow.
Why in the world do we have error codes?  The whole point of 
having exceptions is that it avoids having to pass error codes
around, which is very finicky and error prone.

The reason Mitzi and I both gravitate to putting things in constructors
is that it's easiest to construct objects in working states.  Here,
the mass matrix (er, metric) is already going to be mutable, so setting
it after the constructor is no big deal.  On the other hand, what you'll
have to do when you construct that sampler is set up a dummy inverse metric
or the sampler won't be in a usable state.  That would be nice to avoid,
but probably won't be too much overhead as this is all double values
and this is just one step.  It just feels sloppy when you're used to keeping
your classes in order on construction.  What we absolutely don't want to do
is construct something in a non-working state, or try to have it in
a non-working state and then check the workability and add a default init
if needed---it's hard to keep things consistent with that kind of logic.

I think the method name "set_nominal_stepsize" could be more
precise.  I take it to be referring the imprecise sense of "nominal",
but can't it just be "set_stepsize" or "set_current_stepsize" to imply
it might change?

And speaking of format, how about moving the "=" to start a line rather
than end?  It's easier to scan what's going on (the goal of all of my
code comments) and it also matches standard math typesetting!

mitzimorris:
> I have an ulterior motive to wean people off of the term "mass matrix" as it's not more intuitive than "metric" but is more limiting on the formal side

"mass matrix" is the term that's used in the manual and in the interfaces, and as the services code is the code which will be called from the interfaces, I think that it makes it easier for the interface writer to quickly grok which argument gets plugged in where.

bob-carpenter:
I think it should be consistent and I also think Michael should decide what that consistent terminology is the way Andrew decides what all of our stat terminology is (despite people continuing to ignore his "sample is a set of draws" distinction).  

We can update the manual and we're about to update all the interfaces to Stan 3, so this is the time to make everything consistent going forward.

mitzimorris:
agreed that "metric" is the general case.
however, the services classes which Micheal refers to are these:

```
find src/stan/services -name "*.hpp" -exec grep -lw "mass_matrix" {} \;
src/stan/services/sample/hmc_nuts_dense_e.hpp
src/stan/services/sample/hmc_nuts_dense_e_adapt.hpp
src/stan/services/sample/hmc_nuts_diag_e.hpp
src/stan/services/sample/hmc_nuts_diag_e_adapt.hpp
src/stan/services/sample/hmc_static_dense_e.hpp
src/stan/services/sample/hmc_static_dense_e_adapt.hpp
src/stan/services/sample/hmc_static_diag_e.hpp
src/stan/services/sample/hmc_static_diag_e_adapt.hpp
```

exactly the Euclidian cases where the metric is either a dense or diagonal mass matrix.

bob-carpenter:
The metric will be unit, dense, or diag, so the top level function names are fine.  We just need to make the argument names all consistent.

The math works out the same either way, but I find the metric way of thinking easier because of the direct connection to distance (same quadratic form as in the multivariate normal).  That may just be because I learned metric topologies in analysis but never got a proper introduction to physics :-)

betanalpha:
@mitzimorris Thanks for making the changes.

@bob-carpenter I'm aware of the benefits of assignment-at-construction but I am trying to maintain the current encapsulation of the code.  In particular, the dynamic implementation of HMC doesn't need to know anything, and hence shouldn't know anything, about the choice of metric/kinetic energy.  

One way I had thought about maintaining the encapsulation and getting assignment-at-construction is give the sampler a constructor that takes in the appropriate ps_point derivation.  The assign-after-construction approach works fine here, however, as the internal point of the sampler is constructed in a valid state so everything will behave okay even if the later assignment were to fail.

betanalpha:
I am working on the current merge disaster, plus some additional renaming.

betanalpha:
I resolved the merge conflicts with develop, but I still want to rename all of the instances of `mass_matrix` to `euclidean_metric`.  I'll take care of that tomorrow.

betanalpha:
Okay, we're good to go.  Thanks @mitzimorris for the extensive unit tests that drastically facilitated the renaming.

PR #2289, opened 2017-04-24
Feature/2224 stan includes

bob-carpenter:
#### Submisison Checklist

#### Summary

Add include mechanism to Stan language.

#### Intended Effect

Allow includes with `#include <file-path>` very much like the ones in C++.  

Saves all the paths to report original file and line number and include history with error reports in both the parser and at run time.  

Checks for circularity to avoid blowing out memory.  

#### How to Verify

Unit tests.  

#### Side Effects

Upstream interfaces will need to specify a sequence of paths in which to find include files.

#### Documentation

Yes.  Manual and code.

#### Reviewer Suggestions

Anyone.

#### Copyright and Licensing

Columbia University


syclik:
Jenkins, retest this please.

src/docs/stan-reference/language.tex
```diff
@@ -1,7 +1,158 @@
...
+#include b.stan
+\end{stancode}
+%
+and \code{b.stan} contains
+%
+\begin{stancode}
+#include a.stan
+\end{stancode}
+%
+The result of processing this file will be empty, becuase
```
syclik: spelling: "becuase"

src/stan/io/program_reader.hpp
```diff
@@ -0,0 +1,263 @@
+#ifndef STAN_IO_PROGRAM_READER_PROGRAM_READER_HPP
```
syclik: please change the header guard to: `STAN_IO_PROGRAM_READER_HPP` (duplicated `PROGRAM_READER_`)

src/stan/io/program_reader.hpp
```diff
@@ -0,0 +1,263 @@
...
+     */
+    class program_reader {
+    public:
+      /**
+       * A pair for holding a path and a line number.
+       */
+      typedef std::pair<std::string, int> path_line_t;
+
+      /**
+       * Orered sequence of path and line number pairs.
```
syclik: typo: Ordered

syclik:
@bob-carpenter, is there an easy way to generate a separate file with the include files? 

It's ok if it requires a call to a second function.

From CmdStan, unless we know what the main `.stan` program calls, we won't know to rebuild the executable on a change to any of the includes without having a list.

wds15:
Should we consider to built some of the rstanarm models using this functionality as a test?

I think rstanarm stuff does make heavy use of includes and would be a good benchmark.

bob-carpenter:
RStanArm presumably handles Boost dependencies through RStan and hence through BH, and BH is not under our control.

bob-carpenter:
Ack.  I hadn't realized this never got merged and now it has comments.  I can address the main comments from @syclik.  If I understand what @syclik is asking for here

https://github.com/stan-dev/stan/pull/2289#issuecomment-299261361

it would be a list of files like this:

```
main.stan
foo/bar/baz.stan
...
```

with one relative path per file (the relative paths of the includes will be based on however they're found)?  I can do that.  It will require us to take another argument to the compiler and parser in the form of a `std::ostream&` and the files will be written one per line in ASCII with forward-slashes as delimeters (or whatever might have been input if it's mixed on Windows) terminated by `\n` with a final `\n`.  

@syclik, could you confirm that's what you wanted?

syclik:
> @syclik, could you confirm that's what you wanted?

@bob-carpenter, yes. If we had that, we could get the makefile in CmdStan to only build the executable when something in the list of files has changed. If not, we'd have to rebuild the executable every time just to be safe or have the user keep track on their own.

We don't need to do it for this pull request if it takes a bit of effort. I can create a new feature request.

mitzimorris:
made minor changes requested per code review.
I could take on suggestion:  https://github.com/stan-dev/stan/pull/2289#issuecomment-29926136, but would prefer to do that in its own issue.

mitzimorris:
filed new issue describing list of included files for make  https://github.com/stan-dev/stan/issues/2324

bob-carpenter:
Thanks!

PR #2295, opened 2017-04-28
Silence unused typedef warnings, new in gcc 4.9.

seantalts:
http://discourse.mc-stan.org/t/new-error-on-stan-develop-during-windows-unit-tests/365/5

#### Submisison Checklist

#### Copyright and Licensing

Please list the copyright holder for the work you are submitting (this will be you or your assignee, such as a university or company): Dual Space LLC

### DIFF
--- a/makefile
+++ b/makefile
@@ -25,7 +25,7 @@ C++11 = false
 ##
 # Set default compiler options.
 ## 
-CFLAGS = -I src -isystem $(EIGEN) -isystem $(BOOST) -isystem $(CVODES)/include -isystem $(MATH) -Wall -DBOOST_RESULT_OF_USE_TR1 -DBOOST_NO_DECLTYPE -DBOOST_DISABLE_ASSERTS -DFUSION_MAX_VECTOR_SIZE=12 -DNO_FPRINTF_OUTPUT -pipe 
+CFLAGS = -I src -isystem $(EIGEN) -isystem $(BOOST) -isystem $(CVODES)/include -isystem $(MATH) -Wall -DBOOST_RESULT_OF_USE_TR1 -DBOOST_NO_DECLTYPE -DBOOST_DISABLE_ASSERTS -DFUSION_MAX_VECTOR_SIZE=12 -DNO_FPRINTF_OUTPUT -pipe -Wno-unused-local-typedefs
 CFLAGS_GTEST = -DGTEST_USE_OWN_TR1_TUPLE
 LDLIBS = 
 LDLIBS_STANC = -Lbin -lstanc


makefile
```diff
@@ -25,7 +25,7 @@ C++11 = false
 ##
 # Set default compiler options.
 ## 
-CFLAGS = -I src -isystem $(EIGEN) -isystem $(BOOST) -isystem $(CVODES)/include -isystem $(MATH) -Wall -DBOOST_RESULT_OF_USE_TR1 -DBOOST_NO_DECLTYPE -DBOOST_DISABLE_ASSERTS -DFUSION_MAX_VECTOR_SIZE=12 -DNO_FPRINTF_OUTPUT -pipe 
+CFLAGS = -I src -isystem $(EIGEN) -isystem $(BOOST) -isystem $(CVODES)/include -isystem $(MATH) -Wall -DBOOST_RESULT_OF_USE_TR1 -DBOOST_NO_DECLTYPE -DBOOST_DISABLE_ASSERTS -DFUSION_MAX_VECTOR_SIZE=12 -DNO_FPRINTF_OUTPUT -pipe -Wno-unused-local-typedefs
```
syclik: @seantalts, I'd leave this line alone and put the particular flag to `make/os_win`. Not all the flags are compatible with every version of the compiler and this might cause problems with other builds.

seantalts: This isn't a windows thing, just a gcc thing - should we make a `make/cc_gcc` or something? 
This flag also appears to be accepted (and ignored?) by clang, for what it's worth.

syclik: I'm worried about different versions of gcc, not necessarily clang. If I remember correctly, bad options will make it kick out.

Re: `make/cc_gcc`, I think I'm not really a fan. I'm looking at the options and there's a serious interaction between OS and compiler. If we go in that direction, what makes sense is to do: `make/win_gcc`, `make/win_clang`, etc. I'd be more inclined to (as much as it hurts saying this) repeat the compiler option in 3 places (`os_*`) and be done with it. We don't change it that often.

seantalts: We don't support versions of gcc from before this option was introduced now, though - we're starting the move to C++11 next week, etc.

syclik: What versions of gcc do we support? Are we explicitly no longer supporting anything under gcc 4.9?

seantalts: I believe so - Windows was the laggard and we're now assuming gcc 4.9.3+ with the C++11 switch.

I'm not really sure how the plan you're proposing deals with the risk that old versions of `gcc` will segfault with options it doesn't know about - seems like per OS or per compiler makefiles don't take into account compiler versions and we need this flag on `gcc` > 4.9 ish (and perhaps want to not have it before 4.9).

seantalts: Okay, good news - I had a friend test with `gcc 4.4.7` and this compiler option and it works fine, i.e. discards the unused option and compiles successfully.

syclik: I don't think it segfaults with unknown compiler options, but great! If it works fine, then I'm fine. I wonder what I was remembering causing problems. Doesn't matter.

syclik:
btw, perhaps it's just best to disable the performance tests.

seantalts:
I've been trying to figure out why they're failing all day! Was pretty sudden though I have been messing about with Jenkins, but it's complaining about not finding an xml file that's definitely where it's looking...

PR #2301, opened 2017-05-04
Feature/issue 2240 update gp chapter

rtrangucci:
#### Submisison Checklist

#### Summary

Updates GP chapter to latest GP coding standards, and prior standards.

#### Reviewer Suggestions

@betanalpha @bob-carpenter @avehtari 

#### Copyright and Licensing

Rob Trangucci

### DIFF
--- a/src/docs/stan-reference/examples.tex
+++ b/src/docs/stan-reference/examples.tex
@@ -6597,7 +6597,7 @@ determined by the prior on \code{sigma}.
 \chapter{Gaussian Processes}\label{gaussian-processes.chapter}
 
 \noindent
-Gaussian process are continuous stochastic processes and thus may be
+Gaussian processes are continuous stochastic processes and thus may be
 interpreted as providing a probability distribution over functions.  A
 probability distribution over continuous functions may be viewed,
 roughly, as an uncountably infinite collection of random variables,
@@ -6605,11 +6605,11 @@ one for each valid input.  The generality of the supported functions
 makes Gaussian priors popular choices for priors in general
 multivariate (non-linear) regression problems.
 
-The defining feature of a Gaussian process is that the distribution of
-the function's value at a finite number of input points is a
-multivariate normal distribution.  This makes it tractable to both fit
-models from finite amounts of observed data and make predictions for
-finitely many new data points. 
+The defining feature of a Gaussian process is that the joint distribution of
+the function's value at a finite number of input points is a multivariate
+normal distribution.  This makes it tractable to both fit models from finite
+amounts of observed data and make predictions for finitely many new data
+points. 
 
 Unlike a simple multivariate normal distribution, which is
 parameterized by a mean vector and covariance matrix, a Gaussian
@@ -6619,13 +6619,14 @@ return a mean vector and covariance matrix which provide the mean and
 covariance of the outputs corresponding to those input points in the
 functions drawn from the process.
 
-Gaussian processes can be encoded in Stan by implementing their mean
-and covariance functions and plugging the result into the Gaussian
-form of their sampling distribution.  This form of model is easy to
-understand and may be used for simulation, model fitting, or posterior
-predictive inference.  More efficient Stan implementation for the
-basic (non-logistic) regression applies a Cholesky-factor
-reparameterization of the Gaussian and computes the posterior
+Gaussian processes can be encoded in Stan by implementing their mean and
+covariance functions and plugging the result into the Gaussian form of their
+sampling distribution, or by using the specialized covariance functions
+outlined below.  This form of model is easy to understand and may be used for
+simulation, model fitting, or posterior predictive inference. A more efficient
+Stan implementation for the GP with a normally distributed outcome marginalizes
+over the latent Gaussian process, and applies a Cholesky-factor
+reparameterization of the Gaussian to compute the likelihood and the posterior
 predictive distribution analytically.
 
 After defining Gaussian processes, this chapter covers the basic
@@ -6659,17 +6660,17 @@ a positive-definite matrix for any input $x$.%
   does not have unconstrained support.}
 
 A popular covariance function, which will be used in the
-implementations later in this chapter, is a generalized, squared
-exponential function,
+implementations later in this chapter, is a generalized, 
+exponented quadratic function,
 \[
 k(x)_{i, j}
-= \eta^2 
+= \alpha^2 
 \exp \left(
-- \rho^2 \sum_{d=1}^D (x_{i,d} - x_{j,d})^2
+- \dfrac{1}{2 \rho^2} \sum_{d=1}^D (x_{i,d} - x_{j,d})^2
 \right)
 + \delta_{i, j} \sigma^2,
 \]
-where $\eta$, $\rho$, and $\sigma$ are hyperparameters defining the
+where $\alpha$, $\rho$, and $\sigma$ are hyperparameters defining the
 covariance function and where $\delta_{i, j}$ is the Kronecker delta
 function with value 1 if $i = j$ and value 0 otherwise; note that this
 test is between the indexes $i$ and $j$, not between values $x_i$ and
@@ -6689,13 +6690,11 @@ The summation involved is just the squared Euclidean distance between
 $x_i$ and $x_j$ (i.e., the $L_2$ norm of their difference, $x_i -
 x_j$). This results in support for smooth functions in the process.
 The amount of variation in the function is controlled by the free
-hyperparameters $\eta$, $\rho$, and $\sigma$.  
+hyperparameters $\alpha$, $\rho$, and $\sigma$.  
 
 Changing the notion of distance from Euclidean to taxicab distance
 (i.e., an $L_1$ norm) changes the support to functions which are
 continuous but not smooth.
-
-
 
 \section{Simulating from a Gaussian Process}
 
@@ -6708,7 +6707,7 @@ The Stan model defines the mean and covariance functions in a
 transformed data block and then samples outputs $y$ in the model using
 a multivariate normal distribution.  To make the model concrete, the
 squared exponential covariance function described in the previous section
-will be used with hyperparameters set to $\eta^2 = 1$, $\rho^2 = 1$,
+will be used with hyperparameters set to $\alpha^2 = 1$, $\rho^2 = 1$,
 and $\sigma^2 = 0.1$, and the mean function $m$ is defined to always
 return the zero vector, $m(x) = {\bf 0}$.  Consider the following
 implementation of a Gaussian process simulator.%
@@ -6723,13 +6722,40 @@ data {
 }
 transformed data {
   vector[N] mu;
+  matrix[N, N] Sigma;
+  mu = rep_vector(0, N);
+  for (i in 1:(N - 1)) {
+    Sigma[i,i] = 1 + 0.1;
+    for (j in (i + 1):N) {
+      Sigma[i, j] = exp(-1/2 * pow(x[i] - x[j], 2));
+      Sigma[j, i] = Sigma[i, j];
+    }
+  }
+  Sigma[N,N] = 1 + 0.1;
+}
+parameters {
+  vector[N] y;
+}
+model {
+  y ~ multi_normal(mu, Sigma);
+}
+\end{stancode}
+%
+The above model can also be written more compactly using the specialized
+covariance function that implements the exponentiated quadratic kernel. 
+%
+\begin{stancode}
+data {
+  int<lower=1> N;
+  real x[N];
+}
+transformed data {
+  vector[N] mu;
   cov_matrix[N] Sigma;
-  for (i in 1:N) 
-    mu[i] = 0;
-  for (i in 1:N) 
-    for (j in 1:N)
-      Sigma[i, j] = exp(-pow(x[i] - x[j], 2)) 
-                    + i == j ? 0.1 : 0.0;
+  mu = rep_vector(0, N);
+  Sigma = cov_exp_quad(x, 1, 1);
+  for (i in 1:N)
+    Sigma[i, i] = Sigma[i, i] + 0.1;
 }
 parameters {
   vector[N] y;
@@ -6742,14 +6768,12 @@ model {
 The input data is just the vector of inputs \code{x} and its size
 \code{N}.  Such a model can be used with values of \code{x} evenly
 spaced over some interval in order to plot sample draws of functions
-from a Gaussian process.  The covariance matrix \code{Sigma} is not
-being computed efficiently here; see
-Section~\refsection{fit-gp} for a better approach.
+from a Gaussian process.  
 
 \subsection{Multivariate Inputs}
 
-Only the covariance function's distance computation needs to change in
-moving from a univariate model to a multivariate model.%
+Only the input data needs to change in moving from a univariate model to a
+multivariate model.%
 %
 \footnote{The model is available in the Stan example model repository;
 see \url{http://mc-stan.org/documentation}.}
@@ -6764,27 +6788,20 @@ data {
 }
 transformed data {
 ...
-      Sigma[i, j] = exp(-dot_self(x[i] - x[j]))
-                    + i == j ? 0.1 : 0.0;
 ...
 \end{stancode}
 %
 The data is now declared as an array of vectors instead of an array of
-scalars; the dimensionality \code{D} is also declared.  The squared
-Euclidean distance calculation is done using the \code{dot\_self}
-function, which returns the dot product of its argument with itself, here
-\code{x[i]~-~x[j]}.
+scalars; the dimensionality \code{D} is also declared.  
 
-In the remainder of the chapter, univariate models will be used for
-simplicity, but any of them could be changed to multivariate in the
-same way as the simple sampling model.  The only extra computational
-overhead from a multivariate model is in the distance calculation,
-which is only done once when the transformed data block is run after
-the data is read.
+In the remainder of the chapter, univariate models will be used for simplicity,
+but any of the models could be changed to multivariate in the same way as the
+simple sampling model. The only extra computational overhead from a
+multivariate model is in the distance calculation.
 
 \subsection{Cholesky Factored and Transformed Implementation}
 
-A much more efficient implementation of the simulation model can be
+A more efficient implementation of the simulation model can be
 coded in Stan by relocating, rescaling and rotating an isotropic unit
 normal variate.  Suppose $z$ is an an isotropic unit normal variate
 \[
@@ -6839,8 +6856,9 @@ univariate distribution for efficiency; this specifies that each
 vector \code{y} is then defined as a generated quantity using a direct
 encoding of the transform described above.
 
-
 \section{Fitting a Gaussian Process}\label{fit-gp.section}
+
+\subsection{GP with a normal outcome}
 
 The hyperparameters controlling the covariance function of a Gaussian
 process can be fit by assigning them priors, then computing the
@@ -6850,8 +6868,8 @@ to have reasonably small values, broad half-Cauchy distributions act
 as quite vague priors which could just as well be uniform over a
 constrained range of values.  The priors on the parameters should be
 defined based on prior knowledge of the scale of the output values
-($\eta$), the scale of the output noise ($\sigma$), and the scale at
-which distances are measured among inputs ($1/\rho$).
+($\alpha$), the scale of the output noise ($\sigma$), and the scale at
+which distances are measured among inputs ($\rho$).
 
 The Stan code%
 %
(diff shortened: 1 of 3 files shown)


rtrangucci:
@betanalpha @avehtari I just updated the GP chapter, it'd be great if you guys could take a look at the Priors for GP Hyperparameters subsection, and let me know if there's anything you'd like to change.

bob-carpenter:
Thanks---I can review this for general form or what not if you need me to, but I think you want one of the GP experts involved as you pinged.

avehtari:
I'll check this in a few days.

betanalpha:
I didn’t differentiate between new and old content, so
I’m sure some of this isn’t actually targeted at Rob’s
contributions…  :-p

When you introduce a code snippet the proceeding sentence
should end in a colon, not a period.  

In general I think we need to restructure this presentation to
elucidate the difference between exact simulation and MCMC
fitting/simulation.  For example, start with the exact simulation
in the generated quantities block based on the analytic
results and then introduce the MCMC version with a
comparative discussion.  Then when talking about fitting
the hyperparameters you can motivate why you need the
MCMC version as the analytic result falls apart.

Or just try to be more explicit in the text with the current
organization.

Multivariate Inputs —

Perhaps worth noting that Gaussian processes start to behave 
poorly for more than a few input dimensions.

Cholesky Factored and Transformed Implementation —


Before discussing the Cholesky factor simulation would it
not be better to do the multi_normal_rng?  Really 17.2
should be done in generated quantities, no?

The confusion is that if simulating analytically in generated 
quantities then there is no difference in performance.  The 
difference is when you’re fitting with MCMC.

Fitting a Gaussian Process — 

In that first sentence I’d add something to the extent of
“We can fit the hyperparameters like we fit any parameter
in Bayesian inference: we add the parameter to the parameters
block and assign a reasonable prior.”

This particular parameterization is a bit tricky because the
scales implied by the squared parameters aren’t the natural
ones to reason about.

Priors for Gaussian Process Parameters —

This seems to conflict with all of the results of our recent
studies, especially about the length scale getting too small!
The emphasis on rho =  1 / l^2 is also confusing.

Logistic Gaussian Process Regression —

When discussing the Cholesky decomposition I’d note the
importance of the posterior geometry and refer to the funnel 
section in the manual as a related issue.

When discussing sigma_sq I’d mention how the noise term
in a latent model takes on a “nugget” interpretation that may
still be useful for computational stability.

rtrangucci:
@betanalpha I'm confused, rho is length-scale in the writeup. Where 

I realized I missed this in the "Analytic form of Joint Predictive Inference" subsection, so I need to change that formula, but beyond that, the "Priors for hyperparamters" was meant to mirror our discussion the other day...

betanalpha:
It explicitly says that small length scales are okay!  That’s is
the opposite of our conclusions the other day.  In particular,
there’s no discussion of zero-avoiding priors.

rtrangucci:
Are you sure you're looking at the right diff? There's a lot in there about
avoiding small length-scales!

This is the first paragraph about priors for lengths-scales:

GPs are a flexible class of priors, and as such, can represent a wide
spectrum
of functions. Their flexibility can lead to numerical problems when
estimating
a model in Stan using Euclidean Hamiltonian Monte Carlo. Length-scales that
are
smaller than the scale of the input data lead to a perfect fit with the
observed data. This can require small step sizes when exploring this region
of the parameter space. In order to limit the amount of posterior mass that
resides in this region, the prior for length-scale should strongly
regularize
away from length-scales that are far smaller than the scale of the input.



Rob

On Fri, May 5, 2017 at 3:48 PM, Michael Betancourt <notifications@github.com

avehtari:
On 2017-05-05 22:23, Michael Betancourt wrote:
> Perhaps worth noting that Gaussian processes start to behave
> poorly for more than a few input dimensions.

I guess you have much stricter definition for "well behaving GP" or much looser 
definition of "a few" than me. GP is a useful in a regression similarly as 
linear models (ie they are likely to be wrong, but useful). I've seen many 
examples with much more dimensions than a few, were GP is better than linear 
models, but the posterior inference is easier than for many other non-linear 
models (like, neural networks and trees).  If you would like to have a comment 
related to GPs and the number of input dimensions, it would be nice if you 
could be more explicit than just "poorly".


Aki

betanalpha:
LOOK SOMETIMES IT’S HARD TO GIT PULL AFTER
CHECKING OUT A BRANCH.  :-(

Sorry, I’ll take a look at your actual changes later tonight
when I get some free time.

rtrangucci:
Ha, that's fine. Your comments were useful anyway because some of them do
pertain to my writeup. Looking forward to your thoughts on the chapter!


Rob

On Fri, May 5, 2017 at 3:55 PM, Michael Betancourt <notifications@github.com

betanalpha:
Wow, so much better!  :-)

Just a few small and largely ignorable comments.

Intro — 

“easy to understand” can come across as patronizing.
At the very least replace it with something like 
“straightforward”.

17.2. Simulating from a Gaussian Process —

In the second model replace

```
for (i in 1:N)
        Sigma[i, i] = Sigma[i, i] + 0.1;
```

with

```
Sigma = Sigma + diag_matrix(rep_vector(0.1, N));
```


Cholesky Factored and Transformed Implementation —

In the second equation the mu on the right hand
side should probably have an explicit (x) argument
to match the covariance function.

17.3. Fitting a Gaussian Process —

The text talks a lot about Cauchy priors when only
one is used.  I’d give sigma a normal and replace
the explanatory paragraph with a reference to the
upcoming prior section.

As before, the diagonal increment loop can be replaced
with a diag_matrix(rep_vector()) call in one line.

Priors for Gaussian Process Parameters —

In the third paragraph I’d add a connecting sentence
noting that the overfitting induces a sharp ridge
in the likelihood that persists in the posterior,
and it’s that ridge that can require infinitely small
step sizes.

In the fourth paragraph I’d add a sentence expounding
on what “small length scales with respect to the input
data” means.  Perhaps flowing from the discussion in
the above paragraph, like “When the length scale falls
below the spacing of the input covariates the functions
supported by the GP will interpolate almost exactly
between the input data.  This provides such a good fit
to the data that the likelihood strongly peaks, inducing 
a sharp ridge that persists in the posterior and frustrates
MCMC exploration.  In order to avoid these small length
scales we need a prior that favors length scales more
appropriate to the input data.”

In the fifth paragraph, the last sentence should be
a _generalized_ inverse Gaussian with p = 1.

When talking about up crossings I’d mention that this
intuition is limited to one-dimensional covariates.

Analytical Form of Joint Predictive Inference —

Should the link to footnote 8 be after the sentence
instead of before it?

bob-carpenter:
> “easy to understand” can come across as patronizing.
> At the very least replace it with something like 
> “straightforward”.

Agreed.  The real problem is when the reader doesn't understand something marked "easy to understand".  A lot of math is "and the next step is obvious" where it takes you a day to fill in the steps.

src/docs/stan-reference/examples.tex
```diff
@@ -6892,25 +6893,34 @@ \section{Gaussian Process Regression}
...
+
+The hyperparameter $\rho$ is the \emph{length-scale}, and corresponds to the
+frequency of the functions represented by the Gaussian process prior with
+respect to the domain. Values of $\rho$ closer to zero lead the GP to represent
+high-frequency functions, whereas larger values of $\rho$ lead to low-frequency
+functions. The hyperparameter $\alpha$ is the \emph{marginal standard
+deviation}. It controls the magnitude of the range of the function represented
+by the GP. If you were to take the standard deviation of many draws from the GP
+prior at a single input $x$ conditional on one value of $\alpha$ one would
+recover $\alpha$.
```
bbbales2: Does this assume $\sigma$ is zero?

src/docs/stan-reference/examples.tex
```diff
@@ -6976,14 +7011,12 @@ \section{Simulating from a Gaussian Process}
...
 
 \subsection{Multivariate Inputs}
 
-Only the covariance function's distance computation needs to change in
-moving from a univariate model to a multivariate model.%
+Only the input data needs to change in moving from a univariate model to a
+multivariate model.%
 %
 \footnote{The model is available in the Stan example model repository;
 see \url{http://mc-stan.org/documentation}.}
```
bbbales2: Sorry if I'm missing something, but I go this link and I just don't find the GP example models. There are a lot of these footnotes repeated through the chapter.

src/docs/stan-reference/examples.tex
```diff
@@ -7073,67 +7099,89 @@ \subsection{Cholesky Factored and Transformed Implementation}
...
-constrained range of values.  The priors on the parameters should be
-defined based on prior knowledge of the scale of the output values
-($\eta$), the scale of the output noise ($\sigma$), and the scale at
-which distances are measured among inputs ($1/\rho$).
-
-The Stan code%
+\subsection{GP with a normal outcome}
+
+The full generative model for a GP with a normal outcome, 
+$y \in \R^N$, with inputs $x \in \R^N$, for a finite $N$:
```
bbbales2: Is N ever infinite?

src/docs/stan-reference/examples.tex
```diff
@@ -7073,67 +7099,89 @@ \subsection{Cholesky Factored and Transformed Implementation}
...
+variable GP, while the latter will be called the marginal likelihood
+GP.
+
+The hyperparameters controlling the covariance function of a Gaussian process
+can be fit by assigning them priors, like we have in the generative models
+above, and then computing the posterior distribution of the hyperparameters
+given observed data. The priors on the parameters should be defined
+based on prior knowledge of the scale of the output values ($\alpha$), the
+scale of the output noise ($\sigma$), and the scale at which distances are
+measured among inputs ($\rho$). See \refsection{} for more information
```
bbbales2: Empty refsection

src/docs/stan-reference/examples.tex
```diff
@@ -7145,87 +7193,325 @@ \section{Fitting a Gaussian Process}\label{fit-gp.section}
...
+  {
+    matrix[N, N] Sigma = cov_exp_quad(x, alpha, rho);
+    matrix[N, N] L_Sigma;
+
+    // diagonal elements
+    for (k in 1:N)
+      Sigma[k, k] = Sigma[k, k] + delta; 
+
+    L_Sigma = cholesky_decompose(Sigma);
+    f = L_sigma * eta;
```
bbbales2: Capitalize the Sigma in L_sigma to match other variables

src/docs/stan-reference/examples.tex
```diff
@@ -7145,87 +7193,325 @@ \section{Fitting a Gaussian Process}\label{fit-gp.section}
...
-expressing \code{1~/~rho\_sq}.  An alternative would be to just divide
-by \code{inv\_rho\_sq}.
+hyperparameters, allowing us to learn the hyperparameters. 
+
+We have used the Cholesky parameterized \distro{MultiNormal} rather than the
+standard \distro{MultiNormal} because it allows us to the
+\code{cholesky\_decompose} function which has been optimized for both small and
+large matrices. When working with small matrices the differences in
+computational speed between the two approaches will not be noticeable, but for
+larger matrices ($N \gtrsim 100$) the Cholesky decomposition version will be faster.
```
bbbales2: For a single draw from multi_normal, is this true? Isn't multi_normal just doing a Cholesky internally to do the solve?

src/docs/stan-reference/examples.tex
```diff
@@ -7145,87 +7193,325 @@ \section{Fitting a Gaussian Process}\label{fit-gp.section}
...
+\end{align} 
+
+we'll likely want to constrain large length-scales as well.  A length scale
+that is larger than the scale of the data yields a GP posterior that is
+practically linear (with respect to the particular covariate) and increasing
+the length scale has little impact on the likelihood. This will introduce
+nonidentiability in our model, as both the fixed effects and the GP will
+explain similar variation. In order to limit the amount of overlap between the
+GP and the linear regression, we should use a prior with a sharper right tail
+to limit the GP to higher-frequency functions. We can use an inverse Gaussian,
```
bbbales2: Is the inverse Gaussian (I'm assuming it's this: https://en.wikipedia.org/wiki/Inverse_Gaussian_distribution) available in Stan? (I wanted to use it but didn't find it)

src/docs/stan-reference/examples.tex
```diff
@@ -7145,87 +7193,325 @@ \section{Fitting a Gaussian Process}\label{fit-gp.section}
...
+length-scales that are small with respect to the input data.  
+
+If we're using the GP as a component in a larger model that includes an overall
+mean and fixed effects for the same variables we're using as the domain for the
+GP, i.e.:
+
+\begin{align}
+  f & \sim \distro{MultiNormal}\left(0, K_{\rho,\alpha}(x)\right) \\
+  y_i & \sim \distro{Normal}(\beta_0 + x_i \beta_{[1:D]} + f_i, \sigma) \, \forall i \in \{1, \dots, N\} \\
+  & x_i^T, \beta_{[1:D]} \in \reals{R}^D,\, x \in \reals^{N \times D}, f \in \
```
bbbales2: At the end of the line, f \in ?

Also \reals{R}^D looks like fancy_R R^D when rendered

src/docs/stan-reference/examples.tex
```diff
@@ -7145,87 +7193,325 @@ \section{Fitting a Gaussian Process}\label{fit-gp.section}
...
+A priori, the closer $\rho_d$ is to zero, the more nonlinear the
+conditional mean in dimension $d$ is.  A posteriori, the actual dependencies
 between $x$ and $y$ play a role.  With one covariate $x_1$ having a
 linear effect and another covariate $x_2$ having a nonlinear effect,
-it is possible that $\rho_1 < \rho_2$ even if the predictive relevance
-of x1 is higher. That is, $\rho$ is related to the relevance, but it
-is more accurate to say that it measures the nonlinearity or the
-expected number of upcrossings \cite[page~80]{RasmussenWilliams:2006}.
+it is possible that $\rho_1 > \rho_2$ even if the predictive relevance
+of x1 is higher \cite[page~80]{RasmussenWilliams:2006}.
```
bbbales2: x1 -> $x_1$

bob-carpenter:
> Also, there's a place in the text that says "NxN unit matrix". I think that needs changed to "NxN identity matrix".

"unit" is a synonym for "multiplicative identity" in algebra;  see https://en.wikipedia.org/wiki/Identity_matrix

bob-carpenter:
@rtrangucci --- let me know when you're done and I'll merge.  Don't kill yourself trying to make it perfect.  And thanks!

bbbales2:
Also, real quick, is there more to the "Multiple-output Gaussian processes" section or is that leftover text?

avehtari:
Note that covariance functions which are positive (semi) definite with L2-norm, are not necessarily positive (semi) definite with L1-norm.

Poisson GP: maybe add the expected count term (which correspnds to centering the data)
'y ~ poisson_log(a + f);'
where a is the log of the expected number of counts.

Logistic Gaussian Process Regression: maybe mention the bias term
'y ~ bernoulli_logit(a + f);'
where a can be data or a parameter, but makes the model perform better in case of inbalanced classes (more observation in one class than in the other)

"Perhaps most importantly, the parameters ρ and α are weakly identified"
 should we have a reference?

rtrangucci:
@betanalpha @avehtari can you guys take a look at the prior recommendations section and the last section on multiple-output GPs? If those look good, I'm happy with the chapter.

betanalpha:
Priors for Gaussian Process Parameters:

Oxford comma all the way!

The first sentence of the second paragraph is awkward.
I’d change it to simply “Perhaps most importantly, the
parameters rho and alpha are only weakly identified
(Zhang, 2004).  The ratio…”

Priors for length-scale:

I thin that this section can be cleaned up a bit.  Perhaps
something like “For length scales below the minimum
spacing of the covariates the GP likelihood plateaus.
Unless regularized by a prior, this flat likelihood induces
considerable posterior mass at small length scales where
the observation variance drops to zero and the functions
supported by the GP being to exactly interpolate between
the input data.  The resulting posterior not only significantly
overfits to the input data, it also becomes hard to accurately
sample using Euclidean HMC.”

After Eqn 17.3: “then we likely”.  Also this entire paragraph
doesn’t quite make sense — you start by saying that you
don’t need further constraints but then go on to talk about
soft regularization with a prior.  Perhaps you meant to say
“further constraints beyond penalizing small length scales”?

When discussing inv gamma you should note it’s shape —
it’s sharp for small values and diffuse for larger values
including an asymmetric regularization.

Lots of extra indentations here — are you adding extra
newlines in the LaTeX source?  A useful trick is to use
comments, “%” to add whitespace to the source without
forcing new paragraphs in the compiled document.

Missing an “is”: “Note, however, that this statistic is only valid…”

I’m not sure I understand the context of that last paragraph.
What are you trying to imply about higher-dimensional GPs?
Or is accidentally cut out from the previous discussion of ARD?

Priors for marginal standard deviation:

Perhaps intro the second paragraph by saying something
like “A half-t or half-Gaussian prior on alpha also has the
benefit of putting nontrivial prior mass around zero.  This
allows the GP support the zero functions and allows the
possibility that the GP won’t contribute to the conditional
mean of the total output."

Multiple Output Gaussian processes:

Is x_i meant to be in R^{N}?

Where is k(x) defined?  I’m guessing it’s related to K_{rho, alpha}
somehow?  Same thing with C — is it a matrix of constants or
a matrix function?

bob-carpenter:
@rtrangucci If this is done, feel free to merge.  There'll be another manual pull request before we release the next version if there are changes you want to make between merging this and the next release.

avehtari:
Sorry it took so long to check priors and multioutput

"below the minimum spacing of the covariates"
 - below the minimum spacing of the covariate values?

The equation after 
"If we're using the GP as a component in a larger model that includes an overall
mean and fixed effects for the same variables we're using as the domain for the
GP, i.e.:"
ends as "f \in", and I guess there should be more?

When mentioning a prior distribution by name, it would be handy to mention also the corresponding function in Stan language.

rtrangucci:
@bob-carpenter Ok, this is good enough to go in. Thanks for all the eyes on this @avehtari @betanalpha @bob-carpenter @bbbales2!

bob-carpenter:
Thanks.  I'll merge it after the tests have all passed.

avehtari:
Now there's a recommendation to use generalized inverse Gaussian distribution as a prior, but then the manual says "This has not yet been implemented in Stan’s math library, but it is possible to implement as a user defined function.". Could we have a pointer to a ready made function code or additional alternative recommendation what to use before ginv-Gaussian internal functions have been added?

bob-carpenter:
OK, I won't merge pending response to https://github.com/avehtari

Shouldn't we be building in that special function?  Maybe @seantalts could implement it to try out the new `operands_and_partials` multivariate stuff.  But we should definitely just put the definition in the manual rather than assuming users will be able to figure it out.

seantalts:
I have yet to make a distribution in the first case but would be happy to take a crack at it. Is it this one? https://en.wikipedia.org/wiki/Generalized_inverse_Gaussian_distribution

betanalpha:
Yes. Wanna go through the derivatives when I get back next week?

seantalts:
Sure! We probably shouldn't wait on me for this PR though, right?

rtrangucci:
I'll add the user-defined function in the manual chapter.


Rob

betanalpha:
No, be chapter should go ahead with the user defined function.  Having a proper density will be helpful for the GP case study.

bob-carpenter:
First step is to figure out whether they want inverse Gaussian or generalized inverse Gaussian and submit the issue up for stan-dev/math.  If it's the former, you should be able to do the derivatives yourself.

rtrangucci:
Here's the density in Stan code:

```
functions {
	real generalized_inverse_gaussian_lpdf(real y, int p, real a, real b) {
		  real lp =
			p / 2 * log(a / b) - log(2 * modified_bessel_second_kind(p, sqrt(a * b)))
			+ (p - 1) * log(y) - (a * x + b / x) * 0.5;
			return lp;
	}
}
```

bob-carpenter:
Nooo!  Don't give users examples of poorly formatted code.  You don't need the intermediate variable;  I also like to line up the additive terms.  So that'd be

```
 real generalized_inverse_gaussian_lpdf(real x, int p, real a, real b) {
   return p * 0.5 * log(a / b)
          - log(2 * modified_bessel_second_kind(p, sqrt(a * b)))
          + (p - 1) * log(x)
          - (a * x + b / x) * 0.5;
 }
```

If you don't want to do this, just drop a note in the next manual issue and I'll deal with it before the next release.  I don't want to hang this thing up forever, but I also don't want poorly formatted and indented code to get into our manual.

avehtari:
Bob, your alignment overflows over the textwidth.

avehtari:
To be more ccurate: Your alignment overflows over the textwidth if you put that as is in the stancode environment in the manual. The stancode environment uses too big font size compared to the width of the text area on manual page.

bob-carpenter:
How about

Are you using the Lucida fonts?  I'm also OK with the emacs standard if the above won't fit:

```
 real generalized_inverse_gaussian_lpdf(real x, int p, real a, real b) {
   return p * 0.5 * log(a / b)
      - log(2 * modified_bessel_second_kind(p, sqrt(a * b)))
      + (p - 1) * log(x)
      - (a * x + b / x) * 0.5;
 }
```

But there has to be some indentation.

bob-carpenter:
Also, I don't see how mine will overflow since all lines are shorter than the original arguments!

avehtari:
A user was also asking some default recommendations for the parameters of GIG.

Aki

avehtari:
I see this, where the code goes to the margin
image

bob-carpenter:
I'd like up `real a` directly under `real x`, following the C++ and Java conventions.

And I'd keep to two indents (I might have gone one too many---formatting in these comments is madness because it doesn't use fixed-width fonts for markdown).

bob-carpenter:
```
functions {
  real generalized_inverse_gaussian_lpdf(real x, int p,
                                         real a, real b) {
    return p * 0.5 * log(a / b)
      - log(2 * modified_bessel_second_kind(p, sqrt(a * b)))
      + (p - 1) * log(x)
      - (a * x + b / x) * 0.5;
 }
}
```

PR #2306, opened 2017-05-08
Memory corruption when using stan_model class in more than one thread

ariddell:
(This is a bug report with a test case in the PR)
The stan_model *class* is not threadsafe. Given two separate instances of a given stan_model, you can't call, say, stan::model::log_prob_grad with them in parallel in different threads. I suspect that there's some unsafe manipulation of the class variables at fault.

The test attached leads to a segfault.

gdb tells me that the problem is memory corruption, "corrupted double-linked list" in ``stan_math/stan/math/rev/core/grad.hpp``:

```
li#1  0x0000555555560bd3 in stan::math::grad (vi=0x55555583dfa0) at lib/stan_math/stan/math/rev/core/grad.hpp:44
44              (*it)->chain();
(gdb) list
39            vi->init_dependent();
40            it_t begin = ChainableStack::var_stack_.rbegin();
41            it_t end = empty_nested()
42              ? ChainableStack::var_stack_.rend() : begin + nested_size();
43            for (it_t it = begin; it < end; ++it) {
44              (*it)->chain();
45            }
46          }
47
48        }
```

The test attached to the PR makes the problem clear, I hope. You should be able to sample using two separate model instances -- but instead you get a segfault.

Being able to sample in parallel in system threads (not processes) is particularly important for PyStan on Windows.

My sense is that this should be easy to fix. Perhaps a variable needs to be an instance variable instead of a class variable? Any ideas?

#### Intended Effect

Segfault

#### How to Verify
```
make test/unit/services/sample/hmc_nuts_diag_e && test/unit/services/sample/hmc_nuts_diag_e
```

#### Reviewer Suggestions
@bob-carpenter @syclik

### DIFF
--- a/make/tests
+++ b/make/tests
@@ -4,7 +4,7 @@
 ##
 LIBGTEST = test/libgtest.a
 GTEST_MAIN = $(GTEST)/src/gtest_main.cc
-CFLAGS_GTEST += -isystem $(GTEST)/include -isystem $(GTEST)
+CFLAGS_GTEST += -isystem $(GTEST)/include -isystem $(GTEST) -std=c++11
 
 ##
 # Build the google test library.
--- a/makefile
+++ b/makefile
@@ -20,7 +20,7 @@ CC = g++
 O = 3
 O_STANC = 0
 AR = ar
-C++11 = false
+C++11 = true
 
 ##
 # Set default compiler options.
--- a/src/test/unit/services/sample/hmc_nuts_diag_e_test.cpp
+++ b/src/test/unit/services/sample/hmc_nuts_diag_e_test.cpp
@@ -4,17 +4,24 @@
 #include <test/test-models/good/optimization/rosenbrock.hpp>
 #include <test/unit/services/instrumented_callbacks.hpp>
 #include <iostream>
+#include <thread>
 
 class ServicesSampleHmcNutsDiagE : public testing::Test {
 public:
   ServicesSampleHmcNutsDiagE()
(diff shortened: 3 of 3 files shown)


bob-carpenter:
The problem is autodiff, not the model class.

What you need to do to work in parallel is to make the singletons for autodiff into thread local variables.  There's a performance penalty, but then they'll be kept separate across threads and the model class will become threadsafe.

We'll be looking at that more with C++11 threading and the desire to parallelize ODE solvers.

ariddell:
Thanks! Glad to know that you understand what's going on.

Is this tracked elsewhere? This blocks at least one PyStan feature.

If it's not tracked elsewhere, should I perhaps close this and open an issue "Autodiff is not threadsafe"?

bob-carpenter:
It should go in the math library.  The fix is to declare the variables as thread local, but that might not do what some users intend, as it doesn't let you compute one piece of a likelihood in one thread, it just lets you run multiple models in parallel.

I think the slowdown is around 10--20% but we should try and profile again.

ariddell:
Great. Is this as easy as adding ``thread_local`` somewhere (I see that
that's a C++11 keyword)? In the short term I could maintain a simple
patched version. (The problem is a bit of a show-stopper for the new
"backend" for PyStan 3.)

In the meantime, I'll close this and write up an issue for stan-math.

bob-carpenter:
See https://github.com/stan-dev/math/issues/551

Issue #2267, opened 2017-04-07
Standalone compilation of functions block


martincerny:
#### Summary:
In order to call Stan functions directly from R, the c++ file must contain:

- double-instantiated only functions, (not templated for double/var)
- R comments in generated code that R needs

To do this, add function `stan/lang/compile_functions.hpp` that takes a program that only consists of functions and compiles to C++ it with the necessary includes.

This is needed to circumvent the current Rstan hack which uses regexes to munge the transpiled hpp code - we want to go _directly_ from Stan program to c++ code that is callable from R without further processing.

#### Description:
Originated in the discussion of #2224 - it would be useful to be able to parse and compile a file that contains only a functions block. The primary use case is to let thos functions then be invoked from R, but it could also help in allowing truly modular Stan programs.

#### Current Version:
v2.14.0

martincerny:
I am trying this as my first task in the Stan codebase (but it seems I cannot assign myself nor add a "feature" label)

bob-carpenter:
I think we have to somehow make you a member of the project.  Also, you want to coordinate with @mitzimorris.  I believe the recent pull request to remove the requirement of a `model { }` block will make this very easy to generate functions only.

The tricky part is going to be finding minimal headers.  I fear there will be dependencies on code that's only in stan-dev/stan rather than just in stan-dev/math, which would be ideal (much more minimal include).  

The other thing we really want out of this function is an ability to (a) control the namespace in which they get generated from the call (`vector<string>` for namespaces would work), and (b) have them instantiated with only `int` and `double` primitive types.  Having both of these will let us generate functions that can be used directly by Rcpp.

martincerny:
Nice codebase you have! Getting into it was easier than I expected.

I have some working code with a rudimentary test which shows that the generated functions can be parsed and instantiated. The feature is not yet done, but I think it is the correct time to ask for some feedback (all tests currently pass).

The changes I made can be found at https://github.com/martincerny/stan/tree/feature/2267-standalone-functions and a comparison to stan/develop can be reviewed here:
https://github.com/stan-dev/stan/compare/develop...martincerny:feature/2267-standalone-functions?expand=1

What it does:
- There is now a function ```stan::lang::compile_functions()``` that runs the normal parser, but generates only the function block. The normal parser already worked with models that contain only the function block, only difference from ```stan::lang::compile()``` is in the way the output is generated: only the functions block is created, it is possible to specify the namespace for the functions block and no class declaration is created.
[...]

Issue #2224, opened 2017-02-04
general include mechanism for Stan programs


bob-carpenter:
#### Summary:

Add a general file include mechanism in the Stan language to support system and user libraries of Stan functions.

#### Description:

Follow C/C++ convention with system libs in braces:

```
#include <gp>
```

and quoted strings for user libs

```
#include "foo/my_gp_funs.stan"
```

The `compiler` class in `stan::lang` and the `bin/stanc` executable will need to be given an ordered search path for system files and for user files.  

We might want to 

* use `@` instead of `#`, 

* relax the line-initial constraint, and 

* remove the need for quotes around paths.

This deals with half of #810 and generalizes #1294 (both of those issues are now closed).

#### Current Version:
v2.14.0

martincerny:
I am actually itching to have a go at implementing this (at least try to :-) ). To be able to think about this clearly - what are the use cases that need to be addressed? I think mostly you want to reuse functions, but I noticed rstanarm also uses includes in the ```data``` section. Is this encouraged? Are there other legitimate/non-hacky use cases for an include mechanism?

In general, I am unsure it is wise to mimic the behavior of C++ preprocessor as it has history of enabling users to write very hard-to-maintain code (especially if the include mechanism gets later extended with compile-time flags similar to ```#define / #ifdef```). My ideal would be closer to "import" in Python/Java. In practice this could translate some of the following design choices:

- allow ```include``` only at specific places (beginning of file/between top-level blocks/at block beginning)
- don't allow a file to be included twice
- included files are somehow encapsulated (e.g. they contain top level blocks)
- included files are self-sufficient: during model development an included file can be checked for syntax correctness separately (e.g. "Check on save" in RStudio) as it will not be dependent on the contents of the top-level file

Still the actual implementation of the include mechanism would be just pasting the file contents at the right place (because that's easier to do), but IMHO it would feel closer to modern languages with only minor modifications to the parser (e.g. allow multiple ```functions``` blocks in a single program).

bob-carpenter:
I've already finished the guts of the raw include mechanism
on this branch:

  https://github.com/stan-dev/stan/tree/feature/2224-stan-includes

There's a new class called io::program_reader whose constructor
reads a named input stream, expands the includes relative to an
ordered sequence of directories in which to search, and keeps a
record of line numbers.  Once constructed it provides functions to 
  
  * returns the whole program after includes, and

  * map a line number in the whole program back to where it
    came from as a stack of includes and line numbers.

[...]

PR #2310, opened 2017-05-18
Feature/2267 standalone functions

martincerny:
#### Submisison Checklist

#### Summary
Add support for standalone function compilation. See #2267. 

There are a lot of outstanding questions about whether this is the right way to implement it. I am copying those from #2267, as they seem to got lost there:

1) The decision to choose mode of compilation (functions only / model) by file extension and the choice of the ```.stanfuncs``` extension are both important and should be discussed by the core team (there could be a need for more compilation modes in the future, especially with regards to #2224).
    - My reasoning behind this decision was: 
      - The mode of compilation should be deducible from the file to be compiled alone and should not require the user to check some external configuration/build/... file. This is good for both reading code and editor support.
       - Deciding what to do with a file based on its extension is a common practice, in particular it is what gcc does.
2) I did some changes/refactoring to some very core files and to the test infrastructure to avoid duplicating code, but I felt very uncomfortable making the changes without supervision from someone who understands the project better.
3) Test functions files now reside in ```test/test-models/good-standalone-functions``` - is this a reasonable place in the tree? (an obvious alternative would be ```test/test-standalone-functions/good```)
    - Note that I cannot put ```.stanfuncs``` files simply in  ```test/test-models/good``` because it is then impossible to distinguish the resulting ```.hpp``` files from each other (and I don't think ```.stanfuncs``` should produce something like ```.hpp-funcs```)
4) I had trouble with the 80 char limit, I don't like some of the indents / line breaks it forced me to do, but I couldn't find a better way - could you please check that the code reads nice and is aligned with what others write? (e.g. the two-line ```for``` cycle in ```stanc_helper.hpp``` line 293).
5) When compiling standalone functions, it is definitely necessary to check whether other blocks are contained in the input file. I see two possible ways of doing this - what would you prefer?
    1) keep the default parser and check whether the members of ```stan::lang::program``` are all empty except ```function_decl_defs_``` (requires maintanence when new types of program blocks are introduced or when the ```program``` struct changes)
    2) create a specific parser variant that only recognizes the functions block (would require maintanance when the parser structure changes and a specific parser for other compilation modes would have to be created as well)
6) Should ```compile_functions()``` give a warning or error when other blocks are present?

#### Intended Effect
Add .stanfuncs as a possible source file extension that then results in compiling a file with instantiations of all templated functions.

#### How to Verify
.stanfuncs files are compiled as part of test build, also tests in ```src/test/unit/lang/parser/functions_standalone_instantiate_test.cpp``` and ```src/test/unit/lang/parser/functions_standalone_test.cpp```

#### Side Effects
Touched a lot of core files, please verify it is OK.

#### Documentation
Is rudimentary and pending approval of the core design.

#### Copyright and Licensing

### DIFF
--- a/src/stan/lang/generator/generate_function_instantiation_body.hpp
+++ b/src/stan/lang/generator/generate_function_instantiation_body.hpp
@@ -6,6 +6,7 @@
 #include <stan/lang/generator/generate_function_instantiation_template_parameters.hpp>
 #include <ostream>
 #include <vector>
+#include <string>
 
 namespace stan {
   namespace lang {
@@ -25,8 +26,8 @@ namespace stan {
      * @param[in] is_log true if function is log probability function
      * @param[in,out] o stream for generating
      */
-    void generate_function_instantiation_body(const function_decl_def& fun, 
-                                    bool is_rng, bool is_lp, bool is_log, 
+    void generate_function_instantiation_body(const function_decl_def& fun,
+                                    bool is_rng, bool is_lp, bool is_log,
                                     const std::string& rng_class,
                                     std::ostream& o) {
       o << "{" << EOL;
@@ -36,8 +37,8 @@ namespace stan {
       }
       generate_function_name(fun, o);
       generate_function_instantiation_template_parameters(
-        fun, is_rng, is_lp, is_log, rng_class, o); 
-        
+        fun, is_rng, is_lp, is_log, rng_class, o);
+
       o << "(";
       for (size_t arg_i = 0; arg_i < fun.arg_decls_.size(); ++arg_i) {
         o << fun.arg_decls_[arg_i].name_;
@@ -54,7 +55,7 @@ namespace stan {
       if (is_rng || is_lp || fun.arg_decls_.size() > 0)
         o << ", ";
       o << "pstream__";
-      
+
       o << ");" << EOL;
       o << "}" << EOL;
     }
--- a/src/stan/lang/generator/generate_function_instantiation_body.hpp
+++ b/src/stan/lang/generator/generate_function_instantiation_body.hpp
@@ -33,7 +33,7 @@ namespace stan {
                                     const std::string& rng_class,
                                     std::ostream& o) {
       o << "{" << EOL;
-      o << "\t";
+      o << "  ";
       if (!fun.return_type_.is_void()) {
         o << "return ";
       }
--- a/src/stan/lang/generator/generate_function_instantiation_body.hpp
+++ b/src/stan/lang/generator/generate_function_instantiation_body.hpp
@@ -1,0 +1,65 @@
+#ifndef STAN_LANG_GENERATOR_GENERATE_FUNCTION_INSTANTIATION_BODY_HPP
+#define STAN_LANG_GENERATOR_GENERATE_FUNCTION_INSTANTIATION_BODY_HPP
+
+#include <stan/lang/ast.hpp>
+#include <stan/lang/generator/generate_function_name.hpp>
+#include <stan/lang/generator/generate_function_instantiation_template_parameters.hpp>
+#include <ostream>
+#include <vector>
+
+namespace stan {
+  namespace lang {
+
+    /**
+     * Generate the body for a function instantiation (e.g., the 
+     * call to the templated function with all templated arguments based on 
+     * double).
+     * Requires precalculated flags for whether it is an RNG, uses the log
+     * density accumulator or is a probability function, to the
+     * specified stream.
+     *
+     * @param[in] fun function declaration
+     * @param[in] is_rng true if function is an RNG
+     * @param[in] is_lp true if function accesses log density
+     * accumulator
+     * @param[in] is_log true if function is log probability function
+     * @param[in,out] o stream for generating
+     */
+    void generate_function_instantiation_body(const function_decl_def& fun, 
+                                    bool is_rng, bool is_lp, bool is_log, 
+                                    const std::string& rng_class,
+                                    std::ostream& o) {
+      o << "{" << EOL;
+      o << "\t";
+      if (!fun.return_type_.is_void()) {
+        o << "return ";
+      }
+      generate_function_name(fun, o);
+      generate_function_instantiation_template_parameters(
+        fun, is_rng, is_lp, is_log, rng_class, o); 
+        
+      o << "(";
+      for (size_t arg_i = 0; arg_i < fun.arg_decls_.size(); ++arg_i) {
+        o << fun.arg_decls_[arg_i].name_;
+        if (arg_i + 1 < fun.arg_decls_.size()) {
+          o << ", ";
+        }
+      }
+      if ((is_rng || is_lp) && fun.arg_decls_.size() > 0)
+        o << ", ";
+      if (is_rng)
+        o << "base_rng__";
+      else if (is_lp)
+          o << "lp__, lp_accum__";
+      if (is_rng || is_lp || fun.arg_decls_.size() > 0)
+        o << ", ";
+      o << "pstream__";
+      
+      o << ");" << EOL;
+      o << "}" << EOL;
+    }
+  }
+}
+
+
+#endif
--- a/src/stan/lang/generator/generate_function_instantiation_body.hpp
+++ b/src/stan/lang/generator/generate_function_instantiation_body.hpp
@@ -24,6 +24,8 @@ namespace stan {
      * @param[in] is_lp true if function accesses log density
      * accumulator
      * @param[in] is_log true if function is log probability function
+     * @param[in] rng_class class of the RNG being used (required by xxx_rng 
+     * functions)
      * @param[in,out] o stream for generating
      */
     void generate_function_instantiation_body(const function_decl_def& fun,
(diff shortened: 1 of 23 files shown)


bob-carpenter:
Header tests failed.  Please let me know when you think this is ready to review and I can look it over.  I can do that before the header tests get fixed if it's otherwise done.  And thanks for submitting a pull request.

martincerny:
My bad, forgot I could run header tests locally. Fixed (hopefully).

bob-carpenter:
Thanks.  I'll get to reviewing this week.

mitzimorris:
I checked out this branch from Martin's repo.  As a test I took the file `src/test/test-models/good-standalone-functions/basic.stanfuncs` and copied it to `src/test/test-models/good/basic.stan`.  Also reformatted it for consistent indentation and placement of open curly brackets, so contents of both files is:
```
functions {
  real my_log1p_exp(real x) {
    return log1p_exp(x);
  }

  vector my_vector_mul_by_5(vector x) {
    vector[num_elements(x)] result = x * 5.0;
    return result;
  }
}
```

I compiled both files using Stan's makefile, which invokes stanc on the files `basic.stanfuncs` and `basic.stan:` 
```
make src/test/test-models/good-standalone-functions/basic.hpp-test
make src/test/test-models/good/basic.hpp-test
```
To compare these files, I edited out the model definition and the model typedef statement which is the last line of the transpiled stan program, then ran `diff`:

```
diff -U 3 src/test/test-models/good/basic.hpp src/test/test-models/good-standalone-functions/basic.hpp
--- src/test/test-models/good/basic.hpp	2017-05-28 11:50:02.000000000 -0400
+++ src/test/test-models/good-standalone-functions/basic.hpp	2017-05-28 11:51:13.000000000 -0400
@@ -1,24 +1,19 @@
 // Code generated by Stan version 2.15.0
 
-#include <stan/model/model_header.hpp>
-
-namespace basic_model_namespace {
+#include <stan/model/standalone_functions_header.hpp>
 
+namespace basic_functions { 
 using std::istream;
 using std::string;
 using std::stringstream;
 using std::vector;
-using stan::io::dump;
 using stan::math::lgamma;
-using stan::model::prob_grad;
 using namespace stan::math;
 
 typedef Eigen::Matrix<double,Eigen::Dynamic,1> vector_d;
 typedef Eigen::Matrix<double,1,Eigen::Dynamic> row_vector_d;
 typedef Eigen::Matrix<double,Eigen::Dynamic,Eigen::Dynamic> matrix_d;
 
-static int current_statement_begin__;
-
 template <typename T0__>
 typename boost::math::tools::promote_args<T0__>::type
 my_log1p_exp(const T0__& x, std::ostream* pstream__) {
@@ -91,6 +86,14 @@
     }
 };
 
+double
+my_log1p_exp(const double& x, std::ostream* pstream__){
+	return my_log1p_exp<double>(x, pstream__);
+}
 
+Eigen::Matrix<double, Eigen::Dynamic,1>
+my_vector_mul_by_5(const Eigen::Matrix<double, Eigen::Dynamic,1>& x, std::ostream* pstream__){
+	return my_vector_mul_by_5<double>(x, pstream__);
 }
 
+ } 
```

Two things:

1. why use a different header?
compiled stan program header file include:  
```
#include <stan/model/model_header.hpp>
```
compiled stan functions-only block header file include:
```
#include <stan/model/standalone_functions_header.hpp>
```

The set of includes in  `standalone_functions_header.hpp` is a subset of the files in `model_header.hpp`.  It would be better to just use the `model_header.hpp` file - first off, it's not clear to me that the standalone functions header has all necessary includes, secondly, maintaining two sets of includes is more work.

I reviewed the changes to the code in the `src/stan/lang/` and it seems OK.

However, we need clarification from @bgoodri as to what the comments need to be generated so that the transpiled hpp files can be included directly from R.

mitzimorris:
Please add test models for rng and lp functions.
Suggested rng function test:
```
functions {
  real foo_rng(real mu, real sigma) {
    return normal_rng(mu, sigma);
  }
}
```

bob-carpenter:
Did we also want a test that both ode_integrators work?

bgoodri:
At the top, it needs `// [[Rcpp::depends(rstan)]]`. Then, it needs `// [[Rcpp::export]]` without a semicolon before each function (but not the corresponding functor), as in
```
// [[Rcpp::export]]
inline double
my_function(...)
```
I would also suggest to add a default argument of the null pointer to each `std::ostream* pstream__` argument because that cannot be passed from R. For `_lpdf` and `_lpmf` functions, the trailing `lp_accum__` should be defaulted to zero. The functions that end in `_rng` are difficult. I currently replace the `base_rng__` argument with an `int seed` and then at the first line of the function body, I have
```
static boost::ecuyer1988 base_rng__(seed);
```

sakrejda:
You can pass both the RNG and stream of you have helper functions for
producing them.

martincerny:
Thanks for the feedback, it is appreciated. However, all of the comments so far deal with (in my view) minor technical stuff. Does that mean that you are OK with the high-level structure and the main design decisions behind the way I implemented this feature? Especially the points 1-3 in the pull request description. Any opinion on 5&6 would also be welcome.

@mitzimorris The reduced set of headers was explicitly requested (https://github.com/stan-dev/stan/issues/2267#issuecomment-292600155) and should lead to shorter compilation times - the model includes are IMHO a huge beast. I currently don't have a strong opinion whether this is worth the extra maintanence cost, just wanted to point out the reasoning behind this choice.

martincerny:
@mitzimorris the test for ```_rng```, ```_lp``` and ```_lpdf``` functions was already there, in ```src/test/test-models/good-standalone-functions/special_functions.stanfuncs``` and it is tested that it both compiles and executes (in ```src/test/unit/lang/parser/functions_standalone_test.cpp``` and ```src/test/unit/lang/parser/functions_standalone_instantiate_test.cpp```. Is it sufficient or do you suggest extending it?

@bgoodri I added the RCpp comments and the null default to pstream (the default was added to all functions - even in normal Stan - is that a problem?), however, I cannot reasonably give default the ```lp_accum__``` argument as it is a ```double&```. I have also added a default ```__create_rng(int seed)``` function that is also exported to RCpp (currently for standalone functions only).

martincerny:
OK, so adding default for the ```pstream__``` argument turned out to be harder than expected: if a forward declaration is present, the default needs to be put there and only there. But AFAIK, I cannot check in the generator whether the function I am currently generating has a forward declaration somewhere above. 

So I added the default only to instantiations, but that means that it will not be present for functions which do not need instantiations (e.g. int-only arguments). Maybe it will be better to create a helper function to get a stream as well?

sakrejda:
BTW: this is more or less what the helper function in R would look like, similar for the stream (you could even pass `Rcpp::rcout` from the helper so the output would appear correctly.

```
#include <Rcpp.h>

// [[Rcpp::export]]
Rcpp::XPtr<boost::ecuyer1988> get_rng(const int& seed=0) {
  boost::ecuyer1988 base_rng__(seed);
  Rcpp::XPtr<boost::ecuyer1988> ptr(base_rng__);
  return ptr;
}
```

src/stan/lang/generator/generate_rng_rcpp_helper.hpp
```diff
@@ -0,0 +1,26 @@
...
+     * This should match the default RNG chosen in 
+     * generate_function_instantiation()
+     *
+     * @param[in] o output stream
+     */
+    void generate_rng_rcpp_helper(std::ostream& o) {
+      o << "// [[Rcpp::export]]" << EOL;
+      o << "boost::ecuyer1988 __create_rng(int seed) {" << EOL;
+      o << "\treturn(boost::ecuyer1988(seed));" << EOL;
+      o << "}" << EOL << EOL;
```
mitzimorris: what is different between this and suggestion:
```
  Rcpp::XPtr<boost::ecuyer1988> ptr(base_rng__);
  return ptr;
```

bob-carpenter:
No, we can't do what @sakrejda is suggesting in https://github.com/stan-dev/stan/pull/2310#issuecomment-304935311 because we can't introduce dependencies on R or Rcpp (or any other GPL-ed code) in Stan itself.

src/stan/lang/generator/generate_function_instantiation_body.hpp
```diff
@@ -0,0 +1,68 @@
...
+     * functions)
+     * @param[in,out] o stream for generating
+     */
+    void generate_function_instantiation_body(const function_decl_def& fun,
+                                    bool is_rng, bool is_lp, bool is_log,
+                                    const std::string& rng_class,
+                                    std::ostream& o) {
+      o << "{" << EOL;
+      o << "\t";
+      if (!fun.return_type_.is_void()) {
```
mitzimorris: replace all `\t` with 2 spaces - generated code should contain spaces not tabs - change this everywhere.

sakrejda:
Bob, I'm not suggesting putting Rcpp stuff into Stan code. The point is we
can leave the RNG argument in these generated functions because it's
trivial for rstan to provide a function that fills it in!

bob-carpenter:
@sakrejda  Great --- as long as the helper R function stays in RStan, that's fine.

bob-carpenter:
I'd go and make these changes myself, but I don't know how to work on remote branches.  @seantalts --- you said you'd show me how to do this, but I'm going to be out of town for a week.  Would it be easy for me to somehow change all the `\t` to two spaces?  I think that's the last thing waiting to do before this gets merged.

martincerny:
I've replaced the tabs with spaces. Should be ready to go.

PR #2317, opened 2017-05-31
Feature/issue 512 log

syclik:
#### Submisison Checklist

#### Summary

Implements a logger interface described here on the [wiki](https://github.com/stan-dev/stan/wiki/Logging-Spec#design--prototype).

This does not change the behavior of the output. All output is currently sent to `info`.

#### Intended Effect

This pull request is meant to split out messages meant for the interfaces to report into a logger with different levels.

This pull request puts all the messages at the same level, `info`. I didn't want to make decisions on what are `debug` vs `warn` vs `info` vs `error` vs `fatal` for the time being.

I will follow up with pull requests to CmdStan, RStan, and PyStan that will implement loggers.

**Note:** after implementing this, I believe having the logger implement `operator<<()` will be more trouble than it's worth. If we implemented it that way, we would need to check on every call to flush. I think the right level is whole lines and we'll need the interfaces to determine what to do with each line.

#### How to Verify

Tests. These things all have new tests.

#### Side Effects

This will affect CmdStan, RStan, and PyStan. We'll need to work out the upstream changes.

#### Documentation

Doxygen doc included for the new classes.

#### Reviewer Suggestions

@sakrejda, @ariddell, @seantalts

#### Copyright and Licensing

Stan Group


syclik:
This is ready for review. The upstream test requires CmdStan branch `feature/issue-554-log`; @seantalts, what's the procedure for having the upstream branch be something other than `develop`?

There are now pull requests on all the interfaces that are ready to go once this is in Stan:
- CmdStan: stan-dev/cmdstan#555
- RStan: stan-dev/rstan#419
- PyStan: stan-dev/pystan#346

syclik:
Thanks. 

Regarding flushing: it's up to the implementation. We could write a logger that queues everything up and flushes at once. The stream_logger just implements existing behavior.

seantalts:
@syclik re: testing upstream - I think it might be the same way as before (?) - Go to the Stan Pull Request - Upstream - CmdStan, click Build with Parameters, and put in the appropriate stan_sha ("origin/pr/2317/merge" in this case) and cmdstan_branch ("origin/feature/issue-512-log"). It's running now: http://d1m1s1b1.stat.columbia.edu:8080/job/Stan%20Pull%20Request%20-%20Upstream%20-%20CmdStan/516

I don't think that will integrate into the github API appropriately to change the checkbox on the PR, though. Not sure if we can do much about that.

seantalts:
Couple of design questions:

1. You asked me to look at using `operator<<` and I guess I'm not sure why that wouldn't work? It seems like both PyStan and RStan use the default logger implementation that internally takes streams and `<<`s stuff to them. This is a fairly simple example allowing `<<` with levels: http://www.drdobbs.com/cpp/logging-in-c/201804215?pgno=1 .
2. I think a lot of logging frameworks are implemented with a global logger precisely because logging is often a very orthogonal, cross-cutting and global concern. It has the nice practical benefit that it lets one avoid passing a logger to literally everything. More broadly, is there a reason we don't use an existing logging framework? Boost comes with one, for example: http://www.boost.org/doc/libs/1_64_0/libs/log/doc/html/index.html These things are often set up with battle-tested macros that elide much of the work done accumulating logs and concatenating strings if the level isn't high enough to eventually produce them.

That said, what you have here is great and pretty unobjectionably an improvement over the existing, I would say. So I'm definitely not trying to hold this back, more just giving food for thought.

syclik:
Jenkins, retest this please.

syclik:
@sakrejda and @seantalts, thanks! I'll merge when the tests pass. Doesn't look like there's anything that needs to be fixed.

> 1. You asked me to look at using operator<< and I guess I'm not sure why that wouldn't work? It seems like both PyStan and RStan use the default logger implementation that internally takes streams and <<s stuff to them. This is a fairly simple example allowing << with levels: http://www.drdobbs.com/cpp/logging-in-c/201804215?pgno=1 .

It's not that it can't work. It's just that I couldn't think about how to get it done efficiently. In RStan, there's usually a string that prefixes each line that indicates what chain the message comes from. So far, the message unit is a line, not a chunk. So if we wanted to use something like a stream operator with `std::endl` as a marker for the end of the line, RStan's logger would need to inspect the stream as it comes in and decide how to break apart lines. I'm probably over-valuing existing behavior and we could change that in the future. 

> 2. I think a lot of logging frameworks are implemented with a global logger precisely because logging is often a very orthogonal, cross-cutting and global concern. It has the nice practical benefit that it lets one avoid passing a logger to literally everything. More broadly, is there a reason we don't use an existing logging framework? Boost comes with one, for example: http://www.boost.org/doc/libs/1_64_0/libs/log/doc/html/index.html These things are often set up with battle-tested macros that elide much of the work done accumulating logs and concatenating strings if the level isn't high enough to eventually produce them.

Ah ha. I had originally searched through the different reasonable C++ logging libraries including Boost, Apache's log4cxx, log4cpp, and Pantheios. They all had their issues and for now, I thought our logging needs are fairly simple. We can either replace it after it gets in or write implementations of the logger with one of these libraries.

Here are some of the reasons I didn't use Boost.Log:
- Adds yet another dependency. It's not included in R's bh package, so it would have been more pain for RStan.
- It's not header only. We'd have to build a library for CmdStan, RStan, and PyStan. And for Stan just to test. We haven't used a non-header-only library from Boost yet, so I didn't feel like it's a good enough reason to figure out the build process and integrate that into Stan for this process.

Otherwise, I would definitely have used an existing logging interface!

bob-carpenter:
I'm moving further design discussion back to 

http://discourse.mc-stan.org/t/logging-in-stan-services/300/3

Issue #2300, opened 2017-05-04
Algebra Solver


charlesm93:
#### Summary:
This related to issue #2023  and the pending pull request #2023 . Expose the algebraic function. I'm raising the issue now, because I do not expect the signature for `algebra_solver` to change. Given this is a higher-order function, I will follow the scheme used for the ODE solvers.

#### Description:
Allows users to solve systems of algebraic equations, given an initial guess.

#### Additional Information
If we move to C++11, we could change the signature, notably by removing the x_r and x_i arguments. I expect we'll however still need the parms argument to calculate the Jacobian.

#### Current Version:
v2.15.0

charlesm93:
Following the example of the ODE integrators, I got a working prototype. However, I can't get rid of the argument `std::string& solver_name`, which is equivalent to `std::string& integration_function_name`, in the constructor of the algebra_solver structure (.../lang/ast/node/algebra_solver_def.hpp). Since there is only one solver, I do not need this argument, but I get a bug if I don't include it.

It is not a error at compilation, but when I run the test, all the arguments are "misplaced" -- pushed to the left, as though an additional argument were constructed at the beginning. Which prompts several error messages from the parser.

Can I view how the parser attempts to call the function? As far as I can tell, no hpp file gets produced if the parsing fails.

bob-carpenter:
Is there a public branch where I can look at what you've done?  

The problem is that if the grammar rule doesn't fire, you won't get any output.  You can instrument the grammar rules themselves with print statements, but it's a huge pain because it requires semantic actions (it'll get easier with lambdas in C++11).  You need to make sure the parser children (right hand side of rules) and the semantic coercion that maps the sequence of terminals into the parent category match.

charlesm93:
@bob-carpenter thanks for the pointers. The branch I used is feature/issue-2300-algebra_solver.

bob-carpenter:
Did you make sure to synch the fusion adaptor declarations in `term_grammar_def.hpp`?

```
BOOST_FUSION_ADAPT_STRUCT(stan::lang::algebra_solver,
                           (std::string, solver_name_)
                           (std::string, system_function_name_)
                           (stan::lang::expression, x_)
                           (stan::lang::expression, y_)
                           (stan::lang::expression, dat_)
                           (stan::lang::expression, dat_int_) )

BOOST_FUSION_ADAPT_STRUCT(stan::lang::algebra_solver_control,
                           (std::string, solver_name_)
[...]

PR #2339, opened 2017-06-26
Feature/issue 2300 algebra solver

charlesm93:
#### Submisison Checklist

- [x] Run unit tests: `./runTests.py src/test/unit` (ran `parser_generator_test.cpp`, `ast_test.cpp`, and `algebra_solver_test.cpp`).

#### Summary
Expose algebraic solver (see pull request stan-dev/math#516 in the math repo and issue #2300). The tuning parameters for the solver are optional.

#### How to Verify
Run unit tests: `parser_generator_test.cpp`, `ast_test.cpp`, and `algebra_solver_test.cpp`.

#### Documentation
Doxygen doc. Need to add a section in the user-manual.

#### Reviewer Suggestions
Bob and Michael.

#### Copyright and Licensing

Please list the copyright holder for the work you are submitting (this will be you or your assignee, such as a university or company): Metrum Research Group, LLC


charlesm93:
Finished updating the user manual. Everything should be in place for a review.

bob-carpenter:
I'll review this one.  It's failing tests right now for integration---is that just because the math lib hadn't been merged yet?  If so, you can restart the test.

src/docs/stan-reference/appendices.tex
```diff
@@ -803,6 +805,34 @@ \subsection{Probability Function Naming}
...
 
+\subsection{Algebraic Solver Argument Types and Origins}
+
+The \code{algebra\_solver} function may be used without control
+parameters; in this case
+%
+\begin{itemize}
+\item its first argument refers to a function with signature
+\begin{quote}
+  \code{(~real[], ~real[],~real[],~int[]~:~real[]},
```
bob-carpenter: Why is there a semicolon here?

src/docs/stan-reference/appendices.tex
```diff
@@ -803,6 +805,34 @@ \subsection{Probability Function Naming}
...
+\end{quote}
+\item the remainin four arguments must be assignable to types
```
bob-carpenter: missing "g" --- should be "remaining"

src/docs/stan-reference/examples.tex
```diff
@@ -8046,6 +8046,154 @@ \section{Circular Representations of Days and Years}
...
+can be found in \refchapter{functions-algebraic-solver}.
+%
+Solving any system of algebraic equations can be translated into a root-finding
+problem, that is, given a function $f(x)$, we wish to find $x$ such that
+$f(x) = 0$.
+
+\section{Example: System of Nonlinear Algebraic Equations}
+
+For systems of linear algebraic equations, we recommend solving the system
+using an inverse matrix. The algebraic solver becomes handy when we want
```
bob-carpenter: may want to warn people not to literally use an inverse function if it can be avoided---we'd rather use matrix division

src/docs/stan-reference/examples.tex
```diff
@@ -8046,6 +8046,154 @@ \section{Circular Representations of Days and Years}
...
+to solve nonlinear equations.
+%
+As an illustrative example, we consider a nonlinear system of two equations
+with two unknowns. Our goal is to simultaneously solve all equations for
+$x_1$ and $x_2$, such that the vector function $z$ goes to 0.
```
bob-carpenter: $z$ isn't the function, it's the result of applying the function to $x$.

src/docs/stan-reference/examples.tex
```diff
@@ -8046,6 +8046,154 @@ \section{Circular Representations of Days and Years}
...
+  z_2 &= x_1 x_2 + \theta_2 \\
+  \end{aligned}
+\end{eqnarray}
+%
+
+\section{Coding an Algebraic System}
+
+A system of algebraic equations is coded directly in Stan as a function with a
+strictly specified signature. For example, the nonlinear system given by
+\refequation{algebra.equation} can be coded using the following function
```
bob-carpenter: You don't need the .equation here---make sure LaTeX runs without warnings.

src/docs/stan-reference/examples.tex
```diff
@@ -8046,6 +8046,154 @@ \section{Circular Representations of Days and Years}
...
+in Stan (see \refchapter{functions-programming} for
+more information on coding user-defined functions).
+%
+\begin{stancode}
+real[] system(real[] x,       // unknowns
+               real[] theta,  // parameters
```
bob-carpenter: the second and following arguments should line up under the first

src/docs/stan-reference/examples.tex
```diff
@@ -8046,6 +8046,154 @@ \section{Circular Representations of Days and Years}
...
+               real[] dat_r,    // data (real)
+               real[] dat_i) {  // data (integer)
+  real z[2];
+  z[1] = x[1] - theta[1];
+  z[2] = x[1] * x[2] - theta[2];
+  return z;
```
bob-carpenter: You can code this as

```
return { x[1] - theta[1],
         x[1] * x[2] - theta[2] };
```

src/docs/stan-reference/examples.tex
```diff
@@ -8046,6 +8046,154 @@ \section{Circular Representations of Days and Years}
...
+transformed parameters {
+  real theta[2] = {3, 6};
+  real x[2];
+
+  x = algebra_solver(guess, theta, dat_r, dat_i);
+}
+\end{stancode}
+
+which returns $x = \{3, -2\}$.
+
```
bob-carpenter: Any hint at how good the guess has to be?  Is there something else to control if the guess is known to be wild?

src/docs/stan-reference/examples.tex
```diff
@@ -8046,6 +8046,154 @@ \section{Circular Representations of Days and Years}
...
+\subsection{Data versus Parameters}
+The arguments for the initial guess \code{guess}, the real data \code{dat\_r}, and
+the integer data \code{dat\_i} must be expressions that only involve data or
+transformed data variables. The parameters \code{theta} is the only argument
+which may involve parameters.
+
+\subsection{Length of the Algebraic Function and of the Vector of Unknowns}
+The Jacobian of the solution with respect to the parameters is computed
+using the implicit function theorem, which places certain restrictions. In particular,
```
bob-carpenter: I think you want "imposes" rather than "places".  With "places", you really need to add a prepositional phrase, such as "places restrictions on Jacobian invertibility".

src/docs/stan-reference/examples.tex
```diff
@@ -8046,6 +8046,154 @@ \section{Circular Representations of Days and Years}
...
+
+On the other hand, a system may not have a solution (for a given point in the parameter
+space). In that case, the solver will not converge to a solution. When the solver fails to
+do so, the current metropolis proposal gets rejected.
+
+\section{Control Parameters for the Algebraic Solver}
+%
+The call to the algebraic solver shown above uses the default control settings. The solver
+allows three additional parameters, all of which must be supplied if any of them is
+required.
```
bob-carpenter: should be "if any of them is supplied".

src/docs/stan-reference/examples.tex
```diff
@@ -8046,6 +8046,154 @@ \section{Circular Representations of Days and Years}
...
+%
+\begin{stancode}
+  x = algebra_solver(guess, theta, dat_r, dat_i, 
+                     rel_tol, f_tol, max_steps);
+\end{stancode}
+
+The three control arguments are relative tolerance, function tolerance, and maximum
+number of steps. Their default values are respectively \code{1e-10} ($10^{-10}$), \code{1e-6}
+($10^{-6}$), and \code{1e3} ($10^3$).
+
```
bob-carpenter: Explain up front whether the tolerances require both to be satisfied or only one (AND or OR?).  Also, what happens when you hit max steps?  I hope a rejection is thrown with a warning message.  This should be explained here with the function.  Mention earlier that there are default tolerances and max steps whose values are blah-blah-blah and whose behavior is described later.

src/docs/stan-reference/examples.tex
```diff
@@ -8046,6 +8046,154 @@ \section{Circular Representations of Days and Years}
...
+\subsection{Tolerance}
+%
+The relative and function tolerances control the accuracy of the solution generated by
+the solver. Relative tolerances are relative to the solution value. The function tolerance
+is the norm of the algebraic function, once we plug in the proposed solution. This norm
+should go to 0 (equivalently, all elements of the vector function are 0). It helps to think about this
+geometrically. Ideally the output of the algebraic function is at the origin; the norm measures
+deviations from this ideal. Users should remember the norm scales with the square root of
```
bob-carpenter: the ideal being zero?

src/docs/stan-reference/examples.tex
```diff
@@ -8046,6 +8046,154 @@ \section{Circular Representations of Days and Years}
...
+$f$'s length.
```
bob-carpenter: For L2, the norm is square root of the sum of squares, or pythagorean theorme, so it'd scale with length.  what norm are you talking about here?

src/docs/stan-reference/examples.tex
```diff
@@ -8046,6 +8046,154 @@ \section{Circular Representations of Days and Years}
...
+
+Smaller relative tolerances produce more accurate solutions but require more computational time.
+
+\subsubsection{Sensitivity Analysis} \ \\
+The tolerances should be set low enough that setting them lower does not change the
+statistical properties of posterior samples generated by the Stan program.
+
+\subsection{Maximum Number of Steps}
+%
+The maximum number of steps can be used to stop a runaway simulation. This can arise in
```
bob-carpenter: Explain what happens when you hit the limit.

src/docs/stan-reference/functions.tex
```diff
@@ -3424,6 +3424,127 @@ \chapter{Mixed Operations}\label{mixed-operations.chapter}
...
+another function as one of its arguments. The only other functions
+in Stan which share this feature are the ordinary differential equation
+solvers (see section~\ref{functions-ode-solver.chapter}). Ordinary Stan
+functions do not allow functions as arguments.
+
+Second, some of the arguments of the algebraic solvers are restricted
+to data only expressions. These expressions must not contain variables
+other than those declared in the data or transformed data blocks. Ordinary
+Stan functions place no restriction on the origin of variables in their argument
+expressions.
```
bob-carpenter: The ODE solvers place these restrictions, but those are also not ordinary functions.

Mitzi's finishing the pull request to have a way to mark user-defined functions as having data-only arguments so that these functions can be called safely within user-defined functions with compile-time rather than run-time errors.

src/docs/stan-reference/functions.tex
```diff
@@ -3424,6 +3424,127 @@ \chapter{Mixed Operations}\label{mixed-operations.chapter}
...
+                      real[] x_r, int[] x_i)
+\end{stancode}
+%
+The algebraic system function should return the value of the algebraic
+function which goes to 0, when we plug in the solution to the algebraic
+system.
+
+The argument of this function are:
+\begin{itemize}
+\item \farg{unknown}, the unknowns we wish to solve for
```
bob-carpenter: I think you should use `f(y, theta, x_r, x_i)` consistently to match the ODE discussion.

As is, the first mention above has variable names `x, theta, dat_r, dat_i` and the one right here has `unkown, theta, x_r, x_i`.

src/docs/stan-reference/functions.tex
```diff
@@ -3424,6 +3424,127 @@ \chapter{Mixed Operations}\label{mixed-operations.chapter}
...
+
+\begin{description}
+%
+   \fitemthreelines{real[~]}{algebra\_solver}%
+                  {function \farg{algebra\_system}}%
+                  {real[] \farg{initial\_guess}}%
+                  {real \farg{theta}, real[] \farg{x\_r}, int[] \farg{x\_i}}%
+                  {Solves the algebraic system, given an initial guess,
+                    using the Powell Hybrid algorithm with the unsupported
+                    implementation from Eigen.}
```
bob-carpenter: I'd get rid of the " with the unsupported implementation from Eigen".  It's in Stan now and we're supporting it.

bob-carpenter: I'd also use `y\_init` rather than `initial\_guess` to make it clear which variable we're dealing with and to keep it shorter.  I could live with `y\_initial\_guess`.  Otherwise, this is consistent with the recommendation above.

src/docs/stan-reference/functions.tex
```diff
@@ -3424,6 +3424,127 @@ \chapter{Mixed Operations}\label{mixed-operations.chapter}
...
+                   
+\fitemfourlines{real[~]}{algebra\_solver}%
+                  {function \farg{algebra\_system}}%
+                  {real[] \farg{initial\_guess}}%
+                  {real \farg{theta}, real[] \farg{x\_r}, int[] \farg{x\_i}}%
+                  {real \farg{rel\_tol}, real \farg{f\_tol}, int \farg{max\_steps}}%
+                  {Solves the algebraic system, given an initial guess,
+                    using the Powell Hybrid algorithm with the unsupported
+                    implementation from Eigen with additional control
```
bob-carpenter: same comment as above

src/docs/stan-reference/functions.tex
```diff
@@ -3424,6 +3424,127 @@ \chapter{Mixed Operations}\label{mixed-operations.chapter}
...
+\item \farg{algebra\_system}: function literal refering to a function
+  specifying the system of algebraic equations with signature described
+  in \refsection{functions-algebra-function}:
+  \begin{quote}
+    (real[], real[], real[], int[]):real[]
+  \end{quote}
+  The arguments represent (1) unknown, (2) parameters, (3) real data, and
+  (4) integer data, and the return value contains the value of the algebraic
+  function, which goes to 0 when we plug in the solution to the algebraic system,
+\item \farg{initial\_guess}: initia guess for the solution, type \code{real[]},
```
bob-carpenter: change name,  add "l" after "initia".

bob-carpenter: That's "referring" with two "r"s.

src/docs/stan-reference/functions.tex
```diff
@@ -3424,6 +3424,127 @@ \chapter{Mixed Operations}\label{mixed-operations.chapter}
...
+
+\subsection{Return Value}
+The return value for the algebraic solver is an array of type \code{real[]}, with values
+which, when plugged in as the \code{unknowns} make the algebraic function go to 0.
+
+\subsection{Sizes and Parallel Arrays}
+Certain sizes must match. In particular, the following groups are of the same size:
+%
+\begin{enumerate}
+  \item unknowns passed into the system function, algebraic function returned by the system
```
bob-carpenter: This is confused.  The second group isn't things that have to be the same sizes.  I'd skip the enumeration and say

```
Certain sizes have to be consistent. The initial guess, return value of the solver, and return value of the algebraic function must all be the same size. 

The parameters, real data, and integer data will be passed from the solver directly to the system function. 
```

src/stan/lang/ast/fun/has_non_param_var_vis_def.hpp
```diff
@@ -93,6 +93,17 @@ namespace stan {
         || boost::apply_visitor(*this, e.theta_.expr_);
     }
 
+    bool has_non_param_var_vis::operator()(const algebra_solver& e) const {
+      // if any vars, return true  -- CHECK: nonlinearity?
```
bob-carpenter: Shouldn't this also test the parameters theta?

src/stan/lang/ast/fun/has_non_param_var_vis_def.hpp
```diff
@@ -93,6 +93,17 @@ namespace stan {
...
+      return boost::apply_visitor(*this, e.y_.expr_);
+    }
+
+    bool has_non_param_var_vis::operator()(const algebra_solver_control& e)
+      const {
+      // if any vars, return true
+      return boost::apply_visitor(*this, e.y_.expr_);
```
bob-carpenter: Same comment about parameters theta.

src/stan/lang/ast/fun/has_var_vis_def.hpp
```diff
@@ -69,6 +69,16 @@ namespace stan {
         || boost::apply_visitor(*this, e.theta_.expr_);
     }
 
+    bool has_var_vis::operator()(const algebra_solver& e) const {
+      // only y may contain vars
+      return boost::apply_visitor(*this, e.y_.expr_);
```
bob-carpenter: Same comment again---shouldn't this also be checking theta?

src/stan/lang/ast/fun/has_var_vis_def.hpp
```diff
@@ -69,6 +69,16 @@ namespace stan {
...
+    }
+
+    bool has_var_vis::operator()(const algebra_solver_control& e) const {
+      // only y may contain vars
+      return boost::apply_visitor(*this, e.y_.expr_);
```
bob-carpenter: Same comment again.

src/stan/lang/ast/fun/is_nil_vis.hpp
```diff
@@ -36,6 +38,8 @@ namespace stan {
       bool operator()(const variable& x) const;  // NOLINT(runtime/explicit)
       bool operator()(const integrate_ode& x) const;  // NOLINT
       bool operator()(const integrate_ode_control& x) const;  // NOLINT
+      bool operator()(const algebra_solver& x) const;  // NOLINT
```
bob-carpenter: I don't think these NOLINT markers are necessary, but don't worry about it for this pull request---not your problem.

src/stan/lang/ast/fun/var_occurs_vis_def.hpp
```diff
@@ -62,6 +62,14 @@ namespace stan {
       return false;  // no refs persist out of integrate_ode_control() call
     }
 
+    bool var_occurs_vis::operator()(const algebra_solver& e) const {
+      return false;
```
bob-carpenter: why isn't this and the next function checking recursively into y and theta?  isn't this just looking for expressions with variables through which derivatives need to be taken?

src/stan/lang/ast/node/algebra_solver.hpp
```diff
@@ -0,0 +1,64 @@
...
+      /**
+       * Construct a default algebra solver node.
+       */
+      algebra_solver();
+
+      /**
+       * Construct an algebraic solver.
+       *
+       * @param system_function_name name of ODE system
+       * @param x initial guess for solution
```
bob-carpenter: Whatever you decide, this should be consistent.  I'd prefer as stated above to use theta for parameters, x for ata, and y for solutions and initial guesses.

src/stan/lang/ast/node/algebra_solver_control.hpp
```diff
@@ -0,0 +1,87 @@
...
+namespace stan {
+  namespace lang {
+
+    struct expression;
+
+    /**
+     * Structure for an algebra solver statement with control
+     * parameters for the solver.
+     */
+    struct algebra_solver_control {
```
bob-carpenter: This could've probably been implemented as a subclass of `algebra_solver` to reduce all the duplicated code.  We probably didn't do that for ODEs, either.

src/stan/lang/ast/node/algebra_solver_control.hpp
```diff
@@ -0,0 +1,87 @@
...
+       */
+      expression dat_int_;
+
+      /**
+       * Relative tolerance (real).
+       */
+      expression rel_tol_;
+
+      /**
+       * Function tolerance (real).
```
bob-carpenter: Absolute tolerance?  That's the language you used elsewhere.

src/stan/lang/ast/node/algebra_solver_control.hpp
```diff
@@ -0,0 +1,87 @@
...
+       * Construct a default algebra solver object with control.
+       */
+      algebra_solver_control();
+
+      /**
+       * Construt an algebraic solver with control parameters with
+       * the specified values.
+       *
+       * @param system_function_name name of algebraic solver
+       * @param x initial guess for solution
```
bob-carpenter: same comment about keeping naming of variables consistent

src/stan/lang/ast/node/expression_type_vis_def.hpp
```diff
@@ -43,6 +43,16 @@ namespace stan {
       return expr_type(DOUBLE_T, 2);
     }
 
+    expr_type
+    expression_type_vis::operator()(const algebra_solver& e) const {
+      return expr_type(VECTOR_T, 0);
```
bob-carpenter: Doesn't it return a real value?  All the doc seemed to indicate a `real[]` return, but this is a `vector` return.

src/stan/lang/ast/node/expression_type_vis_def.hpp
```diff
@@ -43,6 +43,16 @@ namespace stan {
...
+    }
+
+    expr_type
+    expression_type_vis::operator()(const algebra_solver_control& e) const {
+      return expr_type(VECTOR_T, 0);
```
bob-carpenter: same question

src/docs/stan-reference/examples.tex
```diff
@@ -8046,6 +8046,154 @@ \section{Circular Representations of Days and Years}
...
+space). In that case, the solver will not converge to a solution. When the solver fails to
+do so, the current metropolis proposal gets rejected.
+
+\section{Control Parameters for the Algebraic Solver}
+%
+The call to the algebraic solver shown above uses the default control settings. The solver
+allows three additional parameters, all of which must be supplied if any of them is
+required.
+%
+\begin{stancode}
```
bob-carpenter: examples of types and sizes would help here

src/stan/lang/grammars/semantic_actions_def.cpp
```diff
@@ -1546,6 +1551,143 @@ namespace stan {
...
+          .is_defined(alg_fun.system_function_name_, system_signature)) {
+        error_msgs << "first argument to "
+                   << "algebra_solver"
+                   << " must be the name of a function with signature"
+                   << " (vector, vector, real[], int[]) : vector ";
+        pass = false;
+      }
+
+      // test regular argument types
+      if (alg_fun.x_.expression_type() != expr_type(VECTOR_T, 0)) {
```
bob-carpenter: If these really are vectors, which the code seems to say they are, all the doc has to be fixed to match.

src/docs/stan-reference/examples.tex
```diff
@@ -8046,6 +8046,154 @@ \section{Circular Representations of Days and Years}
...
+\section{Calling the Algebraic Solver}
+%
+Let's suppose $\theta = \{3, 6\}$. To call the algebraic solver, we need to
+provide an initial guess. A good guess can significantly speed up the solver and
+even determine whether the solver converges or not.
+%
+The following code returns the solution to our nonlinear algebraic system:
+%
+\begin{stancode}
+transformed data {
```
bob-carpenter: It looks like later on that these system states must be vectors, not real arrays.  You need to go back and update all the doc to match.

src/stan/lang/grammars/semantic_actions_def.cpp
```diff
@@ -1546,6 +1551,143 @@ namespace stan {
...
+      sys_arg_types.push_back(expr_type(DOUBLE_T, 1));  // dat
+      sys_arg_types.push_back(expr_type(INT_T, 1));  // dat_int
+      function_signature_t system_signature(sys_result_type, sys_arg_types);
+      if (!function_signatures::instance()
+          .is_defined(alg_fun.system_function_name_, system_signature)) {
+        error_msgs << "first argument to "
+                   << "algebra_solver"
+                   << " must be the name of a function with signature"
+                   << " (vector, vector, real[], int[]) : vector ";
+        pass = false;
```
bob-carpenter: I'd be inclined to return here or use else-if.  Otherwise, we'll get piled up error messages with no endlines between them.  Generally, Stan only tries to report the first spot of trouble.

src/docs/stan-reference/appendices.tex
charlesm93: The semicolon is not part of the code, but of the sentence in which the code is quoted.

Same syntax was used for integrate_ode_*, line 844:
```
\begin{itemize}
\item its first argument to refer to a function with signature
\begin{quote}
 \code{(real,~real[],~real[],~real[],~int[])~:~real[]},
\end{quote}
```

src/docs/stan-reference/examples.tex
charlesm93: good point. I remember in highschool my math teacher emphasizing the difference between f and f(x).

src/docs/stan-reference/examples.tex
charlesm93: If you don't mind, I'll stick to z, since the variable is referred to in the text.

src/docs/stan-reference/examples.tex
charlesm93: I got mixed up with the ODE integrators. Yes, because the algebraic solver is based on Eigen's dogleg, the returned object, initial guess and parameters are vectors. The real and integer data are vectors.

src/docs/stan-reference/examples.tex
charlesm93: Short answer: no. It seems to vary a lot on a case by case basis. Some equations present no pathologies, so the guess doesn't really matter, other systems will be strongly affected by it. I would follow up @betanalpha, @wds15 , and @billgillespie.

src/docs/stan-reference/examples.tex
charlesm93: Yes. We want the algebraic function to go to 0, this is what finding the root means.

> Ideally the output of the algebraic function is at the origin

Not to be confused with "the output the algebraic solver is at the origin".

src/docs/stan-reference/examples.tex
charlesm93: The norm we get from Pythagoras' theorem. An alternative would have been to use the norm for individual elements of z. That is require || z[i] || < ftol, for all i. If there are n elements in the vector, this constraint very roughly translates to || z || < ftol * sqrt(n). This what the thought I was trying to express. For clarity, I'll simply state:

> As the length of z increases, a certain function tolerance becomes an increasingly difficult criteria to meet, since the deviations from every individual element of the vector add up.

bob-carpenter:
+  \code{(~real[], ~real[],~real[],~int[]~:~real[]},

Ah, I see the issue now---it' missing a close )  and the colon's for the result type.

src/docs/stan-reference/functions.tex
charlesm93: > The ODE solvers place these restrictions, but those are also not ordinary functions.

Right, the ODE solvers are mentioned as non-ordinary function in the previous paragraph, so hopefully they won't be any confusion there.

bob-carpenter:
> Short answer: no. It seems to vary a lot on a case by case basis. Some equations present no pathologies, so the guess doesn't really matter, other systems will be strongly affected by it. 

Could you say that in the doc?  And if it fails, will it just hit
the iteration limit without finding an answer?

bob-carpenter:
> The norm we get from Pythagoras' theorem.

That scales up linearly with length of the inputs.

f(3, 4) = sqrt(3^2 + 4^2) = 5

f(6, 8) = sqrt(6^2 + 8^2) = 10

> For clarity, I'll simply state:
> 
> As the length of z increases, a certain function tolerance becomes an increasingly difficult criteria to meet, since the deviations from every individual element of the vector add up.

That sounds good.

src/docs/stan-reference/functions.tex
charlesm93: > I'd get rid of the " with the unsupported implementation from Eigen". It's in Stan now and we're supporting it.

My point is not to caution the function is unsupported, rather to state where the function we built the algebraic solver on comes from.

charlesm93: I think I'll go with `y\_guess`.

src/stan/lang/ast/fun/has_non_param_var_vis_def.hpp
charlesm93: it does. y_ is theta, sorry about the confusing notation. These are the terms we used in the grant proposal, when explaining the implicit function theorem:
x -- unknown
y -- auxiliary parameters
z -- returned value by the algebraic function

src/stan/lang/ast/fun/var_occurs_vis_def.hpp
charlesm93: I'm not sure. I followed the example for the ode integrator here. Judging by the comment:

```
// no refs persist out of integrate_ode_control() call
```
I would say this only matters if one of the argument is passed by ref and modified. Adding the comments for the algebraic solver.

src/stan/lang/ast/node/algebra_solver_control.hpp
charlesm93: Absolute tolerance is misleading. I'll change it to function tolerance.

src/stan/lang/ast/node/algebra_solver.hpp
charlesm93: sorry about that, will adopt the convention used in the manual:
y - initial guess / solution
theta - parameters
x_r and x_i for data

src/stan/lang/ast/node/expression_type_vis_def.hpp
charlesm93: Yes, the doc was wrong. It does return a vector (primarily because the solver is based on a function from Eigen).

src/stan/lang/grammars/semantic_actions_def.cpp
charlesm93: I think it's helpful to see all the errors in one go. And these errors still correspond to "one spot", since they are all caused by one call to the function.

charlesm93:
> That scales up linearly with length of the inputs.

We have different interpretation of the word length, hence our disagreement. I used it to mean the number of elements in a vector, or dimension.

bob-carpenter:
> 
> I'd get rid of the " with the unsupported implementation from Eigen". It's in Stan now and we're supporting it.
> 
> My point is not to caution the function is unsupported, rather to state where the function we built the algebraic solver on comes from.

Understood, but we don't want to make implementations part of the documentation unless we absolutely have to.

bob-carpenter:
> it does. y_ is theta, sorry about the confusing notation. These are the terms we used in the grant proposal, when explaining the implicit function theorem:
> x -- unknown
> y -- auxiliary parameters
> z -- returned value by the algebraic function

Whatever you do, make it consistent.  I think it'd also help if it is consistent with the ODE solver, but that's a bonus.

bob-carpenter:
> 
> I think it's helpful to see all the errors in one go. And these errors still correspond to "one spot", since they are all caused by one call to the function.

OK.  They'll stack up if you're missing one, but there are a finite number of them so it should be OK.  Just make sure if there are such multiple error reports that they wind up with some kind of std::endl separator so they don't all just get smashed together on one line.

bob-carpenter:
> On Aug 7, 2017, at 6:25 PM, Charles Margossian <notifications@github.com> wrote:
> 
> That scales up linearly with length of the inputs.
> 
> We have different interpretation of the word length, hence our disagreement. I used it to mean the number of elements in a vector, or dimension.

OK---try to clarify.  This is always confusing ground.  The basic size function in Stan is size() to avoid this confusion :-)  R uses length() for what Stan calls size().

charlesm93:
> >  Short answer: no. It seems to vary a lot on a case by case basis. Some equations present no pathologies, so the guess doesn't really matter, other systems will be strongly affected by it.
> 
> Could you say that in the doc?  And if it fails, will it just hit
> the iteration limit without finding an answer?

I think this should do it:

This varies on a case-by-case basis, but in general
a good guess will speed up the solver and, in pathological cases, even determine
whether the solver converges or not. If the solver does not converge, the metropolis
proposal gets rejected and a warning message, stating no acceptable solution has
been found and that the maximum number of steps has been reached, is issued.

bob-carpenter:
Let me know when this is ready for final review and I can give it a pass.

charlesm93:
Hi Bob, this should be ready for a final pass. As far as I can tell, the only thing I didn't address was the following comment.

> This could've probably been implemented as a subclass of algebra_solver to reduce all the duplicated code. We probably didn't do that for ODEs, either.

As you write, this wasn't done for the ODEs. If needed, I could find some time to look into it tomorrow.

bob-carpenter:
OK---I'll take a look tomorrow.  No need to factor into a class and a subclass.  I'm not even sure if it would work.

PR #2343, opened 2017-07-02
Added signatures, tests, and docs for append_array function (Issue 481)

bbbales2:
#### Submisison Checklist

#### Summary

Added append_array function hooks into Stan (https://github.com/stan-dev/math/issues/481)

For the docs I woulda simplified the definitions with a T parameter as is used elsewhere in the manual, but these functions don't support appending arrays of arrays so I wanted to be clear (usually T can be array of arrays in all the other places it comes up)

#### How to Verify

./runTests.py src/test/unit/lang/parser/matrix_functions_test.cpp

#### Documentation

Added under "Array Concatenation" in src/docs/stan-reference/functions.tex

#### Reviewer Suggestions

@seantalts or @bob-carpenter

#### Copyright and Licensing

Please list the copyright holder for the work you are submitting (this will be you or your assignee, such as a university or company): University of California, Santa Barbara

### DIFF
--- a/src/docs/stan-reference/functions.tex
+++ b/src/docs/stan-reference/functions.tex
@@ -1744,38 +1744,6 @@ used to extract the dimensions of vectors and matrices.
   type \code{real[4,3]} then \code{size(x)} is 4.}
 \end{description}
 
-%
-\section{Array Concatenation}\label{array-concatenation.section}
-
-Arrays of integers, reals, vectors, row\_vectors, and matrices can
-be concatenated together using the following operations.
-
-\begin{description}
-\fitem{int[]}{append\_array}{int[] \farg{x}, int[] \farg{y}}{
-  Return the concatenation of two arrays of integers in the order
-  of the arguments. }
-%
-\fitem{real[]}{append\_array}{real[] \farg{x}, real[] \farg{y}}{
-  Return the concatenation of two arrays of reals in the order
-  of the arguments. }
-%
-\fitem{vector[]}{append\_array}{vector[] \farg{x}, vector[]
-  \farg{y}}{ Return the concatenation of two arrays of vectors
-  in the order of the arguments. The vectors must all have the same
-  number of rows.}
-%
-\fitem{row\_vector[]}{append\_array}{row\_vector[] \farg{x}, row\_vector[]
-  \farg{y}}{ Return the concatenation of two arrays of row\_vectors
-  in the order of the arguments. The row\_vectors must all have the same
-  number of columns.}
-%
-\fitem{matrix[]}{append\_array}{matrix[] \farg{x}, matrix[]
-  \farg{y}}{ Return the concatenation of two arrays of matrices
-  in the order of the arguments. The matrices must all have the same
-  dimensions.}
-\end{description}
-%
-
 \section{Array Broadcasting}\label{array-broadcasting.section}
 
 The following operations create arrays by repeating elements to fill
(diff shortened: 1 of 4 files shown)


src/docs/stan-reference/functions.tex
```diff
@@ -1744,6 +1744,37 @@ \section{Array Size and Dimension Function}
...
+  \farg{y}}{ Return the concatenation of two arrays of row\_vectors
+  in the order of the arguments. The row\_vectors must all have the same
+  number of columns.}
+%
+\fitem{matrix[]}{append\_array}{matrix[] \farg{x}, matrix[]
+  \farg{y}}{ Return the concatenation of two arrays of matrices
+  in the order of the arguments. The matrices must all have the same
+  dimensions.}
+\end{description}
+%
```
bob-carpenter: Is there a reason `append_array` doesn't just work for any type `T[]` to return `T[]`?  It's OK to doc that way and then list the types for `T`.  Here, it should be easy to implement this for arbitrary arrays.

src/stan/lang/function_signatures.h
```diff
@@ -38,6 +38,9 @@ add("add", MATRIX_T, DOUBLE_T, MATRIX_T);
 for (size_t i = 0; i < base_types.size(); ++i) {
   add("add", base_types[i], base_types[i]);
 }
+for (size_t i = 0; i < base_types.size(); ++i)
+  add("append_array", expr_type(base_types[i], 1), expr_type(base_types[i], 1),
```
bob-carpenter: I think this can be generalized from just the base types to a broad set of types.

bbbales2:
Sure, I'll change it to support all the arrays-of types as well. It should work fine. The only reason I didn't include them was cause there weren't explicit tests in the math library for them. But assuming copy constructors work, it'll be fine :p.

bbbales2:
Alright nevermind, I lied. tldr; I think this should go in with the limited types. I updated the docs to make them shorter.

It seems like the signature:

```c++
template <typename T1, typename T2>
inline typename std::vector<typename return_type<T1, T2>::type>
append_array(const std::vector<T1>& x, const std::vector<T2>& y)
```

Should work. However, ```return_type``` returns the promoted scalar type.

So if ```T1 == std::vector<std::vector<double> >```, ```T2 == std::vector<std::vector<var> >```, then the return type is ```std::vector<var>``` not ```std::vector<std::vector<var> >```.

The promotion needs to be there to allow mixing of data/params.

The crudest way I see to get this to work is to manually add signatures ```std::vector<std::vector<T1> >```, ```std::vector<std::vector<std::vector<T1> > >``` etc.

It should also be possible to build a return_type metaprogram that keeps the ```std::vectors``` intact so the original signature works.

Very likely there's some alternative I missed, but both of these (and probably any alternative) go back to the Math library. The difficulty in any case is figuring out what the tests should be (to keep the combinatorics under control). The implementation code, either way, is probably pretty simple.

Your call, but I think this is useful as is.

bob-carpenter:
What's needed is the metaprogram to compute return type.  Then the rest is trivial.

I agree that it's worth doing even if it lacks generality.

bbbales2:
Let's go ahead and include this as is. It is useful. I'll update the append_array issue with details on the vector return types and put that on my list of things to do.

bbbales2:
I guess this can wait now. I've opened a pull req in the math lib (https://github.com/stan-dev/math/pull/575) to address (https://github.com/stan-dev/stan/pull/2343#pullrequestreview-48164940). Everything should still work here, just once that goes through we can support any type instead of just a subset.

PR #2372, opened 2017-08-03
Fixes to GP chapter

rtrangucci:
#### Submisison Checklist

- [x] Generate doc: `make docs`

#### Summary

Fixes typos in GP chapter models

#### Intended Effect

Make code identical to example model repo's GP code.

#### Side Effects

None

#### Reviewer Suggestions

@bob-carpenter 

#### Copyright and Licensing

Rob Trangucci


rtrangucci:
@seantalts I should have fixed the merge conflicts, I'll wait and see how the tests go.

rtrangucci:
@bob-carpenter finally fixed merge conflicts thanks to @seantalts and the tests are now passing, so this is ready for review when you have a chance. I'd like to get it in for the 2.17 release.

bob-carpenter:
Can you have @seantalts review it.  I'm out all day and I think it's going to need to happen soon.  But then I think Jenkins is hammered with everyone trying to get stuff in, so it may be delayed further.

seantalts:
I'm happy to review it. Going to mostly assume Rob has the math right barring any typos.

src/docs/stan-reference/examples.tex
```diff
@@ -7000,12 +7000,10 @@ \section{Simulating from a Gaussian Process}
   real x[N];
 }
 transformed data {
-  vector[N] mu;
-  cov_matrix[N] K;
-  mu = rep_vector(0, N);
-  K = cov_exp_quad(x, 1, 1);
-  for (i in 1:N)
-    K[i, i] = K[i, i] + 0.1;
+  matrix[N, N] K = cov_exp_quad(x, 1.0, 1.0);
```
seantalts: Why do we prefer matrix[N, N] to cov_matrix[N] here?

rtrangucci:
It's mostly updates to the GP code which mirror the code I just pushed to example-models.

Rob

src/docs/stan-reference/examples.tex
```diff
@@ -7725,41 +7740,41 @@ \subsubsection{Analytical Form of Joint Predictive Inference}
...
-      K_div_y1 = mdivide_right_tri_low(K_div_y1',L_K)';
-      k_x1_x2 = cov_exp_quad(x1, x2, alpha, rho);
-      f2_mu = (k_x1_x2' * K_div_y1);
-      v_pred = mdivide_left_tri_low(L_K, k_x1_x2);
-      cov_f2 = cov_exp_quad(x2, alpha, rho) - v_pred' * v_pred;
-      diag_delta = diag_matrix(rep_vector(delta,N2));
-
-      f2 = multi_normal_rng(f2_mu, cov_f2 + diag_delta);
-    }
+functions{
```
seantalts: Maybe space between s and {?

src/docs/stan-reference/examples.tex
```diff
@@ -7725,41 +7740,41 @@ \subsubsection{Analytical Form of Joint Predictive Inference}
...
+    vector gp_pred_rng(real[] x2,
```
seantalts: Did you mean to add two spaces before everything here?

rtrangucci: That looks off, it should only be 2 spaces in every tab. I can go change that. Good catch.

src/docs/stan-reference/examples.tex
rtrangucci: Yeah, I'll change that. Thanks!

src/docs/stan-reference/examples.tex
bob-carpenter: I'm guessing it's to avoid having an issue with the `cov_matrix` constraint check being too strict.  I think it's better  behaved now.  I'd rather see this declared as `cov_matrix` if that's computationally stable and documented if it's not (yet).

rtrangucci: I guess it'd be fine to use `cov_matrix` in the data block because the constraint checking will only get called once. I never use `cov_matrix` in a model because I'm usually generating the covariance matrix in the `model` block or in a local block in the `transformed parameters` block. If we were to have a covariance matrix defined in the `transformed parameter` I would strongly advocate against using `cov_matrix` because K + diag_matrix(1e-9) will be positive definite by construction and checking covariance matrix constraint checking is redundant.

rtrangucci: @bob-carpenter where and how is the `cov_matrix` constraint checked?

bob-carpenter: At the end of the transformed data and transformed parameter blocks, all constraints are validated.  If transformed data fails, the algorithm never gets off the ground.  If transformed parameters fail, it results in a rejection.

bob-carpenter: And that's a very good point about efficiency.  I'll drop a note to that effect into the efficiency chapter.

rtrangucci: Great, I'll leave the declaration as a `matrix` in. I don't want to expose readers to `cov_matrix` only to use `matrix` in all of the inference examples. Seems confusing to the user.

rtrangucci:
Thanks for reviewing @seantalts. Think it's ready to merge if no one has any other objections.

Issue #2336, opened 2017-06-20
next manual, 2.17


bob-carpenter:
#### Summary:

This is the issue for suggesting fixes for the Stan manual.  Please just add suggestions as comments rather than opening new issues.

#### Current Version:
v2.16.0

bob-carpenter:
HMM Viterbi example problem in `examples.tex`.

Originally reported in https://github.com/stan-dev/example-models/issues/31#issuecomment-309627744

* [ ] fix
* [ ] thank https://github.com/frobnitzem

luisdamiano:
**[programming.tex](https://github.com/stan-dev/stan/blob/develop/src/docs/stan-reference/programming.tex) line 185 of the reference manual v2.16.0**
I believe it only makes sense if "kappa * theta" becomes "kappa * phi" instead, especially since alpha is a k-sized vector.

luisdamiano:
**[programming.tex](https://github.com/stan-dev/stan/blob/develop/src/docs/stan-reference/programming.tex) line 185 of the reference manual v2.16.0**
I believe it only makes sense if "kappa * theta" becomes "kappa * phi" instead, especially since alpha is a k-sized vector.

aaronjg:
in the GP section
"In this case, an inverse gamma, inv_gamma_lpdf in Stan’s language, will work well as it has a sharp
left tail that puts negligible mass on length-scales, but a generous right tail, allowing for large length-scales."

needs an adjective to describe length-scales in the first clause. Perhaps:
"In this case, an inverse gamma, inv_gamma_lpdf in Stan’s language, will work well as it has a sharp
left tail that puts negligible mass on **infinitesimal** length-scales, but a generous right tail, allowing for large length-scales."

bob-carpenter:
OK, I'll fix that.  I'll mention that gamma, inverse gamma and lognormal are all "zero avoiding" in the sense that the limit of the density at zero is zero and estimates near zero will be pushed away.  Then I'll cross-reference the discussion in the regression chapter, which cites Andrew's and Vince's papers.

bob-carpenter:
* [ ] add note to `int_step()` saying that it differs in behavior at 0 from `step()`
* [ ] add similar note to `step()`

aaronjg:
In the GP section on page 247 and 255, the example code multiplies by 1/2. It seems like this would just round to 1 given integer division. Everywhere else in the manual uses multiplication by 0.5 instead.

aaronjg:
[...]

PR #2400, opened 2017-09-22
Fixing the integration variable

mapio:
In the definition of the Cumulative Distribution Functions the integration variable was incorrectly defined as $\theta$ instead of $y%.

### DIFF
--- a/src/docs/stan-reference/distributions.tex
+++ b/src/docs/stan-reference/distributions.tex
@@ -118,7 +118,7 @@ F_Y(y)
 \ = \
 \mbox{Pr}[Y < y]
 \ = \
-\int_{-\infty}^y p(y \, | \, \theta) \ \mathrm{d}\theta.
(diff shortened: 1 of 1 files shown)


src/docs/stan-reference/distributions.tex
```diff
@@ -118,7 +118,7 @@ \section{Cumulative Distribution Functions}
 \ = \
 \mbox{Pr}[Y < y]
 \ = \
-\int_{-\infty}^y p(y \, | \, \theta) \ \mathrm{d}\theta.
+\int_{-\infty}^y p(y \, | \, \theta) \ \mathrm{d}y.
```
bob-carpenter: This should really be a different bound variable `y'`.  I can just update this in the next round of updates.

bob-carpenter:
I'll just fix this with the next manual update and thank you for the correction in the acknowledgements.  Thanks!

bob-carpenter:
Oops, forgot the reference:

https://github.com/stan-dev/stan/issues/2336#issuecomment-331960060

PR #2414, opened 2017-10-05
Jenkinsfile and pipelines

seantalts:
#### Copyright and Licensing

Columbia University

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -22,11 +22,12 @@ def setup(String pr, Boolean failOnError = true) {
     return script
 }
 
-def mailDevs(String label) {
+def mailBuildResults(String label) {
     emailext (
         subject: "[StanJenkins] ${label}: Job '${env.JOB_NAME} [${env.BUILD_NUMBER}]'",
         body: """${label}: Job '${env.JOB_NAME} [${env.BUILD_NUMBER}]': Check console output at ${env.BUILD_URL}""",
-        recipientProviders: [[$class: 'DevelopersRecipientProvider']]
+        recipientProviders: [[$class: 'RequesterRecipientProvider']],
+        to: "${env.CHANGE_AUTHOR_EMAIL}, stan-buildbot@googlegroups.com"
     )
 }
 
@@ -34,11 +35,20 @@ def runTests(String testPath) {
     "runTests.py -j${env.PARALLEL} ${testPath} || echo ${testPath} failed"
 }
 
+def updateUpstream(String upstreamRepo) {
+    if (env.BRANCH_NAME == 'develop') {
+        sh "curl -O https://raw.githubusercontent.com/stan-dev/ci-scripts/master/jenkins/create-${upstreamRepo}-pull-request.sh"
+        sh "sh create-${upstreamRepo}-pull-request.sh"
+    }
+}
+
+
 pipeline {
     agent none
     parameters {
         string(defaultValue: '', name: 'math_pr', description: "Leave blank "
-                + "unless testing against a specific math repo pull request.")
+                + "unless testing against a specific math repo pull request, "
+                + "e.g. PR-640.")
     }
     options { skipDefaultCheckout() }
     stages {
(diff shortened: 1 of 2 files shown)


seantalts:
@syclik sorry for spam on this one. I think it's ready for review now.

seantalts:
(ignore the failed "default" test suite - I canceled it since the only change is the Jenkinsfile and you should be able to see all the same results in the pr-merge link to the pipeline'd job)

Jenkinsfile
```diff
@@ -0,0 +1,164 @@
...
+                        bat setupCC()
+                        bat "make -j${env.PARALLEL} test-headers"
+                    }
+                    post { always { deleteDir() } }
+                }
+                stage('Unit') { 
+                    agent any
+                    steps {
+                        unstash 'StanSetup'
+                        sh setupCC(false)
```
syclik: Why is the argument `false` here?

Jenkinsfile
```diff
@@ -0,0 +1,164 @@
...
+    if (env.BRANCH_NAME == 'develop') {
+        sh "curl -O https://raw.githubusercontent.com/stan-dev/ci-scripts/master/jenkins/create-${upstreamRepo}-pull-request.sh"
+        sh "sh create-${upstreamRepo}-pull-request.sh"
+    }
+}
+
+pipeline {
+    agent none
+    parameters {
+        string(defaultValue: '', name: 'math_pr', description: "Leave blank "
```
syclik: Is the idea to provide the changed math branch to this Stan PR?

Mind describing how to deal with a math branch that breaks Stan or a Stan branch that breaks CmdStan?

Jenkinsfile
seantalts: Some tests will fail with `-Werror`, for example the multiple translation units test fails (at least on Jenkins) with:
```
[Unit] clang: error: test/unit/multiple_translation_units1.o: 'linker' input unused
[Unit] clang: error: test/unit/multiple_translation_units2.o: 'linker' input unused
```

(from http://d1m1s1b1.stat.columbia.edu:8080/job/Stan%20Pipeline/view/change-requests/job/PR-2414/23/console)

Though it doesn't fail like that for me locally...

Jenkinsfile
seantalts: Yeah, so that if you wanted to run the Stan suite (likely on its own PR or branch) against a specific math PR, you could do that here in a way that's pretty similar to how we currently do it ("Build with parameters" and, now, enter just the github PR number). CmdStan has two similar parameters, `stan_pr` and `math_pr` that you enter in a similar way on the build with parameters screen:
image


I need to add a doc string to those parameters in CmdStan, woops. 

So this pipelines thing does not immediately seem to be able to improve the situation regarding these multi-repo breaking changesets :(

seantalts: Stan and Math could also take parameters indicating which CmdStan (and Stan) PRs to build against when running the upstream tests, which is maybe slightly better because then the test results show up against the lower repo's pull request (on github) as that check or X and it runs them all at once. But there's no way to automatically do that, either, and these parameters are necessary because these jobs are triggered by the tests downstream with the parameters filled in - this is new. Currently we have a sort of more basic single-job "upstream" test defined (e.g. Upstream Stan in Math: http://d1m1s1b1.stat.columbia.edu:8080/job/Math%20Pull%20Request%20-%20Upstream%20-%20Stan/) but what I have happening here is triggering a complete build of the upstream pipeline (for now under this "downstream tests" "job" shown on this page here: http://d1m1s1b1.stat.columbia.edu:8080/job/CmdStan/)

syclik: Great! Thanks for the clarificaion

Jenkinsfile
syclik: Makes sense. Perhaps that's using the wrong compiler? Seems like clang instead of clang++. But we can leave that for another day.

seantalts: The line above that error shows `clang++` was the thing actually called:
```
[Unit] clang++ -Werror -Wall -I . -isystem lib/stan_math/lib/eigen_3.3.3 -isystem lib/stan_math/lib/boost_1.64.0 -isystem lib/stan_math/lib/cvodes_2.9.0/include -std=c++1y -DBOOST_RESULT_OF_USE_TR1 -DBOOST_NO_DECLTYPE -DBOOST_DISABLE_ASSERTS -DBOOST_PHOENIX_NO_VARIADIC_EXPRESSION -Wno-unused-function -Wno-uninitialized -I src -isystem lib/stan_math/ -DFUSION_MAX_VECTOR_SIZE=12 -Wno-unused-local-typedefs -stdlib=libc++ -Wno-unknown-warning-option -Wno-tautological-compare -Wsign-compare -DNO_FPRINTF_OUTPUT -pipe  -c -shared -o test/unit/libmultiple.so test/unit/multiple_translation_units1.o test/unit/multiple_translation_units2.o
```

I am a bit confused about this, but these jobs didn't have `-Werror` in the existing Jenkins jobs either, so for now I think we're preserving status quo in this area.

syclik:
Ok. (I thought I saw a message about clang instead of clang++.)

Makes sense.

PR #2423, opened 2017-10-20
fix Backus-Naur form gramma of print and reject

jan-glx:
#### Summary
Reference (manual) only.

The old grammar would contain `print("Hallo " "world!")`  but not `print("Hello ", "world!")`  while only the latter is actually accepted by stan. Should be a comment in #2336 . sorry, next time.

#### Submission Checklist

#### Copyright and Licensing

CC-BY 4.0 (https://creativecommons.org/licenses/by/4.0/)

### DIFF
--- a/src/docs/stan-reference/appendices.tex
+++ b/src/docs/stan-reference/appendices.tex
@@ -715,8 +715,8 @@ atomic_statement_body
    | 'target' '+=' expression
    | 'break'
    | 'continue'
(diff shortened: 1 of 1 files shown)


bob-carpenter:
Thanks --- I'll merge as soon as it passes tests (this is way overkill, but we're still working on streamlining testing for specific things that we know won't break the code).

I'll also thank you in the manual:   https://github.com/stan-dev/stan/issues/2336#issuecomment-338399246)

PR #2438, opened 2017-11-17
Feature/foreach loops

VMatthijs:
#### Submission Checklist

#### Summary
Adds foreach loops to loop over arrays and matrices.

#### Intended Effect
Make it easier to loop over arrays and matrices.

#### How to Verify
Run `break_continue_test.cpp` and `for_statements_bad_indices_test.cpp` for type checking and error checking and look at generated code for `break-continue.stan`.

#### Copyright and Licensing

Matthijs Vákár


VMatthijs:
@bob-carpenter (or @mitzimorris ) , I've now got it to generate the right code. However, two things still need to happen and I think I could use some guidance on those:
1) type checking
2) overloading of the keyword 'for' (currently all loops have distinct keywords).

The first should be done with semantic actions, I think. The second should involve moving the semantic action which adds an identifier to the variable map until after the type checking has succeeded (while keeping a way to reference the identifier which was originally parsed by storing it in some way using a semantic action, presumably).

For 1), could reuse the existing semantic actions `infer_array_expr_type` and `infer_vec_or_matrix_expr_type`? They seem to do something very similar to checking that something is an array or a matrix.

For 2), do you have any suggestions on how I might store the result of a parse using a semantic action?

More generally, do you have any advice on how to get familiar with boost::spirit::qi? I've been reading the documentation, but their examples are all much simpler than what you're doing, it seems.

Thanks!

mitzimorris:
w/r/t question 1, yes, semantic actions is where type checking happens and if you can reuse or generalize existing functions, by all means do so.

> advice on how to get familiar with boost::spirit::qi?

searching specifically on site:http://boost-spirit.com sometimes pulls up blogposts with examples and comments of how to do complicated things.
agreed that the docs are maddeningly trivial.

VMatthijs:
Thanks, Mitzi!

Any words of wisdom on how it normally deals with overloaded keywords? In particular, I'd like to use the keyword "for" for both kinds of foreach-loops (as well as for normal for loop). I can't seem to get that working though. Therefore, I've temporarily used the keywords "fora" and "form" for array and matrix foreach loops.

If I try to change "fora" to "for", the parser starts to complain. I don't see why the rules should clash, though.

I'm afraid I'm going to have to ask you guys for some more help understanding the code base. I think I'm a bit stuck as is.

bob-carpenter:
The parser is top-down, left to right.  It'll backtrack.  So the only thing you need to do is overload is write two rules that start with "for".  

If you use > rather >> for sequencing, it sets up an "expectation" for the next item to parse.  When expectations fail, the whole parse fails and you get a message indicating where it failed.  Otherwise, it'll backtrack, though you have to be careful there if there are side effects in rules that affect the parent category in a rule (by changing `_val`)---those effects will persist over backtracking.

So I was suggesting changing the current rule to push off the check for integer type until after you got the range `:` in the original rule.  To do that, you need to add a local variable to the rule declaration and set it when you get the identifier, then check that it's an integer after you get the `:` range operator.  

  That should set up an expectation that fails.  That is, once you get the `:`, then you want to fail altogether if you didn't have an integer.  You can do that by using an `eps` (epsilon, matches nothing) with an expectation, that is `> eps[action taking _pass as reference]`  The point here is that if you set `_pass` to false in the semantic action because the epsilon was under an expectation (the `>`), the whole parse will fail.  And any message you wrote out during the semantic action will propagate back to the user and the little carat that says where the error is at will point right up to the next character to consume.

src/stan/lang/grammars/statement_grammar_def.hpp
```diff
@@ -192,12 +204,47 @@ namespace stan {
       for_statement_r
```
VMatthijs: Thanks, @bob-carpenter ! Is this the sort of use of a local variable you mean? My intention is to store the identifier in `_a`. I'll look into the difference between `>` and `>>`. Thanks for explaining. Could we maybe plan a time on Monday or Tuesday (or Wednesday before 3:30pm) to sit down together and go through some compiler code? It'd be nice to get to the point where I understand it well enough myself to be productive independently. Wednesday will be my last day in NYC for this trip sadly. Also, do you still want to try to meet Andrew together?

src/stan/lang/grammars/semantic_actions_def.cpp
```diff
@@ -1464,6 +1464,48 @@ namespace stan {
...
+                                         const scope& var_scope,
+                                         bool& pass, variable_map& vm,
+                                         std::stringstream& error_msgs) const {
+      std::size_t numdims = expr.expression_type().num_dims();
+      pass = !(vm.exists(name)) && (numdims > 0);
+      if (!pass)
+        error_msgs << "ERROR: loop variable already declared."
+                   << " variable name=\"" << name << "\"" << std::endl;
+      else
+        vm.add(name, base_var_decl(name, std::vector<expression>(), expr.expression_type().type()),
```
VMatthijs: Hi @mitzimorris , @bob-carpenter , do you have any suggestions on how to get a vector of dimension sizes out of an expression? To use `base_var_decl` to add an array element to the var map, we need to specify such a vector. I can only find how to get the number of dimensions out of an expression. Thanks!

VMatthijs: (It seems to me like this is not possible. Should I just declare the variable without without sizes for its dimensions? How would I do that? Presumably I want to pass a vector with length `numdims - 1`, but I'm not sure what it should contain.)

mitzimorris: at compile time, the size of the dimensions is unknown - you only know the number of dimensions

VMatthijs: Thanks, Mitzi! I am wondering then what the following constructor expects as its second argument:
`      /**
       * Construct a base variable declaration with the specified
       * name, dimensions, and base type.  The definition is set to
       * nil. 
       *
       * @param name name of variable
       * @param dims dimensions of variables
       * @param base_type base expression type of variable
       */
      base_var_decl(const std::string& name,
                    const std::vector<expression>& dims,
                    const base_expr_type& base_type);

mitzimorris: helper function `infer_var_dims_type` ??
(line 167 of semantic_actions_def.hpp)

VMatthijs: ( it lists 
```     /**
       * Dimension sizes for variable.
       */
      std::vector<expression> dims_;```
)

**Comment by bob-carpenter** [comment]
You'll have the expressions that determine the dimensions at runtime.

For the foreach loops, you shouldn't need any dimension sizes.  Any code that gets generated should use

matrix<T, R, C> x;
for (int i = 0; i < x.size(); ++i)
  ... x(i) is value ...

**Comment by bob-carpenter** [comment]
>        * Dimension sizes for variable.
>        */
>       std::vector<expression> dims_;

These are the expressions in the AST for the dimensions (probably just array dimensions, but maybe all the dimensions including matrix or vector dimensions in some cases).  At runtime, the code they generate will denote the dimension sizes when executed.

**Comment by VMatthijs** [comment]
Thanks, @bob-carpenter ! If the `dims_` are only determined at runtime, do you think I should just initialise using
``` base_var_decl(name,
                           std::vector<expression>(numdims),
                           sometype);
``` 
?

VMatthijs:
Hi @mitzimorris , @bob-carpenter , This can be code reviewed, as far as I'm concerned. I do need to create some documentation for the manual still. Am in the process of doing that now.

VMatthijs:
Hi @mitzimorris , I've run the generated C++ code on a bunch of examples and it does precisely what it should. I think this is ready for review.

bob-carpenter:
@mitzimorris Are you going to review this or should I?

bob-carpenter:
Jenkins, retest this please.

bob-carpenter:
Jenkins, test this please.

bob-carpenter:
Does anyone know the incantation to make Jenkins test this?  It didn't seem to respond to the previous two requests (or it's very very slow, as it's been over a minute now).

VMatthijs:
Jenkins, ok to test.

VMatthijs:
Jenkins retest this please

mitzimorris:
Jenkins, retest this please.

mitzimorris:
Jenkins tests are failing in the integration test:

```
[Integration] ./src/test/test-models/good/break-continue.hpp: In function ‘int break_continue_model_namespace::foo(const int&, std::ostream*)’:
[Integration] ./src/test/test-models/good/break-continue.hpp:116:24: error: unused variable ‘v’ [-Werror=unused-variable]
[Integration]              for (auto& v : vs) {
```

investigating...

mitzimorris:
this test fails:
```
make src/test/test-models/good/break-continue.hpp-test
```

mitzimorris:
@VMatthijs, were any compiler flags set in your `make/local` file?  if so, they'd need to be added to the appropriate file in the `make` dir.

VMatthijs:
Hi @mitzimorris , I'll get onto this tomorrow and will let you know when it's passing Jenkins.

VMatthijs:
Sorry for the delay. I'd still like to make a small improvement before I think this can go in.

mitzimorris:
questions about changes made in last commit: https://github.com/stan-dev/stan/pull/2438/commits/1217a7651106bc3b97491c94cee478e5e0cabf5f

why are statments like:

```
 for (v in vs) break; 		
 for (v in vs) continue;
```

no longer legal?  is this what we want?  has it been documented?

VMatthijs:
They are still legal. The Stan compiler doesn't complain. It's just that the C++ compiler gives a warning if we  have a foreach loop where we do not use the loop variable. The tests were failing before because of this clang warning, which is a bit silly.
However, I'd still like to make a minor change before this is ready for final review. I feel I could generate nicer code for the matrix loop.

bob-carpenter:
No rush at all!

VMatthijs:
Great! I'm trying to get the GLMs into Stan first. By the way, for that purpose, I'd like to make a small change to both the math library GLMs and add the interfaces. Can that go in one pull request?

bob-carpenter:
You can generate a void usage of a variable to avoid that error.  We do it elsewhere in the code.  See:

```
stan/src/stan/lang/generator/generate_void_statement.hpp
```

bob-carpenter:
We try to keep the GitHub comments on topic.  This one should go in the GLM.  The language request has to be made after the math library request is working.  If the math library breaks the language or algorithms somehow, then it's trickier.

VMatthijs:
Hi @mitzimorris , I've been made the changes I wanted to make, have run enough models to test and sanity check to feel that this is ready for review. Hope the course is going well!

VMatthijs:
But, @bob-carpenter , isn't that warning something we want? If someone writes a foreach loop over an array or a matrix and doesn't refer to the loop variable, shouldn't that give a warning?

bob-carpenter:
@VMatthijs --- We don't want C++ compiler errors trickling through to users.  It'd be nice to try to catch and throw a parser warning in Stan if the foreach loop variable wasn't used in the body, but it's not necessary for this pull request.

VMatthijs:
OK, I've made the change, @bob-carpenter . This is ready for review whenever is good for you, @mitzimorris .

mitzimorris:
thanks Matthijs - will review ASAP

VMatthijs:
Thanks, @mitzimorris ! There's no rush. Just wanted to let you know it's done. :-)

mitzimorris:
@VMatthijs do you want me to merge this?

VMatthijs:
Hi @mitzimorris , That'd be great! I just did some last proof reading of the manual entry and it looks good now. :-)

Issue #2450, opened 2017-12-29
fix effective sample size computation and documentation


avehtari:
#### Summary:
Function effective_sample_size stan/mcmc/chains.hpp uses inferior approach to truncate the autocorrelation series causing reported N_eff to be inaccurate when N_eff close to N or N_eff>N. 
In addition the documentation docs/stan-reference/inference.tex has some errors.
Fix: use Geyer's initial positive sequence in the code and update the documentation.

#### Description:

**Code: effective_sample_size function in stan/mcmc/chains.hpp**

The current code truncates autocorrelation series based on the first
negative autocorrelation, which gives wrong results when the true
first lag autocorrelation is negative or close to zero. Summary of
experiments demonstrating this can be found in Discourse.

- [ ] Fix: Use Geyer's initial positive sequence (Geyer, 1992), which gives
more accurate results. Change the code comment to refer to Stan
reference model for the details of the algorithm used.

This issue is for N_eff computations using the function effective_sample_size in stan/mcmc/chains.hpp
If some interfaces or packages use their own implementation, they should be fixed, too.

**Documentation: docs/stan-reference/inference.tex Section 29.4**

Stan reference Section 29.4 describes a variogram-based estimator for
autocorrelations (same as in BDA3).  The code uses fast Fourier
transform to compute autocorrelations. Based on the experiments,
computing autocorrelations directly is also more accurate for higher
lags than the variogram estimator given in BDA3. There are other
variogram estimators, but since the current autocorrelations
computation is accurate enough and FFT based computation is faster,
there is no need to look for alternative variogram estimators.

- [ ] Fix: Give equations based on autocorrelations instead of variograms.

Documentation says: "Stan does not (yet) do the paired expectations
because NUTS almost by construction avoids the negative
autocorrelation regime. Thus terminating at the first negative
autocorre- lation is a reasonable approximation for stopping when the
noise in the autocorrelation estimator dominates." Based on the
experiments summarized at Discourse, this does not hold when N_eff is
close to N or larger than N.

- [ ] Fix: State that Geyer's initial positive sequence is used as described
in BDA3.

- [ ] Fix other errors in Section 29.4.

#### Additional Information:
See discussion and examples of current and preferred behavior in discourse thread http://discourse.mc-stan.org/t/n-eff-bda3-vs-stan/2608. The latest experiments in comment http://discourse.mc-stan.org/t/n-eff-bda3-vs-stan/2608/21

#### Current Version:
v2.17.1

PR #2451, opened 2017-12-29
Bugfix/issue 2450 effective sample size

avehtari:
#### Submission Checklist

#### Summary
Use Geyer's initial positive sequence in the code and update the documentation.
Fixes issue #2450 

#### Copyright and Licensing

Please list the copyright holder for the work you are submitting (this will be you or your assignee, such as a university or company): Aki Vehtari


avehtari:
Tests are failing because the new version returns slightly different result for n_eff. I can change the test to use new values, but I'll wait first other potential comments.

avehtari:
Based on the experiments and discussion at StanCon, I added Geyer's initial monotone sequence estimator. It will reduce the chance of getting rare too low n_eff estimates, and thus has slightly smaller overall variation than using Geyer's initial positive sequence estimator.

avehtari:
Oh, I need to update the documentation to mention initial monotone sequence estimator.

avehtari:
> This is just doc and it all looks fine. Thanks.

Changes in chains.hpp is not just doc.

syclik:
@bob-carpenter, there’s also code in the PR. Did you want to review the code too? (If not, I can handle that part of it.)

bob-carpenter:
Don't know how I missed that.  Sure, @syclik, go ahead and review.

src/stan/mcmc/chains.hpp
```diff
@@ -254,22 +256,39 @@ namespace stan {
...
         Eigen::VectorXd rho_hat_t(n_samples);
         rho_hat_t.setZero();
-        double rho_hat = 0;
-        int max_t = 0;
-        for (int t = 1; (t < n_samples && rho_hat >= 0); t++) {
-          Eigen::VectorXd acov_t(chains);
-          for (int chain = 0; chain < chains; chain++) {
-            acov_t(chain) = acov(chain)(t);
+        double rho_hat_even = 1;
+        double rho_hat_odd = 0;
```
syclik: Remove this declaration + definition. Move it both down to where it's first used, line 264.

So, it should look more like:
```
double rho_hat_even = 1;
double rho_hat_odd = 1 - (mean_var - mean(acov_t)) / var_plus;
```

src/stan/mcmc/chains.hpp
```diff
@@ -254,22 +256,39 @@ namespace stan {
...
+        Eigen::VectorXd acov_t(chains);
+        for (int chain = 0; chain < chains; chain++)
+          acov_t(chain) = acov(chain)(1);
+        rho_hat_odd = 1 - (mean_var - mean(acov_t)) / var_plus;
+        rho_hat_t(1) = rho_hat_odd;
+        // Geyer's initial positive sequence
+        int max_t = 1;
+        for (int t = 1;
+             (t < (n_samples - 2) && (rho_hat_even + rho_hat_odd) >= 0);
```
syclik: minor: don't need the outer parenthesis. It's all clear since it's on its own line. (the old version had the whole for loop in a single line, so it helped readability) Feel free to do what you think is more readable.

src/stan/mcmc/chains.hpp
```diff
@@ -254,22 +256,39 @@ namespace stan {
...
+        double rho_hat_even = 1;
+        double rho_hat_odd = 0;
+        Eigen::VectorXd acov_t(chains);
+        for (int chain = 0; chain < chains; chain++)
+          acov_t(chain) = acov(chain)(1);
+        rho_hat_odd = 1 - (mean_var - mean(acov_t)) / var_plus;
+        rho_hat_t(1) = rho_hat_odd;
+        // Geyer's initial positive sequence
+        int max_t = 1;
+        for (int t = 1;
```
syclik: Please switch the indexes to start at `t = 2`. If you look in the loop, there's no direct use of `t`, it's always `t + 1` or `t + 2`. (The only place it's used as `t` is within the condition in the for loop, which can be changed easily)

src/stan/mcmc/chains.hpp
```diff
@@ -254,22 +256,39 @@ namespace stan {
...
+             (t < (n_samples - 2) && (rho_hat_even + rho_hat_odd) >= 0);
+             t += 2) {
+          for (int chain = 0; chain < chains; chain++)
+            acov_t(chain) = acov(chain)(t+1);
+          rho_hat_even = 1 - (mean_var - mean(acov_t)) / var_plus;
+          for (int chain = 0; chain < chains; chain++)
+            acov_t(chain) = acov(chain)(t+2);
```
syclik: spacing: `t + 2` instead of `t+2`

src/stan/mcmc/chains.hpp
```diff
@@ -254,22 +256,39 @@ namespace stan {
...
+          rho_hat_odd = 1 - (mean_var - mean(acov_t)) / var_plus;
+          if ((rho_hat_even + rho_hat_odd) >= 0) {
+            rho_hat_t(t+1) = rho_hat_even;
```
syclik: spacing:`t + 1` instead of `t+1`

src/stan/mcmc/chains.hpp
```diff
@@ -254,22 +256,39 @@ namespace stan {
...
+            rho_hat_t(t+2) = rho_hat_odd;
```
syclik: spacing

src/stan/mcmc/chains.hpp
```diff
@@ -254,22 +256,39 @@ namespace stan {
...
           }
-          rho_hat = 1 - (mean_var - mean(acov_t)) / var_plus;
-          if (rho_hat >= 0)
-            rho_hat_t(t) = rho_hat;
-          max_t = t;
+          max_t = t+2;
```
syclik: spacing

src/stan/mcmc/chains.hpp
```diff
@@ -254,22 +256,39 @@ namespace stan {
...
         }
-        double ess = chains * n_samples;
-        if (max_t > 1) {
-          ess /= 1 + 2 * rho_hat_t.sum();
+        // Geyer's initial monotone sequence
+        for (int t = 3; t <= (max_t - 2); t += 2) {
```
syclik: remove the parenthesis here: `t <= (max_t - 2)`
the order of operations in C++ is a known thing and there shouldn't be ambiguity

src/stan/mcmc/chains.hpp
```diff
@@ -254,22 +256,39 @@ namespace stan {
...
+          if ((rho_hat_t(t+1) + rho_hat_t(t+2)) > (rho_hat_t(t-1) + rho_hat_t(t))) {
```
syclik: spacing. also remove the parens on the left and right of the `>`

src/stan/mcmc/chains.hpp
```diff
@@ -254,22 +256,39 @@ namespace stan {
...
+            rho_hat_t(t+1) = (rho_hat_t(t-1) + rho_hat_t(t)) / 2;
+            rho_hat_t(t+2) = rho_hat_t(t+1);
+          }
```
syclik: I believe there's an error here. If you're only using the initial monotone sequence, then I believe you need (not sure how to describe this better):
```
else {
  rho_hat_t(t + 1) = 0;
  rho_hat_t(t + 2) = 0;
}
```
all the way out to `max_t`. I believe if the monotone sequence length is less than `max_t`, the elements beyond the monotone sequence still has possibly non-zero values in it.

The alternative is to wrap this condition into the loop above this one and only go out as long as it is monotone.

src/stan/mcmc/chains.hpp
avehtari: Indexing was to follow BDA3 to make it easier to compare code and explanation. Due to adding monotone sequence, the indexing in manual is now different from BDA3. Should I follow your recommendation or try to make code readable compared to equations in documentation?

src/stan/mcmc/chains.hpp
avehtari: monotone sequence length is `max_t`

avehtari: thanks @syclik , see two of my comments inline. I'll update tests

avehtari:
thanks @syclik , see two of my comments inline. I'll update tests

src/stan/mcmc/chains.hpp
syclik: I see. Let me look at BDA3 and I’ll get back to you.

avehtari: I think it's better to look at the updated stan/src/docs/stan-reference/inference.tex as it has now different indexing than BDA3. The initial positive sequence was easy enough to describe without having explicit symbol for the paired sums and Andrew wanted then leave that out from BDA3. Monotone sequence is easier to describe with explicit symbol for paired sums, and the updated inferences.tex uses that. The code is closer to description in inferences.tex but doesn't have the explicit variable for the paired sum, so the indexing is still different.

avehtari:
I updated n_eff test with values from Geyer's imse values using the same blocker csv's

src/stan/mcmc/chains.hpp
```diff
@@ -254,22 +256,39 @@ namespace stan {
...
+          max_t = t + 2;
         }
-        double ess = chains * n_samples;
-        if (max_t > 1) {
-          ess /= 1 + 2 * rho_hat_t.sum();
+        // Geyer's initial monotone sequence
+        for (int t = 3; t <= max_t - 2; t += 2) {
+          if (rho_hat_t(t + 1) + rho_hat_t(t + 2) >
+              rho_hat_t(t - 1) + rho_hat_t(t)) {
+            rho_hat_t(t + 1) = (rho_hat_t(t - 1) + rho_hat_t(t)) / 2;
```
syclik: I'm pretty sure this logic is incorrect. I believe if `rho_hat_t(t + 1) + rho_hat_t(t + 2) > rho_hat_t(t - 1) + rho_hat_t(t)`, we want the `rho_hat_t` from `t + 1` until the end (or just out to `max_t`) to be 0.

I'm not sure why we're filling in the value with half of the sum of the larger value.

I think the right thing to do is to set `rho_hat_t(t + 1)` and `rho_hat_t(t + 2)` to 0 and we can leave the loop as-is.

syclik: Also, should we start at `t = 1`? Is there any chance that the sum of the second pair is bigger than the first pair?

src/docs/stan-reference/inference.tex
```diff
@@ -579,80 +579,63 @@ \subsection{Definition of Effective Sample Size}
...
-negative autocorrelation is a reasonable approximation for stopping
-when the noise in the autocorrelation estimator dominates.
-
-Stan carries out the autocorrelation computations for all lags
-simultaneously using Eigen's fast Fourier transform (FFT) package with
-appropriate padding; see \citep{Geyer:2011} for more detail on using
-FFT for autocorrelation calculations.
+where
+\[
+\hat{\tau} = 1 + 2 \sum_{t=1}^{2m+1} \hat{\rho}_t = -1 + 2 \sum_{t'=0}^{m} \hat{P}_{t'},
```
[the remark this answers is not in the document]
avehtari: The first sum starts from t=1, so rho_0 is not included and rho_0=1. The second sum starts from t'=0, and the P_0=rho_0+rho_1, and thus the sum includes rho_0 and as the sum is multiplied by 2, rho_0 is there twice, and we need -1. This as in Geyer (1992).

src/stan/mcmc/chains.hpp
avehtari: The logic is correct. Geyer (1992) uses only the paired sums in description of monotone sequence. I can rewrite this to use pairs sums, if you think this is too confusing. If a paired sum is larger than the previous paired sum, we set it to the value of the previous paired sum. When we do this sequentially, the series of the paired sums is non-increasing. Since we care only about the paired sums, it really doesn't matter what are the actual values for rho's as long as the paired sum is correct. I just decided to divide the paired sum equally for two rho's in the sum. There is a very small probability that second pair is larger than the first pair, and even then the difference has to be small.

syclik: I finally got it. Yes, the logic is right. It's not confusing at all.

I was having trouble picking it out of the wording: "The initial monotone sequence is obtained by further reducing $Pˆt′$ to the minimum of the preceding ones so that the estimated sequence is monotone."

I had misread the text, so if the original summed pair sequence was something like `(1.0, 0.9, 0.2, 0.8, 0.7, 0.0)`, I thought it should have been `(1.0, 0.9, 0.2, 0.0, 0.0, 0.0)`, but after re-reading, your code produces the right thing: `(1.0, 0.9, 0.2, 0.2, 0.0)`.

src/docs/stan-reference/inference.tex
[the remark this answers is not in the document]
syclik: Thanks! I missed that the index changed to starting at 0. (I don't know why it's any changing that index, but ok.)

PR #2471, opened 2018-02-20
Feature/issue 2467 vectorized signatures rn gs

roualdes:
#### Submission Checklist

#### Summary

 Vectorize the *_rng distribution functions following stan-dev/math#605 stan-dev/math#722.

#### Intended Effect

Vectorize the following functions

```
beta_rng
cauchy_rng
chi_square_rng
double_exponential_rng
exp_mod_normal_rng
exponential_rng
frechet_rng
gamma_rng
gumbel_rng
inv_chi_square_rng
inv_gamma_rng
logistic_rng
lognormal_rng
pareto_rng
pareto_type_2_rng
rayleigh_rng
scaled_inv_chi_square_rng
skew_normal_rng
student_t_rng
weibull_rng
```

#### How to Verify

Double check updated tests in  `stan/src/test/test-models/good/function-signatures/distributions/rngs.stan` and/or run

`./runTests.py src/test/unit/lang/parser/distribution_functions_test.cpp`

I didn't run `src/test/unit`, but the tests (and updated tests) from `src/test/unit/lang` do pass.

#### Side Effects

None

#### Documentation

All *_rng distribution functions have s added to their arguments to indicate these function willingly accept vectors.

#### Copyright and Licensing


src/test/test-models/good/function-signatures/distributions/rngs.stan
```diff
@@ -15,57 +15,97 @@ generated quantities {
...
-  v <- multi_normal_rng(mu,Sigma);
-  // v <- multi_normal_prec_rng(mu,Sigma);
-  v <- multi_normal_cholesky_rng(mu,L);
-  v <- multi_student_t_rng(3.0,mu,Sigma);
-  Sigma <- wishart_rng(3.0,Sigma);
-  Sigma <- inv_wishart_rng(3.0,Sigma);
-  Sigma <- lkj_corr_rng(3,2.5);
-  L <- lkj_corr_cholesky_rng(3,3.0);
-  // DEPRECATE:  Sigma <- lkj_cov_rng(Sigma, mu, alpha, 2.0);
+  v = normal_rng(mu, 2.0);
```
bob-carpenter: If the vectorization allows arbitrary vector types, we really want to test for

* int
* real
* real[]
* int[] ? probably not accepted
* vector
* row_vector

in every position.  But that's 36 cases, just for the normal, and it'd be 216 for student-t.  I'm not sure what the right thing to do here is.  I'm going to make completion optional here.

src/test/test-models/good/function-signatures/distributions/rngs.stan
```diff
@@ -15,57 +15,97 @@ generated quantities {
...
+  v = lognormal_rng(mu, 3.7);
+  v = scaled_inv_chi_square_rng(mu, 3.0);
+  v = gamma_rng(2.4, theta);
+  v = inv_gamma_rng(alpha, 3.4);
+  v = weibull_rng(alpha, 5.3);
+  v = frechet_rng(mu, 8.2);
+  v = pareto_rng(alpha, 3.9);
+  v = pareto_type_2_rng(mu, alpha, 8.1);
+  v = beta_rng(theta, 0.4);
+
```
bob-carpenter: This is missing the scalar / vector case, e.g., `normal_rng(1.0, mu)`.  This form has to be added, even if all the variants allowed for `mu` are not inumerated.

src/docs/stan-reference/distributions.tex
```diff
@@ -1117,7 +1117,7 @@ \subsubsection{Stan Functions}
 \end{description}
 %
 \begin{description}
-\fitem{real}{normal\_rng}{real \farg{mu},  real \farg{sigma}}{Generate a normal
+\fitem{real}{normal\_rng}{reals \farg{mu},  reals \farg{sigma}}{Generate a normal
```
bob-carpenter: The return type and behavior under vectorization needs to be specified.  It needs to answer at least these questions:

* What is return type if all arguments are `real`?
* Can one argument be `real` and the other `vector`?  If so, what's the return type?
* Same question for `vector` and `row_vector` and for `vector` and `real[]` (array of real)?

The answers should be written up in a section on vectorized RNGs.  It can go with the rest of the probability function vectorization descriptions.  But it should be in its own section.  Then these definitions should provide a short link, as in "see \refsec{vec-sec-name} for argument and return type details."

src/stan/lang/function_signatures.h
```diff
@@ -123,24 +123,43 @@ for (size_t i = 0; i < vector_types.size(); ++i) {
...
   }
 }
-add_binary("beta_rng");
+for (size_t i = 0; i < primitive_types.size(); ++i) {
+  for (size_t j = 0; j < primitive_types.size(); ++j) {
+    add("beta_rng", expr_type(double_type()), primitive_types[i], primitive_types[j]);
+  }
+ }
+for (size_t i = 0; i < primitive_types.size(); i++) {
+  add("beta_rng", expr_type(vector_type()), primitive_types[i], expr_type(int_type(), 1U));
```
bob-carpenter: Are `int[]` not allowed with other vector arguments?

src/stan/lang/function_signatures.h
```diff
@@ -123,24 +123,43 @@ for (size_t i = 0; i < vector_types.size(); ++i) {
...
+  add("beta_rng", expr_type(vector_type()), expr_type(int_type(), 1U), primitive_types[i]);
+ }
+for (size_t i = 1; i < vector_types.size(); ++i) {
+  for (size_t j = 0; j < primitive_types.size(); ++j) {
+    add("beta_rng", expr_type(vector_type()), primitive_types[j], vector_types[i]);
+    add("beta_rng", expr_type(vector_type()), vector_types[i], primitive_types[j]);
+  }
+  add("beta_rng", expr_type(vector_type()), vector_types[i], expr_type(int_type(), 1U));
+  add("beta_rng", expr_type(vector_type()), expr_type(int_type(), 1U), vector_types[i]);
```
bob-carpenter: Must fix:  The vector types include all the primitive types.  So this will add a two integer argument version with a vector return type.  The loops should cover only the required cases.  

There are a few ways to do this.  One would be to define a vector-only container of types, which is what I think you were thinking that `vector_types` was.

Another would be to define a vectorized return type function that would let you just write this all as a big double loop that would test, as in:

```
for (const auto& t : vector_types)
  for (const auto& u : vector_types)
    add("beta_rng", rng_return_type<double>(t, u), t, u);
```

where that `rng_return_type` would return `double` (its template arg so the same thing will work for vectorized `int` forms) if both the arguments were primitives.  I would prefer this latter form as it's much less code everywhere.  Something like:

```
expr_type
rng_return_type(const expr_type& t, const expr_type& u) {
  return is_primitive(t) && is_primitive(u)
    ? expr_type(double_type()) : expr_type(vector_type());
}
```
The `expr_type()` constructors may not be necessary if they're defined as implicit.  This assumes there's somehting like an `is_primitive(t)` test on the types.  If not, that'll have to be written, too.

src/test/test-models/good/function-signatures/distributions/rngs.stan
roualdes: Correct.  We probably should test all cases.  I'm leaning towards laying out all tests, if you won't mind the file bloat.  For now, I'm going to consider your other comments first.

src/stan/lang/function_signatures.h
roualdes: Please correct, if wrong.  My going assumption was that `expr_type(int_type(), 1U)` meant int[].  In which case, lines 140 and 141 have int[] with other vector arguments.

bob-carpenter: Yes, that's right, `expr_type(int_type(), 1U)` gives you int arrays.  The whole type system's a bit more opaque than it should be and not well abstracted for specifying types for these kinds of polymorphic functions.   Sorry it's such a pain to deal with.

src/test/test-models/good/function-signatures/distributions/rngs.stan
bob-carpenter: file bloat for tests is ok

src/stan/lang/function_signatures.h
bob-carpenter: Oh, that is sneaky.  Please don't do that kind of thing.  Code readability has to be priority number one for maintainability.  Better to make your own collection, because the vector types collection might change ordering.

roualdes:
This already looks much cleaner.  Though, I don't quite know where the new function rng_return_type should go.  For now, I put it in src/stan/lang/ast/sigs/function_signatures_def.hpp. 

Tests and docs have been updated.

src/stan/lang/ast/sigs/function_signatures_def.hpp
```diff
@@ -224,6 +224,32 @@ namespace stan {
           expr_type(double_type()));
     }
 
+    template<typename T>
+    expr_type function_signatures::rng_return_type(const expr_type& t) {
+      T return_type;
+      return t.is_primitive() ?
```
bob-carpenter: You can move the test down and just use its value (which will be 0 or 1).

```
return exp_type(return_type, !t.is_primitive());
```

If you think it's clearer, it could also be `t.is_primitive() ? 0 : 1`.

bob-carpenter:
It's OK there for now.  We'll probably eventually want to move it to one of the meta directories, but there's going to be a big re-org going forward that should simplify that when it happens.

bob-carpenter:
Hmm.  Now I can see my own review.  No idea what's going on, as it only popped up after I left the last comment.  I'm pretty sure it wasn't there before.

bob-carpenter:
The location for the return type function is fine for now.  We like to try to break functions out into their own files, but it's no big deal here.

src/docs/stan-reference/distributions.tex
```diff
@@ -258,7 +260,31 @@ \subsection{Evaluating Vectorized Functions}
...
+has two or more arguments.  For instance, the following is allowed in a
+generated quantities block.
+
+\begin{stancode}
+    real x[3];
+    vector[3] mu;
+    x = normal_rng(mu, 3);
+\end{stancode}
 
+The result of a vectorized PRNG function depends on the dimensions of
```
bob-carpenter: Perfect---thanks for adding this.

src/docs/stan-reference/distributions.tex
```diff
@@ -258,7 +260,31 @@ \subsection{Evaluating Vectorized Functions}
   y[n] ~ normal(mu[n], sigma);
 \end{stancode}
 
+\subsection{Evaluating Vectorized PRNG Functions}\label{vectorization-prng.subsection}
+
+Some PRNG functions accept vector arguments.  Such functions are
+indicated by the plural pseudo-types \code{reals} or \code{ints}.  As
+in the vectorized arguments subsections of
+Section~\ref{prob-vectorization.section}, the pseudo type \code{reals}
```
bob-carpenter: You don't need to use it here, but there's a macro that lets you do `\refsec{prob-vectorization}`.  It inserts the `.section` markup and the `section~` bit.

src/docs/stan-reference/distributions.tex
```diff
@@ -1165,8 +1191,8 @@ \subsubsection{Stan Functions}
 \end{description}
 %
 \begin{description}
-\fitem{real}{exp\_mod\_normal\_rng}{real \farg{mu}, real \farg{sigma},
-real \farg{lambda}}{Generate a exponentially modified normal variate
+\fitem{real}{exp\_mod\_normal\_rng}{reals \farg{mu}, reals \farg{sigma},
```
bob-carpenter: This is the thing where the return type needs to be something more generic.  It's not going to be `real` unless both arguments are `real`.  So I was suggesting something like a generic `R` and a link to the section describing RNG vectorization in the doc.

src/stan/lang/ast/sigs/function_signatures.hpp
```diff
@@ -257,6 +257,41 @@ namespace stan {
        */
       void add_quaternary(const::std::string& name);
 
+      /**
+       * Determine the return type of distributions' RNG function
+       * based on the primitive-ness of the arguments; if argument
+       * has dimension greater than 1 should return a vector_type.
+       *
+       * @param t type of first argument
+       */
```
bob-carpenter: There should be a @return doc, too.  It winds up looking a bit redundant with the first line for simple functions.  

"primitiveness" doesn't have a hyphen. 

It's size, not dimension that has to be greater than 0.  But I'd just say "if both arguments are scalar, the return type is int or real depending on the distribtuion.  Otherwise, the return type is `int[]` for discrete distributions and `real[]` for continuous ones.  These aren't vector types in the Stan sense, but 1D arrays.

 If the argument has dimension greater than 0, it should return a vector type.  But I'd say that differently---if the argument is non-scalar, it will return `vector`.

src/stan/lang/ast/sigs/function_signatures_def.hpp
```diff
@@ -224,6 +224,32 @@ namespace stan {
...
+        ? expr_type(return_type) : expr_type(return_type, 1U);
+    }
+
+    template<typename T>
+    expr_type function_signatures::rng_return_type(
+                            const expr_type& t,
+                            const expr_type& u,
+                            const expr_type& v) {
+      T return_type;
+      return t.is_primitive() && (u.is_primitive() && v.is_primitive())
```
bob-carpenter: This would be clearer if it's chained to just be

```
return rng_return_type(rng_return_type(t, u), v);
```

Though not relevant if you update, you don't need parens around `a && b && c`.  By default, it'll be left associative and resolve as `(a && b) && c`, which is fine.

src/stan/lang/function_signatures.h
```diff
@@ -123,24 +131,28 @@ for (size_t i = 0; i < vector_types.size(); ++i) {
     }
   }
 }
-add_binary("beta_rng");
+for (const auto& t : all_vector_types) {
```
bob-carpenter: Thanks --- this all looks much cleaner!

roualdes:
Thank you for your time on the comments and for the feedback on my coding.  Just pushed the latest changes.  Let me know what your next thoughts are.

roualdes:
@bob-carpenter, it looks like some checks were not successful because they timed out.  I'm happy to do something about this, just let me know what.  Thanks.

bob-carpenter:
We need to restart the timed-out jobs.  Travis is getting very finicky these days.

roualdes:
Dang, timed-out again.  Would you mind restarting these, when you get a chance?  Thanks.

bob-carpenter:
If you want to go ahead and make the doc fixes, I might not get to it for a couple days.  It's the same comment as I had last time---the return types aren't clear in the doc.  Also, you want to use compound declare-define whenever possible.

bob-carpenter:
I'll update the doc now and resubmit.  Otherwise, this looks great.

bbbales2:
Thanks for doing this @roualdes! (and @bob-carpenter ofc., but that's implied)

bob-carpenter:
I was going to do this now, but now I'm confused about how to modify someone else's branch.  Sean showed me how to do it once and I didn't take good enough notes.  

What needs to happens is for those signatures `reals foo(reals, reals)`, the return type is more subtle.  It's `real` if both arguments are `real` and `vector` (?) or something else otherwise.  Once that change goes in, I'll merge.  I was going to figure out what the return type was by looking at the signatures, but I'm stuck on the Git process.  I'm sure I'll get that sorted out if you'd rather hang on and let me finish this doc.  I was suggesting pointing in each function's definition to a section that explains how the return types work for vectorized RNGs.

bob-carpenter:
OK, I think I'm in business.  Mitzi walked me through the instructions in the [Git Book](https://git-scm.com/book/en/v2/Git-Basics-Working-with-Remotes),  which are the same as those [provided by Sean](http://discourse.mc-stan.org/t/dealing-with-prs-from-forks/3156). I added them our [GitHub process page](https://github.com/stan-dev/stan/wiki/Dev:-Git-Process).

bob-carpenter:
Turns out there were only a couple missing cases.  This looks great.  When it passes, it can be merged.

Issue #2462, opened 2018-01-26
effective_sample_size() to POD types for RStan/PyStan usage


bob-carpenter:
#### Summary:

The current interface is

```
double effective_sample_size(const Eigen::Matrix<Eigen::VectorXd,  Dynamic, 1>& samples) const;
```

which is suboptimal for interfaces because of its use of Eigen.  I don't think it's good for Stan itself because of its unconventional use of Eigen matrices.

How about something more standard like

```
double effective_sample_size(const std::vector<std::vector<double>>& samples) const;
```

Lower level, we'd get even better memory locality with

```
double effective_sample_size(const double* samples, int num_chains, int num_draws) const;
```

For the interfaces, is it better if we actually take in a 3D input and do this for each parameter?

```
double effective_sample_size(const double* samples, int num_params, 
            int num_chains, int num_draws) const;
```
#### Description:

This'll make it easier for RStan and PyStan to use the effective sample size calculations based on Stan.

To maintain backward compatibility, we just need a new function

@bgoodri and @ariddell --- would this be enough to let you use it in RStan?  

#### Current Version:
v2.17.1

bgoodri:
For R, any of these would be fine.

ariddell:
Ditto for Python.

Thanks for getting to this! It will be a relief to have this long-standing todo checked off.

roualdes:
@bob-carpenter, just submit a pull request for this.  It might not be 100% what you're looking for, but it's a good start as it has the right data structures and accurate calculations.  I'll let you continue discussion here or in the pull request.  Thanks.

bob-carpenter:
It may be a week or two before I can knock this off, so if someone else wants to do it, it shouldn't be difficult.  We usually try to open-ended discussions in Discourse, big functional specs on Wikis, and technical specs as part of the issue refinement and PR reviews.

roualdes:
With https://github.com/stan-dev/stan/pull/2575 now merged, should we close this?

PR #2477, opened 2018-02-27
Feature/issue 2462 eff sample size pod types

roualdes:
#### Submission Checklist

- [ x] Run cpplint: `make cpplint`
- [ x] Declare copyright holder and open-source license: see below

#### Summary

Add plain old data structure interface to effective_sample_size.

#### Intended Effect

New interface for Rstan and Pystan.

#### How to Verify

New public method effective_sample_size_pod() is added so as to allow unit tests.  Run

`./runTests.py src/test/unit/mcmc/chains_test.cpp`

#### Side Effects

None intended.

#### Documentation

None, but happy to add.  Don't know where.

#### Copyright and Licensing


bob-carpenter:
Thanks!  I'll review.

Forget that last response, I thought you were asking me to just create the PR.

src/stan/mcmc/chains.hpp
```diff
@@ -292,6 +291,91 @@ namespace stan {
...
+       * reference manual section "Effective Sample Size".
+       *
+       * Current implementation takes the minimum number of samples
+       * across chains as the number of samples per chain.
+       *
+       * @param std::vector<std::vector<double> > samples
+       * @return
+       */
+      double
+      effective_sample_size(const std::vector<std::vector<double> >& samples)
```
bob-carpenter: We have to make sure with @bgoodri and @ariddell that this will work for both RStan and PyStan.  I was proposing a POD solution with just this signature:

```
double effective_sample_size(const double* draws, int n_chains, int n_draws_per_chain);  
```

I don't think we can use 2D arrays because we don't know sizes statically.

where the chains were stored one after the other.

src/stan/mcmc/chains.hpp
```diff
@@ -292,6 +291,91 @@ namespace stan {
...
+       * across all kept samples.
+       *
+       * The implementation is close to the effective sample size
+       * description in BDA3 (p. 286-287).  See more details in Stan
+       * reference manual section "Effective Sample Size".
+       *
+       * Current implementation takes the minimum number of samples
+       * across chains as the number of samples per chain.
+       *
+       * @param std::vector<std::vector<double> > samples
```
bob-carpenter: The parameters and return need textual descriptions.  They don't need to have the type duplicated---that's already in the signature.  

We try to reserve the "sample" for a set of draws and not use "sample" for a single draw.  So this should either be singular "sample" or plural "draws".  Or it could be `chains`.

src/stan/mcmc/chains.hpp
```diff
@@ -292,6 +291,91 @@ namespace stan {
         return ess;
       }
 
+      /**
+       * Returns the effective sample size for the specified parameter
+       * across all kept samples.
+       *
+       * The implementation is close to the effective sample size
+       * description in BDA3 (p. 286-287).  See more details in Stan
```
bob-carpenter: "BDA3" isn't an adequate citation.  Not everyone's familiar with Andrew's cutesy abbreviations for things.  I would just point to the Stan manual here.

src/stan/mcmc/chains.hpp
```diff
@@ -292,6 +291,91 @@ namespace stan {
...
+       * reference manual section "Effective Sample Size".
+       *
+       * Current implementation takes the minimum number of samples
```
bob-carpenter: There's no minimums taken here.  It all just uses the cross-chain mean and variance estimates to adjust within chain estimates.  But I wouldn't even try to say that, I'd just point to the manual.  You can include a URL to the online version of the manual:

http://mc-stan.org/users/documentation

src/stan/mcmc/chains.hpp
```diff
@@ -292,6 +291,91 @@ namespace stan {
...
+        // need to generalize to each jagged samples per chain
+        int n_samples = samples[0].size();
+        for (int chain = 1; chain < chains; chain++) {
+          n_samples = std::min(n_samples,
+                               static_cast<int>(samples[chain].size()));
+        }
+
+        std::vector<std::vector<double> > acov(chains);
+        for (int chain = 0; chain < chains; chain++) {
+          stan::math::autocovariance(samples[chain], acov[chain]);
```
bob-carpenter: This probably got copied from elsewhere, but we don't need `stan::` if we're in the `stan` namespace, so this can just be `math::autocovariance`.

src/stan/mcmc/chains.hpp
```diff
@@ -292,6 +291,91 @@ namespace stan {
...
+        }
+
+        std::vector<double> chain_mean(chains);
+        std::vector<double> chain_var(chains);
+        for (int chain = 0; chain < chains; chain++) {
```
bob-carpenter: We prefer `++chain` to `chain++` when the pre-increment value isn't needed.  There are a bunch of these.

src/stan/mcmc/chains.hpp
```diff
@@ -292,6 +291,91 @@ namespace stan {
...
+          chain_mean[chain] = stan::math::mean(samples[chain]);
+          chain_var[chain] = acov[chain][0]*n_kept_samples/(n_kept_samples-1);
+        }
+
+        double mean_var = stan::math::mean(chain_var);
+        double var_plus = mean_var*(n_samples-1)/n_samples;
+        if (chains > 1)
+          var_plus += stan::math::variance(chain_mean);
+        std::vector<double> rho_hat_t(n_samples);
+        std::fill(rho_hat_t.begin(), rho_hat_t.end(), 0);
```
bob-carpenter: You can just use the initializing constructor syntax here, which will also be more efficient.  So that'd just be:

```
std::vector<double> rho_hat_t(n_samples, 0);
```

src/stan/mcmc/chains.hpp
```diff
@@ -292,6 +291,91 @@ namespace stan {
...
+        for (int t = 3; t <= max_t - 2; t += 2) {
+          if (rho_hat_t[t + 1] + rho_hat_t[t + 2] >
+              rho_hat_t[t - 1] + rho_hat_t[t]) {
+            rho_hat_t[t + 1] = (rho_hat_t[t - 1] + rho_hat_t[t]) / 2;
+            rho_hat_t[t + 2] = rho_hat_t[t + 1];
+          }
+        }
+        double sum_rho_hat_t = std::accumulate(rho_hat_t.begin(),
+                                               rho_hat_t.end(),
+                                               0.);
```
bob-carpenter: [Optional] If you're going to have a `0.`, I'd suggest `0.0` is more readable.

src/stan/mcmc/chains.hpp
```diff
@@ -292,6 +291,91 @@ namespace stan {
...
+        double ess = chains * n_samples;
```
bob-carpenter: I don't like this pattern of assigning the wrong value, then correcting it.  I think this would be clearer as just:

```
return chains * n_samples / (1 + 2 * sum_rho_hat_t);
```

src/stan/mcmc/chains.hpp
```diff
@@ -292,6 +291,91 @@ namespace stan {
...
+        }
+        // Geyer's initial monotone sequence
+        for (int t = 3; t <= max_t - 2; t += 2) {
+          if (rho_hat_t[t + 1] + rho_hat_t[t + 2] >
+              rho_hat_t[t - 1] + rho_hat_t[t]) {
+            rho_hat_t[t + 1] = (rho_hat_t[t - 1] + rho_hat_t[t]) / 2;
+            rho_hat_t[t + 2] = rho_hat_t[t + 1];
+          }
+        }
+        double sum_rho_hat_t = std::accumulate(rho_hat_t.begin(),
```
bob-carpenter: Using `_t` for a non-typedef will be confusing as that's the convention we use for typedefs.  You don't need to put the argument on this, so I'd suggest just changing all `rho_hat_t` to `rho_hat`, including `sum_rho_hat_t`.

src/stan/mcmc/chains.hpp
```diff
@@ -701,6 +785,18 @@ namespace stan {
         return effective_sample_size(index(name));
       }
 
+      double effective_sample_size_pod(const int index) const {
+        std::vector<std::vector<double> > samples(num_chains());
+        for (int chain = 0; chain < num_chains(); chain++) {
+          Eigen::VectorXd Esample = this->samples(chain, index);
+          samples[chain].resize(Esample.size());
+          for (int n = 0; n < Esample.size(); n++) {
```
bob-carpenter: This can be done with an iterator copy.

```
std::copy(&Esample[0], &Esample[0] + Esample.size(), samples[chain].begin());
```

But one of the reasons I was suggesting using POD (plain old data) types is that then you don't need this copy, you can just send in the pointers if you use `std::vector<double*>` as the type.  I'm not quite sure what the best thing to do here is.  Given everything else going on, the copies shouldn't be too expensive.
Eventually, we'd want to chase all of this through and change the underlying types of member variables so we don't have to copy.  Is the original result stored in this calss in a `VectorXd`?

src/stan/mcmc/chains.hpp
```diff
@@ -701,6 +785,18 @@ namespace stan {
         return effective_sample_size(index(name));
       }
 
+      double effective_sample_size_pod(const int index) const {
+        std::vector<std::vector<double> > samples(num_chains());
+        for (int chain = 0; chain < num_chains(); chain++) {
+          Eigen::VectorXd Esample = this->samples(chain, index);
```
bob-carpenter: Local variables shouldn't have upper-case names.  We reserve that for types in our own code (not that we've followed this everywhere.

src/stan/mcmc/chains.hpp
roualdes: Right, a POD solution.  I backed away from that after thinking that the [original issue](https://github.com/stan-dev/stan/issues/2462) had some agreement that the std::vector signature would suffice and finding so many stan::math functions based on std::vectors.  But, you're right, confirmation would be good.  I'll wait until we hear from @bgoodri and/or @ariddell.  Happy to rewrite as you all prefer.

ariddell:
We probably want to do some superficial performance tests here. For
example, we might want to make sure that this is not 5x slower than the
current RStan and PyStan implementations due to memory copying. As an
example, we might consider eight schools with 20,000 draws and 4 chains.

(I remember trying to write this exact function several years ago and
recall this being an issue.)

bob-carpenter:
Sure, if you want to clean it up as a whole, that'd be fantastic. But don't feel it's your responsibility to clean any of the things up that you didn't add.

bob-carpenter:
@ariddell I want to write this without having to do any copying.  How about a signature like this:
```
double effective_sample_size(const std::vector<const double*>& chains);
```
That will require at most a single pointer copy per chain from whatever the underlying data structure is.  For both `Eigen::Matrix` vectors and row vectors and std::vector` vector types, we can use `&x[0]` to produce that `const double*` pointer.  Or we could write a generic metaprogram.

ariddell:
I like that signature. That looks like it might work well for everyone.

roualdes:
This might be getting above my pay grade, but here it goes anyway.  The problem I see with 

`double effective_sample_size(const std::vector<const double*>& chains);`

or anything that doesn't have a single chain's samples stored in a std::vector is the function `stan::math::autocovariance` which takes as its first argument a std::vector.  Hence, for each chain, we'd have to copy from double* to a std::vector.  As far as I can tell, this is exactly what we're trying to avoid.

Instead, what do you say to the following signature?

`double effective_sample_size(const std::vector<std::vector<double>* >& chains);`

bob-carpenter:
My proposal is to fix the external services interfaces so they're as easy to use as possible from the interfaces.  So this should really be up to @ariddell and @bgoodri.

Then when we settle on the front-end types, we fix the back-end implementations so they don't involve any copying.  This is all just a matter of finagling container types.  In fact, if we can support C-style pointer arrays, standard vectors, and Eigen vectors using a template and `double operator[](int i)`.  It's easy to map a contiguous block of `double` values to an Eigen type using `Eigen::Map`---it doesn't copy.

ariddell:
I agree with the spirit of Bob's comment. It would be very nice if this
particular PR could (superficially) check and see if the code can be
called successfully via RStan or PyStan (doesn't matter which).

bob-carpenter:
I wasn't trying to be that ambitious.  I just wanted to query @ariddell and @bgoodri to make sure the interface would be usable by RStan and PyStan.  Perhaps the best thing to do will be to have a reference to container type.

```
template <typename C>
real effective_sample_size(const std::vector<const C&>& chains) const;
```
where `C` implements 
```
double operator[](size_t i) const;
size_t size() const;
```
That requirement to implement `size()` removes the ability to use plain old data types.  If we need to support that, then we need a second argument for the size, too.  We could even go totally generic and use something like:

```
template <typename C>
double effective_sample_size(const C& sample, size_t chains, size_t iterations) const;
```
where we only need `C` to be such that `sample[i][m]` returns the `double` value of draw `m` from chain `i`.  


None of these are at all hard to implement---just need to decide on one.




Presumably any of the code can be called.  I just want to make it convenient for both RStan and PyStan and CmdStan to call without having to copy the entire posterior sample multiple times to do it.  CmdStan can use anything.  We can also rewrite the functions to be very generic, as in:

ariddell:
I'm +1 for the non-template signature you proposed. We can always switch
to the more flexible signature later.

Sticking with simple types when possible seems like a good idea in the
interfaces because the code needs to be easily understood by those who
are not C++ experts (i.e., interface writers).

bob-carpenter:
> On Mar 2, 2018, at 6:59 PM, Allen Riddell <notifications@github.com> wrote:
> 
> I'm +1 for the non-template signature you proposed. We can always switch
> to the more flexible signature later.
> 
> Sticking with simple types when possible seems like a good idea in the
> interfaces because the code needs to be easily understood by those who
> are not C++ experts (i.e., interface writers).

Which non-template signature?  You mean vector of vector of double or
vector of array?  Neither of those are plain old data types because
of std::vector.  The plain old data type version would be double** for
flexible member placement or double* for memory locality.

We only have three interfaces that use the C++.  I don't think we have
to design much beyond that.  CmdStan will be easy to adapt to any decision
as there are no external type-interface constraints, so it's really just
down to what is going to be easiest for PyStan and RStan.  Ben seems to
be OK with any of these, but I would like to minimize copying as much as
possible.

ariddell:
Either one is fine. I prefer the first one:

- ``double effective_sample_size(const std::vector<const double*>& chains);``
- ``double effective_sample_size(const double**& chains);``

bob-carpenter:
Thanks, @ariddell---that was the answer I was looking for.

bob-carpenter:
@roualdes:

Could you rewrite this with this signature:

```
double effective_sample_size(const std::vector<const double*>& chains);
```

No pressure, of course---we can get it done one way or another.

roualdes:
I'm happy to give it a try.  After some research over this weekend, I've got two thoughts going.  If you get a minute, any feedback is much appreciated.

1) We need to accept the number of draws as an argument as well, hence

`double effective_sample_size(const std::vector<const double*>& chains, int num_draws);`

2) We need to overload `stan::math::autocorrelation` and `stan::math::autocovariance` to play nicely with `Eigen::Map(&chains[0], num_draws)`.  This should is possible since `Eigen::FFT` is overloaded for both `std::vector` and `Eigen::VectorXd` -- your LingPipe blog post on autocorrelations calculated with FFTs helped me find this.  

As a side benefit, this should allow the current working `effective_sample_size` to avoid its expensive copying from `Eigen::VectorXd` to `std::vector` and back.

If I get 2) working, is there a protocol to create a pull request under stan-dev/math that references this issue?

ariddell:
Does C++11 not allow for avoiding a copy? Can we use a move constructor?
Here's something that might be related:
https://stackoverflow.com/questions/10402241/c11-move-constructor

bob-carpenter:
> On Mar 5, 2018, at 5:17 AM, Allen Riddell <notifications@github.com> wrote:
> 
> Does C++11 not allow for avoiding a copy?

Not implicitly.

> Can we use a move constructor?

No, because we're dealing with something that's not a class.  Move semantics is for moving the responsibility for memory and is managed among a closely related set of types.  There's no way to do a move from an Eigen type to a std::vector or vice versa, given the encapsulation of memory under the RAII pattern.  You can see this by looking at the move constructor for std::vector:

http://www.cplusplus.com/reference/vector/vector/vector/

Now if all these types were under our control we could do something, but they're not.

ariddell:
This makes sense. Thanks for the explanation.

I think a small performance hit is worth paying if it means eliminating
the code duplication (triplication) we have. Hopefully it will not be
too much slower with < 100 params and <10,000 draws per chain.

bob-carpenter:
I agree that removing code duplication is the number one priority---it'll improve reliability, make the code more maintainable, and make it more consistent across interfaces.  In this case, we won't even need to sacrifice performance.

roualdes:
I just pushed my latest efforts, alongside a new [stan-dev/math pull request](https://github.com/stan-dev/math/pull/790).  Some notes:

1) std::vector does not appear to get copied within Eigen's FFT, however I still chose to go with Eigen data structures.

2) The signature your after works, but I use many Eigen data structures behind it.  Hopefully this is OK, as I believe they are faster than hand written for loops around std::vector.

3) Eigen::Map does show up, but I don't think any copying is done, what with MatrixBase templates and no assignments to Eigen::VectorXd.  You'll find these details on the stan-dev/math side.  Please correct me if you disagree about the copying.

4) In the previous effective_sample_size function there is a comment about unequal numbers of draws across the chains.  The new implementation assumes equal number of draws amongst the chains.

5) I at least tried to correct all the suggestions you had in all previous commits.

bob-carpenter:
For (2), Eigen data structures aren't any faster than std::vector in and of themselves (though Eigen has the advantage of not constructing initial membrs, which means it's badly behaved in that it can have uninitialized memory, but it's more efficient).  Compound operations like matrix multiplication are much more efficiently written in Eigen than naive loops.

For (3), Eigen Map itself only calls the constructor and copies a pointer.  Whenever a Map is assigned to an actual matrix type, a copy will happen.  This is true in general with expression templates.  

For (4), we never did settle on a way to do ragged chains, so assuming everything's the same size is fine for now.

For (5), thanks!

bob-carpenter:
See: http://discourse.mc-stan.org/t/analysis-api/3486/5

for more discussion.

roualdes:
It's worth mentioning, this won't pass its tests until it can use the code from https://github.com/stan-dev/math/pull/790.  However, I respect betanalpha's idea (linked above) that, in a broader context, this sort of code should be less piecemeal.  So I'm going to patiently wait on this until directed otherwise.  Until then, on to some other "good first issue"s. Thanks.

roualdes:
Replaced by https://github.com/stan-dev/stan/pull/2530

PR #2481, opened 2018-03-07
[Jenkins] Integration tests are always essentially make-only.

seantalts:
#### Copyright and Licensing

Columbia University

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -34,8 +34,10 @@ def mailBuildResults(String label, additionalEmails='') {
     )
 }
 
-def runTests(String testPath) {
-    sh "./runTests.py -j${env.PARALLEL} ${testPath} --make-only"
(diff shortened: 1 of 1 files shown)


syclik:
What does the make only flag do?

And yes, these will fail to compile when something is wrong. The output of google test just confirms that it has indeed compiled correctly.

seantalts:
I will give you 3 guesses 😂

syclik:
My hope: it runs make without actually running the test executable. 

My fear: it just prints the output of make without running make. 

My third guess: it turns on a makerbot in the office.

seantalts:
Hahahaha. Those were great. The hope was correct! But I realized that the
integration tests are both .PHONY targets that mostly just make sure the
models can be compiled, so they don't need to have a separate `--make-only`
step (and the one I had added was causing things to take twice as long in
this stage because the target was .PHONY and always runs).

PR #2491, opened 2018-03-13
Feature/2440 map rect

bob-carpenter:
#### Submission Checklist

#### Summary

Add `map_rect` function to language with manual doc.  

#### Intended Effect

Clear way for MPI in addition to implementing the function.

#### How to Verify

Unit tests.  Or pull out CmdStan and run an example.

Without map:

```
data {
  int y[12];
  real x[12];
}
parameters {
  vector[2] beta;
}
model {
  beta ~ normal(0, 1);
  y ~ bernoulli_logit(beta[1] + beta[2] * to_vector(x));
}
```

With map:

```
functions {
  // logistic regression lpmf
  vector lr(vector beta, vector theta, real[] x, int[] y) {
    real lp = bernoulli_logit_lpmf(y | beta[1] + to_vector(x) * beta[2]);
    return [lp]';
  }
}
data {
  // N = 12 data points
  int y[12];
  real x[12];
}
transformed data {
  // K = 3 shards
  int ys[3, 4] = { y[1:4], y[5:8], y[9:12] };
  real xs[3, 4] = { x[1:4], x[5:8], x[9:12] };
  vector[0] theta[3];
}
parameters {
  vector[2] beta;
}
model {
  beta ~ normal(0, 1);
  target += sum(map_rect(lr, beta, theta, xs, ys));
}
```

And here's some data:

```
x <- 
c(-0.0950975784183072, 0.439459508843735, 0.456727631025774, -0.682256473834409, -0.136250176411917,
-0.249300158948066, 0.23367488898078, 0.838578027991903, 1.03590847440499, -1.3742813878688, -0.160658830059433,
-0.588358254227975)
y <- 
c(1, 0, 0, 1, 1, 0, 0, 0, 0, 1, 0, 0)
```

#### Side Effects

No.

#### Documentation

Yes, lots in the manual.

#### Copyright and Licensing

Columbia University


wds15:
I am happy to check that the code generates everything needed to make map_rect work in its serial and mpi version, but I am hesitant to review more than that as I am really not an expert in Stan and its DSL facilities.

bob-carpenter:
Thanks.  The function in the math library is clean enough that we should only need someone reviewing this who knows the parser.

wds15:
So this looks great, but I am missing one detail which is a must for the MPI version. In order to make the MPI communication work we have to tell the boost serialization library a few things. This is most conveniently done using macros. What is currrently missing for the above logistic regression is this macro call which needs to be placed in the top level namespace:

```
STAN_REGISTER_MAP_RECT(63, logistic_model_namespace::lr_functor__)
```

The definition of this macro is empty whenever we compile in serial mode, but the macro will be redefined whenever we are in MPI mode. [See here](https://github.com/stan-dev/math/blob/feature/concept-mpi-2/stan/math/prim/mat/functor/map_rect_mpi.hpp#L44).

An example which works [is here](https://github.com/stan-dev/cmdstan/blob/feature/proto-mpi/examples/mpi/mpi_defs3.hpp#L31).

bob-carpenter:
Sure, I can generate code like that once I understand it.

* is anywhere after the functor definitions OK?

* is there any signfiicance to the identifier, such as matching the call?  If the same functor is mapped over different data, what do I generate?

* is there overhead to including this for all functors, even if they don't use MPI?

wds15:
Let me number your questions:

1. Anywhere after the functor def is fine. It only has to be in the root namespace.

2. You generate such a call for every call_id/functor pair for which map_rect is called. Examples for a program with functor foo:
   - multiple calls to `map_rect<123, foo>` with the same data => only a single macro call
   - call to foo with different data through `map_rect`; so `map_rect<123, foo>` & `map_rect<456, foo>`  => multiple calls to the macro, one per call_id+functor

3. I think there is a tiny bit of overhead with this, yes. These calls trigger that the types are registered in some boost serialization global thing. So these calls feed into a lookup table which boost serialization generates in some singleton. However, I think that you can neglect the overhead here. I would not be worried if you do this for all user-defined functors if this makes your life a lot easier.

bob-carpenter:
Thanks.  Given the global namespace constraint and the requirement to generate matching macros, I'll need to accumulate the calls in the program and then dump out at the end.  So I'll only generate the ones we need.

bob-carpenter:
@wds15 --- I added the code to generate the macro declarations.  Would you mind checking the target looks right for MPI?

wds15:
Cool! Looks great. I just compiled the logistic example with and without MPI. It is getting really simple to use MPI now. The Stan language works and the build stuff is also taken care of.

mitzimorris:
are the unit tests actually passing here?
I merged this into develop because I assumed the unit tests had passed.
in fact, running unit tests fail - 

command:  `./runTests.py src/test/unit`
fails with message:

```
make: *** No rule to make target `src/test/unit/version'.  Stop.
rm test/unit/multiple_translation_units.o
make -j1 test/unit/multiple_translation_units src/test/unit/version src/test/unit/callbacks/writer src/test/unit/callbacks/logger src/test/unit/callbacks/stream_writer src/test/unit/callbacks/interrupt src/test/unit/callbacks/tee_writer src/test/unit/callbacks/stream_logger src/test/unit/optimization/bfgs src/test/unit/optimization/bfgs_update src/test/unit/optimization/bfgs_minimizer src/test/unit/optimization/lbfgs_update src/test/unit/optimization/bfgs_linesearch src/test/unit/io/writer src/test/unit/io/reader src/test/unit/io/empty_var_context src/test/unit/io/program_reader src/test/unit/io/chained_var_context src/test/unit/io/cmd_line src/test/unit/io/dump src/test/unit/io/random_var_context src/test/unit/io/stan_csv_reader src/test/unit/io/validate_zero_buf src/test/unit/io/read_line src/test/unit/io/util failed
exit now (03/23/18 21:23:32 EDT)
```

mitzimorris:
nevermind?
did `make math-revert` and things started running OK.

src/stan/lang/ast/node/map_rect.hpp
```diff
@@ -0,0 +1,118 @@
...
+   */
+  map_rect(const map_rect& mr);
+
+  /**
+   * Assign the member variables of the specified rectangular map
+   * object to this object's member variables.  This operator and the
+   * copy constructor are defined to avoid incrementing the call
+   * identifier accidentally.
+   *
+   * @paramp[in] mr rectangular map to assign
```
seantalts: Shouldn't these be "@param" or is "@paramp" a thing? I think it caused some new compiler warnings on develop, though not sure why the PR didn't show up with warnings.

mitzimorris:
most definitely should be "@param"
don't know why this passed doxygen before merging.
I fixed this on the tuples-step-1 branch before doing the PR.

bob-carpenter:
Yes, it should be @param

seantalts:
I'll look into how the PR tests didn't catch it...

seantalts:
So it turned out to be just a warning from doxygen that didn't fail the command. The warning did get parsed by the GCC warnings parser, but apparently those warnings don't fail builds. Should I make them fail builds (or mark them unstable or something?)?

syclik:
Yes. Make those warnings fail the build.

PR #2530, opened 2018-05-28
Feature/issue 2462 effective_sample_size interface for rstan/pystan

roualdes:
… old effective_sample_size to new stan-dev/math autocovariance fns

#### Submission Checklist

#### Summary

Add new interface to effective_sample_size for RStan and PyStan.

Replace https://github.com/stan-dev/stan/pull/2477

#### Intended Effect

New interface.

#### How to Verify

make clean && ./runTests.py -j2 src/test/unit/mcmc/chains_test.cpp

#### Side Effects

None intended.

#### Documentation

Inline as function doc.

#### Copyright and Licensing

### DIFF
--- a/src/stan/mcmc/chains.hpp
+++ b/src/stan/mcmc/chains.hpp
@@ -177,14 +177,14 @@ namespace stan {
 
       static Eigen::VectorXd autocorrelation(const Eigen::VectorXd& x) {
         using std::vector;
-        using stan::math::index_type;
+        using math::index_type;
         typedef typename index_type<vector<double> >::type idx_t;
 
         std::vector<double> ac;
         std::vector<double> sample(x.size());
         for (int i = 0; i < x.size(); i++)
           sample[i] = x(i);
-        stan::math::autocorrelation(sample, ac);
+        math::autocorrelation(sample, ac);
 
         Eigen::VectorXd ac2(ac.size());
         for (idx_t i = 0; i < ac.size(); i++)
@@ -194,14 +194,14 @@ namespace stan {
 
       static Eigen::VectorXd autocovariance(const Eigen::VectorXd& x) {
         using std::vector;
-        using stan::math::index_type;
+        using math::index_type;
         typedef typename index_type<vector<double> >::type idx_t;
 
         std::vector<double> ac;
         std::vector<double> sample(x.size());
         for (int i = 0; i < x.size(); i++)
           sample[i] = x(i);
-        stan::math::autocovariance(sample, ac);
+        math::autocovariance(sample, ac);
 
         Eigen::VectorXd ac2(ac.size());
         for (idx_t i = 0; i < ac.size(); i++)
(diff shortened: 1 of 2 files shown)


roualdes:
@bob-carpenter, when you have some time, would you please provide some feedback on the following items?  Thanks.

1) The (old) method `double effective_sample_size(const Eigen::Matrix<Eigen::VectorXd, Dynamic, 1> samples) const` no longer copies the chains twice.  However, the new interface `double effective_sample_size(const std::vector<const double*>& chains, int num_draws) const` makes one copy, which I can't seem to avoid. 

If I redo the calculations within the new interface, instead of reusing the code as in this PR, I can avoid that one copy.  Which would you prefer: a) duplicate code so as to avoid one copy, or b) reuse code and suffer a single copy, for easy maintenance?

2) Surely, I'm missing something, but how exactly is RStan/PyStan supposed to access this new private method `double effective_sample_size(const std::vector<const double*>& chains, int num_draws) const`?

3) The function that enables testing is named `effective_sample_size_new`, but surely there's a better name than this.

bob-carpenter:
There is still a namespace level using.  Every `using` has to be encapsulated in a function.

bob-carpenter:
1.  Does the version currently in develop copy?  If not, the new version should not do any copying, even if it requires code duplication.  [Sorry---couldn't follow what you meant by old and new here.]

2. They should access the new method however they accessed the old method.  It doesn't need to be private if the privacy is the issue.

3.  `effective_sample_size_new` shouldn't be in the API.  What was the goal of implementing the same function with different names?  

Stepping back, if all these functions are under the hood, why are we updating them?  Is this being done at the request of the interface developers?

roualdes:
Good thought, stepping back.  The intent of issue https://github.com/stan-dev/stan/issues/2462 is for RStan and PyStan to be able to outsource their effective sample size calculations.  As it stands both RStan and PyStan call `stan::math::autocovariance` and then do the remaining calculations internally.  

When I first wrote this PR, it seemed logical to overload `effective_sample_size` within the `chains` class since this issue came up in Stan.  This no longer seems like a good idea to me, because I don't see how RStan/PyStan will access a member (public or private) of the `chains` class without instantiating the entire class.  Instantiating the `chains` class works for CmdStan, but forcing this class on RStan or PyStan is too much.

All this is to say that Michael's [analysis API](http://discourse.mc-stan.org/t/analysis-api/3486) idea seems like the right move, inclusive of his not piecemeal suggestion.  Since it's clear that I've only recently understood the broader issue here, I'm going to step back again, see if I can grok more of the I/O concerns contained in the analysis API thread on discourse, and then try to bring forward a solution to the analysis API idea.

If you'd like me to close this PR, in an attempt to reduce clutter, please let me know.

bob-carpenter:
I'll close the PR here.  

I completely agree there should be some kind of posterior analysis API that lives in stan-dev/stan.  

Are there functions that make sense to implement in stan-dev/math?  These could then be used by an analysis API.  It'd also be great to have stable implementations of correlation, covariance, and autocorrelation exposed through the Stan language.

Issue #2549, opened 2018-06-14
Tests crash clang++-6.0


bgoodri:
#### Summary:

One of the tests causes `clang++-6.0` to crash

#### Description:

One of the tests causes `clang++6.0` to crash

#### Reproducible Steps:

```
make src/test/test-models/good/vec-expr/vec_expr_big.hpp-test
```

#### Current Output:

```
find: ‘stan’: No such file or directory
find: ‘stan’: No such file or directory
clang++-6.0 -Wall -I . -isystem lib/stan_math/lib/eigen_3.3.3 -isystem lib/stan_math/lib/boost_1.66.0 -isystem lib/stan_math/lib/cvodes-3.1.0/include -isystem lib/stan_math/lib/idas-2.1.0/include -std=c++1y -DBOOST_RESULT_OF_USE_TR1 -DBOOST_NO_DECLTYPE -DBOOST_DISABLE_ASSERTS -DBOOST_PHOENIX_NO_VARIADIC_EXPRESSION -Wno-unused-function -Wno-uninitialized -I src -isystem lib/stan_math/ -DFUSION_MAX_VECTOR_SIZE=12 -Wno-unused-local-typedefs -stdlib=libc++ -DNO_FPRINTF_OUTPUT -pipe  -c -O0 -include src/test/test-models/good/vec-expr/vec_expr_big.hpp -o /dev/null test/test-model-main.cpp
#0 0x00007f29a615cb6a llvm::sys::PrintStackTrace(llvm::raw_ostream&) (/usr/lib/llvm-6.0/bin/../lib/libLLVM-6.0.so.1+0x8edb6a)
#1 0x00007f29a615aea6 llvm::sys::RunSignalHandlers() (/usr/lib/llvm-6.0/bin/../lib/libLLVM-6.0.so.1+0x8ebea6)
#2 0x00007f29a615afc3 (/usr/lib/llvm-6.0/bin/../lib/libLLVM-6.0.so.1+0x8ebfc3)
#3 0x00007f29a9397f50 __restore_rt (/lib/x86_64-linux-gnu/libpthread.so.0+0x11f50)
[...]

PR #2550, opened 2018-06-14
Feature/timer

### LINKED ISSUE #2070: time-based refreshes
#### Summary:

Stan should output progress every X seconds rather than every X% of iterations

#### Description:

Stan should output progress every X seconds rather than every X% of iterations

https://groups.google.com/forum/#!searchin/stan-dev/%22time-based%22%7Csort:relevance/stan-dev/za8SWSVugTo/TCE8t6peFwAJ

#### Reproducible Steps:

Estimate a model

#### Current Output:

```
SAMPLING FOR MODEL 'lm' NOW (CHAIN 1).

Chain 1, Iteration:    1 / 2000 [  0%]  (Warmup)
Chain 1, Iteration:  200 / 2000 [ 10%]  (Warmup)
Chain 1, Iteration:  400 / 2000 [ 20%]  (Warmup)
Chain 1, Iteration:  600 / 2000 [ 30%]  (Warmup)
Chain 1, Iteration:  800 / 2000 [ 40%]  (Warmup)
Chain 1, Iteration: 1000 / 2000 [ 50%]  (Warmup)
Chain 1, Iteration: 1001 / 2000 [ 50%]  (Sampling)
Chain 1, Iteration: 1200 / 2000 [ 60%]  (Sampling)
Chain 1, Iteration: 1400 / 2000 [ 70%]  (Sampling)
Chain 1, Iteration: 1600 / 2000 [ 80%]  (Sampling)
Chain 1, Iteration: 1800 / 2000 [ 90%]  (Sampling)
Chain 1, Iteration: 2000 / 2000 [100%]  (Sampling)
 Elapsed Time: 0.795202 seconds (Warm-up)
               0.421444 seconds (Sampling)
               1.21665 seconds (Total)
```

#### Expected Output:

```
 Elapsed Time: 0.795202 seconds (Warm-up)
               0.421444 seconds (Sampling)
               1.21665 seconds (Total)
```
 if X > 1.21665

#### Additional Information:

The current refresh behavior is sufficiently stupid and isolated from anything statistical that we should just get rid of it immediately without deprecation messages. I would be in favor of reinterpreting the integer passed to the `refresh` argument as a number of seconds. That way any existing code that specifies `refresh` would continue to work, although we probably need a higher default than 10 seconds. Alternatively, we could introduce a `refresh_seconds` argument and make people specify that explicitly to get the new behavior but then they won't get the new behavior unless they RTFM. Or we could introduce a `refresh_seconds`  and just output a nonfatal message if the user specifies `refresh` saying that `refresh` is ignored.

#### Current Version:
v2.12.0

bgoodri:
#### Submission Checklist

- [?] Run unit tests: `./runTests.py src/test/unit`

#### Summary

Use time difference rather than iteration difference to determine whether to send "Iteration: ..." to the logger.

Also, check whether `refresh > 0` before printing "Elapsed Time: ..."

Also, print the estimated time to completion by dividing the number of iterations to go by the ratio of the number of interations completed since the last refresh to the refresh interval.

#### Intended Effect

Get Andrew to shut up

Fixes #2070 

#### How to Verify

Look at a run that takes more than `refresh` seconds

#### Side Effects

Departs from past behavior. Any (misguided) tool that is parsing the progress based on iteration differences might break.

Also, the default value of `refresh` used by all the interfaces are not particularly appropriate when (re)interpreted in seconds. 100 seconds is too long as a default. 30 sounds about right.

#### Documentation

None

#### Copyright and Licensing

Please list the copyright holder for the work you are submitting (this will be you or your assignee, such as a university or company): Trustees of Columbia University


bgoodri:
Jenkins, retest this please.

bgoodri:
Jenkins, test this please.

bgoodri:
Jenkins, WTF is wrong with you?

bbbales2:
Hahahahaha, the chat thing got disabled. Look for “retry” in here: http://discourse.mc-stan.org/t/new-jenkins-jobs-tutorial/2383

bgoodri:
Possibly there is some test in CmdStan that is broken either by expecting a fixed number of entries in the logger or by passing `refresh = 0` and breaking because it no longer prints the Elapsed Time thing at the end. But looking at the log, there is no actual error message.

syclik:
That sounds plausible. If it's doing that, part of the motivation was to lock down the output so it didn't break RStan (and also PyStan). We don't have a well-defined output format for this stuff, but if it changes, then all 3 interfaces need to change with it.

bgoodri:
The code that generates the progress is the same, but it now occurs a
non-fixed number of times. There were some tests in Stan that assumed the
output was coming a fixed number of times and / or was coming irrespective
of whether refresh = 0 that I relaxed. But I can't even see which test in
CmdStan needs a similar fix.

bob-carpenter:
I also poked around in here and can't find what failed.  I pulled
up a console log that was very brief and just said it got a return error.

Have you tried running the upstream tests locally to see if you get more information?

bgoodri:
I can't make it through the entire unit tests due to https://github.com/stan-dev/stan/issues/2549 but I don't think that is what Jenkins is experiencing.

bob-carpenter:
There should be issues created for something this major and the PR should reference the issue.  I think it saves time on false starts for development.

Can we maintain backward compatibility by making time a second condition on iteraiton output?  That is, we would have two refresh values, N for minimum iterations and T for minimum seconds between iterations.  That is, we wait for N iterations *and* T seconds to go by.   Then if we set N = 0, the refresh will be totally controlled by time if there's a new iteration to report.  And if we set T = 0, everything is backward compatible and controlled by iteration.  Backard compatible default behavior would have T = 0.  Turning off output would then be turning T or N up very high (though that's awkward and we could use a negative number, too).

30s seems like an awful long time before the first refresh.  What I want is something with a growing span, say a refresh at 1s, 10s, 1m, and every 5m thereafter or something like that.  But even that would depend on how long the job's running.  I can't imagine ever wanting less than once/hour updates.  And I'm not really going to be actively watching after a minute.

bgoodri:
There is a referenced issue: https://github.com/stan-dev/stan/issues/2070

We could certainly get more fancy, but what in the current behavior of the progress logger is worth preserving? I have never thought the strong presumption for backwards compatibility extended to how things look as long as the draws are the same as they were before.

bgoodri:
Also, the progress logger does not include the chain ID currently, but for most of Stan's history it did. Then it went away, came back, and went away again. No one complained that doing so broke their homemade progress parser, so my guess is there are no homemade progress parsers to break.

bob-carpenter:
> There is a referenced issue: #2070

Thanks.  Didn't see that earlier.


> We could certainly get more fancy, but what in the current behavior of the progress logger is worth preserving?

1.  I think it's worth preserving backward compatibility unless there's a compelling reason to abandon it.  This PR would change the meaning of "refresh = N".  I don't think we have to do that, so I don't think we should.

2.  Removing a way to have deterministic behavior.  Not only did default behavior change, but the default behavior went from deterministic in the reporting of iterations to non-deterministic.  It was always non-deterministic on reporting run time.  So this probably isn't an issue, but I'd like to check with downstream clients.  


> I have never thought the strong presumption for backwards compatibility extended to how things look as long as the draws are the same as they were before.

This is what I'd like to ask those managing downstream clients about. I take it you're OK with it for RStan.

bob-carpenter:
> On Jun 16, 2018, at 1:43 AM, bgoodri <notifications@github.com> wrote:
> 
> Also, the progress logger does not include the chain ID currently, but for most of Stan's history it did. Then it went away, came back, and went away again.

It should include the chain ID.  My guess is that the changes were accidental, but maybe not.  We should be testing this kind of thing.


> No one complained that doing so broke their homemade progress parser, so my guess is there are no homemade progress parsers to break.

Thanks---that's a really helpful reference point.

bgoodri:
The backwards compatibility standard you are applying is inconsistent with the standard of backward compatibility that has been applied in the past. In addition to the chain ID hokey pokey, we have changed the number of digits in the print output, changed the way n_eff is calculated, and are changing the way R_hat is calculated. In each of those situations, we didn't introduce another flag that defaults to false in order to keep doing things the old way unless the user flips it on; we forced people to do things the new way because it is better. 

The progress reports is even less consequential than those things. The only harm is if someone has an old script where they specify `refresh = 1000` or something _in order to keep the progress output minimal_ and then they run their old script again, which gets reinterpreted as "refresh every 1000 seconds", then the progress output will be even more minimal than what they might have been expecting based on Stan's past behavior. But that isn't a big deal.

bob-carpenter:
> On Jun 17, 2018, at 2:05 AM, bgoodri <notifications@github.com> wrote:
> 
> The backwards compatibility standard you are applying is inconsistent with the standard of backward compatibility that has been applied in the past.

It's OK for us to change standards.  What we need to do is make them more clear at any given time.

I do not think the right thing to do is try to enforce bit-level release-over-release backward compatibility.


> In addition to the chain ID hokey pokey, we have changed the number of digits in the print output, changed the way n_eff is calculated, and are changing the way R_hat is calculated. In each of those situations, we didn't introduce another flag that defaults to false in order to keep doing things the old way unless the user flips it on; we forced people to do things the new way because it is better.

It might be helpful to separate out the concerns here.

I think the ID "hokey pokey" was a bug that got through due to insufficient testing.  

The n_eff thing changed the estimator, not what's being estimated.  So nothing in the API changes meaning.

The R-hat replacement's a work in progress.  If it ever does come up, I will be OK deprecating R-hat support and adding a second diagnostic that we recommend.  I'm also OK changing the default output in various ways.

My concerns about this particular pull request are:

1.  Backward compatibility

refresh=N changes meaning from N iterations to N seconds

2.  Non-deterministic behavior

With the same seed, get different number of iteration updates.

I'm OK with (2) but wanted to check with Sebastian, since he's been the one that's been most constrained by "reproduciblity" at this level.  My guess is that you (Ben) are right and moving to a non-deterministic behavior is OK.

I'm less OK with (1), which is why I think the simple expedient of allowing control by both iterations and time makes sense.  Like with potential R-hat replacements, I think the responsible thing to do would be to deprecate the per-iteration method and add the timed method and then remove per-iteration in a major release.

> The progress reports is even less consequential than those things. The only harm is if someone has an old script where they specify refresh = 1000 or something in order to keep the progress output minimal and then they run their old script again, which gets reinterpreted as "refresh every 1000 seconds", then the progress output will be even more minimal than what they might have been expecting based on Stan's past behavior. But that isn't a big deal.

I agree the danger is minimal here.  The "harm" in my mind is the perceived (and actual) instability of our interfaces.  

I'd like to hear from more devs here than just the two of us.

syclik:
I'll review the comments so far and put up some thoughts tomorrow.

syclik:
I just saw the discussion here:
http://discourse.mc-stan.org/t/timed-iteration-updates-and-backward-compatibility-standards/4552/12

I'll post a response there tomorrow.

bob-carpenter:
@bgoodri --- any idea what's up with this failure?  We're about to release 2.18 and this would be nice to include if it's just something like the tests needing to be kicked off again.

bgoodri:
No. This is what I was complaining to @seantalts about: Jenkins alleges that the CmdStan tests failed but if you open up the log there is no actual error message or test failure. 

But Daniel is not agreeing with the PR anyway, so we'll probably have to take it up for 2.19.

seantalts:
Weird, the same bug was happening on a Math PR. I added a workaround to the Jenkinsfile in Math but not here. I will open a PR.

src/stan/services/util/generate_transitions.hpp
```diff
@@ -22,8 +23,9 @@ namespace stan {
        * @param[in] finish end iteration number used for printing messages
        * @param[in] num_thin when save is true, a draw will be written to the
        *   mcmc_writer every num_thin iterations
-       * @param[in] refresh number of iterations to print a message. If
-       *   refresh is zero, iteration number messages will not be printed
+       * @param[in] refresh initial number of seconds between progress messages.
```
syclik: @bgoodri, was this really the intended behavior? The refresh rate gets doubled?

syclik:
I'm working on fixing it so it can handle this, but I'm really confused.

Do you really want "1" second refresh to mean it'll get printed at times:
- 1 sec
- 3 sec
- 7 sec
- 15 sec
- 31 sec 

and so on?

bgoodri:
Yeah, or something like that. Marginal utility of progress reports is
decreasing so we should not print that many of them.

bob-carpenter:
I think it's good to provide this kind of early indication that everything is running.  30 seconds is a long time to wait for first feedback by default.

A factor of four-ish would work for a default

1s
4s
16s
64s
4.25m
17m
64m

and then I think capping at 64m makes sense for a long running job, but then I don't have much experience with those other than ones running on my desktop.

bob-carpenter:
I think we're OK merging this now after discussions of what pieces of the API need to remain stable.  Is there something that needs to be done to get this working with tests?

bgoodri:
@syclik has a different branch now that does the same thing in practice
https://discourse.mc-stan.org/t/development-of-time-based-refresh/5988

rok-cesnovar:
Closing stale stan-dev/stan pull requests. This one  has been stale for 2+ years. Feel free to reopen once the conflicts are resolved and if you wish to proceed here.

PR #2566, opened 2018-07-02
Update BNF of Stan language

### LINKED ISSUE #2564: Errors in BNF description of Stan
#### Summary:

While working on a [tree-sitter](https://github.com/jrnold/tree-sitter-stan) grammar for Stan, I found some places where the BNF does not match the behavior of Stan's parser.

-    The exponent literal is missing an optional sign: `exp_literal ::= ('e' | 'E') ?('+' | '-') integer_literal`
-    The numeric literal rules could be written as 
      ```
      numeric_literal ::= integer_literal | real_literal
      integer_literal ::= [0-9] [0-9]*
      real_literal ::= integer_literal '.' [0-9]* ?exp_literal
        | '.' [0-9]* ?exp_literal
        |  integer_literal exp_literal
      ```
      The current production rules suggest that leading `0`s aren't valid, but they are parsed fine.
-    The `lhs` rule should allow for multiple indexes
-    The `range` rule should be adjusted (perhaps with additional expressions) to exclude certain operators (as in the `expression7` grammar in the parser).

I'll submit a pull request with these changes.

#### Current Version:
v2.17.1

jrnold:
Fixes #2564

#### Submission Checklist

#### Summary

Updates the BNF of the Stan modeling languages in the documentation. This only changes documentation.

#### Copyright and Licensing

Please list the copyright holder for the work you are submitting (this will be you or your assignee, such as a university or company): Jeffrey B. Arnold

### DIFF
--- a/src/docs/stan-reference/appendices.tex
+++ b/src/docs/stan-reference/appendices.tex
@@ -610,12 +610,12 @@ var_decls_statements ::= '{' var_decl* statement* '}'
 {
 \small
 \begin{Verbatim}[fontsize=\small]
-function_decl ::= unsized_return_type identifier '(' unsized_types ')'
-                 statement
+function_decl ::= return_type identifier '(' parameter_decl % ',' ')'
+                  statement
 
-unsized_return_type ::= 'void' | unsized_type
-unsized_type ::= (basic_type ?unsized_dims)
-unsized_types ::= unsized_type % ','
+return_type ::= 'void' | unsized_type
+parameter_decl ::= ?'data' unsized_type identifier
+unsized_type ::= basic_type ?unsized_dims
 basic_type ::= 'int' | 'real' | 'vector' | 'row_vector' | 'matrix'
 unsized_dims ::= '['  ','*  ']'
 \end{Verbatim}
--- a/src/docs/stan-reference/appendices.tex
+++ b/src/docs/stan-reference/appendices.tex
@@ -590,7 +590,7 @@ e.g., \code{A\{6\}} is equivalent to a sequence of six copies of \code{A}.
 
 {\small
 \begin{Verbatim}
-program ::= ?functions ?data ?tdata ?params ?tparams model ?generated
+program ::= ?functions ?data ?tdata ?params ?tparams ?model ?generated
 
 functions ::= 'functions' function_decls
 data ::= 'data' var_decls
@@ -680,6 +680,7 @@ range_expression ::= range_expression arithmeticInfixOp range_expression
 common_expression ::= real_literal
                     | variable
                     | '{' expressions '}'
+                    | '[' expressions ']'
                     | function_literal '(' ?expressions ')'
                     | function_literal '(' expression ?('|' expression % ',') ')'
                     | integrate_ode '(' function_literal (',' expression){6} ')'
@@ -688,7 +689,7 @@ common_expression ::= real_literal
                     | integrate_ode_bdf
                       '(' function_literal (',' expression){6|9} ')'
                     | algebra_solver
-                    '('function_literal (',' expression){4|7} ')'
+                    '(' function_literal (',' expression){4|7} ')'
                     | '(' expression ')'
 
 prefixOp ::= ('!' | '-' | '+' | '^')
--- a/src/docs/stan-reference/appendices.tex
+++ b/src/docs/stan-reference/appendices.tex
@@ -671,7 +671,7 @@ expression ::= expression `?` expression `:` expression
              | expression postfixOp
              | common_expression
 
-range_expression ::= range_expression mathInfixOp range_expression
+range_expression ::= range_expression arithmeticInfixOp range_expression
                    | prefixOp range_expression
                    | range_expression postfixOp
                    | range_expression '[' indexes ']'
@@ -691,15 +691,15 @@ common_expression ::= real_literal
                     '('function_literal (',' expression){4|7} ')'
                     | '(' expression ')'
 
-prefixOp ::= '!' | '-' | '+' | '^'
+prefixOp ::= ('!' | '-' | '+' | '^')
 
-postfixOp ::= '\\''
+postfixOp ::= '\''
 
-infixOp ::= mathInfixOp | logicalInfixOp
+infixOp ::= arithmeticInfixOp | logicalInfixOp
 
-mathInfixOp ::= ('+' | '-' | '*' | '/' | '%' | '\\' | '.*' | './')
+arithmeticInfixOp ::= ('+' | '-' | '*' | '/' | '%' | '\' | '.*' | './')
 
-logicalInfixOp :: ('||' | '&&' | '==' | '!=' | '<' | '>' | '<=' | '>=')
+logicalInfixOp :: ('||' | '&&' | '==' | '!=' | '<' | '<=' | '>' | '>=')
 
 index ::= ?(expression | expression ':' | ':' expression
         | expression ':' expression)
--- a/src/docs/stan-reference/appendices.tex
+++ b/src/docs/stan-reference/appendices.tex
@@ -725,19 +725,17 @@ function_literal ::= identifier
 \begin{Verbatim}[fontsize=\small]
 statement ::= atomic_statement | nested_statement
 
-atomic_statement ::= atomic_statement_body ';'
-
-atomic_statement_body ::=  lhs assignment_op expression
-   | expression '~' identifier '(' expressions ')' ?truncation
-   | function_literal '(' expressions ')'
-   | 'increment_log_prob' '(' expression ')'
-   | 'target' '+=' expression
-   | 'break'
-   | 'continue'
-   | 'print' '(' (expression | string_literal) % ',' ')'
-   | 'reject' '(' (expression | string_literal) % ',' ')'
-   | 'return' expression
-   | ''
+atomic_statement ::=  lhs assignment_op expression ';'
+   | expression '~' identifier '(' expressions ')' ?truncation ';'
+   | function_literal '(' expressions ')' ';'
+   | 'increment_log_prob' '(' expression ')' ';'
+   | 'target' '+=' expression ';'
+   | 'break' ';'
+   | 'continue' ';'
+   | 'print' '(' (expression | string_literal) % ',' ')' ';'
+   | 'reject' '(' (expression | string_literal) % ',' ')' ';'
+   | 'return' expression ';'
+   | ';'
 
 assignment_op ::= '<-' | '=' | '+=' | '-=' | '*=' | '/=' | '.*=' | '/*='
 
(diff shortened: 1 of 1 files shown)


src/docs/stan-reference/appendices.tex
```diff
@@ -663,35 +664,55 @@ \subsection{Expressions}
...
-             | integrate_ode_rk45
-               '(' function_literal (',' expression){6|9} ')'
-             | integrate_ode_bdf
-               '(' function_literal (',' expression){6|9} ')'
-             | algebra_solver
-             '('function_literal (',' expression){4|7} ')'
-             | '(' expression ')'
+             | common_expression
+
+range_expression ::= range_expression mathInfixOp range_expression
```
bob-carpenter: There is no restriction on what can be a range expression other than that integers require integers and that all variables have to be have been introduced.  The only reason things like odes aren't allowed is that they don't return scalars.  So I think this is easier to do with a side condition than in the grammar.

One thing that's definitely missing here is functions, which are allowed in ranges.

src/docs/stan-reference/appendices.tex
```diff
@@ -663,35 +664,55 @@ \subsection{Expressions}
...
+                    '('function_literal (',' expression){4|7} ')'
+                    | '(' expression ')'
+
+prefixOp ::= '!' | '-' | '+' | '^'
+
+postfixOp ::= '\\''
+
+infixOp ::= mathInfixOp | logicalInfixOp
+
+mathInfixOp ::= ('+' | '-' | '*' | '/' | '%' | '\\' | '.*' | './')
```
bob-carpenter: Thanks for writing these out.  There's also a backslash for left division.  There should be a complete list in the manual under the precedence table.

src/docs/stan-reference/appendices.tex
```diff
@@ -663,35 +664,55 @@ \subsection{Expressions}
...
+                      '(' function_literal (',' expression){6|9} ')'
+                    | algebra_solver
+                    '('function_literal (',' expression){4|7} ')'
+                    | '(' expression ')'
+
+prefixOp ::= '!' | '-' | '+' | '^'
+
+postfixOp ::= '\\''
+
+infixOp ::= mathInfixOp | logicalInfixOp
```
bob-carpenter: How about "arithmetic" instead of "math"?

src/docs/stan-reference/appendices.tex
```diff
@@ -663,35 +664,55 @@ \subsection{Expressions}
...
+logicalInfixOp :: ('||' | '&&' | '==' | '!=' | '<' | '>' | '<=' | '>=')
 
 index ::= ?(expression | expression ':' | ':' expression
-            | expression ':' expression)
+        | expression ':' expression)
 
 indexes ::= index % ','
 
-integer_literal ::= 0 | [1-9] [0-9]*
+integer_literal ::= [0-9]+
```
bob-carpenter: The original one is correct.  Integer literals may not have leading zeroes, e.g., "007" is not an integer literal.

src/docs/stan-reference/appendices.tex
```diff
@@ -663,35 +664,55 @@ \subsection{Expressions}
...
 
-real_literal ::= integer_literal ?('.' [0-9]*) ?exp_literal
+real_literal ::= integer_literal '.' [0-9]* ?exp_literal
```
bob-carpenter: Much better.  Thanks for fixing.

src/docs/stan-reference/appendices.tex
```diff
@@ -724,7 +745,7 @@ \subsection{Statements}
 
 truncation ::= 'T' '[' ?expression ',' ?expression ']'
 
-lhs ::= identifier ?('[' indexes ']')
+lhs ::= identifier ('[' indexes ']')*
```
bob-carpenter: I didn't even realize the language supported `a[2][3] = 5;`, but I just verififed and the new version's correct.  Thanks for updating.

src/docs/stan-reference/appendices.tex
```diff
@@ -645,15 +645,16 @@ \subsection{Variable Declarations}
 
 range_constraint ::= ?('<' range '>')
 
-range ::= 'lower' '=' expression ',' 'upper' = expression
-        | 'lower' '=' expression
-        | 'upper' '=' expression
+range ::= 'lower' '=' range_expression ',' 'upper' = range_expression
```
bob-carpenter: See below---the current `range_expression` isn't right and I'd suggest just treating the value as a semantic constraint rather than trying to pack it into the syntax.

src/docs/stan-reference/appendices.tex
jrnold: But what about [expression07_grammar_def.hpp](https://github.com/stan-dev/stan/blob/9d30c0be79f73c4d7a4d1542fce304ba1935f314/src/stan/lang/grammars/expression07_grammar_def.hpp).  Any operataor with lower precedence than `+`|`-` can't be inside a range expression in order to resolve the ambiguity of `<` and `>`.

src/docs/stan-reference/appendices.tex
jrnold: Sorry, that was just sloppiness.

src/docs/stan-reference/appendices.tex
jrnold: Agreed.

src/docs/stan-reference/appendices.tex
jrnold: Neither did I.  But it's somewhere in all those tests.

src/docs/stan-reference/appendices.tex
jrnold: But this parses and compiles.
```
library("rstan")
foo <- "transformed data {
  int a = 007;
} model {
  
}"
stan_model(model_code = foo)
```

jrnold: And so does all this
```
library("rstan")
foo <- "transformed data {
  real alpha = 0e7;
  real bravo = 0;
  int charlie = 0;
  int delta = 0000000;
  real echo = 00.;
  int foxtrot = 007;
} model {
  
}"
stan_model(model_code = foo)
```

So it seems that at the moment the current implementation doesn't care about the leading zeros. Unless it is treating them as reals and then casting them back into integers?

src/docs/stan-reference/appendices.tex
jrnold: The inclusion of the `common_expression` rule in `range_expression` means that includes function expressions.

src/docs/stan-reference/appendices.tex
jrnold: Rather, it was trying to figure out exactly how many backslashes needed to escape things.

src/docs/stan-reference/appendices.tex
jrnold: Fair enough.  I wrote it out that way because it was the easiest way to get the GLR parser that tree-sitter uses to correctly parse the range constraints and it seemed like the Spirit parser implements it as a separate grammar rather than just a semantic constraint.

jrnold:
I also think that the BNF is missing an expression for vector expressions as in Sec 3.3 of the 2.17 manual.
```
row_vector[2] rv2= [ 1, 2 ];
vector[3] v3 = [ 3, 4, 5 ]';
```

src/docs/stan-reference/appendices.tex
bob-carpenter: Really?  That was purely by accident.  I meant to restrict to C++ literals.  Here's how the code gets compiled:

```
  stan::math::assign(alpha,0.0);
...
  stan::math::assign(echo,0.0);
...
  stan::math::assign(foxtrot,7);
```

So we must be using Boost somewhere to convert the funny numbers with leading zeroes to integers on the insde of the AST, then generating back out in standardized way.

What should we do about this?  I suppose it doesn't hurt anything other than aesthetically to allow all those crazy non-standard literals.

P.S.  Model blocks are optional now!

src/docs/stan-reference/appendices.tex
bob-carpenter: I originally missed that the last option in the range grammar was to call the common grammar.  But I thought the range expressions were more restrictive.

This works:

```
real<lower = exp(-1.9), upper = 2.9 + 1.3> x;
```
but doesn't seem to be allowed in the specific range expression grammar.   

But you're right that it's not a free-for-all. I had forgotten that this isn't allowed (until reminded by your next comment):

```
  real<lower = 1 > 2> y;
```
I gave up before figuring out how to do the backtracking for this case.

bob-carpenter:
I pulled out the manual into a bookdown document with the BNF.  The PR is still hung up due to upstream SNAFUs on Jenkins.  So anything you do here I'm going to have to diff and then merge it by hand into the new BNFs.

The new BNFs are in the branch behind this PR:

https://github.com/stan-dev/stan/pull/2558

in

https://github.com/stan-dev/stan/blob/feature/2418-manual-ref-html/src/docs/reference-manual/syntax.Rmd

src/docs/stan-reference/appendices.tex
jrnold: I was also surprised that `model` blocks don't require `{}` and could be followed immediately by a single statement.

jrnold: So what blocks are actually required now? Are they all optional in the grammar, with any issues arising in the sematics?

Since the leading 0's don't cause any harm, I don't see why not. Is not having a leading `0` in C++ literals due to octal literals?

src/docs/stan-reference/appendices.tex
jrnold: I think it does. Since `range_expression` includes the`common_expression` expression, it also includes `identifier '(' expression ')'`, which would handle the lower bound,`exp(-1.9)`.  The `range_expression` rule includes the expression `range_expression arthmeticInfixOp range_expression` which would match `2.9 + 1.3` in the upper bound.   

Handling range expressions was the main/only reason that I separated the infix operators into logical (`||`, `&&`, `<`, `<=`, `>=`, `>`) operators (not allowed in range expressions) and arithmetic (the other higher precedence operators allowed in range expressions). From scanning the code and empirical trial and error, this is how I understood how it is being handled by the Stan parser.

jrnold: And even though it's a "hack", there's really no expressivity lost since this equivalent way of writing that variable declaration is valid
```
real<lower = (1 > 2)> y;
```

jrnold:
Regarding bookdown: Awesome!

I'll update this branch just to have my work somewhere, and when #2558 is merged I'll rebase this PR and make those additional edits.

bob-carpenter:
I went ahead and OK-ed the changes.  All of the tests are fouled up upstream somewhere, but when everything clears up and this passes, please feel free to merge when you've updated to your satisfaction.

bob-carpenter:
It may be the octals---don't know.  I don't think Java allows it, either, but they also have octal literals.  I've never seen it allowed, but I just tried R, and sure enough, it allows

a <- 007

bob-carpenter:
The braces in the model block should be required.  It's a bug that they're not that Mitzi is going to fix.

bob-carpenter:
D'oh.  My parser broke and I wasn't reading that as function name arguments.  Thanks for being patient with me here.

PR #2570, opened 2018-07-07
Sampling is able to initialize when there is an exception in transformed parrameters

### LINKED ISSUE #2562: Exception in transformed parameters stops sampling
#### Summary:
An exception in transformed parameters stops sampling from continuing. It should catch the exception and continue to the next iteration, if possible.

#### Description:
When constraints are violated, it seems like the exception terminates Stan. We should catch it and move to the next iteration. It's odd that it doesn't print the messages to the logger. Perhaps there's some buffering happening.

#### Reproducible Steps:
I need to provide a minimal example.

#### Current Output:
No prints from the program get printed, but then I see this:

```
Exception: foo_model_namespace::write_array: theta[1] is -20.6479, but must be greater than or equal to 0  (in 'foo.stan' at line 10)
```

#### Expected Output:
printout + continue to sample.

#### Additional Information:
I'm seeing slightly different behavior in CmdStan and RStan, so I don't know what's up. I think this is probably being handled improperly by Stan, though.

#### Current Version:
v2.17.1

syclik:
#### Submission Checklist

#### Summary

This pull request allows initialization of a Stan program to continue when there's a violation of constraints in the transformed parameters block (or a `reject()` statement). This fixes #2562, reverting behavior to pre-v2.17.

#### Intended Effect

This is meant to make running Stan easier. When there's an exception thrown, the initialization procedure treats that as a bad starting point and retries.

#### How to Verify

There are updated tests in `src/test/unit/services/util/initialize_test.cpp`. In the new tests, there is a mock model that throws in the `write_array()` method of the model, which is where the transformed parameters are validated. The new tests make sure that the initialization is retried 100 times.

#### Side Effects

Some additional things in the pull request:
- updated .gitignore for *.gch (precompiled headers)
- added new test to `random_var_context_test.cpp` to verify that `random_var_context` will now be constructed in a bad state.

#### Documentation

None needed. 

#### Copyright and Licensing

Generable

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -56,3 +56,5 @@ stan.kdev4
 
 # local make include
 /make/local
+
+*.gch
--- a/src/stan/services/util/initialize.hpp
+++ b/src/stan/services/util/initialize.hpp
@@ -85,31 +85,40 @@ std::vector<double> initialize(Model& model,
   int MAX_INIT_TRIES = is_fully_initialized || init_zero ? 1 : 100;
   int num_init_tries = 0;
   for (; num_init_tries < MAX_INIT_TRIES; num_init_tries++) {
-    stan::io::random_var_context
-        random_context(model, rng, init_radius, init_zero);
+    std::stringstream msg;
+    try {
+      stan::io::random_var_context
+          random_context(model, rng, init_radius, init_zero);
 
-    if (!any_initialized) {
-      unconstrained = random_context.get_unconstrained();
-    } else {
-      stan::io::chained_var_context context(init, random_context);
+      if (!any_initialized) {
+        unconstrained = random_context.get_unconstrained();
+      } else {
+        stan::io::chained_var_context context(init, random_context);
 
-      std::stringstream msg;
-      try {
         model.transform_inits(context,
                               disc_vector,
                               unconstrained,
                               &msg);
-      } catch (const std::exception& e) {
-        if (msg.str().length() > 0)
-          logger.info(msg);
-        logger.info(e.what());
-        throw;
       }
+    } catch (std::domain_error& e) {
       if (msg.str().length() > 0)
         logger.info(msg);
+      logger.info("Rejecting initial value:");
+      logger.info("  Error evaluating the log probability"
+                  " at the initial value.");
+      logger.info(e.what());
+      continue;
+    } catch (std::exception& e) {
+      if (msg.str().length() > 0)
+        logger.info(msg);
+      logger.info("Unrecoverable error evaluating the log probability"
+                  " at the initial value.");
+      logger.info(e.what());
+      throw;
     }
+
+    msg.str("");
     double log_prob(0);
-    std::stringstream msg;
     try {
       log_prob = model.template log_prob<false, true>
                  (unconstrained, disc_vector, &msg);
--- a/src/stan/services/util/initialize.hpp
+++ b/src/stan/services/util/initialize.hpp
@@ -13,202 +13,202 @@
 #include <vector>
 
 namespace stan {
-  namespace services {
-    namespace util {
+namespace services {
+namespace util {
 
-      /**
-       * Returns a valid initial value of the parameters of the model
-       * on the unconstrained scale.
-       *
-       * For identical inputs (model, init, rng, init_radius), this
-       * function will produce the same initialization.
-       *
-       * Initialization first tries to use the provided
-       * <code>stan::io::var_context</code>, then it will generate
-       * random uniform values from -init_radius to +init_radius for missing
-       * parameters.
-       *
-       * When the <code>var_context</code> provides all variables or
-       * the init_radius is 0, this function will only evaluate the
-       * log probability of the model with the unconstrained
-       * parameters once to see if it's valid.
-       *
-       * When at least some of the initialization is random, it will
-       * randomly initialize until it finds a set of unconstrained
-       * parameters that are valid or it hits <code>MAX_INIT_TRIES =
-       * 100</code> (hard-coded).
-       *
-       * Valid initialization is defined as a finite, non-NaN value
-       * for the evaluation of the log probability and all its
-       * gradients.
-       *
-       * @param[in] model the model
-       * @param[in] init a var_context with initial values
-       * @param[in,out] rng random number generator
-       * @param[in] init_radius the radius for generating random values.
-       *   A value of 0 indicates that the unconstrained parameters (not
-       *   provided by init) should be initialized with 0.
-       * @param[in] print_timing indicates whether a timing message should
-       *   be printed to the logger
-       * @param[in,out] logger logger for messages
-       * @param[in,out] init_writer init writer (on the unconstrained scale)
-       * @throws exception passed through from the model if the model has a
-       *   fatal error (not a std::domain_error)
-       * @throws std::domain_error if the model can not be initialized and
-       *   the model does not have a fatal error (only allows for
-       *   std::domain_error)
-       * @return valid unconstrained parameters for the model
-       */
-      template <class Model, class RNG>
-      std::vector<double> initialize(Model& model,
-                                     stan::io::var_context& init,
-                                     RNG& rng,
-                                     double init_radius,
-                                     bool print_timing,
-                                     stan::callbacks::logger& logger,
-                                     stan::callbacks::writer&
-                                     init_writer) {
-        std::vector<double> unconstrained;
-        std::vector<int> disc_vector;
+/**
+ * Returns a valid initial value of the parameters of the model
+ * on the unconstrained scale.
+ *
+ * For identical inputs (model, init, rng, init_radius), this
+ * function will produce the same initialization.
+ *
+ * Initialization first tries to use the provided
+ * <code>stan::io::var_context</code>, then it will generate
+ * random uniform values from -init_radius to +init_radius for missing
+ * parameters.
+ *
+ * When the <code>var_context</code> provides all variables or
+ * the init_radius is 0, this function will only evaluate the
+ * log probability of the model with the unconstrained
+ * parameters once to see if it's valid.
+ *
+ * When at least some of the initialization is random, it will
+ * randomly initialize until it finds a set of unconstrained
+ * parameters that are valid or it hits <code>MAX_INIT_TRIES =
+ * 100</code> (hard-coded).
+ *
+ * Valid initialization is defined as a finite, non-NaN value
+ * for the evaluation of the log probability and all its
+ * gradients.
+ *
+ * @param[in] model the model
+ * @param[in] init a var_context with initial values
+ * @param[in,out] rng random number generator
+ * @param[in] init_radius the radius for generating random values.
+ *   A value of 0 indicates that the unconstrained parameters (not
+ *   provided by init) should be initialized with 0.
+ * @param[in] print_timing indicates whether a timing message should
+ *   be printed to the logger
+ * @param[in,out] logger logger for messages
+ * @param[in,out] init_writer init writer (on the unconstrained scale)
+ * @throws exception passed through from the model if the model has a
+ *   fatal error (not a std::domain_error)
+ * @throws std::domain_error if the model can not be initialized and
+ *   the model does not have a fatal error (only allows for
+ *   std::domain_error)
+ * @return valid unconstrained parameters for the model
+ */
+template <class Model, class RNG>
+std::vector<double> initialize(Model& model,
+                               stan::io::var_context& init,
+                               RNG& rng,
+                               double init_radius,
+                               bool print_timing,
+                               stan::callbacks::logger& logger,
+                               stan::callbacks::writer&
+                               init_writer) {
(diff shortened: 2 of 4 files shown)


seantalts:
Travis was broken by something downstream in Math - submitted https://github.com/stan-dev/stan/pull/2573 to fix it, sorry about that.

.gitignore
```diff
@@ -56,3 +56,5 @@ stan.kdev4
 
 # local make include
 /make/local
+
+*.gch
```
bob-carpenter: [optional]

Looks like this file is missing a newline.  I don't know if that's a big deal or not.

src/stan/services/util/initialize.hpp
```diff
@@ -13,202 +13,211 @@
...
+ * parameters.
+ *
+ * When the <code>var_context</code> provides all variables or
+ * the init_radius is 0, this function will only evaluate the
+ * log probability of the model with the unconstrained
+ * parameters once to see if it's valid.
+ *
+ * When at least some of the initialization is random, it will
+ * randomly initialize until it finds a set of unconstrained
+ * parameters that are valid or it hits <code>MAX_INIT_TRIES =
```
bob-carpenter: What is `MAX_INIT_TRIES`?  Is it something the user can set?  If not, then I don't think it should be given a name like this that makes it look like a property that a user could set.

src/stan/services/util/initialize.hpp
```diff
@@ -13,202 +13,211 @@
...
+ * 100</code> (hard-coded).
+ *
+ * Valid initialization is defined as a finite, non-NaN value
+ * for the evaluation of the log probability and all its
```
bob-carpenter: "log probability" -> "log density function" (or "log probability desnity function")

src/stan/services/util/initialize.hpp
```diff
@@ -13,202 +13,211 @@
...
+    is_fully_initialized &= init.contains_r(param_names[n]);
+    any_initialized |= init.contains_r(param_names[n]);
+  }
 
-        int MAX_INIT_TRIES = is_fully_initialized || init_zero ? 1 : 100;
-        int num_init_tries = 0;
-        for (; num_init_tries < MAX_INIT_TRIES; num_init_tries++) {
-          stan::io::random_var_context
-            random_context(model, rng, init_radius, init_zero);
+  bool init_zero = init_radius <= std::numeric_limits<double>::min();
```
bob-carpenter: [required]

This needs documentation.  The name `init_zero` also needs to change as it's not testing for zero, it's testing for smaller than smallest normalized floating point number.  That means there are values other than 0 that this passes for.

[optional]

This could just change to `== 0` and the only thing you'd miss is unnormalized small numbers and it wouldn't need any doc or name changes.

src/stan/services/util/initialize.hpp
```diff
@@ -13,202 +13,211 @@
...
-            logger.info("");
-            std::stringstream msg1;
-            msg1 << "Gradient evaluation took " << deltaT << " seconds";
-            logger.info(msg1);
+        model.transform_inits(context,
+                              disc_vector,
+                              unconstrained,
+                              &msg);
+      }
+    } catch (std::domain_error& e) {
```
bob-carpenter: Did someone take a pass to make sure we only throw domain errors when we want to reject and keeping going?  I thought we still had Boost errors of various types and other kinds of `std::exception` like range errors and things that shouldn't stop sampling.

src/stan/services/util/initialize.hpp
```diff
@@ -13,202 +13,211 @@
...
 
-            std::stringstream msg2;
-            msg2 << "1000 transitions using 10 leapfrog steps"
-                 << " per transition would take"
-                 << " " << 1e4 * deltaT << " seconds.";
-            logger.info(msg2);
+    msg.str("");
+    double log_prob(0);
+    try {
+      log_prob = model.template log_prob<false, true>
```
bob-carpenter: [question]

The Jacobian template parameter depends on the algorithm.  

Is the initialization code shared or is this just initialization for MCMC?

src/stan/services/util/initialize.hpp
```diff
@@ -13,202 +13,211 @@
...
+                 (unconstrained, disc_vector, &msg);
+      if (msg.str().length() > 0)
+        logger.info(msg);
+    } catch (std::domain_error& e) {
+      if (msg.str().length() > 0)
```
bob-carpenter: [suggestion]

This pattern is everywhere, and while it's simple, I'd rather not see this test everywhere.  Can we just have the `info()` mthod ignore empty messages in general?  If we really want a blank line, we can pass `info(" ")` or `info("\n")` if it doesn't build one in automatically.

If not that, then maybe wrap this all up in a function that does the test so this pattern isn't cut-and-paste dozens of times.

src/stan/services/util/initialize.hpp
```diff
@@ -13,202 +13,211 @@
...
 
-            logger.info("Adjust your expectations accordingly!");
-            logger.info("");
-            logger.info("");
-          }
-          if (gradient_ok)
-            break;
-        }
+    for (size_t i = 0; i < gradient.size(); ++i) {
+      if (gradient_ok && !boost::math::isfinite(gradient[i])) {
```
bob-carpenter: This can only print once and it'd be much cleaner if there was an is_finite function.

```
template <class C>
bool is_finite(const C& gradient) {
  return boost::math::is_finite(sum(gradient));
}
```
or write it out in a loop if you're worried about efficiency.  Then this code is just:

```
if (!is_finite(gradient))
  do something about it
...
```

The major pain in pulling things out into functions is the extra doc and tests that some reviewers require.

src/stan/services/util/initialize.hpp
```diff
@@ -13,202 +13,211 @@
...
+        logger.info("  Stan can't start sampling from this"
+                    " initial value.");
+        gradient_ok = false;
+      }
+    }
+    if (gradient_ok && print_timing) {
+      logger.info("");
+      std::stringstream msg1;
+      msg1 << "Gradient evaluation took " << deltaT << " seconds";
+      logger.info(msg1);
```
bob-carpenter: [future request]

Our loggers should just take these, as in:

```
logger.info() << "Gradient evaluation took " << ...;
```

Or we should have a polyadic info function that could look like:

```
logger.info("Gradient evaluation took ", deltaT, " seconds");
```

src/stan/services/util/initialize.hpp
```diff
@@ -13,202 +13,211 @@
...
+      std::stringstream msg2;
+      msg2 << "1000 transitions using 10 leapfrog steps"
+           << " per transition would take"
+           << " " << 1e4 * deltaT << " seconds.";
+      logger.info(msg2);
 
-        init_writer(unconstrained);
-        return unconstrained;
-      }
+      logger.info("Adjust your expectations accordingly!");
```
bob-carpenter: [optional]
I strongly prefer `logger.info("Adjust ... accordingly!\n\n");` without two extra empty lines.  I really think we should get rid of empty logger messages.

src/stan/services/util/initialize.hpp
```diff
@@ -13,202 +13,211 @@
...
+      logger.info("");
+      logger.info("");
+    }
+    if (gradient_ok)
```
bob-carpenter: This should return from here---there's no more work to do, then you won't need a confusing test in the next block.

```
if (gradient_ok) {
  init_writer(unconstrained);
  return unconstrained;
}
```

Then the rest of the logic simplifies.

src/stan/services/util/initialize.hpp
```diff
@@ -13,202 +13,211 @@
...
+      break;
+  }
 
+  if (num_init_tries == MAX_INIT_TRIES) {
```
bob-carpenter: Now you don't need to test  any more.  We've returned if we found a decent solution with finite gradient, right?

I think there should be a separate print for the init zero condition and the same tips should be given.  So this all reduces to something like:

```
if (init_zero) {
  logger.info("initialization at zero fialed after 100 attempts");
} else {
  std::stringstream msg;
  msg << "Initialization between (-" << init_radius ...
  logger.info(msg);
}
msg << " Try specifying initial values,"
                    " reducing ranges of constrained values,"
                    " or reparameterizing the model.";
logger.info(msg);
throw std::domain_error("Initialization failed");
```

src/stan/services/util/initialize.hpp
```diff
@@ -13,202 +13,211 @@
...
+          << " " << MAX_INIT_TRIES <<  " attempts. ";
+      logger.info(msg);
+      logger.info(" Try specifying initial values,"
+                  " reducing ranges of constrained values,"
+                  " or reparameterizing the model.");
     }
+    throw std::domain_error("Initialization failed.");
   }
+
+  init_writer(unconstrained);
```
bob-carpenter: Does `init_writer` modify `unconstrained`?  I have no idea what this is doing here, so some doc may help.

.gitignore
syclik: fixed

src/stan/services/util/initialize.hpp
syclik: fixed

src/stan/services/util/initialize.hpp
syclik: I'll leave an inline comment. There's also a function-level comment:
>  * When the <code>var_context</code> provides all variables or
>  * the init_radius is 0, this function will only evaluate the
>  * log probability of the model with the unconstrained
>  * parameters once to see if it's valid.
> ...
> * @param[in] init_radius the radius for generating random values.
> *   A value of 0 indicates that the unconstrained parameters (not
>  *   provided by init) should be initialized with 0.

I'm sure I could doc it better; what documentation do I need to leave behind so this is clear?

syclik: I changed the name to `is_initialized_with_zero` and I think there isn't much I can do with an inline comment now. Just let me know if there's anything else that I can add to the doc.

src/stan/services/util/initialize.hpp
syclik: I did a while ago. After this catch, we catch all `std::exception`, which include Boost errors.

I haven't seen any behavior stop sampling where it shouldn't. Have you? Or have you seen it reported anywhere? We can take a pass over every throw if you want.

src/stan/services/util/initialize.hpp
syclik: This is shared across all of the algorithms. I'll fix that.

The behavior shouldn't change: the initialization just checks that the log density can be evaluated and the gradients can also be evaluated. The only time we'd have different behavior is if the Jacobian term is infinite or throws when the rest of it doesn't (either for the log density or the gradients).

syclik: And... fixed.

src/stan/services/util/initialize.hpp
syclik: The pattern is everywhere. I'll create an issue and we'll take a pass.

syclik: Created an issue.

src/stan/services/util/initialize.hpp
syclik: Definitely not worried about efficiency here. That looks simpler.

I don't find it to be a pain to split things into functions with doc and tests if it's reusable. Here, I think we have `sum` and `isfinite` in the math library, so we'll just use that.

src/stan/services/util/initialize.hpp
syclik: I just created a new issue for this.

The first implementation was a compromise and was only meant to be there as a stepping stone. Now that it's in, it's a lot easier to change safely.

src/stan/services/util/initialize.hpp
syclik: I don't mind getting rid of empty lines, but we need the empty logger message. This is the behavior that allows RStan to print what chain messages came from. We assume each message is a single line and we don't have the loggers scan for `\n` to split on that. We could, but currently, we don't.

src/stan/services/util/initialize.hpp
syclik: Indeed. That is much simpler.

src/stan/services/util/initialize.hpp
syclik: No, it does not. It just outputs the initialization. This is used for ADVI and it really is necessary if we want to think about restarting.

src/stan/services/util/initialize.hpp
syclik: Yes, we should. The current design was a compromise and a stepping stone. 

I'd like move forward now that this is in.

syclik:
@bob-carpenter, thanks for leaving the detailed comments! I fixed most of it. Some of it I deferred by creating new issues. I think some of it was legitimately outside the scope here.

seantalts:
Looks like the tests are passing too. Once we get this in, I think I can release 2.18.

PR #2583, opened 2018-07-12
Corrected multi_normal_rng, multi_normal_cholesky_rng, and multi_stud…

bbbales2:
…ent_t_rng signatures and updated docs

#### Submission Checklist

#### Summary

Addresses https://github.com/stan-dev/stan/issues/2582

#### Intended Effect

Fixed multivariate rng signatures and added docs

#### How to Verify

```
git clone --recursive https://github.com/stan-dev/performance-tests-cmdstan.git
cd cmdstan/stan
git checkout bugfix/issue-2582-vectorized-rng-signature-errors
cd ../../
./runPerformanceTests.py stat_comp_benchmarks/benchmarks/gp_regr
```

#### Side Effects

None

#### Documentation

Improved!

#### Copyright and Licensing

### DIFF
--- a/src/stan/lang/function_signatures.h
+++ b/src/stan/lang/function_signatures.h
@@ -768,15 +768,20 @@ add("multi_gp_cholesky_lpdf", expr_type(double_type()), expr_type(matrix_type())
     }
   }
 }
-for (const auto& t : all_vector_types) {
-  add("multi_normal_rng", rng_return_type<double_type>(t), t, expr_type(matrix_type()));
- }
-for (const auto& t : all_vector_types) {
-  add("multi_normal_cholesky_rng", rng_return_type<double_type>(t), t, expr_type(matrix_type()));
- }
-for (const auto& t : all_vector_types) {
-  add("multi_student_t_rng", rng_return_type<double_type>(t), expr_type(double_type()), t, expr_type(matrix_type()));
- }
+
+add("multi_normal_rng", expr_type(vector_type()), expr_type(vector_type()), expr_type(matrix_type()));
+add("multi_normal_rng", expr_type(vector_type(), 1U), expr_type(vector_type(), 1U), expr_type(matrix_type()));
+add("multi_normal_rng", expr_type(vector_type()), expr_type(row_vector_type()), expr_type(matrix_type()));
+add("multi_normal_rng", expr_type(vector_type(), 1U), expr_type(row_vector_type(), 1U), expr_type(matrix_type()));
+add("multi_normal_cholesky_rng", expr_type(vector_type()), expr_type(vector_type()), expr_type(matrix_type()));
+add("multi_normal_cholesky_rng", expr_type(vector_type(), 1U), expr_type(vector_type(), 1U), expr_type(matrix_type()));
+add("multi_normal_cholesky_rng", expr_type(vector_type()), expr_type(row_vector_type()), expr_type(matrix_type()));
+add("multi_normal_cholesky_rng", expr_type(vector_type(), 1U), expr_type(row_vector_type(), 1U), expr_type(matrix_type()));
+add("multi_student_t_rng", expr_type(vector_type()), expr_type(double_type()), expr_type(vector_type()), expr_type(matrix_type()));
+add("multi_student_t_rng", expr_type(vector_type(), 1U), expr_type(double_type()), expr_type(vector_type(), 1U), expr_type(matrix_type()));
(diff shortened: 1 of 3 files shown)


bbbales2:
I have a question for @bob-carpenter or @syclik on this one.

The current signatures for the multi_normal_rngs are (the other two are similar):
```
vector multi_normal_rng(vector mu, matrix Sigma)
vector multi_normal_rng(row_vector mu, matrix Sigma)
vectors multi_normal_rng(vectors mu, matrix Sigma)
vectors multi_normal_rng(row_vectors mu, matrix Sigma)
```

Do you think the two signatures that take in row_vectors as their means should return row_vectors?

Like:

```
row_vector multi_normal_rng(row_vector mu, matrix Sigma)
row_vectors multi_normal_rng(row_vectors mu, matrix Sigma)
```

I didn't notice this until I was looking at the interfaces now. It seems more natural this way. I think it's programmatically cleaner the way it is though (always returning vector(s)). All the other univariate rngs keep the output type fixed and don't try to associate it with the input type (all vector outputs are std::vectors).

If you think the rngs should output row_vectors when the means are row_vectors, then I think we should just make a pull req. that removes the multivariate_rng interfaces for now and let's the rest of the stuff go in 2.18. Otherwise this should be good.

syclik:
The first argument should be vector or vectors only. No row vector or
matrix. And that naturally limits the return types.

bbbales2:
In that regard, my goal for vectorizing the rngs was imitating the _lpdfs/_lpmfs (to make models convenient). multi_normal_lpdf has these signatures:

```
real multi_normal_lpdf(vectors y | vectors mu, matrix Sigma)
real multi_normal_lpdf(vectors y | row_vectors mu, matrix Sigma)
real multi_normal_lpdf(row_vectors y | vectors mu, matrix Sigma)
real multi_normal_lpdf(row_vectors y | row_vectors mu, matrix Sigma)
```

Since row_vectors are allowed there I think they should be allowed in the _rngs.

src/stan/lang/function_signatures.h
```diff
@@ -768,15 +768,20 @@ add("multi_gp_cholesky_lpdf", expr_type(double_type()), expr_type(matrix_type())
...
-  add("multi_normal_rng", rng_return_type<double_type>(t), t, expr_type(matrix_type()));
- }
-for (const auto& t : all_vector_types) {
-  add("multi_normal_cholesky_rng", rng_return_type<double_type>(t), t, expr_type(matrix_type()));
- }
-for (const auto& t : all_vector_types) {
-  add("multi_student_t_rng", rng_return_type<double_type>(t), expr_type(double_type()), t, expr_type(matrix_type()));
- }
+
+add("multi_normal_rng", expr_type(vector_type()), expr_type(vector_type()), expr_type(matrix_type()));
```
bob-carpenter: Is vectorization not implemented for arrays of covariance matrices/Cholesky factors?  If that was intentional because it's not implemented in the math library, please open a math library issue to complete the implementations with arrays of covariance matrices.

bob-carpenter:
I spoke too soon.  I'll merge when tests pass.

syclik:
Does multi_normal actually have those signatures? We decided we weren't
going to allow that for the multivariate distribution lpdf/lpmf, but if
it's there, we should have rngs that match. (At some point we had decided
to only use vectors and matrices so we don't get into ambiguous territory
with mixing vectors and row vectors. But if it's in, it's in.)

bob-carpenter:
Yes, this matches the lpdfs.  So I think the solution here is the best we can do.

I hadn't realized this has actually been merged, either.  But I think the decision was that too many users failed to transpose row vectors pulled out of matrices, so it'd be easier to use with row vector signatures.

We definitely didn't want to try to broadcast row vectors or vectors into matrices.  We were also worried specifically about the case this issue raises, which is that if you have something vectorized with multiple arguments, it's not clear what the return type should be if the two arguments are vectors.

bbbales2:
I went combing through the original pull req. just to make sure everything else was in line. Docs for the vectorized binomial rng were missing. With apologies to Travis and Jenkins, I just added em' here: https://github.com/stan-dev/stan/pull/2583/commits/cc4cc11df90269e03201be2345e875c704b01788

bob-carpenter:
I don't understand that link.  Let me know if there's more I'll need to merge other than the original PR.

bbbales2:
Those are the changes are in this PR, there isn't a second one. I just added them after you approved it. It's 2 lines of text to the doc. I didn't want to open a 2nd PR for something super tiny like that.

roualdes:
Thanks @bbbales2 for your help on this one.

PR #2598, opened 2018-08-06
Revert "Update expected values for performance test"

mitzimorris:
Reverts stan-dev/stan#2538


syclik:
We shouldn't revert these.

These values should have changed with:  stan-dev/math#874. That was merged into math.

mitzimorris:
closing without merging  - original merge of PR was done before mathlib update, but everything should be good now.

PR #2635, opened 2018-09-11
Add argument to stanc to allow include paths to be passed

jjramsey:
#### Submission Checklist

#### Summary

This implements part of [the solution recommended by syclik](https://github.com/stan-dev/cmdstan/issues/633#issuecomment-406593430) to allow CmdStan to use user-provided include flags. In particular, it allows stanc to accept a flag `--include_paths` that has a comma-separated list of include paths as an argument. Since `boost::tokenizer` is used to parse this argument, commas that are *not* intended as a separator (because they are part of an unfortunately named path name) can be escaped. Also, both single and double quotes may be used to quote a path.

#### Copyright and Licensing

Please list the copyright holder for the work you are submitting (this will be you or your assignee, such as a university or company): Myself, J. J. Ramsey.


src/stan/command/stanc_helper.hpp
```diff
@@ -228,6 +229,35 @@ inline int stanc_helper(int argc, const char* argv[],
...
+      boost::escaped_list_separator<char> extra_path_els("\\", // escape char
+                                                         ",",  // delimiter
+                                                         "\"'" // quote chars
+                                                         );
+      
+      boost::tokenizer<
+        boost::escaped_list_separator<char>
+        > extra_path_tokenizer(extra_path_str, extra_path_els);
+
+      std::cout << "Include paths:\n";
```
mitzimorris: why is this printing to `std::cout`?  is this just to check your work during development?

bob-carpenter: Nothing should ever print to `std::cout` in `stan-dev/stan`.  This causes undefined behavior in our interfaces.  Instead, if printing is really required, then it needs to use an argument output stream that will be configured higher up.

mitzimorris: the include path names are used in file `stan/io/program_reader.hpp`.   as it stands, the `read` method (https://github.com/stan-dev/stan/blob/7206bd9b0e2d5d585f54e0a8c1e14efca5f1350b/src/stan/io/program_reader.hpp#L219) searches all include paths in turn - lines 255-270.  if the include paths are bad, it will report and error - line 272.  to detect bad include paths, you'd need to change the code so that there's a record of the paths that were tried which can then be dumped into the error message.

mitzimorris: correct - if there are paths, add to message.

jjramsey: Just FYI, unit tests have been added, and I followed the recommendation to only list the include paths if there is an error.

PR #2649, opened 2018-09-27
Fixing broken travis build

syclik:
#### Submission Checklist

#### Summary
This should fix the broken Travis build.

Here are the different things that are being done in this PR:
1. changing make/local to set `CXX` instead of `CC`.
2. Changing the `-stdlib=libc++` to be added to `CXXFLAGS_OS` instead of `LDFLAGS_OS`. Some of these failures occurred because the `LDFLAGS_OS` are only added to the linking step, but not the compilation step (from `.cpp` to `.o`). It's needed there.
3. Adding `-pthread` to `CXXFLAGS_OS` for `g++-4.9`. I looked at the logs and it's in the build options that work pre-merge of the PR.
4. Changing the dependency generation flags. Perhaps this should be based on compiler (for a future issue): `DEPFLAGS_OS = -M -E`.

#### Intended Effect
Fix travis.

#### How to Verify
Hopefully it'll pass. I'll put up a link when it does.

#### Side Effects
None.

#### Documentation
None.

#### Copyright and Licensing

Generable

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -68,47 +68,47 @@ matrix:
           PARALLEL=2
     - <<: *linux_gcc
       env:
-        - MATRIX_EVAL="CXX=g++-4.9"
+        - MATRIX_EVAL="CXX=g++-4.9 && CXXFLAGS_OS=-pthread"
           TESTFOLDER="src/test/unit/callbacks src/test/unit/io"
           PARALLEL=1
     - <<: *linux_gcc
       env:
-        - MATRIX_EVAL="CXX=g++-4.9"
+        - MATRIX_EVAL="CXX=g++-4.9 && CXXFLAGS_OS=-pthread"
           TESTFOLDER="src/test/unit/lang/*.cpp"
           PARALLEL=1
     - <<: *linux_gcc
       env:
-        - MATRIX_EVAL="CXX=g++-4.9"
+        - MATRIX_EVAL="CXX=g++-4.9 && CXXFLAGS_OS=-pthread"
           TESTFOLDER="src/test/unit/lang/ast"
           PARALLEL=1
     - <<: *linux_gcc
       env:
-        - MATRIX_EVAL="CXX=g++-4.9"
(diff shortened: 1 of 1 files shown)


syclik:
@seantalts this is ready to review. 

I think I might need to rethink the naming of the make variable, but it’ll have to be done at the math level. Anyway, this PR fixes Travis on develop, which was the goal. Please tag anything odd or if there’s a better way to accomplish the same thing.

.travis.yml
```diff
@@ -33,82 +35,82 @@ matrix:
     - <<: *linux_clang
       env:
           # For Travis's Ubuntu 14.04, the libstdc++ is broken with clang
-        - MATRIX_EVAL="CXX=clang++-3.8 && LDFLAGS_OS=-stdlib=libc++"
+        - MATRIX_EVAL="CXX=clang++-3.8 && CXXFLAGS_OS=-stdlib=libc++"
```
seantalts: I think the `-stdlib=libc++` flag is only for the linker, right?

syclik: That’s what I thought too before this PR, but I learned something new. When you try to comple from a `.cpp` to a `.o`, it uses that flag to bring in the right includes. Without it, it can't compile. 
Perhaps we're supposed to include a flag for the path to the headers and leave the flag for linking only?

syclik: And... I didn't include key info. Adding it to the `CXXFLAGS_OS` adds the flags when it's compiling from `.cpp` to `.o` and also when it links an executable. After reading a bunch of stack overflow posts including this one: https://stackoverflow.com/questions/19774778/when-is-it-necessary-to-use-use-the-flag-stdlib-libstdc, I think it's the right thing to do... 

And... more info. It looks like the `-stdlib=libc++` is an actual compiler flag, not just a linker flag: https://libcxx.llvm.org/docs/UsingLibcxx.html#using-libc-on-linux. To add it as a link option, it's the `-lc++abi` thing. on that line.

PR #2719, opened 2018-12-22
Issue/2706 cleanup docs

mitzimorris:
#### Submission Checklist

#### Summary
Removes directory `src/docs` and all make tasks for documentation.

#### Intended Effect
Keep all user-facing documentation in repo `docs`.  This simplifies build and testing requirements for this repo.

#### How to Verify
Check makefile tasks.  Check that there are no longer make tasks `make manual`, `make docs`, `make clean-manual`, `make clean-docs`

#### Copyright and Licensing

Columbia University


seantalts:
I think Bob should probably review this - I don't know anything about the docs.

mitzimorris:
@seantalts and @bob-carpenter - this needs to go in for 2.19 release.

mitzimorris:
@bob-carpenter or @VMatthijs or @seantalts please review so this can go in for 2.19

Bob and Matthijs - everything that was here has been captured to docs - just checked.

seantalts:
I thought we had decided not to do this at some point? How is the Stan reference going to stay in sync with the code?

mitzimorris:
the doxygen stuff is still there.
only the markdown files are gone.

seantalts:
What about the function reference?

mitzimorris:
the functions reference is in the `docs` repo - this is a consequence of putting all docs online.

do we want to roll the docs back into the mono repo?  totally doable...  docs urls become `mc-stan.org/stan/...`. instead of `mc-stan.org/docs`

seantalts:
I don't know, to me it seems like it would be easier to have the actual functions reference and anything else that's likely to be really sensitive to the version number in the repo w/ the code. I think we could still have a docs repo that just hosts stuff online? I don't know. Like what's the plan for releases if they stay in a different repo?

seantalts:
(the release plan is extremely important as there is a release tomorrow and I'm performing it and I have no idea what to do about docs now).

mitzimorris:
don't understand distinction between functions reference and other manuals.
for the release, we need to generate the 2_19 docs from the docs repo.
we can bundle the pdfs for the 2_19 docs with the release.
the srcs in the docs repo are up-to-date and ready to go for 2.19.  in particular, syntax for affine transform is there.
running the docs python build script with MAJOR/MINOR args 2 19 will generate pdfs.
once the release is out, we need to then go into the website - mc-stan.github.io and update the documentation page.
happy to help with all of these tasks.

seantalts:
I mean the functions manual and Stan language reference are different from the "user's guide" or whatever we're calling all of the sections on modeling, right? I think they might tie more closely to the version of Stan than the user's guide does. An example might be that we don't want to confuse someone running 2.16 into thinking they have access to "for each" loops, or something similar. I don't know how important that is. @bob-carpenter do you have a sense? 

My goals here are to 1) minimize incoming support requests 2) make it easy to tie a version to anything we release and 3) make releasing easier as time goes on, not harder. In mostly that order, I think. I'll also acknowledge that I don't have that good of a sense for how the docs are typically used, especially now that we're also releasing an HTML version.

mitzimorris:
the problem is that there are cross-references between the manuals.
at one point it was all the same LaTeX doc.
when we went to markdown things were split into 3 documents.
the 3 docs should be in the same repo - they share styling, bibliography, and again, cross-refs.

seantalts:
Do you think you could add a Jenkins stage that runs the script on the right repo when we merge to master on Stan?

mitzimorris:
OK, will try.

bob-carpenter:
> On Mar 17, 2019, at 6:13 PM, seantalts <notifications@github.com> wrote:
> 
> I mean the functions manual and Stan language reference are different from the "user's guide" or whatever we're calling all of the sections on modeling, right? I think they might tie more closely to the version of Stan than the user's guide does.

That's right.

The problem with tying the docs into the rest of the code is
that they now require not only R, but pandoc installed in
order to build them.  

> An example might be that we don't want to confuse someone running 2.16 into thinking they have access to "for each" loops, or something similar. I don't know how important that is. @bob-carpenter do you have a sense?

I thought Mitzi had some plan for release versioning
of the docs on the web site separate from the releases.


> My goals here are to 1) minimize incoming support requests 2) make it easy to tie a version to anything we release and 3) make releasing easier as time goes on, not harder. In mostly that order, I think. I'll also acknowledge that I don't have that good of a sense for how the docs are typically used, especially now that we're also releasing an HTML version.

I agree with your three priorities, but somewhere in the
mix needs to be (?) make contributing code easier, not harder.

> On Mar 17, 2019, at 6:19 PM, Mitzi Morris <notifications@github.com> wrote:

> the problem is that there are cross-references between the manuals.

I think detailed cross-references should be *eliminated* and replaced
with references to the top-level doc with no direct link other
than to the page of documentation on the web site.  

The reference manual should be self-contained.  

The functions reference can make a top-level citation of the reference 
manual.  

These two need to be kept up to date with releases.

The user's guide can make a top-level citation of the reference manual
or names of chapters.  This should ideally be kept up with
releases, but it's not critical.

The methodology "book" should be pulled out from the user's
guide.  It can make top-level citations to other things.
This one does *not* need to be kept up to date with new
releases and there's even an advantage in keeping it lagging
a bit for users who haven't installed new versions.

- Bob

seantalts:
> The problem with tying the docs into the rest of the code is
that they now require not only R, but pandoc installed in
order to build them.

Why is this a problem for tying them into the code? For Jenkins, I can dockerize the build now that I learned how to do this in a better way (that doesn't involve circleCI). That might also be an easy way for developers to make the docs, actually. Are there other issues involved there?

I agree and/or defer to you on everything else you've said. 

I know there is also the URL issue - I don't personally mind if the Stan user's guide is at `/stan` instead of `/docs`, though... and same for cmdstan at `/cmdstan`.

bob-carpenter:
> On Mar 19, 2019, at 4:43 AM, seantalts <notifications@github.com> wrote:
> 
> The problem with tying the docs into the rest of the code is
> that they now require not only R, but pandoc installed in
> order to build them.
> 
> Why is this a problem for tying them into the code? For Jenkins, I can dockerize the build now that I learned how to do this in a better way (that doesn't involve circleCI). That might also be an easy way for developers to make the docs, actually. Are there other issues involved there?

Sure, if you can make it work, that'd be fine by me. 

If we go back to requiring devs to keep the doc up to date with each PR, it'll require everyone to install R and pandoc.  No problem for me, as I have them installed.

Is the users guide going to move, too?  If so, that should wait until after the users guide and "book" are pulled apart.  We needed to keep the book in its own repo because it's *not* going to be kept up to date and Andrew's going to be messing with it in ways that violate all of our testing, coding style, etc., and we don't want that affecting the main stan-dev/stan repo.


> I agree and/or defer to you on everything else you've said.
> 
> I know there is also the URL issue - I don't personally mind if the Stan user's guide is at /stan instead of /docs, though... and same for cmdstan at /cmdstan.

I don't care either, but I do think if things move, we need to forward appropriately forever, or links go stale.

seantalts:
> If we go back to requiring devs to keep the doc up to date with each PR, it'll require everyone to install R and pandoc.

True, but that requirement can be independent of where the documentation lives, right? Though I think that having that requirement + putting it in a separate repo is operationally difficult. If we drop the update manual requirement (but who takes over?) and the manual release can be somehow automated I'd be just about as happy with that even if it lives in a different repo.

> Is the users guide going to move, too? 

That's in `docs` already right? That location makes sense to me.

Issue #2722, opened 2019-01-08
generated .hpp function `get_dims` wrong order of dimensions for arrays of vectors and matrices


mitzimorris:
#### Summary:
in file `src/stan/lang/generator/generate_dims_method.hpp` funcions `generate_var_dims` outputs variable dimensions in wrong order - it should list array dimensions first, then container dimensions.

#### Description:
Example:  for a 2-d array of matrices, array dimensions 4,5  matrix dims 2,3, the dimensions should be listed in this order: 4,5,2,3.  currently the listed order is 2,3,4,5.

A check of the interfaces shows that this function is exposed, but doesn't seem to be used, judging from lack of errors reported on the forums.

#### Reproducible Steps:
this issue was discovered as part of https://github.com/stan-dev/stan/compare/feature/2705-standalone-gqs-constrained-params-inputs.
this branch adds unit test: `src/test/unit/services/sample/standalone_gqs_multidim_test.cpp` which tests this.

#### Current Output:
for Stan program `src/test/test-models/good/services/gq_test_multidim.stan`, which has 2 output variables:
```
matrix<lower=0,upper=1>[2,3] p_ar_mat[4,5];
matrix[2,3] gq_ar_mat[4,5] = ar_mat;
```
generated code is:
```
   void get_dims(std::vector<std::vector<size_t> >& dimss__) const {
        dimss__.resize(0);
        std::vector<size_t> dims__;
        dims__.resize(0);
        dims__.push_back(2);
        dims__.push_back(3);
        dims__.push_back(4);
        dims__.push_back(5);
        dimss__.push_back(dims__);
        dims__.resize(0);
        dims__.push_back(2);
        dims__.push_back(3);
        dims__.push_back(4);
        dims__.push_back(5);
        dimss__.push_back(dims__);
    }
```

#### Expected Output:
```
   void get_dims(std::vector<std::vector<size_t> >& dimss__) const {
        dimss__.resize(0);
        std::vector<size_t> dims__;
        dims__.resize(0);
        dims__.push_back(4);
        dims__.push_back(5);
        dims__.push_back(2);
        dims__.push_back(3);
        dimss__.push_back(dims__);
        dims__.resize(0);
        dims__.push_back(4);
        dims__.push_back(5);
        dims__.push_back(2);
        dims__.push_back(3);
        dimss__.push_back(dims__);
    }
```

#### Additional Information:
(will be) fixed as par of PR for feature/2705

#### Current Version:
v2.18.1

[...]

Issue #2723, opened 2019-01-09
add constructor to class `stan/io/array_var_context` which takes as input an Eigen::Matrix vector of reals


mitzimorris:
#### Summary:
constructor method to class `stan/io/array_var_context` for real-valued vars only has params:
```
 /**
       * Construct an array_var_context from only real value arrays.
       *
       * @param names_r  names for each element
       * @param values_r a vector of double values for all elements
       * @param dim_r   a vector of dimensions
       */
```

add constructor methods where `values_r` argument is of type Eigen::VectorXd and Eigen::RowVectorXd

#### Description:
this would allow us to write methods for the standalone generated quantities feature which operate on Eigen matrices directly.  when working against csv output files using the `stan::io::csv_reader`, samples are read into an Eigen matrixXd.  the parameter values in each row must be copied into a `std::vector<double>`.  avoiding the copy step would be a win.

#### Current Version:
v2.18.1

Issue #2705, opened 2018-12-01
add logic to service `standalone_gqs` for constrained parameters as input


mitzimorris:
#### Summary:
add logic so that service `standalone_gqs` can handle constrained parameters as inputs.

#### Description:
the current implementation of the standalone generated quantities service `stan::service::standalone_gqs` takes as input values for the unconstrained parameters.  but the service is designed to work with a set of draws where the parameters are reported as constrained parameters.  we need to add the missing logic to transform the constrained parameters from the unconstrained to the constrained scale so that it will be easy for the interfaces to expose this feature.

#### Additional Information:
Provide any additional information here.

#### Current Version:
v2.18.0

mitzimorris:
implemented by PR #2725

PR #2725, opened 2019-01-11
Feature/2705 standalone gqs constrained params inputs

mitzimorris:
#### Submission Checklist

#### Summary

Changed services method `stan::services::standalone_generate` to take input drawset of constrained parameter values in the form of an `Eigen::MatrixXd`.

The essential processing steps are:

- assemble parallel arrays of parameter name, dimensions via calls to model methods `constrained_param_names`, `get_dims`

- verify that draw set columns match expected number of parameter values

- transform the constrained values to their corresponding unconstrained values
  + create a `var_context` object which contains the constrained parameters by calling `stan::io::array_var_context`
  + call the models `transform_inits` method passing in the `var_context` object.  the unconstrained parameter values are written to the `std::vector<double>` object `unconstrained_params_r`

- call the models `write_array` method passing in `unconstrained_params_r`

This PR fixes the following issues:  #2705, #2722, #2723 

Issue #2722 fixes an error introduced in the refactor of the generator; for n-arrays of type vector, row_vector, and matrix the `get_dims` method in the generated c++ model class lists the variable dimensions in the wrong order - the correct order is  array dimensions first, then the number of rows, columns.  It appears the none of the interfaces depend on the ordering - if this method is used anywhere, it is only used to determine the total number of elements for each variable.

Issue #2723 is a consequence of changing the input drawset from `std::vector<std::vector double> >` to `Eigen::MatrixXd`, which is how the a sample is stored in a `stan::io::stan_csv` object.

#### Intended Effect

Make it easy for the interfaces to call this method; all they need to do is marshall the parameter columns from the set of draws returned by the sampler.

#### How to Verify

Unit tests in `src/test/unit/services/sample`; also added tests to `src/test/io` for `array_var_context`.

#### Side Effects

Fixes bug #2722 

#### Documentation
N/A

#### Copyright and Licensing


src/test/unit/services/sample/standalone_gqs_bad_test.cpp
```diff
@@ -6,47 +6,44 @@
...
+  ServicesStandaloneGQ2() :
+    logger(logger_ss, logger_ss, logger_ss, logger_ss, logger_ss) {}
+
+  void SetUp() {
+    stan::io::empty_var_context context;
+    model = new stan_model(context);
+  }
+  
+  void TearDown() {
+    delete(model);
```
seantalts: Out of curiosity, why is this necessary?

src/test/unit/services/sample/standalone_gqs_bad_test.cpp
```diff
@@ -6,47 +6,44 @@
...
-  draws.push_back(draw1);
-  draws.push_back(draw2);
-  const std::vector<std::vector<double> > cdraws(draws);
-
-  int return_code = stan::services::standalone_generate(model,
-                                                        cdraws,
+  const Eigen::MatrixXd draws(1,1);
+  std::stringstream sample_ss;
+  stan::callbacks::stream_writer sample_writer(sample_ss, "");
+  int return_code = stan::services::standalone_generate(*model,
```
seantalts: Seems like this test should test that the test_gq2 model's print statement executes somehow, right?

src/test/unit/services/sample/standalone_gqs_multidim_test.cpp
```diff
@@ -0,0 +1,81 @@
...
+  std::stringstream sample_ss;
+  stan::callbacks::stream_writer sample_writer(sample_ss, "");
+  int return_code = stan::services::standalone_generate(*model,
+                                    multidim_csv.samples.middleCols<120>(7),
+                                    12345,
+                                    interrupt,
+                                    logger,
+                                    sample_writer);
+  if (return_code != stan::services::error_codes::OK)
+    std::cout << "ERROR: " << logger_ss.str() << std::endl;
```
seantalts: This needs to actually assert failure rather than just print a string

src/test/unit/services/sample/standalone_gqs_multidim_test.cpp
```diff
@@ -0,0 +1,81 @@
...
+  EXPECT_EQ(return_code, stan::services::error_codes::OK);
+  EXPECT_EQ(count_matches("gq_ar_mat",sample_ss.str()),120);
```
seantalts: Shouldn't you also test that the values produced for `gq_ar_mat` are correct?

src/test/unit/services/sample/standalone_gqs_test.cpp
```diff
@@ -2,159 +2,126 @@
...
-    : model(context, &model_log) {}
+  ServicesStandaloneGQ() :
+    logger(logger_ss, logger_ss, logger_ss, logger_ss, logger_ss) {}
+
+ void SetUp() {
+  std::fstream data_stream("src/test/test-models/good/services/bernoulli.data.R",
+                           std::fstream::in);
+  stan::io::dump data_var_context(data_stream);
+  data_stream.close();
+  model = new stan_model(data_var_context);
```
seantalts: PS Bravo figuring out how to run actual compiled model code and do gtest test fixtures 👍

src/test/unit/services/sample/standalone_gqs_test.cpp
```diff
@@ -2,159 +2,126 @@
...
-                                                        cdraws,
+  int return_code = stan::services::standalone_generate(*model,
+                                                        bern_csv.samples.middleCols<1>(7),
                                                         12345,
                                                         interrupt,
                                                         logger,
                                                         sample_writer);
-  // model test_gq.stan gen quantities block has 3 params:  xqg, y_rep.1, y_rep.2
+  if (return_code != stan::services::error_codes::OK)
+    std::cout << "ERROR: " << logger_ss.str() << std::endl;
```
seantalts: (all of these need to assert failure)

src/test/unit/services/sample/standalone_gqs_test.cpp
```diff
@@ -2,159 +2,126 @@
...
+    if (row == 0) {
+      ++row;
+      continue;
+    } 
+    if (row == 1001) {
+      break;
+    } 
+    qois.clear();
+    boost::algorithm::split(qois, line, boost::is_any_of(","));
+    sampler_qoi_ss.str(std::string());
```
seantalts: What does this line do?

src/test/unit/services/sample/standalone_gqs_test.cpp
```diff
@@ -2,159 +2,126 @@
...
-                                                        logger,
-                                                        sample_writer);
-  EXPECT_EQ(return_code, stan::services::error_codes::DATAERR);
-}
+  std::stringstream sampler_qoi_ss;
+  std::vector<std::string> qois;
+  std::istringstream f(sample_ss.str());
+  std::string line;    
+  size_t row = 0;
+  while (std::getline(f, line)) {
```
seantalts: All of this functionality here may be worth breaking out into a named function that helps encapsulate what it's doing. It seems here like you're trying to sort of parse part of a CSV?

src/test/test-models/good/services/gq_test_multidim.stan
```diff
@@ -0,0 +1,21 @@
+data {
+  matrix<lower=0,upper=1>[2,3] ar_mat[4,5];
+}
+parameters {
+  matrix<lower=0,upper=1>[2,3] p_ar_mat[4,5];
```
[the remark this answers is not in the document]
mitzimorris: the data block is a distraction.
I'm not sure what I was thinking when I set up this model.
removing data block, data file, and modifying test accordingly.

src/test/unit/services/sample/standalone_gqs_bad_test.cpp
mitzimorris: not sure what the scope of the question is - why is TearDown necessary?
https://www.ibm.com/developerworks/aix/library/au-googletestingframework.html
section on Test Fixtures:

> You can do deallocation of resources in TearDown or the destructor routine. However, if you want exception handling you must do it only in the TearDown code because throwing an exception from the destructor results in undefined behavior.

That said, it's here because I copied it from elsewhere in our unit test suite (advi?)

src/test/unit/services/sample/standalone_gqs_bad_test.cpp
mitzimorris: not relevant for this feature.
the only reason there's a print statement is to have a non-empty GQ block.

src/test/unit/services/sample/standalone_gqs_multidim_test.cpp
mitzimorris: that test is there on line 78 - the test will fail.
it was useful during development - if the test fails, you get immediate feedback on why.
if this is a problem for Travis or other CI tools, I can remove it.

src/test/unit/services/sample/standalone_gqs_test.cpp
mitzimorris: lines 91, 92 written per this SO answer:
https://stackoverflow.com/questions/834622/best-way-to-empty-stringstream

src/test/unit/services/sample/standalone_gqs_bad_test.cpp
seantalts: I guess I'm not sure why we need to delete the model - does its destructor not get automatically called when a new one is created?

src/test/unit/services/sample/standalone_gqs_bad_test.cpp
seantalts: What's this test testing?

src/test/unit/services/sample/standalone_gqs_multidim_test.cpp
seantalts: Ah, I see. I would have expected the testing framework to provide the exact same service...

src/test/unit/services/sample/standalone_gqs_test.cpp
mitzimorris: all of them test the return code:

`EXPECT_EQ(return_code, stan::services::error_codes::<appropriate-code-for-test>);`

src/test/unit/services/sample/standalone_gqs_multidim_test.cpp
seantalts: Probably best to just take them out and use the testing framework to figure out where test failures happen.

src/test/unit/services/sample/standalone_gqs_bad_test.cpp
mitzimorris: https://github.com/stan-dev/stan/blob/1812edf5274792006ed104388cc4a1d9a469af3d/src/stan/services/sample/standalone_gqs.hpp#L86

don't bother to run standalong generated quantities if no variables declared in the gq block.

src/test/unit/services/sample/standalone_gqs_multidim_test.cpp
mitzimorris: will do.

src/test/unit/services/sample/standalone_gqs_test.cpp
mitzimorris: added function to file `src/test/unit/util.hpp` called `match_csv_columns`.

src/test/unit/services/sample/standalone_gqs_multidim_test.cpp
mitzimorris: added that test, using newly refactored utility method `match_csv_columns`.

src/test/unit/services/sample/standalone_gqs_bad_test.cpp
syclik: If you're calling `new`, you should call `delete`. I understand that this is just test code, but there's no reason we should be leaking memory.

@seantalts, I believe the `SetUp()` method is setting up a model by calling `new` because we don't have default constructor for the model. So we're actually allocating memory for the object as we construct it.

seantalts: Ah okay, that makes sense. Sorry about that. I didn't know that trick and was assuming `new` was just a deprecated way of calling the default constructor, haha (we don't use it much!). My bad.

syclik: nah... it's a good question. For our use on the autodiff stack, we redefine `operator new` so it grabs memory from our managed memory. Here, we're not using that `new`.

We don't do much with pointers for this reason... we need to keep track of every exit and make sure we've dealt with the memory properly. I haven't checked into C++11 initialization... there may be a way to hold onto the memory and then later initialize with the constructor without a default constructor. I'm not sure, though. (I thought I saw that while looking at the testing framework.)

mitzimorris: re-instated TearDown fns, per this discussion.

PR #2730, opened 2019-01-24
make read_stan_csv more robust; affects CmdStan.

### LINKED ISSUE #2729: read_stan_csv has issues reading valid output due to boost::lexical_cast exception
#### Summary:
CmdStan fails to read properly generated csv files in `bin/stansummary` (sometimes).

#### Description:
We currently use `boost::lexical_cast` to convert from a string to C++ primitive. It turns out this is not very robust.

An isolated case that throws an exception:
```
double d = boost::lexical_cast<double>("6.77055e-309");
```

this will work:

```
double d;
std::stringstream("6.77055e-309") >> d;
```

We currently use `std::stringstream` to output the values, so this might be a better way to parse.

#### Reproducible Steps:

I was able to reproduce using CmdStan and the eight_schools example. Rerunning that example repeatedly and then trying to run `bin/stansummary` on the csv file will eventually create an example csv file that will not be parsed by `bin/stansummary`.

#### Current Output:
From CmdStan:
```
$ bin/stansummary output.csv
libc++abi.dylib: terminating with uncaught exception of type boost::exception_detail::clone_impl<boost::exception_detail::error_info_injector<boost::bad_lexical_cast> >: bad lexical cast: source type value could not be interpreted as target
Abort trap: 6
```

This is due to one of the values in the csv file being `6.77055e-309`.

#### Expected Output:

Normal output from `stansummary`. Something like...
```
Inference for Stan model: eight_schools
1 chains: each with iter=(1000); warmup=(0); thin=(1); 1000 iterations saved.

Warmup took (0.046) seconds, 0.046 seconds total
Sampling took (0.066) seconds, 0.066 seconds total

                    Mean     MCSE   StdDev     5%       50%   95%  N_Eff  N_Eff/s    R_hat
lp__            -3.9e+01  1.5e-01  2.6e+00    -44  -3.9e+01   -36    287     4339  1.0e+00
accept_stat__    8.8e-01  7.5e-03  2.0e-01   0.41   9.6e-01  1.00    696    10524  1.0e+00
stepsize__       3.1e-01  4.7e-16  3.3e-16   0.31   3.1e-01  0.31   0.50      7.6      nan
treedepth__      3.6e+00  1.9e-02  5.2e-01    3.0   4.0e+00   4.0    724    10962  1.0e+00
...
```

#### Additional Information:
I've isolated the code to `src/stan/io/stan_csv_reader.hpp`

#### Current Version:
v2.18.1

syclik:
#### Submission Checklist

#### Summary

This fixes #2729.

This makes reading Stan csv files a little more robust. It turns out that `boost::lexical_cast<>` isn't always able to parse what's written out by `std::stringstream`.

This PR also fixes a small bug where the `chain_id` was not being read from the output file and always set to `1`. The tests changed to reflect the fix.

#### Intended Effect

Make CmdStan's `bin/stansummary` a little more robust.

#### How to Verify

There's a new test in `src/test/unit/io/stan_csv_reader_test.cpp`. To run it:

```
./runTests.py src/test/unit/io/stan_csv_reader_test.cpp
```

This reads from a Stan csv file that was generated using CmdStan for the eight schools example. The csv file is located at `src/test/unit/io/test_csv_files/eight_schools.csv`.

If you run the test without the changes, it will fail. With the changes, it parses it fine.

#### Side Effects

It's possible that this fails silently. There's different behavior with using `std::stringstream` as opposed to `boost::lexical_cast`.

When the generated csv is valid, the current implementation will be better.

When the generated csv is completely junk, this implementation will silently parse the file whereas the old implementation would fail with a boost exception. I think it'd be completely obvious when it's failed, but it would still complete the process.

#### Documentation

None. I just replaced uses of `boost::lexical_cast` with calls to `std::stringstream`.

#### Copyright and Licensing

Generable


syclik:
@seantalts or @mitzimorris, have a few minutes to review this or assign to someone else? (I'm tagging you cause it's a CmdStan / Stan thing.)

src/test/unit/io/stan_csv_reader_test.cpp
```diff
@@ -64,7 +65,7 @@ TEST_F(StanIoStanCsvReader,read_metadata1) {
   EXPECT_FALSE(metadata.save_warmup);
   EXPECT_EQ(4085885484U, metadata.seed);
   EXPECT_FALSE(metadata.random_seed);
-  EXPECT_EQ(1U, metadata.chain_id);
+  EXPECT_EQ(0U, metadata.chain_id);
```
syclik: This was another bug that I found while fixing this. It turned out we weren't reading the `chain_id` properly from the header info.

PR #2766, opened 2019-06-09
Faster indexing

seantalts:
#### Submission Checklist

- [] Run unit tests: `./runTests.py src/test/unit`

#### Summary
Since we're already doing our own bounds-checking we can avoid it in `rvalue` (and probably many other places).

#### Intended Effect
Performance improvement

#### Copyright and Licensing

Columbia University

### DIFF
--- a/src/stan/model/indexing/rvalue.hpp
+++ b/src/stan/model/indexing/rvalue.hpp
@@ -22,7 +22,7 @@ namespace stan {
      * a nil index list, which just returns the value.
      *
      * Types:  T[] : T
-     * 
+     *
      * @tparam T Scalar type.
      * @param[in] c Value to index.
      * @return Input value.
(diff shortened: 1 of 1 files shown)


syclik:
So... the `get_base1()` function is slower than direct indexing using `[]` or `()`. I don't know what to do about it due to:
- if we don't pre-check, we could actually seg fault. In CmdStan, it's not such a problem -- the program's dead at that point anyway. In RStan or PyStan, this could cause R / Python to crash without recovery.

Should we introduce a different function that doesn't check bounds that's unsafe and expects the user to do the right thing?

seantalts:
How is `get_base` different from `rvalue`, do you know? I haven't been using it in the new compiler yet at all so it might become a non-issue soon enough.

syclik:
I don't know off the top of my head. I don't even know if there is a difference between get_base1 and get_base

PR #2785, opened 2019-07-08
Feature/2759 model base

bob-carpenter:
#### Submission Checklist

#### Summary

Add `stan::model::model_base` base class and and `stan::model::model_crtp` static adapter class for constructing models.   

`model_base` defines pure virtual methods for all functionality used by the services and algorithms.

Model classes are now generated to extend `model_crtp` rather than `prob_grad`.  They also define a factory function to create new instances.  The implementation of the CRTP ensures that all the existing template functions *and* virtual functions are implemented in `model_base` by delegating to the implementation's template function.  (There's an exmaple in the tests for `model_crtp`.)

#### Intended Effect

Speed up compile time.  Specifically, allow functions like the services to be instantiated with  `model_base` and compiled in their own translation units.  Then models are compiled in their own translation unit and linked.

#### How to Verify

Check the matching branch in `stan-dev/cmdstan` that implements CmdStan using the new pattern . You can verify there that there is no slowdown.

Otherwise, there are unit tests for the instantiation and inheritance of the base classes and all other tests still pass.

#### Side Effects

This change maintains backward compatibility in that no functionality is lost from generated model code, which can still be used as before.  All template programs were left intact---they can simply be instantiated with `base_model` rather than being hard coded for `base_model`.  

The main side effect is that all model classes `foo` now contain virtual functions where there used to be non-virtual functions.  As none of these are in tight loops, there is negligible effect on speed.

The `model_name()` function was changed from `static` to a virtual member function so that it could be used as part of the base model specification.  Nothing in the services use the model name, but it might get used in RStan, PyStan, etc.  

#### Documentation

Yes.  The goal is to have the `model_base` fully documented so that clients can understand Stan's requirements for models.

Also added documentation for `prob_grad` that was previously missing.  

#### Copyright and Licensing

Columbia University


bob-carpenter:
> Result: 0.98587122603

Does this mean it's a bit slower?  There's an outlier, so now I'm curious about aggregation method:

> stat_comp_benchmarks/benchmarks/arK/arK.stan, 0.78)

To save everyone the trouble of looking it up, there's nothing special about this particular Stan program:

```
data {
  int<lower=0> K;
  int<lower=0> T;
  real y[T];
}
parameters {
  real alpha;
  real beta[K];
  real<lower=0> sigma;
}
model {
  alpha ~ normal(0, 10);
  beta ~ normal(0, 10);
  sigma ~ cauchy(0, 2.5);
  for (t in (K+1):T) {
    real mu;
    mu = alpha;
    for (k in 1:K)
      mu = mu + beta[k] * y[t - k];
    y[t] ~ normal(mu, sigma);
  }
}
```

wds15:
Yes... values smaller than 1 mean its slower...bigger than 1 and you are speeding up. 0.78 for ark seems a lot (but that model runs super fast).

bob-carpenter:
> Yes... values smaller than 1 mean its slower

Thanks.


> 0.78 for ark seems a lot (but that model runs super fast).

I can't imagine it runs faster per log density eval than eight schools.  I just kicked off test again after merging latest math updates.  I should've just restarted the benchmark rather than pushing.  Too late now, I think.

bob-carpenter:
I ran 20 iterations of the old and new model code for `arK.stan` from the benchmarks repo and here are the results:

```
# OLD
2.79759, 2.81587, 2.80003, 2.80729, 2.81272, 2.80357, 2.80182, 2.83747, 2.84059, 2.79978, 2.82007, 2.78515, 2.80253, 2.84306, 2.80799, 2.81918, 2.81565, 2.84919, 2.82136, 2.77996
# NEW
2.75693, 2.80307, 2.82399, 2.82531, 2.81886, 2.78611, 2.82901, 2.80264, 2.77021, 2.77916, 2.76571, 2.78418, 2.79107, 2.81126, 2.75938, 2.8003, 2.82604, 2.79684, 2.83821, 2.80198
```

bob-carpenter:
Hmm.  The outlier is still an outlier despite not having different run time on my machine (Mac OS X, clang++).   I have no idea why separate translation units would slow down this specific Stan program on one platform and not another.

How are these tests being run and on what platform?  Is the config somewhere public?

seantalts:
You can see the config here: https://github.com/stan-dev/performance-tests-cmdstan/blob/master/Jenkinsfile#L102  which mostly sets up and calls https://github.com/stan-dev/performance-tests-cmdstan/blob/master/compare-git-hashes.sh.

It runs on the Mac pro (which does sometimes show performance degradations other Macs don't) with clang++-6.0. arK seems to run *very* fast and can have somewhat variable timing, though it got exactly the same ration the past two times, so who knows. 

I'm pretty confused though - this had equal compilation speed to `develop`, so that suggests to me it isn't actually testing the full thing, right? Or does that not matter?

bob-carpenter:
> this had equal compilation speed to develop, so that suggests to me it isn't actually testing the full thing, right?

This PR just adds the virtual base class and changes the way code is generated to use it.  The speedup will come in CmdStan when the services can be compiled in their own translation units.  That'll also need to be benchmarked.

> with clang++-6.0. arK seems to run very fast and can have somewhat variable timing,

I found very stable timing (as shown in the previous comment) running a single chain with the same seed.

> Mac pro (which does sometimes show performance degradations other Macs don't) 

Is the Mac pro buggy, or are those degradations from our code?

seantalts:
In the one case I saw (one of the TLS PRs) it showed a degradation that didn't show up on other macs (but maybe did on Linux with GCC?) when the code was changed to use a pointer instead of a reference, I believe. So I think the code was all as intended, the debate was about whether we cared about the performance on that CPU which maybe doesn't support some instruction or otherwise doesn't trigger optimal codegen from LLVM.

I think I was thinking of `arma` that had more variable performance, that one only takes 600ms. arK seems to take about 2000ms, which is a bit longer. 

Perhaps the debate is again over whether we care about the old Mac Pro CPU. I personally care much more about the total time it takes to execute a model, which would mean that this is still a massive win for arK as its compilation time decreases by something like 25 seconds while its runtime increases by only about 400ms. I would argue (perhaps in a separate thread) that this is the metric we should all be considering (well, the most-measurable one and most applicable here - perhaps the best one is actually "total time to develop and run a Stan model," which is affected by a large multiplier of write-compile-run-debug feedback loop speed).

bob-carpenter:
I don't care much about that CPU, but it may be indicative of others.

As for total time, I think it depends on use case.  If you have something like a precompiled model that everyone uses (rstanarm, prophet, etc.), then you really don't care much about compilation time.  And if it's in some kind of server setting, you probably do care about performance, but maybe not at a 1% or 2% level except for enormous tasks.

seantalts:
True, rstanarm definitely deserves its own performance tests! cc @bgoodri and @jgabry :P 

I think the wisdom in most of the software engineering shops I've worked in is that deployed workload resource consumption is worth very little compared with consuming the developer's time, even for models that go into production. If we end up with specific models that need to be faster in a corporate production environment, I think I and others would welcome the consulting income or github PRs to deal with that, haha. Of course there's a balance to be struck - I would just argue we could move a bit towards faster development and at the expense of more runtime performance than we are losing here.

bob-carpenter:
Also no difference on GCC 6 (other than it runs in 2.3s in GCC 6 compared to 2.7s in clang from Xcode).

OLD (GCC 6):
```
2.33129, 2.30224, 2.51232, 2.32546, 2.32826, 2.32524, 2.34958, 2.29206, 2.34273, 2.33053, 2.30137, 2.52292, 2.29986, 2.29951, 2.31221, 2.30466, 2.33785, 2.32555, 2.29266, 2.31779
```

NEW (GCC 6):
```
2.31826, 2.35278, 2.36406, 2.32456, 2.32519, 2.3421, 2.3993, 2.36443, 2.30421, 2.3311, 2.31127, 2.34374, 2.38383, 2.32886, 2.3493, 2.33989, 2.35524, 2.38073, 2.35985, 2.32206
```

seantalts:
Sounds good to me.

bob-carpenter:
This is ready to review.

bob-carpenter:
This is ready to review now.

syclik:
@bob-carpenter, it looks like the compile process has changed for the tests in Stan. It seems like we would want to do it for multiple reasons:
1. be able to verify this works without needing CmdStan
2. catch problems if future changes break this (I can't see how, but what can break will)
3. speed up compilation times for Stan tests. (!!!). The integration tests would all get faster, right?

I can help sort that out unless @mitzimorris wants to.

bob-carpenter:
> On Jul 10, 2019, at 11:06 AM, Daniel Lee <notifications@github.com> wrote:
> 
> @bob-carpenter, it looks like the compile process has changed for the tests in Stan.

I didn't change anything in stan-dev/stan around building models.
Everything should remain completely backward compatible.

> It seems like we would want to do it for multiple reasons:
> 
> 	• be able to verify this works without needing CmdStan

Verify what works?  I didn't remove any existing tests.  It's just
as likely to keep working as what we had before. 

> 	• catch problems if future changes break this (I can't see how, but what can break will)

Break what?  

> 	• speed up compilation times for Stan tests. (!!!). The integration tests would all get faster, right?
> I can help sort that out unless @mitzimorris wants to.

I'd rather wait until after 2.20 so it doesn't slow down this PR.

mitzimorris:
I agree with Bob that getting this PR in for 2.19 should take priority.
especially since at some point the mono-repo should remove the issue of whether to test via CmdStan or Stan.  that said, I would be happy to do a follow-on PR to address makefile issues.

wds15:
I looked over this code and overall I think the approach is sound and makes sense. What I struggle with most in my head is the naming of `model_base` and `model_crtp`. On one hand I dislike naming things like a programming pattern, but maybe that's fine (wikipedia does the same in their example on CRTP). However, where we could maybe improve is to name the crtp bit `mode_base_crtp`. This way it makes it clear that this is still a `model_base` - it's just the CRTP part of it.

The compilation times indeed speedup seriously as I have verified on the cmdstan branch of this which goes with it. I would be ok to me with having this PR going into Stan and not take advantage of the speedups in the Stan repository in terms of faster test compilation times; we can address that later or just wait for the monorepo.

The restriction which we gain out of this approach is that any templated function which needs to be part of a model will have to be put into the CRTP thing where we have to flesh out all template variations.

I am happy to take another deep dive into this later today. What makes me a little hesitant to waive this PR is that I never have worked on Stan services / etc.. So I do lack a holistic overview probably. Do others think this is a problem or can someone else also take a look?

seantalts:
Maybe @betanalpha can take a quick look at the services part?

betanalpha:
The services aren’t modified in this PR — everything 
still templates on a model instance, only now the model 
instance that instantiates the template is a wrapper 
around the base and CRTP classes.  No?

wds15:
@betanalpha I think the answer is yes. The way I got is that we have now

`model_base` : virtual only class which has `log_prob` and `write_array` spelled out in all the different variations admitted by our templates
`model_crtp` : This thing is the base class for model instances. It provides an actual implementation of the virtual functions using the templated functions from the model instance itself.

So nothing changes for the services. It's just that adding templated functions to the model instance class gets a bit more tedious. 

At least this is my understanding.

bob-carpenter:
> On Jul 15, 2019, at 4:57 AM, Michael Betancourt <notifications@github.com> wrote:
> 
> The services aren’t modified in this PR — everything 
> still templates on a model instance, only now the model 
> instance that instantiates the template is a wrapper 
> around the base and CRTP classes. No?

Yes.  It's still backward compatible in the sense that you can still template on a more specific instance.  All the calls still work through the templated versions, which are what the model_crtp makes available.

bob-carpenter:
I'm OK changing the name to model_base_crtp or model_base_impl (but it's not a full implementation).

wds15:
`model_base_crtp` is great.

src/test/unit/model/model_base_crtp_test.cpp
```diff
@@ -0,0 +1,147 @@
...
+  EXPECT_FLOAT_EQ(4, bm.log_prob_jacobian(params_r_v, msgs).val());
+  EXPECT_FLOAT_EQ(5, bm.log_prob_propto(params_r, msgs));
+  EXPECT_FLOAT_EQ(6, bm.log_prob_propto(params_r_v, msgs).val());
+  EXPECT_FLOAT_EQ(7, bm.log_prob_propto_jacobian(params_r, msgs));
+  EXPECT_FLOAT_EQ(8, bm.log_prob_propto_jacobian(params_r_v, msgs).val());
+
+  // test template version from base class reference
+  // long form assignment avoids test macro parse error with multi tparams
+  double v1 =  bm.template log_prob<false, false>(params_r, msgs);
+  EXPECT_FLOAT_EQ(1, v1);
```
wds15: You could use extra brackets to avoid the assignment like this:

`EXPECT_FLOAT_EQ(1, (bm.template log_prob<false, false>(params_r, msgs)));`

Same applies to the model_base_test.cpp

src/stan/model/prob_grad.hpp
```diff
@@ -2,46 +2,91 @@
...
-     * The <code>prob_grad</code> class represents the basic parameter
-     * holders for a model.  The command <code>bin/stanc</code> builds
-     * Models extending this base helper class.
+     * Base class for models, holding the basic parameter sizes and
+     * ranges for integer parameters.
      */
     class prob_grad {
-    protected:
+     protected:
+      // TODO(carpenter): roll into model_base; remove int members/vars
```
wds15: This is not related to this PR, but rather to my ingorance... but why is the number of parameters in a model not declared `const`? These never change.

src/stan/model/model_base_crtp.hpp
```diff
@@ -0,0 +1,216 @@
...
+  /**
+   * Destroy this class.  This is required to be virtual to allow
+   * subclass references to clean up superclasses, but is otherwise a
+   * no-op.
+   */
+  virtual ~model_base_crtp() { }
+
+  inline double log_prob(Eigen::VectorXd& theta,
+                         std::ostream* msgs) const override {
+    return static_cast<const M*>(this)
```
wds15: Would it be possible to declare a reference to the parent class once? So you could have the member

`const M& model_;`

This would be initialized with

`model_(static_cast<const M&>(*this))`

With that you can avoid all those `static_cast` calls and simply do `model_.template ...`.

That's easier to read to me. Does that make sense?

bob-carpenter:
Comments inline.

> I have one more question: Do we still need the using stan::model::prob_grad; line in the generated hpp files? 

No, we can get rid of it.  It's a one-line change.


> +  double v1 =  bm.template log_prob<false, false>(params_r, msgs);
> +  EXPECT_FLOAT_EQ(1, v1);
> 
> You could use extra brackets to avoid the assignment like this:
> 
> EXPECT_FLOAT_EQ(1, (bm.template log_prob<false, false>(params_r, msgs)));

Cool!


> Same applies to the model_base_test.cpp
> 
> In src/stan/model/prob_grad.hpp:
> 
> >       */
>      class prob_grad {
> -    protected:
> +     protected:
> +      // TODO(carpenter): roll into model_base; remove int members/vars
> 
> This is not related to this PR, but rather to my ingorance... but why is the number of parameters in a model not declared const? These never change.

Because they get set in a method.  To be constant, they need to be set in an initializer, so that'd have to be something like num_params_r_(...). But we can't calculate the number of parameters until we read the data in, and that's not constant because it can't be easily broken down into initializers.


> In src/stan/model/model_base_crtp.hpp:
> 
> > +   *
> +   * @param[in] num_params_r number of real unconstrained parameters
> +   */
> +  explicit model_base_crtp(size_t num_params_r) :
> +      model_base(num_params_r) { }
> +
> +  /**
> +   * Destroy this class.  This is required to be virtual to allow
> +   * subclass references to clean up superclasses, but is otherwise a
> +   * no-op.
> +   */
> +  virtual ~model_base_crtp() { }
> +
> +  inline double log_prob(Eigen::VectorXd& theta,
> +                         std::ostream* msgs) const override {
> +    return static_cast<const M*>(this)
> 
> Would it be possible to declare a reference to the parent class once? So you could have the member
> 
> const M& model_;
> 
> This would be initialized with
> 
> model_(static_cast<const M&>(*this))
> 
> With that you can avoid all those static_cast calls and simply do model_.template ....
> 
> That's easier to read to me. Does that make sense?

I'd rather not change this as the way I'm doing it is the idiomatic way to use the CRTP.  Adding a model_ member variable adds another pointer of memory, which is probably no big deal here, but is not standard.  I imagine it's more of a problem for register allocation and optimization as we now have two pointers, this and model_, that are the same pointer.

src/stan/model/model_base_crtp.hpp
wds15: Ok... if that's the idiomatic CRTP way.. fine.

betanalpha:
Yeah, none of this touches the services or algorithms code.

src/stan/model/model_base.hpp
```diff
@@ -0,0 +1,612 @@
...
+ * <p><i>Implementation Details:</i> The reason there are so many
+ * overloads of the `log_prob` and `write_array` methods is that
+ * template methods cannot be declared virtual.  This class extends
+ * `stan::model::prob_grad` in order to define sizing for the number
+ * of unconstrained parameters;  thus it is not a pure virtual base
+ * class.
+ *
+ *<p>The approach to defining models used by the Stan language code
+ * generator is use the curiously recursive template base class defined
+ * in the extension `stan::model::model_base_crtp`; see that class's
```
syclik: minor: you can drop the text after the semicolon. Here's what the rendered doxygen looks like:
image

(The `stan::model::model_base_crtp` link takes you to the right place.)

src/stan/model/model_base_crtp.hpp
```diff
@@ -0,0 +1,216 @@
...
+ * methods to their untemplated virtual counterparts declared in
+ * `model_base`.
+ *
+ * The derived class `M` is required to implement the following two
+ * pairs of template functions,
+ *
+ * ```
+ * template <bool propto, bool jacobian, typename T>
+ * T log_prob(std::vector<T>& params_r,
+ *              std::vector<int>& params_i,
```
syclik: minor: mind lining up the `std::vector`s? (remove two spaces from L24 and L25)

Untitled

src/stan/model/model_base_crtp.hpp
```diff
@@ -0,0 +1,216 @@
...
+ *
+ * <p>The derived class `M` must be declared following the curiously
+ * recursive template pattern, for example, if `M` is `foo_model`,
+ * then `foo_model` should be declared as
+ *
+ * ```
+ * class foo_model : public stan::model::model_base_crtp<foo_model> { ... };
+ * ```
+ *
+ * The so-called "curious recursion involves the class type that is
```
syclik: Typo: missing a double quotation mark. ("curious recursion")

src/stan/model/model_base_crtp.hpp
```diff
@@ -0,0 +1,216 @@
...
+ * template <bool propto, bool jacobian, typename T>
+ * T log_prob(Eigen::Matrix<T, -1, 1>& params_r,
+ *            std::ostream* msgs = 0) const;
+ * ```
+ *
+ * and
+ *
+ * ```
+ *  template <typename RNG>
+ *    void write_array(RNG& base_rng,
```
syclik: Also alignment here... delete two spaces in L36-42.

image

src/stan/model/model_base_crtp.hpp
```diff
@@ -0,0 +1,216 @@
...
+ *                     std::vector<double>& params_r,
+ *                     std::vector<int>& params_i,
+ *                     std::vector<double>& vars,
+ *                     bool include_tparams = true,
+ *                     bool include_gqs = true,
+ *                     std::ostream* msgs = 0) const;
+ *
+ *   template <typename RNG>
+ *    void write_array(RNG& base_rng,
```
syclik: Remove a space from L45 to L50.

syclik:
Also... I tried to get the compile times down for the tests. It didn't look like there was anything obvious that was fully linked, so I don't think there's anything to do. (@bob-carpenter, you were right... my thoughts on making things faster were misguided and you probably saw that when I asked the question.)

wds15:
Looks like we are all good. Addressing @syclik comments would be great... and I can approve once more when you touch the code again.

src/stan/lang/generator/generate_class_decl.hpp
```diff
@@ -18,7 +18,9 @@ namespace stan {
      */
     void generate_class_decl(const std::string& model_name,
                              std::ostream& o) {
-      o << "class " << model_name << " : public prob_grad {" << EOL;
+      o << "class " << model_name
+        << " : public stan::model::model_base_crtp<" << model_name << "> {"
```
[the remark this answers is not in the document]
bob-carpenter: Done.

src/stan/model/model_base.hpp
bob-carpenter: Done.

src/stan/model/model_base_crtp.hpp
bob-carpenter: Done.

src/stan/model/model_base_crtp.hpp
bob-carpenter: fixed.

src/stan/model/model_base_crtp.hpp
bob-carpenter: done

src/stan/model/model_base_crtp.hpp
bob-carpenter: done.

bob-carpenter:
I updated the doc as @syclik requested.  Is there more you'd like me to do here, @wds15?

syclik:
Thanks, @bob-carpenter! And @wds15 for reviewing.

Issue #2776, opened 2019-06-26
stan::io::json parser is 10x slower than python's ujson


seangoogle:
#### Summary:
I dumped a 300k x 100 matrix into a json file and had python's ujson load it (5.6s) and then loaded it in CmdStan (67.3s). Seems like we could switch to a 3rd party json parser or another format for better speed. 

For comparison, reading the equivalent .data.R rdump file takes 36s (still not great).

#### Current Version:
v2.19.1

bob-carpenter:
Is that using the same JSON format for output?  

Is that just I/O or does it include conversion to some Python data format?  That is, is the bottleneck in our I/O or in our var_context?  

Is it something we can pull out a C++ implementation from and use in Stan (license and interface)?

bob-carpenter:
Thanks for testing, by the way---it seems rude now that I look at it to just fire back a bunch of cross-exam questions!

seangoogle:
I'm not sure I understand your first question. It's the same exact file. 

The ujson module includes conversion into Python dictionaries and lists, etc. I don't know where the Stan bottleneck is.

There are many fast C++ json parsers. I suspect some of them could be included. I don't have opinions on which ones to use.

bob-carpenter:
There weren't when we wrote this one.

> On Jul 3, 2019, at 1:00 PM, seangoogle <notifications@github.com> wrote:
> 
> I'm not sure I understand your first question. It's the same exact file.

Yes.  That answered my question.


> The ujson module includes conversion into Python dictionaries and lists, etc. I don't know where the Stan bottleneck is.

My guess is either conversion to Python or the basic ASCII to double conversion, because profiling showed our C++ code spends almost all of its time in ASCII to double conversion.


> There are many fast C++ json parsers. I suspect some of them could be included. I don't have opinions on which ones to use.

We couldn't find any back in 2014 or so when Mitzi and I wrote the event-based one that's there now.  All the open-source ones we could find built DOM-like trees and had very limited handling of floating point.

rok-cesnovar:
Just throwing a few references here: 

- this benchmark covers a lot of the libraries https://github.com/miloyip/nativejson-benchmark

- this paper by Microsoft researchers claim that Mison can parse to speeds up to 2GB/s http://www.vldb.org/pvldb/vol10/p1118-li.pdf
[...]

PR #2793, opened 2019-07-27
User RapidJSON to parse JSON input

rok-cesnovar:
#### Submission Checklist

- [x] Declare copyright holder and open-source license

#### Summary

This PR addresses issue #2776 and is still a work in progress. Its a WIP because I have some one questions before I do a proper cleanup and add missing error checks and unit tests. Basically I would like to get a green light from everyone with an opinion on this.

What this PR currently does: 
- speeds up JSON parsing (for 200MB file we get from 35s to 2.5s, for 2GB we get from 350 to 25s)
- JSON syntax error checking (throws on JSON syntax error, same as the old parser)

What is still missing: 
- ~~add stan specific error checking (RapidJSON already does JSON syntax checking, but things like: dont allow nested objects, JSON should start as an object or array, etc.~~ (added this)
- extract the line where the error in JSON happend (can currently only get offset from start)
- ~~unit tests (do we need to add any or would the ones for the existing parser suffice?)~~ (They suffice)
- ~~remove unnecessary files from the RapidJSON folder~~

Open questions:
- ~~what would be the right location for this lib? (if we decide this is going to move to cmdstan this is less of an issue here)~~ (in `/lib`, see discussion)
- ~~is MIT license OK for a 3rd party library?~~ Yes
- ~~should this be used only when a user specifies a flag? If not, should this just replace the old parser?~~ (should replace the existing parser)

All the relevant changes are in `src/stan/io/json/rapidjson_parser.hpp` where the callbacks from the RapidJSON parser call the existing data handler in the same way as the old parser.

#### Intended Effect

Speedup JSON parsing in cmdstan.

#### How to Verify
 
Switch Stan to the feature/rapidjson branch and run your model with JSON data input.

#### Side Effects

There should not be any, but this is a WIP. ~~Tests on memory consumption should definitely be done.~~ (see post below for memory test)

#### Copyright and Licensing

Please list the copyright holder for the work you are submitting (this will be you or your assignee, such as a university or company): Rok Češnovar, University of Ljubljana


rok-cesnovar:
These are the results of my memory usage test. I logged the memory usage of the cmdstan executable every 50ms using `logglm () { while sleep 0.05; do ps -p  `pidof glm` -o pcpu= -o pmem= ; done; }` (glm was the name of the executable). This also tracked the memory usage after the I/O.

mem_usage

Judging by this, I think there arent any differences in the memory usage of the existing parser and RapidJSON.

seantalts:
>is MIT license OK for a 3rd party library?

We already have OpenCL, which is MIT, so I'd say yes.

> should this be used only when a user specifies a flag? If not, should this just replace the old parser?

I imagine it would just replace the old one - @bob-carpenter can you comment on that? Was there anything special about the old JSON parser that we'd need to replicate?

> what would be the right location for this lib? (if we decide this is going to move to cmdstan this is less of an issue here)

I think the `lib` folder makes sense to me. Though I'm not sure why the JSON stuff is in Stan and not CmdStan architecturally.

rok-cesnovar:
> Was there anything special about the old JSON parser that we'd need to replicate?


Compared to how RapidJSON acts by default I had to make the following adjustments to be consistent with the existing parser(no changes to the RapidJSON code to be clear):

- turn on the so-called relaxed-JSON parsing (NaN and Inf are not allowed by the strict JSON standard)
- modified all the error messages so that they match the existing parser
- the existing parser does not treat a JSON with 2+ objects on the top level (`{ "a": 5}{"b": 6}`) as an error, just takes the first one; RapidJSON treats this as a syntax error, so now after we reach the end of a top level object we stop parsing manually to get the same behaviour. I would suggest treating it as a syntax error, but this isnt a strong opinion

> Though I'm not sure why the JSON stuff is in Stan and not CmdStan architecturally.

Bob mentioned this in the issue thread yes. Do we want to merge this first and move the code after or vice-versa. I can volunteer to do that.

seantalts:
That all sounds good to me. What's the move look like? Sort of feels like we might as well just add the new stuff to CmdStan and then delete the old stuff from Stan in two separate PRs.

rok-cesnovar:
In the Stan meeting @mitzimorris said that the JSON parser should live inside the Stan repository. I am not going to attempt to give an exact quote (dont want to get anything wrong) on the reasons for that. So I kindly ask her to repeat it here.

mitzimorris:
the json parser is part of stan::io as it constructs a stan::io::var_context object.

I wrote the first version of the JSON parser in 2014, FWIW.  discussed with Rok during the Stan meeting and would be happy to do the code review.

mitzimorris:
> unit tests (do we need to add any or would the ones for the existing parser suffice?)

I think the existing tests suffice - they go through all legal and illegal combinations and test 1D, 2D, and 3D arrays.

'''
> /github/stan-dev/cmdstan/stan/src/test/unit/io/json> grep '"' *.cpp | grep "test_" | wc -l
'''

rok-cesnovar:
@mitzimorris  Thank you for the review!

Upon removing the existing parser code, I realized not all tests were run with RapidJSON. So I had to change a few things since your review:
- removed a semicolon in a test that followed an object "{"a": 5};". RapidJSON treats this as a syntax error
- I change some of the expected error texts (see json_parser_test.cpp file). Some texts are more descriptive and some are less descriptive than with the existing parser.
- RapidJSON does not throw an error when parsing "[ \"\\uDC00\\uDFFFF\" ]" or "[ \"\\uE000\\uDFFF\" ]". Is this an issue if it does not? If yes, I will look into this.
- the default behavior of RapidJSON in the case where an integer is too big is to try and cast it as a double. I presumed this is not going to fly with Stan. I changed the RapidJSON code to not attempt that, will also do a PR on their end to make this an option flag to the parser, but not sure if they will want it.
- another default behavior of RapidJSON that I disliked is rounding to 0.0 in case of a small number outside the range of double. I changed it so that it throws in that case, like the existing parser.

rok-cesnovar:
In order for the upstream tests to pass the makefiles in cmdstan would probably need to be updated right? Does that mean that I should open a cmdstan PR first?

bob-carpenter:
Given that there aren't any dependencies other var_context, it could live in stan-dev/stan.

I think the question to ask is whether any interface other than CmdStan would use a C++ JSON parser.  If so, then it should go in stan, if not, I'd think it should go in CmdStan.  RStan and PyStan provide their own implementations of var_context, but I think PyStan may use the R dump parser in stan-dev/stan.

mitzimorris:
> whether any interface other than CmdStan would use a C++ JSON parser.

CmdStan is the only interface currently using the JSON parser.  both Python and R have several JSON parser libraries.  I checked the PyStan code which doesn't use the Rdump parser - it re-implements this functionality directly in Python.

Give this, I agree, this code could be added directly to CmdStan, and then we could do a follow-up  PR to remove current JSON parser (and tests) from `stan::io`.  In which case, would the appropriate location be new directory `cmdstan/src/cmdstan/io`?  and for the ujson lib, add directory, `cmdstan/lib` - @bob-carpenter does this sound OK?

> PyStan may use the R dump parser in stan-dev/stan

looked at repo PyStan, and no, no dependencies on code in stan::io.

bob-carpenter:
> On Aug 2, 2019, at 1:46 PM, Rok Češnovar <notifications@github.com> wrote:
> 
> @mitzimorris Thank you for the review!
> 
> Upon removing the existing parser code, I realized not all tests were run with RapidJSON. So I had to change a few things since your review:
> 
> 	• removed a semicolon in a test that followed an object "{"a": 5};". RapidJSON treats this as a syntax error
> 	• I change some of the expected error texts (see json_parser_test.cpp file). Some texts are more descriptive and some are less descriptive than with the existing parser.
> 	• RapidJSON does not throw an error when parsing "[ "\uDC00\uDFFFF" ]" or "[ "\uE000\uDFFF" ]". Is this an issue if it does not? If yes, I will look into this.

This is presumably a test for handling conjugate character encoding in UTF-8.  

What we want to do about this depends on whether we want to check that Stan files are encoded in legal UTF-8.

Stan requires the program itself to be written in ASCII, but suggests UTF-8 may be used for the content of comments as those will not be read.

> 	• the default behavior of RapidJSON in the case where an integer is too big is to try and cast it as a double. I presumed this is not going to fly with Stan.

I'm not sure the exact boundary conditions matter much here.  If Stan's looking for an integer and gets a double value, it'll still raise an error.  It doesn't seem unreasonable for the parser to try to make sense of 
1234567890123456789012345678901234567890 by representing it as a double.  


> I changed the RapidJSON code to not attempt that, 

Did you modify their code or how it's used?  Is there a way to do this without modifying their API?


> will also do a PR on their end to make this an option flag to the parser, but not sure if they will want it.

That's nice.  I don't know if they'll want it as you can argue
a lot of ways about what the boundary behavior should be.


> 	• another default behavior of RapidJSON that I disliked is rounding to 0.0 in case of a small number outside the range of double. I changed it so that it throws in that case, like the existing parser.

We tried to be careful about boundary conditions and throw errors, but I'm not sure that's the right behavior for floating point.  What if someone gives us a 30-digit number like 0.123456789012345678901234567890?  We'll probably just throw away the unused digits.  Arguably, rounding to 0 does the same thing.

mitzimorris:
just filed CmdStan issue https://github.com/stan-dev/cmdstan/issues/719 to move this work directly into CmdStan

rok-cesnovar:
Thank you! Will post answers to Bob's questions and comments on that issue.

Issue #2752, opened 2019-04-21
rhat and ess updates in analyze


avehtari:
#### Summary:
In analyze service
- [ ] Offer interfaces a unified calculation of rhat
- [x] Update effective sample size calculation
- [x] Updating ess requires a small change in the autocovariance computation

#### Description:

#2569 and #2575 moved effective sample size computation from chains.hpp to analyze service.
Move rhat calculation in the same way. In addition make minor changes to the effective sample size and rhat calculation. These changes make the computations more robust, but in most cases affect the computed values only a little. Especially these changes don't include rank normalization, folding or quantiles (these features will be added later).

The reference implementation in R is at https://github.com/stan-dev/rstan/blob/develop/rstan/rstan/R/monitor.R

- [x] autocovariance: use "biased" estimate as recommended by Geyer (1992)
- [ ] rhat: for constant finite values return 1, for non-finite values return NaN
- [x] ess: 1) use split chains, 2) add half rho of the next even lag after the truncation, 3) negative values and maximum ess are set chains*n_samples*log10(chains*n_samples)

#### Reproducible Steps:
Refrence values from R implementation
```
upath <- system.file('unitTests', package='rstan')
f1 <- file.path(upath, 'testdata', 'blocker1.csv')  
f2 <- file.path(upath, 'testdata', 'blocker2.csv')  
fit <- read_stan_csv(c(f1,f2))
monitor(fit)
sims <- as.array(fit)
rh <- rhat(sims[,,1])
checkEquals(rh, 1.007794, tolerance = 0.001); 
eb <- ess_bulk(sims[,,1])
checkEquals(eb, 465.3917, tolerance = 0.001); 
et <- ess_tail(sims[,,1])
checkEquals(et, 349.1353, tolerance = 0.001); 
em <- ess_mean(sims[,,1])
checkEquals(em, 467.8447, tolerance = 0.001); 
```
Same `blocker1.csv` and `blocker2.csv` files are in stan repo at `src/test/unit/mcmc/test_csv_files/` and have been used in unit tests in stan repo, too.

EDIT: updated ess_mean reference value.

avehtari:
ping @roualdes

roualdes:
Thanks, @avehtari.  I think `autocovariance` is in `stan-dev/math`.  I'll start this week to work on these.

avehtari:
>  I think autocovariance is in stan-dev/math

[...]

Issue #2569, opened 2018-07-05
Analysis API: compute effective sample size


roualdes:
#### Summary:

Offer interfaces a unified calculation of effective sample size.

#### Description:

This issue is solely for the calculation of ess for a single parameter across multiple chains.

Many details were discussed on the [Discourse: Analysis API](http://discourse.mc-stan.org/t/analysis-api/3486) thread.  Design details:

- route goes in `stan-dev/stan:src/stan/analyze/mcmc/compute_ess.hpp`
- signature is `compute_effective_sample_size(std::vector<double*> draws, std::vector<size_t> sizes) ` where each chain's draws are stored in a contiguous block of memory
- convenience function `compute_effective_sample_size(std::vector<double*> draws, size_t size)`
- calculation and I/O handled separately.  Only calculation, here.
- minimize copying
- stan-dev/stan is the appropriate place for ess calculation (and other diagnostics) as it is more closely related to sampling algorithms than the stan-dev/math library
- start with ess, integrate it all the way through, then build out other diagnostics 
- long term goal to remove ess from chains class as OO structure is not necessary 

#### Current Version:
v2.17.1

roualdes:
Looking at `effective_sample_size` within the chains class, will [`n_kept_samples`](https://github.com/stan-dev/stan/blob/67298e04b073dcdb625f125b636d7a77401251c7/src/stan/mcmc/chains.hpp#L248) ever be different from [`n_samples`](https://github.com/stan-dev/stan/blob/67298e04b073dcdb625f125b636d7a77401251c7/src/stan/mcmc/chains.hpp#L234)?  If so, do we need to accept `n_kept_samples` as an argument to `compute_effective_sample_size`?

bob-carpenter:
Not sure if this answers your question, but the number of draws before thinning isn't relevant for n_eff calculations.  

We should be using "draws" not "samples" in the code if you want to fix that while you're at it---we'd like to use "sample" to mean a sequence of draws.

riddell-stan:
It seems like `draws` is the preferred label. Do we want to change the 
parameter num_samples (CmdStan) to num_draws? Or at least add a note in 
the documentation that "num_samples" uses deprecated terminology?

roualdes:
I believe the answer to my question is, no, `n_kept_samples` and `n_samples` will never be different.  My best efforts indicate that `develop:effective_sample_size` is only ever given kept draws, as @bob-carpenter suggested.

[...]

PR #2794, opened 2019-08-06
Feature/issue 2752 update rhat and ess (part II)

roualdes:
#### Submission Checklist

#### Summary

Re issue https://github.com/stan-dev/stan/issues/2752: offer interfaces a unified calculation of rhat, and add checks in both calculations of ESS and Rhat to better match the reference implementation in R at https://github.com/stan-dev/rstan/blob/develop/rstan/rstan/R/monitor.R.  

Re Analysis API (see https://github.com/stan-dev/stan/issues/2569 and/or https://github.com/stan-dev/stan/pull/2575) this PR moves potential_scale_reduction to analyze service.  To be more consistent with the discussions in issue https://github.com/stan-dev/stan/issues/2752, the splitting of split_potential_scale_reduction now follows the design of the splitting of the effective sample size calculations.

#### Intended Effect

Provide a unified calculation of split Rhat.

#### How to Verify

Tests added in `src/test/unit/analyze/mcmc`.

#### Side Effects

None intended.

#### Documentation

Inline.

#### Copyright and Licensing

Please list the copyright holder for the work you are submitting (this will be you or your assignee, such as a university or company): California State University, Chico


betanalpha:
I’m a bit slammed but I’ll try to review by the end of next week!

src/stan/analyze/mcmc/compute_effective_sample_size.hpp
```diff
@@ -35,6 +37,30 @@ namespace analyze {
       num_draws = std::min(num_draws, sizes[chain]);
     }
 
+    if ( num_draws < 3 ) {
```
betanalpha: This edge case should be noted in the function documentation.  Perhaps also a note here about why less than three (e.g. variance cannot be estimated for less than three draws).

src/stan/analyze/mcmc/compute_effective_sample_size.hpp
```diff
@@ -35,6 +37,30 @@ namespace analyze {
...
+      return std::numeric_limits<double>::quiet_NaN();
+    }
+
+    Eigen::VectorXd draw_val = Eigen::VectorXd::Random(num_chains);
```
betanalpha: `draw_val` needs to be documented.  The random initialization is also dangerous as it can yield non-reproducible results -- I recommend going with (1, 2, 3, ..., num_chains) to give distinct but determinism values.

src/stan/analyze/mcmc/compute_effective_sample_size.hpp
```diff
@@ -35,6 +37,30 @@ namespace analyze {
...
+          return std::numeric_limits<double>::quiet_NaN();
+        }
+      }
+
+      if ( draw.isApproxToConstant(draw(0)) ) {
+        draw_val(chain) = draw(0);
+      }
+    }
+
+    if ( draw_val.isApproxToConstant(draw_val(0)) ) {
```
betanalpha: Do we not want to return NaN here?  If everything is constant then the asymptotic variance, and hence ESS, isn't particularly well-defined.

src/stan/analyze/mcmc/compute_potential_scale_reduction.hpp
```diff
@@ -0,0 +1,162 @@
...
+   * @param draws stores pointers to arrays of chains
+   * @param sizes stores sizes of chains
+   * @return potential scale reduction for the specified parameter
+   */
+  inline
+  double compute_potential_scale_reduction(std::vector<const double*> draws,
+                                           std::vector<size_t> sizes) {
+    int num_chains = sizes.size();
+    size_t num_draws = sizes[0];
+    for ( int chain = 1; chain < num_chains; ++chain ) {
```
betanalpha: The function doc says the code assumes that the chains are all of equal size, but this code doesn't.  Either the function doc should change ("chains are trimmed from the back to match the length of the shortest chain") or this code should check equal size and throw a NaN or an exception if they're not the same.

src/stan/analyze/mcmc/compute_potential_scale_reduction.hpp
```diff
@@ -0,0 +1,162 @@
...
+          return std::numeric_limits<double>::quiet_NaN();
+        }
+      }
+
+      if ( draw.isApproxToConstant(draw(0)) ) {
+        draw_val(chain) = draw(0);
+      }
+    }
+
+    if ( draw_val.isApproxToConstant(draw_val(0)) ) {
```
betanalpha: Same comment as above -- if everything is constant then do we want a NaN here instead of a "perfect" value?

src/stan/analyze/mcmc/compute_potential_scale_reduction.hpp
```diff
@@ -0,0 +1,162 @@
...
+    }
+
+    Eigen::VectorXd chain_mean(num_chains);
+    Eigen::VectorXd chain_var(num_chains);
+
+    for ( int chain = 0; chain < num_chains; chain++ ) {
+      Eigen::Map<const Eigen::Matrix<double, Eigen::Dynamic, 1>>
+        draw(draws[chain], sizes[chain]);
+      chain_mean(chain) = draw.mean();
+      chain_var(chain) = (draw.array() - chain_mean(chain)).square().sum()
```
betanalpha: This method of computing the empirical variance is prone to numerical inaccuracy.  We should use a Welford accumulator -- these are implemented in Boost but the code is short enough that it can be done by hand here.  See for example https://github.com/stan-dev/math/blob/502487c511594ccb93eb979df5b8fe163becb417/stan/math/prim/mat/fun/welford_var_estimator.hpp.

src/stan/analyze/mcmc/compute_effective_sample_size.hpp
roualdes: A valid point and I agree.  I was going for consistency with the original issue's reference implementation: https://github.com/stan-dev/rstan/blob/8caa8728669ae2df75167202e08dff930d4d7c64/rstan/rstan/R/monitor.R#L213

Should we over rule the consistency argument in favor of returning NaN here?

src/stan/analyze/mcmc/compute_potential_scale_reduction.hpp
roualdes: Same as above.  Should we over rule the consistency argument in favor of returning NaN here?

src/stan/analyze/mcmc/compute_potential_scale_reduction.hpp
roualdes: Nice.  

Is it too much for this PR to make this same upgrade in stan-dev/stan's [analyze/mcmc/autocovariance.hpp](https://github.com/stan-dev/stan/blob/f372f98065461b1d7b4b4fe3d1b0bd984138887a/src/stan/analyze/mcmc/autocovariance.hpp#L87) also?

betanalpha: Could you make an issue and then put together a PR for updating this code and the code in autocovariance.hpp?  That way we can push this PR as is and contain the better variance estimation to its own PR.

src/stan/analyze/mcmc/compute_potential_scale_reduction.hpp
betanalpha: Let me start a Discourse thread appealing for opinion on this.  Worst case I'll approve for the consistency of the current behavior and we'll fix with a separate PR.

betanalpha:
Sorry for the delay -- I was hoping that the Discourse thread would evolve but it seems to have come to an uneasy equilibrium.  For now I'm going to request that you return `NaN` for Rhat and ESS when the chains are constant and we can handle the management of all of the possible cases in the MCMC standard error code (for example returning SE=0 if variance is finite but ESS=NaN, returning SE=NaN if both variance and ESS are NaN, etc) and check functions (reporting that NaN means constant values and users are responsible for verifying whether that's expected or not).

Once that small change is made I'll approve.

Replacing the variance calculations with accumulators can be pushed to a new issue/PR.

Thanks for your patience!

src/stan/analyze/mcmc/compute_effective_sample_size.hpp
roualdes: Done

src/stan/analyze/mcmc/compute_effective_sample_size.hpp
roualdes: (1, 2, 3, ... num_chains), check.  Would you like an inline comment or for `draw_val` to be mentioned in the function documentation?

src/stan/analyze/mcmc/compute_potential_scale_reduction.hpp
roualdes: Made explicit the trimming.

src/stan/analyze/mcmc/compute_potential_scale_reduction.hpp
roualdes: Will do.

src/stan/analyze/mcmc/compute_effective_sample_size.hpp
betanalpha: Typically inline if it doesn't affect the outcome of the function evaluation (and hence is just an implementation detail) and in the function doc if it does affect output values so that the client is aware of any expected edge cases.

betanalpha:
Just waiting on the draw_vals doc and then we're good to go!

roualdes:
Yesterday, I discovered an intermittent error in the tests within the file `split_chains_test.cpp`.  I believe my previous test was using incorrect logic.  Would you please give `split_chains_test.cpp` a look over?  The update should be correct now, but as always a second set of eyes rarely hurts.  Thanks.

PR #2800, opened 2019-08-13
Feature/issue 2799 robust no u turn

### LINKED ISSUE #2799: Binary recursive no-u-turn criterion misses u-turn
#### Summary:

The current no-u-turn checks can miss a u-turn when two halves of a discrete trajectory straddle the desired termination point.

#### Description:

For approximately iid normal targets the no-u-turn criterion can miss the optimal integration time if the step size is configured in a way that allows discrete trajectories to straddle the desired termination point.  This results in "resonances" that drop performance for various steps sizes.  See https://discourse.mc-stan.org/t/nuts-misses-u-turns-runs-in-circles-until-max-treedepth/9727/14 for more discussion.

#### Current Version:
v2.20.0

betanalpha:
#### Submission Checklist

#### Summary

Resolves #2799.

#### Intended Effect

Adds additional no-u-turn checks across subtrees to avoid missing u-turns for approximately iid normal models. 

#### How to Verify

See https://discourse.mc-stan.org/t/nuts-misses-u-turns-runs-in-circles-until-max-treedepth/9727/36?u=betanalpha.

#### Side Effects

Decrease in antithetic behavior for component means in correlated models.

#### Documentation

Inline.

#### Copyright and Licensing

Michael Betancourt

### DIFF
--- a/src/test/unit/mcmc/hmc/nuts/base_nuts_test.cpp
+++ b/src/test/unit/mcmc/hmc/nuts/base_nuts_test.cpp
@@ -149,9 +149,12 @@ TEST(McmcNutsBaseNuts, build_tree_test) {
 
   stan::mcmc::ps_point z_propose(model_size);
 
-  Eigen::VectorXd p_sharp_left = Eigen::VectorXd::Zero(model_size);
-  Eigen::VectorXd p_sharp_right = Eigen::VectorXd::Zero(model_size);
+  Eigen::VectorXd p_begin = Eigen::VectorXd::Zero(model_size);
+  Eigen::VectorXd p_sharp_begin = Eigen::VectorXd::Zero(model_size);
+  Eigen::VectorXd p_end = Eigen::VectorXd::Zero(model_size);
+  Eigen::VectorXd p_sharp_end = Eigen::VectorXd::Zero(model_size);
   Eigen::VectorXd rho = z_init.p;
+  
   double log_sum_weight = -std::numeric_limits<double>::infinity();
 
   double H0 = -0.1;
@@ -170,15 +173,18 @@ TEST(McmcNutsBaseNuts, build_tree_test) {
   stan::callbacks::stream_logger logger(debug, info, warn, error, fatal);
 
   bool valid_subtree = sampler.build_tree(3, z_propose,
-                                          p_sharp_left, p_sharp_right, rho,
+                                          p_sharp_begin, p_sharp_end, 
+                                          rho, p_begin, p_end,
                                           H0, 1, n_leapfrog, log_sum_weight,
                                           sum_metro_prob, logger);
 
   EXPECT_TRUE(valid_subtree);
 
   EXPECT_EQ(init_momentum * (n_leapfrog + 1), rho(0));
-  EXPECT_EQ(1, p_sharp_left(0));
-  EXPECT_EQ(1, p_sharp_right(0));
+  EXPECT_EQ(1.5, p_begin(0));
+  EXPECT_EQ(1, p_sharp_begin(0));
+  EXPECT_EQ(1.5, p_end(0));
+  EXPECT_EQ(1, p_sharp_end(0));
 
   EXPECT_EQ(8 * init_momentum, sampler.z().q(0));
   EXPECT_EQ(init_momentum, sampler.z().p(0));
@@ -207,9 +213,12 @@ TEST(McmcNutsBaseNuts, rho_aggregation_test) {
 
   stan::mcmc::ps_point z_propose(model_size);
 
-  Eigen::VectorXd p_sharp_left = Eigen::VectorXd::Zero(model_size);
-  Eigen::VectorXd p_sharp_right = Eigen::VectorXd::Zero(model_size);
+  Eigen::VectorXd p_begin = Eigen::VectorXd::Zero(model_size);
+  Eigen::VectorXd p_sharp_begin = Eigen::VectorXd::Zero(model_size);
+  Eigen::VectorXd p_end = Eigen::VectorXd::Zero(model_size);
+  Eigen::VectorXd p_sharp_end = Eigen::VectorXd::Zero(model_size);
   Eigen::VectorXd rho = z_init.p;
+  
   double log_sum_weight = -std::numeric_limits<double>::infinity();
 
   double H0 = -0.1;
@@ -228,18 +237,19 @@ TEST(McmcNutsBaseNuts, rho_aggregation_test) {
   stan::callbacks::stream_logger logger(debug, info, warn, error, fatal);
 
   sampler.build_tree(3, z_propose,
-                     p_sharp_left, p_sharp_right, rho,
+                     p_sharp_begin, p_sharp_end, 
+                     rho, p_begin, p_end,
                      H0, 1, n_leapfrog, log_sum_weight,
                      sum_metro_prob, logger);
 
-  EXPECT_EQ(7, sampler.rho_values.size());
+  EXPECT_EQ(7 * 3, sampler.rho_values.size());
   EXPECT_EQ(2 * init_momentum, sampler.rho_values.at(0));
   EXPECT_EQ(2 * init_momentum, sampler.rho_values.at(1));
-  EXPECT_EQ(4 * init_momentum, sampler.rho_values.at(2));
+  EXPECT_EQ(2 * init_momentum, sampler.rho_values.at(2));
   EXPECT_EQ(2 * init_momentum, sampler.rho_values.at(3));
   EXPECT_EQ(2 * init_momentum, sampler.rho_values.at(4));
-  EXPECT_EQ(4 * init_momentum, sampler.rho_values.at(5));
-  EXPECT_EQ(8 * init_momentum, sampler.rho_values.at(6));
+  EXPECT_EQ(2 * init_momentum, sampler.rho_values.at(5));
+  EXPECT_EQ(4 * init_momentum, sampler.rho_values.at(6));
 }
 
 TEST(McmcNutsBaseNuts, divergence_test) {
@@ -255,9 +265,12 @@ TEST(McmcNutsBaseNuts, divergence_test) {
 
   stan::mcmc::ps_point z_propose(model_size);
 
-  Eigen::VectorXd p_sharp_left = Eigen::VectorXd::Zero(model_size);
-  Eigen::VectorXd p_sharp_right = Eigen::VectorXd::Zero(model_size);
+  Eigen::VectorXd p_begin = Eigen::VectorXd::Zero(model_size);
+  Eigen::VectorXd p_sharp_begin = Eigen::VectorXd::Zero(model_size);
+  Eigen::VectorXd p_end = Eigen::VectorXd::Zero(model_size);
+  Eigen::VectorXd p_sharp_end = Eigen::VectorXd::Zero(model_size);
   Eigen::VectorXd rho = z_init.p;
+  
   double log_sum_weight = -std::numeric_limits<double>::infinity();
 
   double H0 = -0.1;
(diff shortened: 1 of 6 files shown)


src/test/unit/mcmc/hmc/nuts/base_nuts_test.cpp
```diff
@@ -228,18 +237,19 @@ TEST(McmcNutsBaseNuts, rho_aggregation_test) {
...
   EXPECT_EQ(2 * init_momentum, sampler.rho_values.at(0));
   EXPECT_EQ(2 * init_momentum, sampler.rho_values.at(1));
-  EXPECT_EQ(4 * init_momentum, sampler.rho_values.at(2));
+  EXPECT_EQ(2 * init_momentum, sampler.rho_values.at(2));
   EXPECT_EQ(2 * init_momentum, sampler.rho_values.at(3));
   EXPECT_EQ(2 * init_momentum, sampler.rho_values.at(4));
-  EXPECT_EQ(4 * init_momentum, sampler.rho_values.at(5));
-  EXPECT_EQ(8 * init_momentum, sampler.rho_values.at(6));
+  EXPECT_EQ(2 * init_momentum, sampler.rho_values.at(5));
+  EXPECT_EQ(4 * init_momentum, sampler.rho_values.at(6));
```
bbbales2: There should be 14 more of these checks since there are now 3x the number of rho_values

betanalpha: There's only one aggregate rho exposed in the `build_tree` function.  The others needed for the new tests are constructed from this aggregate rho on the fly internally and hence not accessible for unit testing.  That internal construction could be repeated externally using the boundary momenta but then there's nothing to compare it to.

betanalpha:
There's currently no validation of the samplers as explicit integration tests because there's not enough functionality to run everything from pure C++ (the main obstruction is tearing out the `var_context` code and replacing it with a clear data access layer and then writing C++ data access layer callbacks so that everything can be run in memory; I have designs that were long ago discussed and agreed upon but no one has had the time to implement it).  In any case I did check this code against some IID models of varying dimension and a correlated Gaussian in which the last bug manifested (using 100000 iterations to be sufficient sensitive to small effects) and everything looked good.

bob-carpenter:
Why not use R dump or JSON format for var contexts?  They're not in memory, but I/O performance shouldn't be an issue for testing.

src/test/unit/mcmc/hmc/nuts/base_nuts_test.cpp
bbbales2: I don't understand this. These values are being returned so why shouldn't they be tested?

bbbales2:
@betanalpha oh okay, got it. Can you e-mail me the models for the test? I guess I should just manually double check that it's all working okay since it's not automated.

@seantalts yo just at-ing you so you're aware of the state of tests. We should probably do something about this eventually.

betanalpha:
They’re all in benchmark repo, https://github.com/stan-dev/stat_comp_benchmarks/tree/master/benchmarks.
The baselines for each expectation value are in https://github.com/stan-dev/stat_comp_benchmarks/tree/master/empirical_results/nuts.

You can run them all or just check the basics like low_dim_corr_gauss and
maybe a 50 or 100 dimensional IID gaussian,

parameters {
  real x[100];
}

model {
  x ~ normal(0, 1);
}

bbbales2:
Got it thanks. I'll run these checks Tuesday or Wednesday (still travelin'). And the extra return values from build_tree?

junpenglao:
Thanks a lot for the discussion and the validation!! 
Just a quick drive by comment: you could skip doing the additional U turn check at depth == 1 within each tree (as it is already been checked), and skip the U turn check altogether at depth == 0 at the tree doubling level (i.e., the outer loop, as it is already been checked during the tree building).

betanalpha:
The test you were looking at checks just the rho aggregation.  The edge momenta and sharp momenta are checked in other tests, including the `diag_e` and `softabs` tests.

junpenglao:
I see - thanks @betanalpha !

betanalpha:
That’s definitely not good.  

Do you get the same result when you run with the diagonal metric?
Let me see if I can reproduce locally.

betanalpha:
Same for a dense metric,

```
1 chains: each with iter=(4096000); warmup=(0); thin=(1); 4096000 iterations saved.

                    Mean     MCSE   StdDev     5%       50%       95%    N_Eff  N_Eff/s    R_hat
z[1]            -3.9e-04  4.8e-04  1.0e+00   -1.6   4.5e-04   1.6e+00  4.3e+06  2.9e+04  1.0e+00
z[2]             3.0e+00  1.0e-03  2.0e+00  -0.29   3.0e+00   6.3e+00  3.8e+06  2.6e+04  1.0e+00
delta_var1      -3.7e-04  9.9e-04  1.4e+00  -1.00  -5.5e-01   2.8e+00  2.1e+06  1.4e+04  1.0e+00
delta_var2      -5.4e-03  4.0e-03  5.6e+00   -4.0  -2.2e+00   1.1e+01  2.0e+06  1.4e+04  1.0e+00
delta_corr       2.8e-05  7.6e-04  1.1e+00   -1.1  -3.4e-01   2.2e+00  2.2e+06  1.5e+04  1.0e+00
```

betanalpha:
I recommend fixing a seed and then using `id` to vary the seed for each run.

betanalpha:
This is weeeeeeeeeird.  For some seeds I do see the effect that you're seeing.  

```
1 chains: each with iter=(16384000); warmup=(0); thin=(1); 16384000 iterations saved.

                    Mean     MCSE   StdDev     5%       50%       95%    N_Eff  N_Eff/s    R_hat
z[1]             4.9e-04  3.4e-04  1.0e+00   -1.6   4.4e-04   1.6e+00  8.4e+06  1.2e+04  1.0e+00
z[2]             3.0e+00  7.3e-04  2.0e+00  -0.28   3.0e+00   6.3e+00  7.4e+06  1.0e+04  1.0e+00
delta_var1      -6.2e-03  4.7e-04  1.4e+00  -1.00  -5.5e-01   2.8e+00  8.9e+06  1.3e+04  1.0e+00
delta_var2      -2.4e-02  2.0e-03  5.6e+00   -4.0  -2.2e+00   1.1e+01  8.2e+06  1.2e+04  1.0e+00
delta_corr      -2.3e-03  4.0e-04  1.1e+00   -1.1  -3.4e-01   2.2e+00  7.8e+06  1.1e+04  1.0e+00
```

What's odd, however, it doesn't look there's some bias that manifests only once the standard error shrinks enough.  Rather it looks like the standard error shrinks and then at some point a bias larger than the standard error suddenly appears.  Then there's the fact that it's an inconsistent bias, and maybe correlated with slightly higher step sizes?

I'm going to go through the code carefully to see if I can see anything.  On the other hand I'm not sure any algorithm has been tested to one _million_ effective samples so if the current version didn't behave as expected then I'd suspect a numerical issue with the estimators themselves.

bob-carpenter:
You mean things like the mean and quantile estimators?

Adding a million numbers will lose a lot of precision if they're all the same sign.  For example, imagine 1e6 numbers between 0 and 1.  You wind up adding numbers on the scale of 5e5 (the expectation) and 1 (the individual numbers being added), which loses more than 5 decimal places of precision (out of a total of 16 or so in double-precision floating point).

betanalpha:
Of course, but the issue is that there is no corresponding
bias in the baseline version of the sampler so it can’t be
something shared between the two versions, such as the
estimator code.

betanalpha:
I can't find anything that looks suspect in the code -- anyone want to take a second look?  I've been poking around at edge cases and places where redundant calculations in the added checks might be vulnerable to insufficient precision introducing errors but I can't find anything inconsistent.

bbbales2:
Here's the results of running overnight.

uturn_metric

> anyone want to take a second look?

I'll make a post in the discourse thread. I will eventually, but this looks like it could be a difficult problem to figure out.

betanalpha:
I think I found it — there’s a bug in how the edge momenta and
sharp momenta are updated before a new expansion/proposal.
As in some of them aren’t updated as they should be, causing 
the new checks to be slightly off (and getting increasingly off 
as the tree depth increases).  I’m testing the seeds that were
previously manifesting the bias now.

betanalpha:
Yup, that did the trick.

```
1 chains: each with iter=(16384000); warmup=(0); thin=(1); 16384000 iterations saved.

                    Mean     MCSE   StdDev     5%       50%       95%    N_Eff  N_Eff/s    R_hat
z[1]             4.9e-04  3.4e-04  1.0e+00   -1.6   4.4e-04   1.6e+00  8.4e+06  1.2e+04  1.0e+00
z[2]             3.0e+00  7.3e-04  2.0e+00  -0.28   3.0e+00   6.3e+00  7.4e+06  1.0e+04  1.0e+00
delta_var1      -6.2e-03  4.7e-04  1.4e+00  -1.00  -5.5e-01   2.8e+00  8.9e+06  1.3e+04  1.0e+00
delta_var2      -2.4e-02  2.0e-03  5.6e+00   -4.0  -2.2e+00   1.1e+01  8.2e+06  1.2e+04  1.0e+00
delta_corr      -2.3e-03  4.0e-04  1.1e+00   -1.1  -3.4e-01   2.2e+00  7.8e+06  1.1e+04  1.0e+00
```

goes to

```
1 chains: each with iter=(16384000); warmup=(0); thin=(1); 16384000 iterations saved.

                    Mean     MCSE   StdDev     5%       50%       95%    N_Eff  N_Eff/s    R_hat
z[1]             9.0e-05  3.5e-04  1.0e+00   -1.6   1.8e-04   1.6e+00  8.0e+06  1.2e+04  1.0e+00
z[2]             3.0e+00  7.3e-04  2.0e+00  -0.29   3.0e+00   6.3e+00  7.5e+06  1.2e+04  1.0e+00
delta_var1       2.6e-04  4.8e-04  1.4e+00  -1.00  -5.5e-01   2.8e+00  8.6e+06  1.3e+04  1.0e+00
delta_var2       2.4e-03  2.0e-03  5.7e+00   -4.0  -2.2e+00   1.1e+01  8.2e+06  1.3e+04  1.0e+00
delta_corr       2.3e-04  4.0e-04  1.1e+00   -1.1  -3.4e-01   2.2e+00  7.6e+06  1.2e+04  1.0e+00
```

I'll push once the tests are updated to the new behavior.

bbbales2:
Great thanks!

betanalpha:
Changes pushed -- I updated `base_nuts_test` with a new instrumented mock sampler that tests the edge momenta that weren't being updated correctly before.  I'm going to rerun the performance tests to verify that hasn't been affected, but given that this only shows up at <Dr. Evil voice>one million iterations</Dr. Evil voice> I don't think there will be much of an affect.

I don't think I've ever seen a sampler tested to this level of precision!  Pretty awesome.

betanalpha:
Performance tests look good -- small differences but the general story is pretty much the same.  I'll post to Discourse.

bob-carpenter:
> On Aug 29, 2019, at 3:11 PM, Michael Betancourt <notifications@github.com> wrote:
> 
> Changes pushed -- I updated base_nuts_test with a new instrumented mock sampler that tests the edge momenta that weren't being updated correctly before. I'm going to rerun the performance tests to verify that hasn't been affected, but given that this only shows up at <Dr. Evil voice>one million iterations</Dr. Evil voice> I don't think there will be much of an affect.
> 
> I don't think I've ever seen a sampler tested to this level of precision! Pretty awesome.

Indeed.  Ben's going to get a reputation for seriously heroic testing (after this and my autodiff test PR).

bbbales2:
Don't jinx us. This looks good. @bob-carpenter can you merge? I'm not authorized.

PR #2805, opened 2019-09-05
Add automated maintanence and cleanup via clang-tools to makefile

SteveBronder:
The clang-tidy and clang-format tools have been added to the
make file so that it's easier to keep all of the code consistent

#### Submission Checklist

- [x] Run unit tests: `./runTests.py src/test/unit`
- [x] Run cpplint: `make cpplint`
- [x] Declare copyright holder and open-source license: see below

#### Summary

This adds

1. make methods for `clang-format` and `clang-tidy` so that it's easy to run them over the files in source. clang-format uses the same `.clang-format` file used in [the math library](https://github.com/SteveBronder/math/blob/develop/.clang-format)

There are only two actual places I made changes in the code

1. in [` src/stan/lang/generator/generate_fun_inst_templ_params.hpp`](https://github.com/stan-dev/stan/compare/develop...SteveBronder:cleanup/clang-format?expand=1#diff-6d4c39775589e2beeb3f9528bd0c0251R1) I removed the macro newline `\` and put a `// NOLINT` at the top
2. in [`src/stan/callbacks/stream_writer.hpp`](https://github.com/stan-dev/stan/compare/develop...SteveBronder:cleanup/clang-format?expand=1#diff-d9a5f54d53833b98b437bbfbb30158c9R26) I made the constructor on line 26 `explicit` since it _can_ accept only one argument

```cpp
stream_writer(std::ostream& output,
                         const std::string& comment_prefix = "")
```

#### Intended Effect

There should be no change to how code is executed. The tools are just so that the stan and math library have similar C++ styles. If that's not a good thing for the code at this level then we can close this PR (it took me like 10 mins to run this so no biggie)

#### How to Verify

From develop you can run `make clang-format` to get the code generated here (minus the two places listed above

#### Side Effects

If there are side effects beyond the two mentioned above then that's bad

#### Copyright and Licensing

Please list the copyright holder for the work you are submitting (this will be you or your assignee, such as a university or company):

By submitting this pull request, the copyright holder is agreeing to license the submitted work under the following licenses: Steve Bronder

PR #2806, opened 2019-09-06
[WIP] Refactor IO

SteveBronder:
#### Submission Checklist

#### Summary

This is a refactor of the `array_var_context` and `reader` classes. This is built off of PR #2805 so it looks like there are a lot of changes rn but it's really just the files which hold these two classes. 

Changes include:

1. Previously in `add_r` and `add_i` we were calculating the inner dimensions of the `dim` vector twice. Once in `validate()` and again in the main loop of both functions. `validate` has been renamed to `validate_dims` and if `validate_dims` throws no errors it gives back a vector with the cumulative sum of the product of the dimensions for each vector inside of dims. This uses a few of the stl algorithms and a lambda so it may look a bit odd.

2. With (1) we can now clean up `add_r` and `add_i`'s loop. The body of the function now loops over names and dim_vec adding them to the table of values.

3. `using` is used rather liberally to clean up some of the loops and types.

4. `vars_*_` are now `unordered_maps` instead of `maps`. I made this change heuristically but would like to test a bit whether this is better. My assumption on this is that we tend to do access in this class more often then insertion and deletion.

5. We now use the braced initializer list style on insertion to the `vars_*_` maps along with perfect forwarding in the floating point vector case. This looks a little wonky but removes a good few temporaries.

```cpp
// tbh I'm really not sure if forwarding this is meaningful
vars_r_[name[i]] = {{std::forward<val_d_t>(values.data()) + dim_vec[i],
                                std::forward<val_d_t>(values.data()) + dim_vec[i + 1]}, dims[i]};
```

6. `vars_*_` are initialized in the constructor initialization list with a number of elements equal to the size of `names_*`. This should save a few reallocations and remove the bad side of unordered maps cost of new insertions at least during initialization.

7. Cleans up the loops so they use the range based style.

8. Made the arguments for the `for` loops in the accessors of `reader` const. This should help the compiler a little bit.

#### Intended Effect

I was trying to understand how some things worked here so did a bit of updating. This should also (hopefully) give us a teeny performance increase. Though I don't expect much

#### How to Verify

This just uses the tests that already existed but if we want to add any additional tests to the reader and array context then I'm happy to add them to this PR.

#### Side Effects

Hopefully none!

#### Documentation

Updating documentation for existing functions.

#### Copyright and Licensing

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -61,3 +61,9 @@ stan.kdev4
 
 #clang-tidy fixes yaml
 .clang-fixes.log
+
+# Visual Studio Settings
+.vscode/*
+
+# clangd compiler database
+compile_commands.json
--- a/makefile
+++ b/makefile
@@ -29,7 +29,7 @@ include make/cpplint                      # cpplint
 include make/tests                        # tests
 include make/clang-tidy
 
-INC_FIRST = -I $(if $(STAN),$(STAN)/src,src)
+INC_FIRST = -I $(if $(STAN),$(STAN)/src,src) -I ./src/
 LDLIBS_STANC ?= -Ltest -lstanc
 
 
--- a/src/stan/io/array_var_context.hpp
+++ b/src/stan/io/array_var_context.hpp
@@ -12,6 +12,8 @@
 #include <utility>
 #include <numeric>
 #include <unordered_map>
+#include <functional>
+#include <type_traits>
 
 namespace stan {
 
--- a/src/stan/io/array_var_context.hpp
+++ b/src/stan/io/array_var_context.hpp
@@ -17,7 +17,6 @@ namespace stan {
 
 namespace io {
 
-
 /**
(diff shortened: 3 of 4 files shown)


SteveBronder:
Crap @serban-nicusor-toptal sorry but it looks like that Jenkinsfile was not correct.

https://jenkins.mc-stan.org/job/Stan/job/PR-2806/7/consoleFull

If you can give me access to this repo I can make a branch and fix the error tmrw (Jenkins won't use Jenkinsfile's from forks)

SteveBronder:
If it's a simple fix I can probably even get it tonight

serban-nicusor-toptal:
@SteveBronder No worries! I have reverted the merge and added some comments for the Jenkinsfile, we can correct PR 2805 and merge it again after if that's okay with you.

Issue #2816, opened 2019-09-11
Use Welford's algorithm for Variance Estimates in src/stan/analyze


roualdes:
#### Summary:
Use Welford's algorithm for variance estimates in `src/stan/analyze/autocovariance.hpp` and `src/stan/analyze/compute_potential_scale_reduction.hpp`.

#### Description:
Welford's algorithm to estimate the variance is much less likely to suffer from catastrophic cancellation than the standard sum of squared deviations from the sample mean.  We should Welford's algorithm where possible in `src/stan/analyze`.

#### Current Version:
v2.20.0

roualdes:
Closed due to #2819

PR #2819, opened 2019-09-17
Feature/issue 2816 Welford's Variance in stan/analyze

roualdes:
#### Submission Checklist

#### Summary
Re Issue https://github.com/stan-dev/stan/issues/2816, use Welford's algorithm for computing the sample variance within `stan/analyze`. 

I added `src/stan/analyze/mcmc/welford_variance.hpp` since there are three separate places the variance is calculated. My rule is to write a function at >= 3 replications of the same code: 2 uses in `compute_potential_scale_reduction.hpp` and 1 in `autocovariance.hpp`.

#### Intended Effect
Reduce risk of catastrophic cancellation within `stan/analyze`'s functions.

#### How to Verify
Tests in `src/test/unit/analyze/mcmc/welford_variance_test.cpp`.

All other tests remain the same and pass as expected.

#### Side Effects
None intended.

#### Documentation
Above new function within `src/stan/analyze/mcmc/welford_variance.hpp`.

#### Copyright and Licensing

Please list the copyright holder for the work you are submitting (this will be you or your assignee, such as a university or company): California State University, Chico


bob-carpenter:
std::pair is much preferred over std::vector where possible because it doesn't require a malloc and has statically resolved data location.

betanalpha:
When going through the ESS bug fixes today I saw that the Boost accumulators were already being used in the covariance/correlation chains methods (https://github.com/stan-dev/stan/blob/develop/src/stan/mcmc/chains.hpp), so it might be easiest to just use those directly instead of duplicating the functionality,

roualdes:
Yes, thanks for catching those bugs in ESS.

My understanding of the Boost accumulators is that we'd need to copy draws into an `accumulator_set`.  But I feel like we've worked a good bit towards avoiding copies.  Is the simplicity of the accumulators worth this copy?

bob-carpenter:
We shouldn't need to copy draws---the accumulators feed them in dynamically.  For mean, variance, covariance, etc., only the sufficient stats get stored.  For quantiles, there's no way to compute them without accumulating min(p, 1 - p) draws for the p quantile.  If we have lots of quantiles, it'll probably be better to just sort rather than to run multiple accumulators.

roualdes:
That was helpful, thank you.  Just pushed updates based on the recommendation to use boost accumulators.

betanalpha:
So the boost accumulators will work on in place memory without a copy?

I’m in Mexico teaching for the next two weeks so if anyone else could 
give this a review I would be much appreciative.  If not I’ll try to give it a
look later this week.

bob-carpenter:
> 
> On Sep 23, 2019, at 4:26 AM, Michael Betancourt <notifications@github.com> wrote:
> 
> So the boost accumulators will work on in place memory without a copy?

Only when possible.  For instance, sum just keeps a running sum.  But if you have to compute a 0.5 quantile, you need to store 50% of the data if it gets streamed in like the accumulators do.  And if you need the 0.05 and 0.95 quantile, you need to store the lowest 5% and highest 5% of the data. So you can wind up storing a lot of intermediate data.  The alternative is to make one copy for quantiles and sort.

The doc hints that if you have multiple quantiles it'll be clever about not storing redundantly, but I haven't tried it.  But even so, the top 10% and bottom 10% and 50% quantiles together will need to store 60% of the data.

To verify, you can run a loop that's 10 billion items long in compuing a variance and see if you run out of memory.  That
is, something like:

for (long i = 0; i < 1e10; ++i)
  acc(i % 3);

betanalpha:
So long as the input values themselves aren’t copied then this is
as ideal as you’re going to get and there’s no reason to not use
the Boost accumulators.

betanalpha:
Just to clarify — I am aware of the memory requirements for the accumulators.
I was just trying to confirm whether or not the boost accumulators copy the
input data locally before running the accumulation, as had been hinted at earlier.

I think we’ve confirmed that the accumulators can be run on in-place memory 
in which case we’re good to go.

roualdes:
Let me know if there's anything else you would like to see on this PR.  Thanks.

PR #2823, opened 2019-09-20
Feature/issue 2474 advi diagnostic

drezap:
#### Submission Checklist

#### Summary
This add the Generalized Pareto Distribution Fit to a univariate sample from an arbitrary distribution (i.e. a vector of doubles).

#### Intended Effect
Add new functionality to services that enables VI dianostics.

#### How to Verify
./runTests.py /src/test/unit/khat_test.cpp

#### Side Effects
N/A

#### Documentation
Zhang and Stephens (2009) A New and Efficient Estimation Method for the Generalized Pareto Distribution.

#### Copyright and Licensing

Please list the copyright holder for the work you are submitting (this will be you or your assignee, such as a university or company): Andre Zapico


drezap:
@SteveBronder might you have time to review this one?

I want to rename it to `gdp_fit` and add some documentation, but is there any test cases you think I can write that could make this break?

SteveBronder:
Idt I know this repo well enough to approve code but I can take a look over it

drezap:
yeah man, any feedback you have would be appreciated. I just translated the code from R, to make the job fast/easy.

I could do a simulation to show that it recovers parameters of the GPD, but wouldn't this be redundant if it's returning the same doubles as the R implementation to a certain precision?

betanalpha:
The “services” folder is meant for public API routes that expose
methods that consume entire models, such as the algorithms and 
gradient diagnostics.

Functionality like the GPD k-hat estimator instead belongs in the 
“analyze” folder with signature similar to that used in the effective 
sample size and Rhat functions (i.e passing values by 
std::vector<double*>).  Because the estimator can be used on 
not only samples but also weights I recommend adding a
“analyze/misc” folder and placing this functionality there.

Note also the the k-hat estimator in the Zhang and Stephens 
paper presumes that the inputs are drawn from a smooth density,
and hence do not work on discrete values such as those arising 
from indicator functions.  To maximize utility the estimator should
be robustified to handle cases where the quantiles fall on top of
each other, although this could be left for a future PR.

avehtari:
For all the planned purposes inputs for khat estimator are continuous until we start sampling discrete parameters. PSIS-paper mentions possibility of using it for continuous importance weight multiplied by potentially discrete generated quantity giving continuous result.

bob-carpenter:
> until we start sampling discrete parameters

That should be OK if this is just measuring the log density of the model.  Because that won't include contributions from where we already do sample discrete parameters, e.g.,

```
generated quantities {
  int y_tilde = bernoulli_logit_rng(alpha + beta * x_tilde);
}
```

avehtari:
> That should be OK if this is just measuring the log density of the model. 

Sorry, if I was imprecise. With parameters I meant parameters defined in the parameters block, and not quantities defined in generated quantities block. I thought these were the definitions.

bob-carpenter:
The base model class treats all the saved variables (params, transformed params, generated quantities) like parameters, though the write_array method lets you specify whether to include transformed params and generated quantities in the output.  

I don't think there's any distinction between the three types of variables in the CmdStan .csv output, nor do I think the distinction is recorded in the R stanfit object.  For example, the latter starts with:

  ..@ model_pars: chr [1:4] "a" "b" "c" "lp__"
  ..@ par_dims  :List of 4
  .. ..$ a   : num(0) 
  .. ..$ b   : num(0) 
  .. ..$ c   : num(0) 
  .. ..$ lp__: num(0) 

when a is a parameter, b is a transformed parameter, and c a generated quantity.

I don't know if this'll matter for k-hat, which I thought was just based on the log densities, so sorry if this is all just a pointless digression.

betanalpha:
There needs to be a clear scope of this functionality for
inclusion into Stan.  The k-hat fit is generically useful,
but would have to be eventually robustified to handle
both continuous and discrete outputs that would arise
from generic fits with both HMC and ADVI that return 
samples.

If the goal is only to use this functionality for importance
sampling then that has to be made very clear and its
location set to match that expectation.  Since we don’t
have generic importance sampling in Stan at the moment
it would have to be limited to analyze/advi with the 
function signature (and doc) clearly specifying that the 
function expects weights and not samples.

avehtari:
Planned scope: As Rhat and ESS for MCMC are moved to analyze, it was considered sensible to move importance sampling based khat and ESS for ADVI and Laplace to analyze, too. khat uses Zhang and estimate. Currently in Stan importance sampling based khat estimate can get only continuous inputs. 

Other possible scopes: GPD is continuous distribution and theory that far enough tail most continuous distributions can be modeled well with GDP holds only for continuous distributions. In a same way as other continuous distributions (like normal) are sometimes used to model discrete data we could sometimes used GPD to model discrete data, but it's then users concern to check that the continuous approximation is sensible. Just providing gpd_fit should be then fine. However, if we advertise khat as a generic diagnostic without mentioning that it is OK only for continuous importance ratios and if someone would have discrete parameter space leading to discrete importance ratios, we could add warning to diagnostic part that this diagnostic can fail if there are many non-unique discrete values. I don't think there is need to have that warning in gdp_fit, in a same way as we don't have warnings for using normal distribution for binary data. In addition of importance sampling, GPD might be useful in analysing tails of marginal distributions, although the experiments demonstrated that it's difficult to have something which would be reliable for automatic warnings. In the case of analysing tails, I think there is no need to have that warning in gdp_fit, and instead the warning would the in the general tail diagnostic function.

Thus, I think we don't need to implement code for robustifing for discrete output in gpd_fit. I can't even imagine what that would as we don't have code in normal distribution functions to robustify them discrete outputs. We can say in documentation that GPD is continuous distribution. We can say in the related diagnostic functions that they assume continuous importance ratios and possible robustification or checking whether the importance ratios or other inputs look like discrete would go in those functions.

betanalpha:
Based on experiments I think it should be sufficient to test how the 
order statistics change with M and then return reasonable defaults 
when the order statistics are degenerate, but if this is intended to 
be purely an ADVI diagnostic then that’s not a concern.

As I said originally if this is meant for ADVI only then it should go in
services/analyize/advi, and the doc should note that it is meant 
for importance sampling improvements based on the variational 
and target densities.  In that case no changes to the Zhang et al
estimator would be needed.

avehtari:
>  if this is meant for ADVI only then it should go in services/analyize/advi,

Not just ADVI. khat is already used for Laplace in rstanarm, too. If this would be in the possible future analyze package, it might be used by loo, too.

betanalpha:
Right now the services support the algorithms implemented in
the core code.  Currently there is just ADVI.  If additional algorithms 
are introduced which also serve as fancy importance distributions 
then the analysis directory can be refactored accordingly.  No
need for premature design here.

bbbales2:
> Not just ADVI. khat is already used for Laplace in rstanarm, too. If this would be in the possible future analyze package, it might be used by loo, too.

rstanarm does Laplace now? I don't see it in the cran docs. Is it in the pipeline?

avehtari:
> Right now the services support the algorithms implemented in the core code.

I thought this was supposed to go to analyze instead of services to support also non-Stan users in their diagnostic needs. I guess I have misunderstood the plan. You can decide where to put gpd_fit.

> rstanarm does Laplace now? I don't see it in the cran docs. Is it in the pipeline?

Different Laplace from what Charles is doing. It seems rstanarm submission to cran was rejected, so it will be there in the future.

bob-carpenter:
> On Sep 30, 2019, at 6:59 PM, Ben Bales <notifications@github.com> wrote:
> ...
> rstanarm does Laplace now? I don't see it in the cran docs. Is it in the pipeline?

> On Sep 30, 2019, at 7:08 PM, Aki Vehtari <notifications@github.com> wrote:
> ... 
> Different Laplace from what Charles is doing. It seems rstanarm submission to cran was rejected, so it will be there in the future.


Specifically, RStan gives you the option to return the inverse Hessian at the mode and the option to return draws from the resulting Laplace approximation to the whole posterior.  It's being implemented directly in RStan using finite diffs of gradients.  

It is not implemented as a service in this repo, nor does CmdStan have the option.  I don't know about PyStan(3).

drezap:
I'm interested in looking at the Laplace code for rstan or rstanarm, I couldn't find it with a quick search. Where can I find the branch?

bob-carpenter:
It's in the release version of rstan.  It's turned on when you have draws set to non-zero in the optimizing function.

betanalpha:
> Right now the services support the algorithms implemented in the core code.
> 
> I thought this was supposed to go to analyze instead of services to support also non-Stan users in their diagnostic needs. I guess I have misunderstood the plan. You can decide where to put gpd_fit.
> 
The goal is for the services to expose diagnostics of the algorithms written
in the core code, facilitating comparisons between the Stan implementations 
and non-Stan implementations.

Right now importance sampling is used just for ADVI so that’s the appropriate
route.  If Laplace, or other deterministic algorithms are introduced to the that 
could benefit from the PSIS correction, are introduced to the core code in the
future then we can readily reorganize the analysis code to include a general 
“cont_is” folder or the like.

PR #2825, opened 2019-09-29
Cleanup a few things and add inline everywhere

SteveBronder:
#### Submission Checklist

- [x] Declare copyright holder and open-source license: Steve Bronder

#### Summary

Was reading through the code this weekend and thought I'd tidy up a bit while I was reading along. This is mostly a cleanup PR with 1 or 2 places that actually need attention. Changes are

1. Most functions are given the `inline` keyword. There are a lot of functions that are just one liners pointing to another function. These tend to be pretty good inline candidates so we do that.

2. variables that could be made `const` are given `const`.

3. If we only defined a copy constructor with values for the members unchanged it was removed. Defining a default copy constructor without a move constructor will cause the class to always copy. Removing the copy constructor let's the compiler construct the default copy and move constructors.

4. Access to an `Eigen::Vector` is done through `coeffRef()` which avoids a bounds checks (see [here](https://github.com/stan-dev/stan/compare/cleanup/remove-hamiltonian-nondefault-constructors?expand=1#diff-b20865daa5090b1668966d1f994bd16fL160) for an example).

5. Replaced references of `.at()` for vectors with `operator[]` as `.at()` causes a bounds check (ex [here](https://github.com/stan-dev/stan/compare/cleanup/remove-hamiltonian-nondefault-constructors?expand=1#diff-f77e0518704ff5f9096751d4bb6e3593L58)).

6. When filling a vector of strings `emplace_back()` is called instead of `push_back()`.

7. When only one non-default constructor existed that set default values for the members we use default member initialization directly in the class instead of declaring those defaults in the initialization list (ala [here](https://github.com/stan-dev/stan/compare/cleanup/remove-hamiltonian-nondefault-constructors?expand=1#diff-4eb84e42fa085ef0ffa363cee96699e9L52) and [here](https://github.com/stan-dev/stan/compare/cleanup/remove-hamiltonian-nondefault-constructors?expand=1#diff-238982cc1442b4ec37ff0b0f6e4e223bL13)). This is more "future safety" than anything so that if we define a new constructor for those classes and left out one initial value than it will at least have a default value to fall back to.

8. If we can use range based initializers for vectors and Eigen classes, we do. (ala [here](https://github.com/stan-dev/stan/compare/cleanup/remove-hamiltonian-nondefault-constructors?expand=1#diff-cb0068efe193585ef119932a99775bb7R79)

**The below needs looked at with scrutiny**

8. There was some very weird indexing happening in [`grad_hess_log_prob.hpp`](https://github.com/stan-dev/stan/compare/cleanup/remove-hamiltonian-nondefault-constructors?expand=1#diff-7940738e6b6bc01877b0b43e26a37ff1) where we would grab a reference to a 'row' in the vector representing the hessian. That was weird so I ripped it out and just do normal indexing. This also showed me that we have a very weird memory access pattern when we fill in that vector representing the hessian. I couldn't figure out a nicer access pattern but also don't know how much we use that particular function so I let it go. Also the diagonal is being incremented twice, idk if that's intended or not.

9. The class [`window_adaption`](https://github.com/stan-dev/stan/compare/cleanup/remove-hamiltonian-nondefault-constructors?expand=1#diff-29b7e8878df34d44fbde2e245fdb0e21) was setting the default value of `adapt_next_window_` to `-1`, but `adapt_next_window_` is declared an `unsigned int` so it's default was actually `MAX_UINT` of the operating system. idk what's intended there but that feels like naughty behaviour. I declared `adapt_next_window_` to be `int` and gave it a value of `-1`. Though I feel like that's wrong as well. What should the default value be there?

10. `fast_vector_copy` and `fast_matrix_copy` of [`ps_point`](https://github.com/stan-dev/stan/compare/cleanup/remove-hamiltonian-nondefault-constructors?expand=1#diff-f77e0518704ff5f9096751d4bb6e3593L24) were removed. A copy is a copy is a copy, idt that method was any faster than what the compiler would do

#### Intended Effect

Very teeny optimizations should help the compiler a bit. 

#### How to Verify

Everything here is just a condiment on the sandwich of optimization so I think all of the normal tests passing are fine.

#### Side Effects

Hopefully none

#### Documentation

Just a refactor but happy to add docs in places

#### Copyright and Licensing

### DIFF
--- a/src/stan/command/stanc_helper.hpp
+++ b/src/stan/command/stanc_helper.hpp
@@ -333,7 +333,9 @@ inline int stanc_helper(int argc, const char* argv[], std::ostream* out_stream,
         out.close();
         break;
       }
-      default: { assert(false); }
+      default: {
+        assert(false);
+      }
     }
 
     if (!valid_input) {
--- a/src/stan/command/stanc_helper.hpp
+++ b/src/stan/command/stanc_helper.hpp
@@ -333,9 +333,7 @@ inline int stanc_helper(int argc, const char* argv[], std::ostream* out_stream,
         out.close();
         break;
       }
-      default: {
-        assert(false);
-      }
+      default: { assert(false); }
     }
 
     if (!valid_input) {
--- a/src/stan/lang/ast/fun/returns_type_vis_def.hpp
+++ b/src/stan/lang/ast/fun/returns_type_vis_def.hpp
@@ -56,10 +56,9 @@ bool returns_type_vis::operator()(const no_op_statement& st) const {
 bool returns_type_vis::operator()(const statements& st) const {
   // last statement in sequence must return type
   if (st.statements_.size() == 0) {
-    error_msgs_
-        << ("Expecting return, found"
-            " statement sequence with empty body.")
-        << std::endl;
+    error_msgs_ << ("Expecting return, found"
+                    " statement sequence with empty body.")
+                << std::endl;
     return false;
   }
   return returns_type(return_type_, st.statements_.back(), error_msgs_);
@@ -97,10 +96,9 @@ bool returns_type_vis::operator()(const break_continue_statement& st) const {
 bool returns_type_vis::operator()(const conditional_statement& st) const {
   // all condition bodies must end in appropriate return
   if (st.bodies_.size() != (st.conditions_.size() + 1)) {
-    error_msgs_
-        << ("Expecting return, found conditional"
-            " without final else.")
-        << std::endl;
+    error_msgs_ << ("Expecting return, found conditional"
+                    " without final else.")
+                << std::endl;
     return false;
   }
   for (size_t i = 0; i < st.bodies_.size(); ++i)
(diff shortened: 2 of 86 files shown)


betanalpha:
This is pretty overwhelming!  It would help if the PR was broken up
by main features so I could focus on on the algorithm files and not
hold up anything else.  

The change to int in window_adaptation should be okay, and I agree
that the initialization to unsigned_int is bad, but I’d have to go through 
the logic carefully to verify.

The fast copies in the ps_point were significant speed ups to the full
copy bad when the code was first implemented.  If the Eigen code
has improved or move semantics avoid unnecessary copies entirely 
then I’m happy to update the code, but I would like to see speed
comparisons (especially with varying dimension) to verify this.

SteveBronder:
> This is pretty overwhelming!

Too much! I'm goofing somewhere upstream! Thinking overnight the few pieces of those PR that mattered should probably be their own PRs. I'll open up those today or tomorrow

> If the Eigen code has improved or move semantics avoid unnecessary copies entirely then I’m happy to update the code, but I would like to see speed comparisons (especially with varying dimension) to verify this.

Sure! Sean's perf test thing is pretty nice for that. I didn't look at the git blame to see when that stuff was made but yeah things have improved a good bit over the last few years

PR #2826, opened 2019-09-30
Cleanup for window_adaption unsigned int inititalization to -1

SteveBronder:
#### Submission Checklist

#### Summary

The class `window_adaption` was setting the default value of it's member `adapt_next_window_` to `-1` when `restart()` was called in the constructor. But `adapt_next_window_` is declared an unsigned int so it's default was actually MAX_UINT of the operating system. idk what's intended there but that feels like naughty behaviour. I declared adapt_next_window_ to be int and gave it a value of -1. Though I feel like that's wrong as well. What should the default value be there? Looking at `end_adaptation_window()` maybe it's zero so for first window starts sliding things forward?

This is adds a perfect forwarding constructor for `window_adaption` when the argument is convertible to a string. Then when something like `window_adaption("my_value")` is called that temp just moves on in to the class without making a string copy.

#### Intended Effect

bugfix!

#### How to Verify

Check it out!

#### Side Effects

Is this function used a lot? Looks like it's used in things like `covar_adaptation`. Should probably figure out a good default value for `adapt_next_window_` or there probably could be side effects somewhere

#### Documentation

Just a bug-fix/cleanup, though happy to add docs if needed

#### Copyright and Licensing

### DIFF
--- a/src/stan/mcmc/windowed_adaptation.hpp
+++ b/src/stan/mcmc/windowed_adaptation.hpp
@@ -11,24 +11,19 @@ namespace mcmc {
 
 class windowed_adaptation : public base_adaptation {
  public:
-  explicit windowed_adaptation(std::string name) : estimator_name_(name) {
-    num_warmup_ = 0;
-    adapt_init_buffer_ = 0;
-    adapt_term_buffer_ = 0;
-    adapt_base_window_ = 0;
+  explicit windowed_adaptation(std::string name) : estimator_name_(name) {}
 
-    restart();
-  }
-
-  void restart() {
+  inline void restart() {
     adapt_window_counter_ = 0;
     adapt_window_size_ = adapt_base_window_;
     adapt_next_window_ = adapt_init_buffer_ + adapt_window_size_ - 1;
   }
 
-  void set_window_params(unsigned int num_warmup, unsigned int init_buffer,
-                         unsigned int term_buffer, unsigned int base_window,
-                         callbacks::logger& logger) {
+  inline void set_window_params(unsigned int num_warmup,
+                                unsigned int init_buffer,
+                                unsigned int term_buffer,
+                                unsigned int base_window,
+                                callbacks::logger& logger) {
     if (num_warmup < 20) {
       logger.info("WARNING: No " + estimator_name_ + " estimation is");
       logger.info("         performed for num_warmup < 20");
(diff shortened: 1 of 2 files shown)


src/stan/mcmc/windowed_adaptation.hpp
```diff
@@ -112,14 +107,14 @@ class windowed_adaptation : public base_adaptation {
...
-  unsigned int adapt_base_window_;
+  unsigned int num_warmup_{0};
+  unsigned int adapt_init_buffer_{0};
+  unsigned int adapt_term_buffer_{0};
+  unsigned int adapt_base_window_{0};
 
-  unsigned int adapt_window_counter_;
-  unsigned int adapt_next_window_;
-  unsigned int adapt_window_size_;
+  int adapt_window_counter_{-1};
```
betanalpha: This should default to 0, not -1.

src/stan/mcmc/windowed_adaptation.hpp
```diff
@@ -112,14 +107,14 @@ class windowed_adaptation : public base_adaptation {
...
+  unsigned int adapt_next_window_{0};
```
betanalpha: This is what you were worried about overflowing, but it is still an `unsigned int`.

SteveBronder: Gaaah dumb on my part

SteveBronder:
> The only time adapt_next_window_ can become negative, and hence overflow, is when adapt_init_window_ = adapt_window_size_ = 0, but in that case the adaptation should be entirely the final window where metric adaptation is turned off and adapt_next_window_ = infty is valid logic.

If you look at the [original constructor](https://github.com/stan-dev/stan/pull/2826/files#diff-29b7e8878df34d44fbde2e245fdb0e21L14) that happens at every initialization (when we call `restart()`). Unsigned ints have well defined rollover behavior, but it's just sort of counter-intuitive logic that reasonable defaults would make go away.

SteveBronder:
Glancing at Eigen's git blame it happened somewhere between 11 months and 4 years ago, so 

 ¯\_(ツ)_/¯

but there's def a nonzero chance they did something worse when you first wrote the fast_copy stuff.

https://github.com/eigenteam/eigen-git-mirror/blame/79646971325e2f5f8f679b1e9002743589623e0b/Eigen/src/Core/util/Memory.h#L523

SteveBronder:
@rok-cesnovar is the goof on travis in this pr because of something in the tbb stuff?

rok-cesnovar:
Looks like it, yeah. The Intel TBB PR on Stan did not touch any travis configuration. We got rid of the Travis on cmdstan, not sure if we can also do that here. Otherwise that should also be done. @seantalts @serban-nicusor-toptal

betanalpha:
I think you put this in the wrong PR!

betanalpha:
Let me know when the Travis issue is resolved.

PR #2827, opened 2019-09-30
Cleanup grad_hess_log_prob so weird indexing is more obvious

SteveBronder:
#### Submission Checklist

#### Summary

1. There was some very weird indexing happening in model/grad_hess_log_prob.hpp where we would grab a reference to a 'row' in the vector representing the hessian. That was weird so I ripped it out and just did normal indexing. This also showed me that we have a very weird memory access pattern when we fill in that vector representing the hessian. I couldn't figure out a nicer access pattern but also don't know how much we use that particular function so I let it go. Also the diagonal is being incremented twice, idk if that's intended or not.

2. `half_epsilon * coefficients[i]` used in the 3rd for loop are both known at the beginning of the function so we move them up

#### Intended Effect

Just a little code cleanup, maybe the next person can figure out a better memory access pattern

#### How to Verify

Tests should run as normal

#### Side Effects

Hopefully none!

#### Documentation

Function was already doc'd

#### Copyright and Licensing

### DIFF
--- a/src/stan/model/grad_hess_log_prob.hpp
+++ b/src/stan/model/grad_hess_log_prob.hpp
@@ -43,22 +43,25 @@ double grad_hess_log_prob(const M& model, std::vector<double>& params_r,
       = {-2 * epsilon, -1 * epsilon, epsilon, 2 * epsilon};
   static const double coefficients[order]
       = {1.0 / 12.0, -2.0 / 3.0, 2.0 / 3.0, -1.0 / 12.0};
-
+  static const double half_epsilon_coeff[order]
+      = {half_epsilon * coefficients[0], half_epsilon * coefficients[1],
+         half_epsilon * coefficients[2], half_epsilon * coefficients[3]};
   double result = log_prob_grad<propto, jacobian_adjust_transform>(
       model, params_r, params_i, gradient, msgs);
   hessian.assign(params_r.size() * params_r.size(), 0);
   std::vector<double> temp_grad(params_r.size());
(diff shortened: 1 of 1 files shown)


SteveBronder:
@syclik would you mind if I tagged you on this? Looking at the git blame I think it's just you and I that have touched this file (me with the clang tidy stuff). Idk how much we use this things so if the answer is not much we can also just close this

bob-carpenter:
Rather than accessing by row, you can access by column then transpose at the end.

syclik:
Weird... I got an email notification with a comment from @bob-carpenter, but I don't see it here:
> Rather than accessing by row, you can access by column then transpose at the end.

SteveBronder:
> Rather than accessing by row, you can access by column then transpose at the end

What was that in reference to? Would be nice to have better mem access here

bob-carpenter:
> > Rather than accessing by row, you can access by column then transpose at the end

> What was that in reference to? Would be nice to have better mem access here

I think it might be more efficient in some cases of large `M` and `N` to fill in memory local order (by column)

Suppose we need to fill a large matrix elementwise

```
matrix b(M, N);
```

We can do that by column  so that the elements of `a` are visited in memory order.  

for (int  n = 0; n < N; ++n)
  for (int m = 0; m < M; ++m)
    do-something(a, m, n)
```

We can then transpose at the end using Eigen's block-based transposition function,


```
matrix b = a.transpose();
```

And it feels like we've filled `b` by row, whereas we really filled `a` by column and transposed.

PR #2828, opened 2019-09-30
[WIP] Removed fast copy funcs from ps_point and dependents

SteveBronder:
#### Submission Checklist

#### Summary

fast_vector_copy and fast_matrix_copy of ps_point were removed. Those functions were doing a resize on the lhs vector and then a `memcopy` in. I'll do a speedtest but I'd be suprised if that beat default behavior

#### Intended Effect

Code removal

#### How to Verify

Performance tests tbd

#### Side Effects

Hopefully none!

#### Copyright and Licensing

### DIFF
--- a/src/stan/mcmc/hmc/hamiltonians/ps_point.hpp
+++ b/src/stan/mcmc/hmc/hamiltonians/ps_point.hpp
@@ -16,55 +16,34 @@ using Eigen::Dynamic;
  * Point in a generic phase space
  */
 class ps_point {
-  friend class ps_point_test;
 
  public:
-  explicit ps_point(int n) : q(n), p(n), V(0), g(n) {}
-
-  ps_point(const ps_point& z)
-      : q(z.q.size()), p(z.p.size()), V(z.V), g(z.g.size()) {
-    fast_vector_copy_<double>(q, z.q);
-    fast_vector_copy_<double>(p, z.p);
-    fast_vector_copy_<double>(g, z.g);
-  }
-
-  ps_point& operator=(const ps_point& z) {
-    if (this == &z)
-      return *this;
-
-    fast_vector_copy_<double>(q, z.q);
-
-    V = z.V;
-
-    fast_vector_copy_<double>(p, z.p);
-    fast_vector_copy_<double>(g, z.g);
-
-    return *this;
-  }
+  explicit ps_point(int n) : q(n), p(n), g(n) {}
 
   Eigen::VectorXd q;
   Eigen::VectorXd p;
-
-  double V;
   Eigen::VectorXd g;
+  double V{0};
 
-  virtual void get_param_names(std::vector<std::string>& model_names,
-                               std::vector<std::string>& names) {
+  virtual inline void get_param_names(std::vector<std::string>& model_names,
+                                      std::vector<std::string>& names) {
+    names.reserve(q.size() + p.size() + g.size());
     for (int i = 0; i < q.size(); ++i)
-      names.push_back(model_names.at(i));
-    for (int i = 0; i < q.size(); ++i)
-      names.push_back(std::string("p_") + model_names.at(i));
-    for (int i = 0; i < q.size(); ++i)
-      names.push_back(std::string("g_") + model_names.at(i));
+      names.emplace_back(model_names[i]);
+    for (int i = 0; i < p.size(); ++i)
+      names.emplace_back(std::string("p_") + model_names[i]);
+    for (int i = 0; i < g.size(); ++i)
+      names.emplace_back(std::string("g_") + model_names[i]);
   }
 
-  virtual void get_params(std::vector<double>& values) {
+  virtual inline void get_params(std::vector<double>& values) {
+    values.reserve(q.size() + p.size() + g.size());
     for (int i = 0; i < q.size(); ++i)
-      values.push_back(q(i));
-    for (int i = 0; i < q.size(); ++i)
-      values.push_back(p(i));
-    for (int i = 0; i < q.size(); ++i)
-      values.push_back(g(i));
+      values.push_back(q[i]);
+    for (int i = 0; i < p.size(); ++i)
+      values.push_back(p[i]);
+    for (int i = 0; i < g.size(); ++i)
+      values.push_back(g[i]);
   }
 
   /**
@@ -73,28 +52,6 @@ class ps_point {
    * @param writer writer callback
    */
   virtual inline void write_metric(stan::callbacks::writer& writer) {}
-
- protected:
-  template <typename T>
-  static inline void fast_vector_copy_(
-      Eigen::Matrix<T, Dynamic, 1>& v_to,
-      const Eigen::Matrix<T, Dynamic, 1>& v_from) {
-    int sz = v_from.size();
-    v_to.resize(sz);
-    if (sz > 0)
-      std::memcpy(&v_to(0), &v_from(0), v_from.size() * sizeof(double));
-  }
-
-  template <typename T>
-  static inline void fast_matrix_copy_(
-      Eigen::Matrix<T, Dynamic, Dynamic>& v_to,
-      const Eigen::Matrix<T, Dynamic, Dynamic>& v_from) {
-    int nr = v_from.rows();
-    int nc = v_from.cols();
-    v_to.resize(nr, nc);
-    if (nr > 0 && nc > 0)
-      std::memcpy(&v_to(0), &v_from(0), v_from.size() * sizeof(double));
-  }
 };
 
 }  // namespace mcmc
--- a/src/stan/mcmc/hmc/hamiltonians/ps_point.hpp
+++ b/src/stan/mcmc/hmc/hamiltonians/ps_point.hpp
@@ -16,7 +16,6 @@ using Eigen::Dynamic;
  * Point in a generic phase space
  */
 class ps_point {
-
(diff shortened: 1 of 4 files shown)


src/stan/mcmc/hmc/hamiltonians/ps_point.hpp
```diff
@@ -16,55 +16,33 @@ using Eigen::Dynamic;
...
+  virtual inline void get_param_names(std::vector<std::string>& model_names,
+                                      std::vector<std::string>& names) {
+    names.reserve(q.size() + p.size() + g.size());
     for (int i = 0; i < q.size(); ++i)
-      names.push_back(std::string("g_") + model_names.at(i));
+      names.emplace_back(model_names[i]);
+    for (int i = 0; i < p.size(); ++i)
+      names.emplace_back(std::string("p_") + model_names[i]);
+    for (int i = 0; i < g.size(); ++i)
+      names.emplace_back(std::string("g_") + model_names[i]);
```
SteveBronder: Are `q`, `p`, and `g` always intended to be the same size? If so we can just put a check for that at the top of this and have one for loop

betanalpha: Yeah, they should all be the same size.

betanalpha:
So what happens without the explicit copy constructor?  Are the implicit constructors start enough to avoid shallow copies?  If so are the implicit deep copies sufficiently fast? 

Also there are points in the code where the base ps_point copy constructor is explicitly called (in order to avoid copying the metric and another higher-level information that isn't needed) -- will this behave as expected?

Could you do a simple C++ example demonstrating equivalent (or reasonably similar) assembly or some empirical tests that the default copy constructors are the same speed?  Sorry for the ask, but there was a significant difference when the fast_copy was first introduced and I want to verify that the implicit defaults have improved enough that the fast_copy's can be dropped.

SteveBronder:
> So what happens without the explicit copy constructor? 

implicit constructor is created!

> Are the implicit constructors start enough to avoid shallow copies? 

I'm guessing by start you mean smart and yes if you ask for a copy it will def give you a copy.

> If so are the implicit deep copies sufficiently fast?

Is it fast? Probably! I'll do a godbolt tmrw to show a bit of this. It would be something like

```cpp
  ps_point(const ps_point& z)
      : q(z.q), p(z.p), V(z.V), g(z.g) {}
```

So it matters what eigen does and then what the compiler does based on what Eigen does. Since q, p, and g are eigen vectors we can look at the eigen code and, through the willy wonka ride of CRTP, see that the Matrix copy constructor calls PlainObjectBase's DenseStorage copy constructor [here](https://github.com/eigenteam/eigen-git-mirror/blob/master/Eigen/src/Core/DenseStorage.h#L353). 

```cpp
T *m_data;
//
    EIGEN_DEVICE_FUNC DenseStorage(const DenseStorage& other)
      : m_data(internal::conditional_aligned_new_auto<T,(_Options&DontAlign)==0>(other.m_rows*other.m_cols))
      , m_rows(other.m_rows)
      , m_cols(other.m_cols)
    {
      EIGEN_INTERNAL_DENSE_STORAGE_CTOR_PLUGIN(Index size = m_rows*m_cols)
      internal::smart_copy(other.m_data, other.m_data+other.m_rows*other.m_cols, m_data);
    }
```

LoL, guess what smart_copy does?? 

```
// std::copy is much slower than memcpy, so let's introduce a smart_copy which
// use memcpy on trivial types, i.e., on types that does not require an initialization ctor.
template<typename T, bool UseMemcpy> struct smart_copy_helper;

template<typename T> EIGEN_DEVICE_FUNC void smart_copy(const T* start, const T* end, T* target)
{
  smart_copy_helper<T,!NumTraits<T>::RequireInitialization>::run(start, end, target);
}

template<typename T> struct smart_copy_helper<T,true> {
  EIGEN_DEVICE_FUNC static inline void run(const T* start, const T* end, T* target)
  {
    IntPtr size = IntPtr(end)-IntPtr(start);
    if(size==0) return;
    eigen_internal_assert(start!=0 && end!=0 && target!=0);
    memcpy(target, start, size);
  }
};

template<typename T> struct smart_copy_helper<T,false> {
  EIGEN_DEVICE_FUNC static inline void run(const T* start, const T* end, T* target)
  { std::copy(start, end, target); }
};
```

So either way it's a memcopy, theirs does a few safety checks pre-flight. So more expensive than a raw memcopy, not as expensive as a memmove.


> Also there are points in the code where the base ps_point copy constructor is explicitly called (in order to avoid copying the metric and another higher-level information that isn't needed) -- will this behave as expected?

Can you show me where?

> Could you do a simple C++ example demonstrating equivalent (or reasonably similar) assembly or some empirical tests that the default copy constructors are the same speed? Sorry for the ask, but there was a significant difference when the fast_copy was first introduced and I want to verify that the implicit defaults have improved enough that the fast_copy's can be dropped.

Yeah sure, I should have time this week. I'll either throw something up on godbolt or use the math performance testing repo Sean set up.

Totes no worries though! Performance things can be super weird so it's good to have tests to make sure we don't do anything bonkers

betanalpha:
> LoL, guess what smart_copy does?
> 
> So either way it's a memcopy, theirs does a few safety checks pre-flight. So more expensive than a raw memcopy, not as expensive as a memmove.
> 
That’s reassuring.  I wonder if that changed after the original 
ps_point was written, hence the performance issue.

> Also there are points in the code where the base ps_point copy constructor is explicitly called (in order to avoid copying the metric and another higher-level information that isn't needed) -- will this behave as expected?
> 
> Can you show me where?
> 
For example https://github.com/stan-dev/stan/blob/develop/src/stan/mcmc/hmc/nuts/base_nuts.hpp <https://github.com/stan-dev/stan/blob/develop/src/stan/mcmc/hmc/nuts/base_nuts.hpp> Line 134 and 146.
> Could you do a simple C++ example demonstrating equivalent (or reasonably similar) assembly or some empirical tests that the default copy constructors are the same speed? Sorry for the ask, but there was a significant difference when the fast_copy was first introduced and I want to verify that the implicit defaults have improved enough that the fast_copy's can be dropped.
> 
> Yeah sure, I should have time this week. I'll either throw something up on godbolt or use the math performance testing repo Sean set up.
> 
> Totes no worries though! Performance things can be super weird so it's good to have tests to make sure we don't do anything bonkers
> 
Thanks!  I’m only being so pedantic because I remember 
having going through all the work of a custom fast copy 
because the copying was so slow.  It was sufficiently long
ago that I wouldn’t be surprised if the default have gotten
much better!

SteveBronder:
Not sure if this is sufficient but here's a little godbolt example of (I think) The minimal stuff you need to get an idea of what's going on

https://godbolt.org/z/c6mxa5

The lhs has the code, top using fast_copy and the other the default constructors, the midcondle has both of the raw ASM, and the rhs has the diff between the two ASM outputs.

A couple things to note:

1. The one that uses the default has less ASM, a semi-okay heuristic is that less assembly, faster code. But in this case I'm not sure if it's as good of a heuristic which I'll detail below
2. At line 300 you can see the copy constructor for the fast_copy constructor. This is called from line 117. In the default constructor method the calls to Eigen::DenseStorage are made directly. It's important to note we are trading off the 3 calls to Eigen::DenseStorage for the 3 calls to fast_copy
3. You can see the call to ps_point::fast_vector_copy on line 443
4. DenseStorage's constructor is defined on line 534, which is the same for both
5. All the deleted code in the middle there is some of the resizing stuff and constructors for the Eigen matrices within the fast_copy constructor

There's a more slimmed down version below where the construction call happens on line 40. It's essentially the same thing, just comparing fast_copy right below there to Eigen::DenseStorage (line 103 on the far rhs)

https://godbolt.org/z/ECP_CM

idk, looking at both it looks like the default constructor is nice looking here in terms of the asm. Is that enough or do you want performance tests as well?

SteveBronder:
Huh, I didn't know you could do this but pretty neat! The link below has a godbolt example where the rhs details for each line which compiler optimizations where hit or missed. 


https://godbolt.org/z/Fea9Vt

betanalpha:
> idk, looking at both it looks like the default constructor is nice looking here in terms of the asm. Is that enough or do you want performance tests as well?

No, this looks sufficient to me, especially with the git blame that indicates
that Eigen indeed updated their copy constructors after the ps_point code
was implemented!

betanalpha:
Ah, may have been too fast with the merge.  If it's still a WIP let me know and I'll revert the merge.

src/stan/mcmc/hmc/hamiltonians/ps_point.hpp
```diff
@@ -16,55 +16,33 @@ using Eigen::Dynamic;
...
-    for (int i = 0; i < q.size(); ++i)
-      names.push_back(model_names.at(i));
-    for (int i = 0; i < q.size(); ++i)
-      names.push_back(std::string("p_") + model_names.at(i));
+  virtual inline void get_param_names(std::vector<std::string>& model_names,
+                                      std::vector<std::string>& names) {
+    names.reserve(q.size() + p.size() + g.size());
     for (int i = 0; i < q.size(); ++i)
-      names.push_back(std::string("g_") + model_names.at(i));
+      names.emplace_back(model_names[i]);
```
SteveBronder: Here I accidentally removed the `std::string("g_");`. I made a revert pr so no worries!

PR #2831, opened 2019-10-09
Revert "[WIP] Removed fast copy funcs from ps_point and dependents"

SteveBronder:
Reverts stan-dev/stan#2828

Michael merged it a bit too early and I had a small bug in the code


SteveBronder:
Tbh I think it would be easier if I make a new PR with that hotfix

Issue #2789, opened 2019-07-10
Step size adaptation targets wrong optimality criterion


betanalpha:
#### Summary:

The step size adaptation targets a suboptimal "acceptance statistic".

#### Description:

The requisite reading for what’s going on is https://arxiv.org/abs/1411.6669.   There we showed that in the original Hybrid Monte Carlo scheme, where the numerical Hamiltonian trajectory is used as a Metropolis proposal, the optimal step size is the one that yields an average Metropolis acceptance probability of 0.8.  Both NUTS and the current implementation of Hamiltonian Monte Carlo used in Stan, however, are not Hybrid Monte Carlo schemes. Instead they sample from an entire trajectory whose length is grown dynamically around the initial point. Because of that the above analysis isn’t immediately applicable.

In order to use the above step size optimization criterion analysis one has to come up with an "acceptance statistic" that acts as a proxy for the Metropolis acceptance probability (incidentally this is which is why it’s called “accept_stat” and not “accept_prob” in the sampler parameters). The current version of Stan pretends that each point in a given numerical trajectory could have been a Metropolis proposal and averages over the corresponding Metropolis acceptance probabilities uniformly.

We do not, however, sample states within a numerical trajectory uniformly in the algorithm. Instead we sample them based on their weights, e^{-H(q, p)}. By weighting the hypothetical Metropolis acceptance probabilities uniformly we over emphasize those states that are unlikely to be chosen, and end up with a conservative proxy acceptance statistic.

#### Current Version:
v2.19.1

PR #2836, opened 2019-10-11
Feature/2789 improved stepsize adapt target

betanalpha:
#### Submission Checklist

# Summary

Addresses #2789 by updating the acceptance statistic used in the step size adaptation.

# Intended Effect

Yields more effective samples per gradient evaluation for models with dimensions more than O(10). Exact threshold depends on correlations in the target density.

This new branch considers each state as a proxy proposal but weights the corresponding acceptance probabilities by the state weight, e^{-H(q, p)}. In this way those states that are likely to be chosen have a greater influence on the proxy acceptance statistic, and any state with vanishing weight is not considered at all. This will always lead to a higher proxy acceptance stat for a given step size, which will cause the adaptation to push toward higher step sizes and, for a fixed integration time, cheaper numerical trajectories.

# How to Verify

Compare effective sample size per gradient evaluation against develop.

See also the suite of test results in the report attached to https://discourse.mc-stan.org/t/request-for-volunteers-to-test-adaptation-tweak/9532/53.

# Side Effects

Pushing the step size to larger values makes the generalized No-U-Turn criterion less accurate. Combined with the multiplicative trajectory expansion this can result in a heavy tail of longer numerical trajectories, especially in lower dimensions.

# Documentation

Inline.

# Copyright and Licensing

Please list the copyright holder for the work you are submitting (this will be you or your assignee, such as a university or company): Michael Betancourt

Code: BSD 3-clause (https://opensource.org/licenses/BSD-3-Clause)
Documentation: CC-BY 4.0 (https://creativecommons.org/licenses/by/4.0/)


betanalpha:
Verified no compromise in sampler accuracy on 1e7 iteration long chains targeting 50 dim iid unit normal, 50 dim iid unit Cauchy (unit_e, checking probability between -1 and 1), and lower dim correlated normal (checking means, variances, and correlations). Everything looks good, complementing the performance improvement tests already run.

@seantalts I’m going to approve this as algo lead, can you give it a formal approval to let me merge it?  Thanks.

seantalts:
You're lucky it's only 5:33pm Anywhere on Earth: https://www.timeanddate.com/time/zones/aoe :P

PR #2845, opened 2019-11-04
[WIP] Inline mcmc methods and a few other cleanups

SteveBronder:
#### Submission Checklist

#### Summary

I'm going through in another PR and using the `require_*` metaprogramming stuff to clean things up. Before I touch any of the mcmc things I wanted to clean the things below up first. There's a couple smallish things here which touch a lot of the `mcmc` stuff

1. A lot of the functions in `mcmc` are inline candidates so I went through and made everything `inline`

2. Instead of declaring defaults during the constructor 

```cpp
base_adapter() : adapt_flag_(false) {}
```

We remove that and use the braced default initialization scheme

```cpp
bool adapt_flag_{false};
```

3. If a destructor is declared in a class, we also declare the move and copy assignments and constructors ala

```cpp
  ~base_mcmc() = default;
  base_mcmc(const base_mcmc& other) = default;
  base_mcmc(base_mcmc&& other) = default;
  base_mcmc& operator=(const base_mcmc&) = default;
  base_mcmc& operator=(base_mcmc&&) = default;
```

4. Make const correct setters and getters for class methods that access or manipulate member values

```cpp
  // accessor
  inline const std::string& param_name(int j) const { return param_names_(j); }
  // mutator
  inline std::string& param_name(int j) { return param_names_(j); }
```

In the same vein, I removed the a lot of the `set_` and get methods such as for `max_num_fixed_point`. `max_num_fixed_point` was defined as

```cpp
  void set_max_num_fixed_point(int n) {
    if (n > 0)
      this->max_num_fixed_point_ = n;
  }
```

Which seems odd. If n <= 0 is bad then shouldn't this throw a warning or error? I opted to just remove it and make a standard accessor and mutator

```cpp
  inline int& max_num_fixed_point() { return this->max_num_fixed_point_; }
  inline const int max_num_fixed_point() const { return this->max_num_fixed_point_; }
```

I did the above for a lot of the set and get methods in mcmc. So now things like `get_mu()` and `set_mu()` just use `mu()` ala

```cpp
// access
auto val = thing.mu()
// mutate
thing.mu() = 12;
```

5. Not really sure why but we were calling the `operator=` for ps_point explicitly in base_hmc, I just change that so it's

```cpp
      this->z_ = z_init; 
```

6. I _h8_ in/out params. The devils params as I personally call them. I fixed `complete_adaptation()` and `learn_stepsize()` so they do a return instead of an in/out

```cpp
this->stepsize_adaptation_.complete_adaptation(this->nom_epsilon_);
```
becomes
```cpp
this->nom_epsilon_ = this->stepsize_adaptation_.complete_adaptation();
```

#### Intended Effect

Just a cleanup

#### How to Verify

Running base tests, open to adding any additional tests if anyone things we need them from this PR

#### Side Effects

There should be no unintended side effects. Perhaps better inlining

#### Documentation

While I'm cleaning these up should I start writing docs or should that go in a separate PR?

#### Copyright and Licensing

### DIFF
--- a/src/stan/command/stanc_helper.hpp
+++ b/src/stan/command/stanc_helper.hpp
@@ -333,7 +333,9 @@ inline int stanc_helper(int argc, const char* argv[], std::ostream* out_stream,
         out.close();
         break;
       }
-      default: { assert(false); }
+      default: {
+        assert(false);
+      }
     }
 
     if (!valid_input) {
--- a/src/stan/command/stanc_helper.hpp
+++ b/src/stan/command/stanc_helper.hpp
@@ -333,9 +333,7 @@ inline int stanc_helper(int argc, const char* argv[], std::ostream* out_stream,
         out.close();
         break;
       }
-      default: {
-        assert(false);
-      }
+      default: { assert(false); }
     }
 
     if (!valid_input) {
--- a/src/stan/lang/ast/fun/returns_type_vis_def.hpp
+++ b/src/stan/lang/ast/fun/returns_type_vis_def.hpp
@@ -56,10 +56,9 @@ bool returns_type_vis::operator()(const no_op_statement& st) const {
 bool returns_type_vis::operator()(const statements& st) const {
   // last statement in sequence must return type
   if (st.statements_.size() == 0) {
-    error_msgs_
-        << ("Expecting return, found"
-            " statement sequence with empty body.")
-        << std::endl;
+    error_msgs_ << ("Expecting return, found"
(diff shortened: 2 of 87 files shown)


bob-carpenter:
I can review this when it's passing tests.

SteveBronder:
ty! I'll ping you when it's ready for review. I still have to add docs to some things I added

SteveBronder:
Actually I think I'm going to break this up into multiple PRs because there's a ton of very teeny changes

PR #2850, opened 2019-11-09
Make the base integrators into CRTP base classes instead of virtual

SteveBronder:
#### Submission Checklist

#### Summary

This makes the base classes `base_integrator` and `base_leapfrog` use the curiously recuring template pattern. This allows the classes to resolve method lookups at compile time instead of at runtime.

This PR is more so to check that people are cool with this pattern before it is done anywhere else in the mcmc folder. Looking over those it seems like the stuff in hamiltonian and pretty much all the other `base_` classes could use this pattern.

#### Intended Effect

Resolve method lookups at compile time

#### How to Verify

This is refactor but happy to add any additional tests if wanted

```bash
./runTests.py -j12 ./src/test/unit/mcmc/
```

#### Side Effects

If a developer wanted to write a method that takes in anything that derives from `base_leapfrog` they use to write

```cpp
template <typename Hamiltonian>
void my_func(stan::mcmc::base_leapfrog<Hamiltonian>, ...)
```

but now they would have to write

```cpp
template <typename Derived, typename Hamiltonian>
void my_func(stan::mcmc::base_leapfrog<Derived<Hamiltonian>, Hamiltonian>
```

I tried making it so that we could just have `Derived` without the hamiltonian on the RHS of the template, but this errors because the `Derived` will be an incomplete state when the crtp base class attempts to see things like the derived classes hamiltonian type. One alt is to use a `traits` metaclass like eigen does, but I'm not sure if that solves the code smell probably. Though if we want to use this style in the future we may want something like that.

https://github.com/eigenteam/eigen-git-mirror/blob/master/Eigen/src/Core/DenseBase.h#L43

#### Documentation

Wrote some docs for the base integrator and leapfrog to explain their `Derived` template args.

#### Copyright and Licensing

### DIFF
--- a/src/stan/command/stanc_helper.hpp
+++ b/src/stan/command/stanc_helper.hpp
@@ -333,7 +333,9 @@ inline int stanc_helper(int argc, const char* argv[], std::ostream* out_stream,
         out.close();
         break;
       }
-      default: { assert(false); }
+      default: {
+        assert(false);
+      }
     }
 
     if (!valid_input) {
--- a/src/stan/command/stanc_helper.hpp
+++ b/src/stan/command/stanc_helper.hpp
@@ -333,9 +333,7 @@ inline int stanc_helper(int argc, const char* argv[], std::ostream* out_stream,
         out.close();
         break;
       }
-      default: {
-        assert(false);
-      }
+      default: { assert(false); }
     }
 
     if (!valid_input) {
--- a/src/stan/lang/ast/fun/returns_type_vis_def.hpp
+++ b/src/stan/lang/ast/fun/returns_type_vis_def.hpp
@@ -56,10 +56,9 @@ bool returns_type_vis::operator()(const no_op_statement& st) const {
 bool returns_type_vis::operator()(const statements& st) const {
   // last statement in sequence must return type
   if (st.statements_.size() == 0) {
-    error_msgs_
-        << ("Expecting return, found"
-            " statement sequence with empty body.")
-        << std::endl;
+    error_msgs_ << ("Expecting return, found"
+                    " statement sequence with empty body.")
+                << std::endl;
     return false;
   }
   return returns_type(return_type_, st.statements_.back(), error_msgs_);
@@ -97,10 +96,9 @@ bool returns_type_vis::operator()(const break_continue_statement& st) const {
 bool returns_type_vis::operator()(const conditional_statement& st) const {
   // all condition bodies must end in appropriate return
   if (st.bodies_.size() != (st.conditions_.size() + 1)) {
-    error_msgs_
-        << ("Expecting return, found conditional"
-            " without final else.")
-        << std::endl;
+    error_msgs_ << ("Expecting return, found conditional"
+                    " without final else.")
+                << std::endl;
     return false;
   }
   for (size_t i = 0; i < st.bodies_.size(); ++i)
--- a/src/stan/lang/ast/fun/returns_type_vis_def.hpp
+++ b/src/stan/lang/ast/fun/returns_type_vis_def.hpp
@@ -56,9 +56,10 @@ bool returns_type_vis::operator()(const no_op_statement& st) const {
 bool returns_type_vis::operator()(const statements& st) const {
   // last statement in sequence must return type
   if (st.statements_.size() == 0) {
-    error_msgs_ << ("Expecting return, found"
-                    " statement sequence with empty body.")
-                << std::endl;
+    error_msgs_
+        << ("Expecting return, found"
+            " statement sequence with empty body.")
+        << std::endl;
     return false;
   }
   return returns_type(return_type_, st.statements_.back(), error_msgs_);
@@ -96,9 +97,10 @@ bool returns_type_vis::operator()(const break_continue_statement& st) const {
 bool returns_type_vis::operator()(const conditional_statement& st) const {
   // all condition bodies must end in appropriate return
   if (st.bodies_.size() != (st.conditions_.size() + 1)) {
-    error_msgs_ << ("Expecting return, found conditional"
-                    " without final else.")
-                << std::endl;
+    error_msgs_
+        << ("Expecting return, found conditional"
+            " without final else.")
+        << std::endl;
     return false;
   }
   for (size_t i = 0; i < st.bodies_.size(); ++i)
--- a/src/stan/lang/grammars/block_var_decls_grammar_def.hpp
+++ b/src/stan/lang/grammars/block_var_decls_grammar_def.hpp
@@ -124,7 +124,7 @@ block_var_decls_grammar<Iterator>::block_var_decls_grammar(
   single_var_decl_r.name("single-element block var declaration");
   single_var_decl_r %= element_type_r(_r1) > identifier_r > opt_def_r(_r1)
                        > eps[validate_single_block_var_decl_f(
-                             _val, _pass, boost::phoenix::ref(error_msgs_))];
+                           _val, _pass, boost::phoenix::ref(error_msgs_))];
 
   element_type_r.name("block var element type declaration");
   element_type_r
@@ -245,13 +245,13 @@ block_var_decls_grammar<Iterator>::block_var_decls_grammar(
       = lit('<')[empty_range_f(_val, boost::phoenix::ref(error_msgs_))]
         > (((lit("lower") > lit('=')
              > expression07_g(_r1)[set_int_range_lower_f(
-                   _val, _1, _pass, boost::phoenix::ref(error_msgs_))])
+                 _val, _1, _pass, boost::phoenix::ref(error_msgs_))])
             > -(lit(',') > lit("upper") > lit('=')
                 > expression07_g(_r1)[set_int_range_upper_f(
-                      _val, _1, _pass, boost::phoenix::ref(error_msgs_))]))
+                    _val, _1, _pass, boost::phoenix::ref(error_msgs_))]))
            | (lit("upper") > lit('=')
               > expression07_g(_r1)[set_int_range_upper_f(
-                    _val, _1, _pass, boost::phoenix::ref(error_msgs_))]))
+                  _val, _1, _pass, boost::phoenix::ref(error_msgs_))]))
         > lit('>');
 
   // _r1 var scope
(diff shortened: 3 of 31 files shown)


bgoodri:
Is this going to cause problems in the future if I don't merge it into StanHeaders right now? Like when I try to upload StanHeaders 2.22 with even more CRTP but rstan 2.21 on CRAN was compiled with all these virtual classes?

wds15:
I don’t think that crtp is an issue...the question is if the Stan model sources change. In any case, I am also worried about incompatibilities...we need RStan in the test pipeline. 

However, for RStan we should keep that other packages using Stan do not use the pre compilation stuff! Otherwise we run in serious trouble. Packages like brms can obviously use precompilAtion as this requires source compatibility at the Stan source level only. We should probably think this through at some time soon (sorry this is off pr)

SteveBronder:
Good Q I didnt think of that! I'm not super familiar with how a precompiled binary expecting virtual methods would interact with the crtp scheme

Shooting from the hip, I think it would be fine? Since the definitions now look like

```cop
inline void begin_update_p(point_type& z, hamiltonian_type& hamiltonian, double epsilon, callbacks::logger& logger) { 
this->derived().begin_update_p(z, hamiltonian, epsilon, logger);
 }
```

So it feels like a method declared virtual should still get called just fine (since by the definition above it has to call the derived classes method).

My only concern is whether something weird would happen with the vptr table of the precompiled binary expecting to see a virtual method

Should we wait to merge this till the 2.20 stuff in rstan is good to go? Once all of that is finalized it will probably be easier to check if this goofs somewhere

SteveBronder:
Also agree with @wds15 would be very cool and good to put rstan in the pipeline. Maybe not a pr failure but a warning posted back to the pr that this breaks something in rstan so we are aware ahead of time

bgoodri:
I was actually worried that if

0. I upload StanHeaders 2.21 to CRAN this week where the services have all these virtual methods
1. I upload rstan 2.21 to CRAN this week with `stan_fit.cpp` compiling all the services into `command()`
2. Stan 2.22 gets released in January or whatever with these PRs
3. I want to upload StanHeaders 2.22 to CRAN but rstan 2.21 on CRAN has `stan_fit.cpp`  that is expecting all these virtual methods that have since been replaced with CRTP stuff

In that case, rstan 2.21 (and thus almost everything else) would not compile against StanHeaders 2.22. 

There actually is a pipeline for RStan but it has never passed because the Jenkins user does not have permission to install packages. Also, it does not check the packages downstream of rstan, which now number 90.

SteveBronder:
> I upload rstan 2.21 to CRAN this week with stan_fit.cpp compiling all the services into command()

Do you mean they are compiled from source by the user or they are compiled by CRAN to binaries? How does CRAN handle updates of dependencies? In my mindplace I'm guessing that CRAN recompiles the binaries if a dependency is updated? I'm looking at the stan_fit.cpp file below and it just includes these so idt there would be an issue. However if there is precompiled stuff that does not get recompiled then yeah this would break the one definition rule and throw an error

https://github.com/stan-dev/rstan/blob/6a38e8e2526d5483680b10a85cdc95f6e6dc1fd4/rstan/rstan/inst/include/rstan/stan_fit.hpp

bgoodri:
CRAN does not recompile binaries when a dependency gets updated. But they
do essentially expect that downstream packages continue to install and pass
their unit checks when a upstream package is updated. So, I'm a bit worried
about putting things into rstan in the next few days that might get broken
in a few months.

SteveBronder:
My main q is, does rstan ever link against the files in integrators twice? Cuz I think the only time this would break is if the following steps happened

1. rstan is compiled
2. stanheaders is updated with this stuff
3. rstan tries to link against the things in mcmc (i.e. like the integrators)

Then the one definition rule is broken and compilation would fail. 

Have their been times in the past where changing a signature in one of these folders caused rstan to have this issue? If yes then my answer is yes

If that's the case then what works best for you? I can slam out the rest of these changes tmrw and probs have them in this week if that works for you (and whomever else)

imo I'd like to have these changes. Pretty sure removing virtualness will give the compiler more optimization opportunities

bgoodri:
I would say if you can get them reviewed in the next day or to, I can just
merge them into Stanheaders 2.21 and we don't have to worry about 2.22.
Otherwise we'll have to manage.

SteveBronder:
Word word I'll try to have these done tmrw, gonna be up to review folks tho for when they actually get merged

SteveBronder:
@serban-nicusor-toptal I can't click the link for jenkins to see what's failing, is jenkins doing stuff atm or is this a goof somewhere?

wds15:
BTW... what is the motivation to use CRTP here? I mean, the virtual function calls are not ideal - true, but it really does not matter at all performance wise (or I would be very surprised if it does make a difference). We spend our time in calculating the log_prob and it's gradient; not virtual function calls of the services.

I am just wondering where this refactor is coming from and why we need it. If "removing virtual function calls" is per se a reason to refactor... ok, so be it; if that's our policy.

bob-carpenter:
We should only be *optimizing* if there's a bottleneck.  I doubt there's one here.  

The CRTP code's more complex, so all things being equal, it needs to be motivated.

I see two benefits from removing virtuals---static dispatch (efficiency) and reduced object size (more efficiency).

CRTP should also play nicely with template parameters on methods, whereas templating doesn't mix with OO/virtuals.

SteveBronder:
> We should only be *optimizing* if there's a bottleneck.  I doubt there's one here.

Yes agree, though I'd argue the virtualness of these could be a bottleneck for the compiler to see other optimizations.


> BTW... what is the motivation to use CRTP here? I mean, the virtual function calls are not ideal - true, but it really does not matter at all performance wise (or I would be very surprised if it does make a difference). We spend our time in calculating the log_prob and it's gradient; not virtual function calls of the services.

> I am just wondering where this refactor is coming from and why we need it. If "removing virtual function calls" is per se a reason to refactor... ok, so be it; if that's our policy.


So, I have decently large'ish Stan model right now that uses the gpu glm method and that's essentially the only thing in the model class. And with Rok's stanc3 PR we have the data already moved over onto the GPU before the model starts doing anything so we don't pay POD data transfer costs.

On each iteration it's spinning up and down about two GBs of memory. That's a lot, and since the gpu stuff is happening for math imo I think most of the time at this point is actually happening here in algorithms. So there's a couple easy'ish layups we can do to make things go faster here

1. Remove virtualness
2. inline all this stuff
3. remove temporaries (by hand or by perfect forwarding)

So this is (1), I have another PR for (2) and (3)

The other bigger thing we could think about is re-writing ps_point, which tends to do the most allocations when I use [`heaptrack`](https://github.com/KDE/heaptrack) on the example programs in stat_comp_benchmark

wds15:
Can u demonstrate the utility of going CRTP with the services?

Sorry for being a pain here, but I can't see how the services get in the way for good performance. I am really happy to be convinced otherwise.

SteveBronder:
I have 3 branches that make the virtual methods into crtp classes, let me see if I can slap all of those together into one and run the benchmarks

SteveBronder:
Eh looks like it's not more performant

```bash
SUCCESS: Gold golds/stat_comp_benchmarks_benchmarks_arK_arK.gold passed.
+ ./comparePerformance.py develop_performance.csv performance.csv
('stat_comp_benchmarks/benchmarks/gp_pois_regr/gp_pois_regr.stan', 1.04)
('stat_comp_benchmarks/benchmarks/low_dim_corr_gauss/low_dim_corr_gauss.stan', 0.82)
('stat_comp_benchmarks/benchmarks/arK/arK.stan', 0.95)
('stat_comp_benchmarks/benchmarks/pkpd/one_comp_mm_elim_abs.stan', 0.97)
('stat_comp_benchmarks/benchmarks/eight_schools/eight_schools.stan', 0.99)
('stat_comp_benchmarks/benchmarks/irt_2pl/irt_2pl.stan', 0.99)
('stat_comp_benchmarks/benchmarks/gp_regr/gp_regr.stan', 0.95)
('performance.compilation', 1.21)
('stat_comp_benchmarks/benchmarks/low_dim_gauss_mix_collapse/low_dim_gauss_mix_collapse.stan', 1.01)
('stat_comp_benchmarks/benchmarks/low_dim_gauss_mix/low_dim_gauss_mix.stan', 0.92)
('stat_comp_benchmarks/benchmarks/sir/sir.stan', 1.04)
('stat_comp_benchmarks/benchmarks/pkpd/sim_one_comp_mm_elim_abs.stan', 1.07)
('stat_comp_benchmarks/benchmarks/garch/garch.stan', 1.14)
('stat_comp_benchmarks/benchmarks/gp_regr/gen_gp_data.stan', 0.94)
('stat_comp_benchmarks/benchmarks/arma/arma.stan', 0.87)
0.993467746661
```

tbh I just don't like virtual methods. They just kind of make things look odd and unless your doing something like PIMPL kind of unnecessary

but yeah if this is going to goof things more than it's worth it's fine to close it

SteveBronder:
It's also that virtual methods can't use templates as nicely

betanalpha:
I’m not surprised that there isn’t a huge performance difference — all of the HMC code
already templates out on specific classes to avoid any virtual table lookups.   Really 
the virtual methods are there partly due to inertia and partly due to serve as a weak 
definition of the object concept.  

The virtual functions could all be dropped without affecting how the code actually 
runs, although we arguably should have proper concept definitions for all of the 
templates to replace them.

If anything the CRTP should slow down compilation slightly, although that might not
be noticeable if that all can be resolved in the initial compile and not the model-dependent
compiles.

SteveBronder:
> The virtual functions could all be dropped without affecting how the code actually
runs, although we arguably should have proper concept definitions for all of the
templates to replace them.

If you drop `virtual` from `transition` in base_mcmc and a few other ones everything goes belly up for me. But since we compile at -03 those might get devirtualized tho'. We could just slap a `final` onto the non `base_*` nuts, classic nuts, etc. to give the compiler more incentive to devirtualize

I can't think of a reason crtp would cause a slowdown, It moves the lookup cost from runtime to compile time. I like the crtp stuff for how it defines the concept of an algorithms interface. like an mcmc class needs to have a method for performing transitions, a nuts algorithm needs a way to build a tree, yada yada

but again if it's going to be a hassle for rstan and company then I'll close this stuff

PR #2856, opened 2019-12-01
Stepsize adaptation and adapters const accessors (NFC)

kedartal:
#### Summary

Add `const noexcept` accessors to the stepsize adaptation class (for accessing delta, kappa, ...) and in the various adapters (for accessing the adaptation object).

#### Intended Effect

No functional change for Stan users.
Allow developers to query the adaptation parameters in contexts where it's natural to have a `const` sampler (logging, serialization, ...).

#### How to Verify

N/A (trivial)

#### Side Effects

None. 

#### Documentation

N/A (trivial)

#### Copyright and Licensing

Tal Kedar

### DIFF
--- a/src/stan/mcmc/stepsize_adaptation.hpp
+++ b/src/stan/mcmc/stepsize_adaptation.hpp
@@ -36,15 +36,15 @@ class stepsize_adaptation : public base_adaptation {
       t0_ = t;
   }
 
-  double get_mu() { return mu_; }
+  double get_mu() const noexcept { return mu_; }
 
-  double get_delta() { return delta_; }
+  double get_delta() const noexcept { return delta_; }
 
-  double get_gamma() { return gamma_; }
+  double get_gamma() const noexcept { return gamma_; }
 
-  double get_kappa() { return kappa_; }
+  double get_kappa() const noexcept { return kappa_; }
 
-  double get_t0() { return t0_; }
+  double get_t0() const noexcept { return t0_; }
 
   void restart() {
     counter_ = 0;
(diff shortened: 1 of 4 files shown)


SteveBronder:
Nice ty!

 Is there a reason you deleted the non-const mutator? You usually have both for const correctness 

https://isocpp.org/wiki/faq/const-correctness#const-overloading

 I think the original intent was to have only the setter methods for setting stuff so yours makes sense in that context

kedartal:
@SteveBronder
In the `stepsize_adaptation` class the `get_XXX` methods return by value, so one couldn't mutate through those methods even before my patch -- there are equivalent `set_XXX` methods for that. I don't see the point in having a non-`const` overload that returns a _copy_ of a class member, what am I missing?
In the three adapter classes I have not removed the non-`const` overloads since those indeed return a reference, which in turn may be mutated.

kedartal:
Created an issue for this at https://github.com/stan-dev/stan/issues/2857

bob-carpenter:
@kedartal is right---it's just returning a copy, so there's no need for a non-const version.  

The others return constant references, which is fine when there weren't accessors before.

@SteveBronder --- I'll let you review this in case I'm missing something.

SteveBronder:
Word! On the road today but I'll look at this tonight

SteveBronder:
This failed because of windows not being able to do something with the tbb, I kicked things off again which will hopefully make it fine

SteveBronder:
@Bob-carpenter this is good to merge. Thought I had write access here but I don't

Issue #1479, opened 2015-06-03
remove unconventional of Eigen from chains.hpp


bob-carpenter:
Right now, `mcmc/chains.hpp` is abusing Eigen to create Eigen vectors of `std::string` and Eigen vectors of Eigen vectors.
- [x] replace Eigen vectors of strings with `std::vector<std::string>`
- [ ] replace Eigen vector of vectors with Eigen matrix
- [ ] review rest of code so that Eigen is only used for vectors or matrices of double values
- [ ] use Eigen::Map where possible to allow for use without allocating memory
- [ ] include an interrupt callback

PR #2863, opened 2019-12-14
Replace Eigen vectors of strings with std::vector<std::string>

mcol:
#### Submission Checklist

#### Summary
This changes the `Eigen::Matrix<std::string, Dynamic, 1>` containers to `std::vector<std::string>` for `stan::mcmc::chains`, as suggested in the first point of #1479. 

#### How to Verify
Current tests should pass.

#### Side Effects
It's not clear to me if other interfaces use this header directly (at least rstan shouldn't be affected, from by inspection), and what guarantees we make about not chaning this.

The only potentially breaking change is that now `param_names()` returns a `const std::vector<std::string>&` rather than `const Eigen::Matrix<std::string, Dynamic, 1>&`, which means that indexing the resulting container will have to use `operator[]` instead of `operator()`.

#### Copyright and Licensing

Marco Colombo

### DIFF
--- a/src/stan/mcmc/chains.hpp
+++ b/src/stan/mcmc/chains.hpp
@@ -47,7 +47,7 @@ using Eigen::Dynamic;
 template <class RNG = boost::random::ecuyer1988>
 class chains {
  private:
-  Eigen::Matrix<std::string, Dynamic, 1> param_names_;
+  std::vector<std::string> param_names_;
   Eigen::Matrix<Eigen::MatrixXd, Dynamic, 1> samples_;
   Eigen::VectorXi warmup_;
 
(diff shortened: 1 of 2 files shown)


src/stan/mcmc/chains.hpp
```diff
@@ -380,10 +378,12 @@ class chains {
       throw std::invalid_argument(
           "add(stan_csv): number of columns in"
           " sample does not match chains");
-    if (!param_names_.cwiseEqual(stan_csv.header).all()) {
-      throw std::invalid_argument(
-          "add(stan_csv): header does not match"
-          " chain's header");
+    for (int i = 0; i < num_params(); i++) {
+      if (param_names_[i] != stan_csv.header(i)) {
+        throw std::invalid_argument(
```
bob-carpenter: [optional]
This is only optional because it's feature creep on the actual issue being fixed.  

We should build up a message including the parameter names that don't match and the index `i`.  That can be done with a stringstream.  It's acceptable in terms of performance to format a message within the body of the conditional where an exception will be thrown.

mcol: I went ahead and changed the error message while I was touching this code.

mcol:
I'm not authorized to merge, and in any case my last push should be reviewed.

bob-carpenter:
You have maintainer permission on the stan-dev repos, so you should be able to merge if there's not an outstanding review required.  GitHub won't let you review your own PR.

PR #2873, opened 2020-01-03
Store ps_points during transition iterations

SteveBronder:
#### Submission Checklist

#### Summary

This is built on top of #2871 

I was running [`heaptrack`](https://github.com/KDE/heaptrack) and found there were a lot of creations of ps_points happening. This makes sense because a ps_point is normally at least 3 eigen matrices equal to the size of the number of parameters and in the dense case also holds an N by N matrix. This pulls out the ps_points from the `transition` and `build_tree` member function of `base_nuts` and stores them as protected members. I think this will help models that either (1) have a large amount of parameters or (2) have a high tree depth 

#### Intended Effect

Reuse `ps_point` in `transition` and `build_tree`

#### How to Verify

Standard tests

#### Side Effects

This adds additional state to `base_nuts` so there is certainly a tradeoff. Though I think it's possible to argue that `base_nuts` was already stateful since the `z_` member is used frequently in `transition`

#### Copyright and Licensing

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,7 @@ before_script:
 linux_clang: &linux_clang
   os: linux
   compiler: "clang-6.0"
+  dist: trusty
   addons:
     apt:
       sources: [ 'ubuntu-toolchain-r-test', 'llvm-toolchain-trusty-6.0' ]
--- a/make/cpplint
+++ b/make/cpplint
@@ -10,4 +10,4 @@ endif
 
 .PHONY: cpplint
 cpplint:
-	@$(PYTHON2) $(CPPLINT)/cpplint.py --output=vs7 --counting=detailed --root=src --extension=hpp,cpp --filter=-runtime/indentation_namespace,-readability/namespace,-legal/copyright,-whitespace/indent,-runtime/reference,-build/c++11 $(shell find src/stan -name '*.hpp' -o -name '*.cpp')
+	@$(PYTHON2) $(CPPLINT)/cpplint.py --output=vs7 --counting=detailed --root=src --extension=hpp,cpp --filter=-runtime/indentation_namespace,-readability/namespace,-legal/copyright,-whitespace/indent,-runtime/reference,-build/c++11,-whitespace/operators $(shell find src/stan -name '*.hpp' -o -name '*.cpp')
--- a/src/stan/io/array_var_context.hpp
+++ b/src/stan/io/array_var_context.hpp
@@ -338,6 +338,21 @@ class array_var_context : public var_context {
   }
 
   /**
+   * Check variable dimensions against variable declaration.
+   * Only used for data read in from file.
+   *
+   * @param stage stan program processing stage
+   * @param name variable name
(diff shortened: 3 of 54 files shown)


betanalpha:
At a higher level have you done any testing to identify circumstances 
where caching improves performance?  

Most of the objects you moved to class members are initialized only
once at the beginning of the transition and should be pretty cheap 
relative to all of the other dynamic allocation that happens during the
tree building.

One important detail is that “z” being used in the current code are
ps_point instances, which do not store metrics or anything else.
They just store the parameter values, momenta values, and gradient 
values which is the minimal set of information to continue evolving 
trajectories from those points but nothing more.  The PR updates 
those to derived ps_points which isn’t needed.

Most importantly the sampler shouldn’t be giving valid inferences 
with the changes in the PR.  Because build_tree is called recursively
there isn’t a single z_propose_final in memory at any given time
but rather many of them.  Forcing all the recursive calls to use the
same memory should overwrite important information and yield 
incorrect samples.

And what’s with the templating of build_tree?  I don’t see how the
added generality is used.

The basic code is designed around the gradient being the most 
expensive computation with the tree building overhead being 
relatively lightweight in comparison.   If the reallocation cost were
significant then I think we’d really want a little memory area 
that the p’s, p_sharp’s, rho’s, etc using the recursive tree building 
could use for preallocated memory.

bob-carpenter:
> ...
> Most of the objects you moved to class members are initialized only
> once at the beginning of the transition and should be pretty cheap 
> relative to all of the other dynamic allocation that happens during the
> tree building.

In general, we should avoid moving things to class members if
it can be avoided.  Especially if it means the class isn't reusable.
Many of our classes, like the model object itself, are designed
to be stateless.


> ...
> The basic code is designed around the gradient being the most 
> expensive computation with the tree building overhead being 
> relatively lightweight in comparison. 

That's certainly what all of our profiling has shown.

SteveBronder:
ack actually I totally misread the heaptrack diagnostics, with the larger model it allocates like 1GB but that's very small relative to other places

SteveBronder:
> In general, we should avoid moving things to class members if
it can be avoided.  Especially if it means the class isn't reusable.
Many of our classes, like the model object itself, are designed
to be stateless.

idt I would call `build_tree` stateless in it's current form. `z_` is carried around through a lot of recursive calls. If two seperate threads started calling this the state of one thread would start effecting the other

betanalpha:
Yeah, the `z_` defined in the `base_hmc` class from which the dynamic HMC code derives does introduce state into the system.  If we wanted to consider parallelization then we would have to modify the code a bit to make it safer, as was discussed when Sebastian suggested building both directions in parallel.  

Personally I'm not against having state that's carried over from iteration to iteration so long as there's no real information persists in that state.  From that perspective having a local memory arena to reduce the allocation overhead for bigger models would be an interesting avenue to explore.  As would making the logger a local member to avoid having to thread it through _everything_ which currently makes the code a pain to read.

PR #2880, opened 2020-01-21
[WIP] Add specialization for getting cols and cleanup of rvalue/lvalue

SteveBronder:
#### Submission Checklist

#### Summary

This PR is for two things

1. In brms issue [833](https://github.com/paul-buerkner/brms/issues/833) Paul found that traversing a matrix row-wise was actually faster. Looking into the code I think it's because `rvalue` has a specialization for pulling out a single row but not a single column. So I added a specializations for pulling out a column or a slice of a column

```stan
A[, M] // index_list(index_omni(), index_uni(M));
A[J:K, M] // index_list(index_min_max(J, K), index_uni(M));
```

I also ran a test program that did a bunch of `vec1 .* vec2 + ...` with min_max slices though heaptrack and found that it was a pretty big culprit for a lot of allocations

image

The flame graph above shows that a large number of allocations happened when pulling from rvalue(). I think in these min_max cases we should be able to just grab a slice of a column and not allocate anything

2. Some of the `rvalue()` code here couldn't take in fixed sized matrices, So I removed the Eigen::Dynamic values in the signature and replaced them with a template value

3. In the case of a nil_index_list for `rvalue` and `assign` I added a perfect forwarding right hand side so that in the case of a temporary the value can be moved into the value to be assigned instead of copied. It maybe good here to have one specialization that does a copy for smaller types while forwarding for larger types

4. We do our own bound checks so I changed instances of `A(i, j)` and `A.coeff(i, j)` with `A.coeffRef(i, j)`. This should remove Eigen's own bounds checking to save a little time

5. The templating here was a bit old school so I tried to update things to use the `require` library instead of boosts stuff

#### Intended Effect

This should give a very small performance bump to Stan code that does a lot of indexing. Though this code touches a lot of models and needs to go through more performance testing

#### How to Verify

I added two new tests to the cases mentioned in (1)

#### Side Effects

Hopefully less allocations. I am concerned that the `auto`'s here may have unintended side effects that need more testing

#### Documentation

New docs for `rvalue` specializations mentioned in (1)

#### Copyright and Licensing

### DIFF
--- a/src/stan/model/indexing/lvalue.hpp
+++ b/src/stan/model/indexing/lvalue.hpp
@@ -26,10 +26,16 @@ namespace model {
  * @param[in] name Name of lvalue variable (default "ANON"); ignored
  * @param[in] depth Indexing depth (default 0; ignored
  */
-template <typename T, typename U>
+template <typename T, typename U, require_not_stan_scalar_t<U>...>
 inline void assign(T& x, const nil_index_list& /* idxs */, U&& y,
                    const char* name = "ANON", int depth = 0) {
   x = std::forward<U>(y);
+}
+
+template <typename T, typename U, require_stan_scalar_t<U>...>
+inline void assign(T& x, const nil_index_list& /* idxs */, const U& y,
+                   const char* name = "ANON", int depth = 0) {
+  x = y;
 }
 
 /**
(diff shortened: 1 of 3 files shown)


SteveBronder:
I think this has to wait on [1628](https://github.com/stan-dev/math/pull/1628) in math

PR #2885, opened 2020-01-31
Allow rvalue and assign to accept eigen expressions

SteveBronder:
#### Submission Checklist

#### Summary

This allows `rvalue` and `assign` to take in generic eigen expressions. This uses a `const` [`Eigen::Ref`](https://eigen.tuxfamily.org/dox/classEigen_1_1Ref.html) on the rhs of `assign` and the input to `rvalue` so that eigen expressions sent into rvalue that need row or column access will first be evaluated. This allows `Eigen::Matrix` inputs to avoid a copy while eigen expressions will be evaluated 

This PR also makes one specialization for the case of `A[, col_index]` for grabbing a column of a matrix. It feels like this should be a regular thing people do and since we do col major index it feels like an easy slice to grab with `A.col()`

#### Intended Effect

Inputs to `rvalue` and `assign` can be more complex eigen expressions

#### How to Verify

Additional tests have been added to rvalue_test and lvalue_test that can be run from

```
./runTests.py ./src/test/unit/model/indexing/
```

This PR also cleans up the names of tests for `assign` to be more descriptive and names the assignee `lhs_x` and variable to be assigned `rhs_y`. I think this makes the test code easier to read 

#### Side Effects

Since everything is templated to remove dynamic. This should allow for Eigen expressions like `colvec * rowvec` to carry forward that the result is known to have 1x1 dims. Once the math library is templated to accept generic eigen expressions we can remove the `.eval()` in a few places here let the returns of `rvalue` be their `auto` deduced type

#### Copyright and Licensing

### DIFF
--- a/src/stan/model/indexing/deep_copy.hpp
+++ b/src/stan/model/indexing/deep_copy.hpp
@@ -41,8 +41,8 @@ inline const T& deep_copy(const T& x) {
  * @return Deep copy of input.
  */
 template <typename EigMat, typename = require_eigen_t<EigMat>>
-inline auto deep_copy(const EigMat& a) {
-  typename EigMat::PlainObject result(a);
+inline auto deep_copy(EigMat&& a) {
+  typename std::decay_t<EigMat>::PlainObject result(a);
   return result;
 }
 
--- a/src/stan/model/indexing/deep_copy.hpp
+++ b/src/stan/model/indexing/deep_copy.hpp
@@ -1,7 +1,8 @@
 #ifndef STAN_MODEL_INDEXING_DEEP_COPY_HPP
 #define STAN_MODEL_INDEXING_DEEP_COPY_HPP
 
-#include <Eigen/Dense>
+#include <stan/math/prim/Eigen.hpp>
+#include <stan/math/prim/meta.hpp>
 #include <vector>
 
 namespace stan {
@@ -39,8 +40,8 @@ inline const T& deep_copy(const T& x) {
  * @param a Input matrix, vector, or row vector.
  * @return Deep copy of input.
  */
-template <typename T, int R, int C>
-inline Eigen::Matrix<T, R, C> deep_copy(const Eigen::Matrix<T, R, C>& a) {
+template <typename EigMat, typename = require_eigen_t<EigMat>>
+inline auto deep_copy(const EigMat& a) {
   Eigen::Matrix<T, R, C> result(a);
   return result;
 }
--- a/src/stan/model/indexing/deep_copy.hpp
+++ b/src/stan/model/indexing/deep_copy.hpp
@@ -1,7 +1,7 @@
 #ifndef STAN_MODEL_INDEXING_DEEP_COPY_HPP
 #define STAN_MODEL_INDEXING_DEEP_COPY_HPP
 
-#include <stan/math/prim/Eigen.hpp>
+#include <stan/math/prim/fun/Eigen.hpp>
 #include <stan/math/prim/meta.hpp>
 #include <vector>
 
@@ -21,7 +21,7 @@ namespace model {
  * @param x Input value.
  * @return Constant reference to input.
  */
-template <typename T>
+template <typename T, typename = require_not_eigen_t<T>>
 inline const T& deep_copy(const T& x) {
   return x;
 }
@@ -42,7 +42,7 @@ inline const T& deep_copy(const T& x) {
  */
 template <typename EigMat, typename = require_eigen_t<EigMat>>
 inline auto deep_copy(const EigMat& a) {
-  Eigen::Matrix<T, R, C> result(a);
+  typename EigMat::PlainObject result(a);
   return result;
 }
 
--- a/src/stan/model/indexing/lvalue.hpp
+++ b/src/stan/model/indexing/lvalue.hpp
@@ -205,7 +205,8 @@ template <typename LhsEigMat, typename I, typename RhsEigMat,
           typename = require_not_eigen_vector_t<RhsEigMat>>
 inline void assign(LhsEigMat& x,
                    const cons_index_list<I, nil_index_list>& idxs,
-                   const RhsEigMat& y, const char* name = "ANON", int depth = 0) {
+                   const RhsEigMat& y,
+                   const char* name = "ANON", int depth = 0) {
   int x_idx_rows = rvalue_index_size(idxs.head_, x.rows());
   math::check_size_match("matrix[multi] assign row sizes", "lhs", x_idx_rows,
                          name, y.rows());
--- a/src/stan/model/indexing/lvalue.hpp
+++ b/src/stan/model/indexing/lvalue.hpp
@@ -137,7 +137,7 @@ inline void assign(LhsEigVec& x, const cons_index_list<I, nil_index_list>& idxs,
  * Assign the specified Eigen matrix at the specified single index
  * to the specified row vector value.
  *
- * Types:  mat[,uni] = rowvec
+ * Types:  mat[uni,] = rowvec
  *
  * @tparam EigMat Type of matrix to be assigned to.
  * @tparam RowVec Type of Eigen Row Vector to be assigned from.
--- a/src/stan/model/indexing/lvalue.hpp
+++ b/src/stan/model/indexing/lvalue.hpp
@@ -118,62 +118,30 @@ inline void assign(Eigen::Matrix<T, 1, Eigen::Dynamic>& x,
  * @throw std::invalid_argument If the value size isn't the same as
  * the indexed size.
  */
-template <typename T, typename I, typename U>
-inline typename boost::disable_if<boost::is_same<I, index_uni>, void>::type
-assign(Eigen::Matrix<T, Eigen::Dynamic, 1>& x,
+template <typename EigVec1, typename EigVec2, typename I,
+  typename = require_not_same_t<index_uni, I>,
+  typename = require_all_eigen_vector_t<EigVec1, EigVec2>>
+inline void assign(EigVec1& x,
        const cons_index_list<I, nil_index_list>& idxs,
-       const Eigen::Matrix<U, Eigen::Dynamic, 1>& y, const char* name = "ANON",
+       const EigVec2& y, const char* name = "ANON",
        int depth = 0) {
   math::check_size_match("vector[multi] assign sizes", "lhs",
                          rvalue_index_size(idxs.head_, x.size()), name,
                          y.size());
+ const Eigen::Ref<const typename EigVec2::PlainObject>&  vec = y;
   for (int n = 0; n < y.size(); ++n) {
     int i = rvalue_at(n, idxs.head_);
     math::check_range("vector[multi] assign range", name, x.size(), i);
-    x(i - 1) = y(n);
+    x(i - 1) = vec(n);
   }
 }
 
-/**
- * Assign the specified Eigen row vector at the specified multiple
- * index to the specified value.
- *
- * Types:   row_vec[multi] <- row_vec
- *
- * @tparam T Scalar type for assigned row vector.
- * @tparam I Type of multiple index.
- * @tparam U Type of value row vector scalar (must be assignable
- * to T).
- * @param[in] x Row vector variable to be assigned.
- * @param[in] idxs Sequence of one multiple index (from 1).
- * @param[in] y Value vector.
- * @param[in] name Name of variable (default "ANON").
- * @param[in] depth Indexing depth (default 0).
- * @throw std::out_of_range If any of the indices are out of bounds.
- * @throw std::invalid_argument If the value size isn't the same as
- * the indexed size.
- */
-template <typename T, typename I, typename U>
-inline typename boost::disable_if<boost::is_same<I, index_uni>, void>::type
-assign(Eigen::Matrix<T, 1, Eigen::Dynamic>& x,
-       const cons_index_list<I, nil_index_list>& idxs,
-       const Eigen::Matrix<U, 1, Eigen::Dynamic>& y, const char* name = "ANON",
-       int depth = 0) {
-  math::check_size_match("row_vector[multi] assign sizes", "lhs",
-                         rvalue_index_size(idxs.head_, x.size()), name,
-                         y.size());
-  for (int n = 0; n < y.size(); ++n) {
-    int i = rvalue_at(n, idxs.head_);
-    math::check_range("row_vector[multi] assign range", name, x.size(), i);
-    x(i - 1) = y(n);
-  }
-}
 
 /**
  * Assign the specified Eigen matrix at the specified single index
  * to the specified row vector value.
  *
- * Types:  mat[uni] = rowvec
+ * Types:  mat[uni,] = rowvec
  *
  * @tparam T Assigned matrix scalar type.
  * @tparam U Type of value scalar for row vector (must be
@@ -187,17 +155,48 @@ assign(Eigen::Matrix<T, 1, Eigen::Dynamic>& x,
  * @throw std::invalid_argument If the number of columns in the row
  * vector and matrix do not match.
  */
-template <typename T, typename U>
+template <typename T, typename RowVec, typename = require_t<is_eigen_row_vector<RowVec>>>
 void assign(Eigen::Matrix<T, Eigen::Dynamic, Eigen::Dynamic>& x,
             const cons_index_list<index_uni, nil_index_list>& idxs,
-            const Eigen::Matrix<U, 1, Eigen::Dynamic>& y,
+            const RowVec& y,
             const char* name = "ANON", int depth = 0) {
   math::check_size_match("matrix[uni] assign sizes", "lhs", x.cols(), name,
                          y.cols());
   int i = idxs.head_.n_;
-  math::check_range("matrix[uni] assign range", name, x.rows(), i);
-  for (int j = 0; j < x.cols(); ++j)  // loop allows double to var assgn
-    x(i - 1, j) = y(j);
+  math::check_range("matrix[uni,] assign range", name, x.rows(), i);
+  const Eigen::Ref<const typename RowVec::PlainObject>&  vec = y;
+  x.row(i - 1) = vec;
+}
+
+/**
+ * Assign the specified Eigen matrix at the specified single column index
+ * to the specified column
+ *
+ * Types:  mat[,uni] = vec
+ *
+ * @tparam T Assigned matrix scalar type.
+ * @tparam U Type of value scalar for row vector (must be
+ * assignable to T).
+ * @param[in] x Matrix variable to be assigned.
+ * @param[in] idxs Sequence of one single index (from 1).
+ * @param[in] y Value row vector.
+ * @param[in] name Name of variable (default "ANON").
+ * @param[in] depth Indexing depth (default 0).
+ * @throw std::out_of_range If any of the indices are out of bounds.
+ * @throw std::invalid_argument If the number of columns in the row
+ * vector and matrix do not match.
+ */
+template <typename T, typename ColVec, typename = require_t<is_eigen_col_vector<ColVec>>>
+void assign(Eigen::Matrix<T, Eigen::Dynamic, Eigen::Dynamic>& x,
+            const cons_index_list<index_omni, cons_index_list<index_uni, nil_index_list>>& idxs,
+            const ColVec& y,
+            const char* name = "ANON", int depth = 0) {
+  math::check_size_match("matrix[uni] assign sizes", "lhs", x.cols(), name,
+                         y.cols());
+  int i = idxs.tail_.head_.n_;
+  math::check_range("matrix[uni,] assign range", name, x.rows(), i);
+  const Eigen::Ref<const typename ColVec::PlainObject>&  vec = y;
+  x.col(i - 1) = vec;
 }
 
 /**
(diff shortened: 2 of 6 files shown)


src/stan/model/indexing/lvalue.hpp
```diff
@@ -118,62 +118,58 @@ inline void assign(Eigen::Matrix<T, 1, Eigen::Dynamic>& x,
...
-  for (int n = 0; n < y.size(); ++n) {
-    int i = rvalue_at(n, idxs.head_);
-    math::check_range("row_vector[multi] assign range", name, x.size(), i);
-    x(i - 1) = y(n);
-  }
+template <typename T, typename RowVec,
+          typename = require_t<is_eigen_row_vector<RowVec>>>
+void assign(Eigen::Matrix<T, Eigen::Dynamic, Eigen::Dynamic>& x,
+            const cons_index_list<index_uni, nil_index_list>& idxs,
+            const RowVec& y, const char* name = "ANON", int depth = 0) {
```
SteveBronder: @t4c1 should assign here be able to assign to an eigen expression? I left it as Eigen::Matrix since I'm not sure if we ever do something like `assign(A.array() * 2, B.array() + 2)`

t4c1: We don't and this example would not work in Eigen anyway. However, we could let it assign to expressions such as transpose or block. But to use such expressions we would probably need some modifications to compiler and possibly stan language.

In other words I don't think we need it yet, but it may come in handy in the future.

SteveBronder:
@t4c1 are you able to review things in this repo? Else maybe @bob-carpenter can if he has time?

src/stan/model/indexing/lvalue.hpp
```diff
@@ -359,27 +347,30 @@ assign(
...
-assign(Eigen::Matrix<T, Eigen::Dynamic, Eigen::Dynamic>& x,
-       const cons_index_list<I1, cons_index_list<I2, nil_index_list> >& idxs,
-       const Eigen::Matrix<U, Eigen::Dynamic, Eigen::Dynamic>& y,
-       const char* name = "ANON", int depth = 0) {
+template <typename LhsEigMat, typename I1, typename I2, typename RhsEigMat,
+          typename = require_eigen_t<LhsEigMat>,
+          typename = require_not_eigen_vector_t<LhsEigMat>,
+          typename = require_any_not_same_t<index_uni, I1, I2>,
+          typename = require_eigen_t<RhsEigMat>,
+          typename = require_not_eigen_vector_t<RhsEigMat>>
```
t4c1: These requires can be again simplified using `all` variants.

src/stan/model/indexing/lvalue.hpp
```diff
@@ -118,62 +96,58 @@ inline void assign(Eigen::Matrix<T, 1, Eigen::Dynamic>& x,
...
 }
 
 /**
- * Assign the specified Eigen row vector at the specified multiple
- * index to the specified value.
+ * Assign the specified Eigen matrix at the specified single index
+ * to the specified row vector value.
  *
- * Types:   row_vec[multi] <- row_vec
+ * Types:  mat[uni,] = rowvec
```
t4c1:
```suggestion
 * Types:  mat[uni] = rowvec
```

src/stan/model/indexing/rvalue.hpp
```diff
@@ -92,58 +70,53 @@ inline T rvalue(const Eigen::Matrix<T, 1, Eigen::Dynamic>& rv,
  * @param[in] depth Depth of indexing dimension.
  * @return Result of indexing vector.
  */
-template <typename T, typename I>
-inline typename boost::disable_if<boost::is_same<I, index_uni>,
-                                  Eigen::Matrix<T, Eigen::Dynamic, 1> >::type
-rvalue(const Eigen::Matrix<T, Eigen::Dynamic, 1>& v,
-       const cons_index_list<I, nil_index_list>& idx, const char* name = "ANON",
-       int depth = 0) {
+template <typename EigVec, typename I,
```
t4c1: Doxygen should include that this also works with row vectors.

src/stan/model/indexing/lvalue.hpp
```diff
@@ -67,39 +67,17 @@ inline void assign(std::vector<T>& x, const nil_index_list& /* idxs */,
  * @param[in] depth Indexing depth (default 0).
  * @throw std::out_of_range If the index is out of bounds.
  */
-template <typename T, typename U>
-inline void assign(Eigen::Matrix<T, Eigen::Dynamic, 1>& x,
+template <typename EigVec, typename Scalar,
```
t4c1: Doxygen should include that this also works with row vectors.

src/stan/model/indexing/lvalue.hpp
```diff
@@ -118,62 +96,58 @@ inline void assign(Eigen::Matrix<T, 1, Eigen::Dynamic>& x,
  * @throw std::invalid_argument If the value size isn't the same as
  * the indexed size.
  */
-template <typename T, typename I, typename U>
-inline typename boost::disable_if<boost::is_same<I, index_uni>, void>::type
-assign(Eigen::Matrix<T, Eigen::Dynamic, 1>& x,
-       const cons_index_list<I, nil_index_list>& idxs,
-       const Eigen::Matrix<U, Eigen::Dynamic, 1>& y, const char* name = "ANON",
-       int depth = 0) {
+template <typename EigVec1, typename EigVec2, typename I,
```
t4c1: Doxygen should include that this alsoworks with row vectors.

src/stan/model/indexing/rvalue.hpp
```diff
@@ -46,36 +46,14 @@ inline T rvalue(const T& c, const nil_index_list& /*idx*/,
  * @param[in] depth Depth of indexing dimension.
  * @return Result of indexing vector.
  */
-template <typename T>
-inline T rvalue(const Eigen::Matrix<T, Eigen::Dynamic, 1>& v,
-                const cons_index_list<index_uni, nil_index_list>& idx,
-                const char* name = "ANON", int depth = 0) {
+template <typename EigVec, typename = require_eigen_vector_t<EigVec>>
```
t4c1: Doxygen should include that this also works with row vectors.

src/stan/model/indexing/rvalue.hpp
```diff
@@ -152,14 +125,41 @@ rvalue(const Eigen::Matrix<T, 1, Eigen::Dynamic>& rv,
...
+                          cons_index_list<index_uni, nil_index_list>>& idx,
+    const char* name = "ANON", int depth = 0) {
+  int n = idx.tail_.head_.n_;
+  math::check_range("matrix[uni] indexing", name, a.rows(), n);
+  return a.col(n - 1).eval();
+}
+
+/**
+ * Return the result of indexing the specified Eigen matrix with a
+ * an omni index and a one single index, returning a vector.
```
t4c1:
```suggestion
 * a single and an omni index, returning a row vector.
```

src/stan/model/indexing/rvalue.hpp
```diff
@@ -234,16 +239,18 @@ inline T rvalue(
...
+          typename = require_eigen_t<EigMat>,
+          typename = require_not_eigen_vector_t<EigMat>>
+inline auto rvalue(
+    const EigMat& a,
+    const cons_index_list<index_uni, cons_index_list<I, nil_index_list>>& idx,
     const char* name = "ANON", int depth = 0) {
   int m = idx.head_.n_;
   math::check_range("matrix[uni,multi] indexing, row", name, a.rows(), m);
-  Eigen::Matrix<T, 1, Eigen::Dynamic> r = a.row(m - 1);
+  Eigen::Matrix<scalar_type_t<EigMat>, 1, EigMat::ColsAtCompileTime> r
```
t4c1: We can avoid one eval here.
```suggestion
  const auto& r
```

src/stan/model/indexing/rvalue.hpp
```diff
@@ -262,21 +269,23 @@ rvalue(
...
+          typename = require_not_same_t<index_uni, I>,
+          typename = require_eigen_t<EigMat>,
+          typename = require_not_eigen_vector_t<EigMat>>
+inline auto rvalue(
+    const EigMat& a,
+    const cons_index_list<I, cons_index_list<index_uni, nil_index_list>>& idx,
     const char* name = "ANON", int depth = 0) {
   int rows = rvalue_index_size(idx.head_, a.rows());
-  Eigen::Matrix<T, Eigen::Dynamic, 1> c(rows);
+  Eigen::Matrix<scalar_type_t<EigMat>, EigMat::RowsAtCompileTime, 1> c(rows);
```
t4c1: We know this is dynamic due to `require_not_eigen_vector_t<EigMat>`.
```suggestion
  Eigen::Matrix<scalar_type_t<EigMat>, Eigen::Dynamic, 1> c(rows);
```

src/stan/model/indexing/rvalue.hpp
```diff
@@ -296,24 +305,27 @@ rvalue(
...
+          typename = require_not_eigen_vector_t<EigMat>>
+inline auto rvalue(
+    const EigMat& a,
+    const cons_index_list<I1, cons_index_list<I2, nil_index_list>>& idx,
+    const char* name = "ANON", int depth = 0) {
   int rows = rvalue_index_size(idx.head_, a.rows());
   int cols = rvalue_index_size(idx.tail_.head_, a.cols());
-  Eigen::Matrix<T, Eigen::Dynamic, Eigen::Dynamic> c(rows, cols);
+  Eigen::Matrix<scalar_type_t<EigMat>, EigMat::RowsAtCompileTime,
+                EigMat::ColsAtCompileTime>
```
t4c1: Again we know these are `Dynamic`. 
```
Eigen::Matrix<scalar_type_t<EigMat>, Eigen::Dynamic, Eigen::Dynamic>
```

src/test/test-models/good/compound-assign/indexed_expr_divide_equals.stan
```diff
@@ -3,5 +3,7 @@ transformed data {
   a /= 5.0;
   print("a: ", a);
   a[1, ] /= 5.0;
+  a[1, 1:2] /= 5.0;
+  a[, 1:2] /= 5.0;
   print("r1 div 5: ", a);
```
t4c1: Are these printed values checked anywhere? If so this change will probably fail the test.

src/test/unit/model/indexing/lvalue_test.cpp
```diff
@@ -21,293 +21,365 @@ using stan::model::index_uni;
...
   double y = 13;
-  assign(xs, index_list(index_uni(3)), y);
-  EXPECT_FLOAT_EQ(y, xs(2));
+  assign(lhs_x, index_list(index_uni(3)), y);
+  EXPECT_FLOAT_EQ(y, lhs_x(2));
 
-  test_throw(xs, index_list(index_uni(0)), y);
-  test_throw(xs, index_list(index_uni(6)), y);
+  assign(lhs_x, index_list(index_uni(3)), y);
+  EXPECT_FLOAT_EQ(y, lhs_x(2));
```
t4c1: Why is this duplicated?

src/stan/model/indexing/deep_copy.hpp
```diff
@@ -39,9 +40,9 @@ inline const T& deep_copy(const T& x) {
  * @param a Input matrix, vector, or row vector.
  * @return Deep copy of input.
  */
-template <typename T, int R, int C>
-inline Eigen::Matrix<T, R, C> deep_copy(const Eigen::Matrix<T, R, C>& a) {
-  Eigen::Matrix<T, R, C> result(a);
+template <typename EigMat, typename = require_eigen_t<EigMat>>
+inline auto deep_copy(const EigMat& a) {
+  typename EigMat::PlainObject result(a);
   return result;
```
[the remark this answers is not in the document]
SteveBronder: I think it's better to leave this as is since the intent of the function is force a copy and in the eigen docs for eval they say

```
Notice that in the case of a plain matrix or vector (not an expression) this function just returns a const reference, in order to avoid a useless copy.
```

https://eigen.tuxfamily.org/dox/classEigen_1_1DenseBase.html#aa73e57a2f0f7cfcb4ad4d55ea0b6414b

Though the function returns by value I think as is it's more clear this is making a copy

src/stan/model/indexing/rvalue.hpp
SteveBronder: Gonna go through and update all the docs

src/stan/model/indexing/rvalue.hpp
SteveBronder: I can change this, the only reason I can think of to use RowsAtCompileTime here is if we started allowed for fixed size matrices

src/test/test-models/good/compound-assign/indexed_expr_divide_equals.stan
SteveBronder: cc @bob-carpenter but I'm pretty sure the test here just checks if the stan programs will compile

SteveBronder:
Actually @t4c1 I was goofing around with this and what if we used `decltype(a.eval())` instead of `Eigen::Ref`? Looking at the below it seems like this gives us what we want in that for actual matrix input types it will be a const reference, but for types that need assignment it will be an actual object

```
--------------
Input Type is: 
  Eigen::Matrix<double, -1, 1>&

decltype(Vector) is: 
 const Eigen::Matrix<double, -1, 1>&
----------------------------
Input Type is: 
  const Eigen::Product<Eigen::Matrix<double, -1, 1>, Eigen::Transpose<Eigen::Matrix<double, -1, 1> >, 0>&&

decltype(Vector * Vector.transpose()) is: 
 const Eigen::Matrix<double, -1, -1>
----------------------------
Input Type is: 
  Eigen::Matrix<double, -1, 1>&

Eigen::Ref<Vector> is: 
 const Eigen::Ref<const Eigen::Matrix<double, -1, 1>, 0, Eigen::InnerStride<1> >&
----------------------------
Input Type is: 
  const Eigen::Product<Eigen::Matrix<double, -1, 1>, Eigen::Transpose<Eigen::Matrix<double, -1, 1> >, 0>&&

Eigen::Ref<Vector * Vector.transpose()> is: 
 const Eigen::Ref<const Eigen::Matrix<double, -1, -1>, 0, Eigen::OuterStride<> >&
--------------
```


https://godbolt.org/z/vYmxdv

I'm trying to read the docs for `Eigen::Ref` and it looks like the `decltype(a.eval())` does essentially the same thing?

https://eigen.tuxfamily.org/dox/classEigen_1_1Ref.html

src/stan/model/indexing/rvalue.hpp
SteveBronder: What about `decltype(auto)` here?

t4c1: I think that should work as well.

t4c1:
> Actually @t4c1 I was goofing around with this and what if we used decltype(a.eval()) instead of Eigen::Ref? 

It would work, but it would also evaluate (copy) objects, such as `Eigen::Map` or `Eigen::Block`. And we do not need copys for those objects. So `Eigen::Ref` is still preferable.

SteveBronder:
@t4c1 this is ready for a review? @bob-carpenter idt @t4c1 has approve permissions here though so could you approve and/or review this as well?

src/stan/model/indexing/lvalue.hpp
t4c1: You missed this one.

src/stan/model/indexing/rvalue.hpp
```diff
@@ -18,28 +18,44 @@ namespace model {
...
  *
  * @tparam T Scalar type.
  * @param[in] c Value to index.
  * @return Input value.
  */
-template <typename T>
-inline T rvalue(const T& c, const nil_index_list& /*idx*/,
-                const char* /*name*/ = "", int /*depth*/ = 0) {
+template <typename T, typename = require_stan_scalar_t<T>>
+inline T rvalue(T&& c, const nil_index_list& /*idx*/, const char* /*name*/ = "",
```
t4c1: Why is here forwarding reference? I think you used pass by value for scalars in other places.

src/stan/model/indexing/rvalue.hpp
```diff
@@ -18,28 +18,44 @@ namespace model {
...
+ * a nil index list, which just returns the value.
+ *
+ * Types:  T[] : T
+ *
+ * @tparam T Scalar type.
+ * @param[in] c Value to index.
+ * @return Input value.
+ */
+template <typename T, typename = require_not_stan_scalar_t<T>>
+inline auto&& rvalue(T&& c, const nil_index_list& /*idx*/,
```
t4c1: `auto` is deduced as a value so this will always return a rvalue reference. `decltype(auto)` is instead deduced as appropriate reference type (no `&&` is needed).
```suggestion
inline decltype(auto) rvalue(T&& c, const nil_index_list& /*idx*/,
```

src/stan/model/indexing/rvalue.hpp
```diff
@@ -249,8 +265,7 @@ inline auto rvalue(
     const char* name = "ANON", int depth = 0) {
   int m = idx.head_.n_;
   math::check_range("matrix[uni,multi] indexing, row", name, a.rows(), m);
-  Eigen::Matrix<scalar_type_t<EigMat>, 1, EigMat::ColsAtCompileTime> r
-      = a.row(m - 1);
+  auto&& r = a.row(m - 1);
```
t4c1: That will not work. Assigning a rvalue to rvalue reference will not extend lifetime of the rvalue so that `Eigen::Row` object will be destroyed after this line is executed, leaving a dangling reference. I was also wrong when I earlier replied to your question saying that `decltype(auto)` would work here. It would not for the same reason. Instead you have to do one of:
```suggestion
  const auto& r = a.row(m - 1);
```
```suggestion
  return rvalue(a.row(m - 1), idx.tail_);
```

bob-carpenter:
The permissions are set across repos for almost everyone and almost every repo.  I have no idea why you can't resolve comments---does one of the participants have to do that?

SteveBronder:
I tend to leave issues open until the reviewer closes them so that's not an issue. If @t4c1 is able to approve here then we are good after I fix up the latest review stuff

src/stan/model/indexing/rvalue.hpp
SteveBronder: I'm cool with changing this to `declar_type(auto)`, from reading the stackoverflow answer below I though that `auto&&` would preserve the returns value type and cv qualifier?

https://stackoverflow.com/a/13242177/2269255

src/stan/model/indexing/rvalue.hpp
SteveBronder: Yep good catch!

src/stan/model/indexing/rvalue.hpp
SteveBronder: I like (3)!

SteveBronder:
@t4c1 would it ever make sense to have something like 

```cpp
template <typename T>
struct eigen_ref {
  using type = std::conditional_t<std::is_rvalue_reference<T>::value,
    Eigen::Ref<typename std::decay_t<T>::PlainObject>&&,
    const Eigen::Ref<const typename std::decay_t<T>::PlainObject>&>;
};

template <typename T>
using eigen_ref_t = typename eigen_ref<T>::type;
```

I'm not really certain how Eigen::Ref interacts with rvalue types. For the case of an `Eigen::Matrix` I think the above would say, "Retain your value type when passing a reference to `Ref`". Though for unevaluated rvalue expressions classes I'm not sure if this would do something weird.

SteveBronder:
Also once this passes tests I'll ping you for another once over

src/stan/model/indexing/rvalue.hpp
t4c1: Right, after some playing around I found out It works the same. While I feel like `decltype(auto)` might be more common in such situations, I will leave it up to you to decide which one to use in your PR.

src/stan/model/indexing/rvalue.hpp
t4c1: I was wrong again. While playing around with https://github.com/stan-dev/stan/pull/2885#discussion_r379287888, I found that [rvalue references also extend the lifetime of what they are pointing at](https://stackoverflow.com/questions/49371012/lifetime-of-rvalue-reference). So no change is necessary, even though I like suggestion 3.

t4c1:
> @t4c1 would it ever make sense to have something like

This boils down to the question, whether `Eigen::Ref<...>&&` makes sense. I think it does not. `Eigen::Ref` does not to allocate any memory on heap, so moving it is no cheaper than copying. 

What I think you want to do is have something like `Eigen::Ref<Expr&&>&&`, to be able to move `Expr` if it is a rvalue matrix. This does not seem to be supported by `Eigen::Ref`.

rok-cesnovar:
This PR is causing Jenkins Unit tests to go in an infinite loop. Restarting the tests worked before in order to get the worker free, but now that there are conflicts that is not doable. It leaves the previous test run alive:
https://jenkins.mc-stan.org/blue/organizations/jenkins/Stan/detail/PR-2885/23/pipeline

@serban-nicusor-toptal please stop this test run whenever you have time. If you wont have time to do it today let me know and I can just resolve the conflicts and introduce a compile error or something.

SteveBronder:
@rok-cesnovar apologies just tried merging to develop to get rid of the conflict

rok-cesnovar:
No need to apologize :) Just wanted to notify you guys as I saw it and Jenkins reported nothing back here.

serban-nicusor-toptal:
Hey, I'll login and fix it in a few minutes.

src/stan/model/indexing/rvalue.hpp
SteveBronder: I'm fine with keeping it 3

src/stan/model/indexing/rvalue.hpp
SteveBronder: I think it's more common to see `decltype(auto)` as the return type so let's leave it as is

SteveBronder:
@t4c1 I think I covered all the last review comments, you want to give this another look?

On a side note I've slowly started working on a design doc related to value types and how to handle move semantics and some of Eigen's peculiars like `Eigen::Ref`. I have a few other things to deal with right now but hope to have something up in the next two weeks or so

SteveBronder:
> This boils down to the question, whether Eigen::Ref<...>&& makes sense. I think it does not. Eigen::Ref does not to allocate any memory on heap, so moving it is no cheaper than copying.

Yeah I agree, reading the Ref docs over more this makes more sense to me

> What I think you want to do is have something like Eigen::Ref<Expr&&>&&, to be able to move Expr if it is a rvalue matrix. This does not seem to be supported by Eigen::Ref. You still can make a separate overload for MatrixXd&&.

Another day we can tinker with this. I think we'd want some sort of clever type trait that does a `MatrixXd&&` when that's the template parameter value and then a Ref otherwise

src/stan/model/indexing/lvalue.hpp
```diff
@@ -137,7 +137,7 @@ inline void assign(LhsEigVec& x, const cons_index_list<I, nil_index_list>& idxs,
  * Assign the specified Eigen matrix at the specified single index
  * to the specified row vector value.
  *
- * Types:  mat[uni,] = rowvec
+ * Types:  mat[,uni] = rowvec
```
t4c1:
```suggestion
 * Types:  mat[uni,] = rowvec
```
I think this should be index before comma.

t4c1:
Hmm, ok it seems I can select approve, but it does not count as approved.

SteveBronder:
Thanks for the catch! I updated that doc so feel free to approve and we will merge when the tests pass

SteveBronder:
@rok-cesnovar

rok-cesnovar:
Just a heads up https://github.com/stan-dev/stanc3/issues/471#issuecomment-590538283

SteveBronder:
Shoot let's revert this then

rok-cesnovar:
If there is any easy fix we can also fix this with a PR.

SteveBronder:
Let me see if I can find a quick fix today else I think it's better revert and then revert revert

Issue #2881, opened 2020-01-22
Doing the decomposition for every momentum draw is inefficient


bbbales2:
#### Summary:

For N parameters, there's an NxN Cholesky computed every time a new momentum is drawn. In reality we only need to recompute that when the metric changes.

#### Description:

Check the code here: https://github.com/stan-dev/stan/blob/develop/src/stan/mcmc/hmc/hamiltonians/dense_e_metric.hpp#L54

That llt() only needs computed when ```z.inv_e_metric_``` changes, not every time ```sample_p``` is called.

#### Reproducible Steps:

Sampling a simple model with a large number of parameters should be sufficient.

```
parameters {
  real x[500];
}

model {
  x ~ normal(0, 1);
}
```

Should do the trick. Run that model with:
```
./test sample num_warmup=0 adapt engaged=0 algorithm=hmc metric=dense_e
```

And compare the time with:
```
./test sample num_warmup=0 adapt engaged=0 algorithm=hmc metric=diag_e
```

And the difference should be noticeable once that cholesky is precomputed.

#### Current Output:
Output is fine, just slow.

#### Expected Output:
Same output.

#### Current Version:
v2.21.0

bob-carpenter:
Yikes.  We had explicitly discussed coding it originally so it didn't have this inefficiency.  I guess that didn't make it to the code.

SteveBronder:
Should it only happen when [`set_metric`](https://github.com/stan-dev/stan/blob/338de974e38c755f9e1ab2a7620afe7cd80f6f21/src/stan/mcmc/hmc/hamiltonians/dense_e_point.hpp#L35) is called? Could also move the metric to a protected or private member so it's not accessed outside of `set_metric`

bbbales2:
> Should it only happen when set_metric is called? Could also move the metric to a protected or private member so it's not accessed outside of set_metric

Yes that makes sense. Right now it's accessed directly by other things: https://github.com/stan-dev/stan/blob/develop/src/stan/mcmc/hmc/nuts/adapt_diag_e_nuts.hpp#L32, but an accessor makes more sense.

yizhang-yiz:
Is a fix under way? This is slows things down quite a bit when I use Radon model for testing.

betanalpha:
The current design is intentional.

[...]

PR #2894, opened 2020-03-06
Don't do inverse metric decomposition every draw

bbbales2:
Fix for Issue #2881 .

This involved switching to setter/getters for interfacing with dense_e_point so I made the change for diag_e_point as well.

I also changed all the set_metric verbage to set_inv_metric.

I didn't know what tests needed added for this. Lemme know.

#### Submission Checklist

#### Copyright and Licensing

### DIFF
--- a/src/stan/mcmc/hmc/hamiltonians/dense_e_point.hpp
+++ b/src/stan/mcmc/hmc/hamiltonians/dense_e_point.hpp
@@ -11,11 +11,13 @@ namespace mcmc {
  * Euclidean manifold with dense metric
  */
 class dense_e_point : public ps_point {
- public:
+private:
   /**
    * Inverse mass matrix.
    */
   Eigen::MatrixXd inv_e_metric_;
+  Eigen::MatrixXd inv_e_metric_llt_matrixU_;
+public:
 
   /**
    * Construct a dense point in n-dimensional phase space
@@ -23,17 +25,40 @@ class dense_e_point : public ps_point {
    *
    * @param n number of dimensions
    */
-  explicit dense_e_point(int n) : ps_point(n), inv_e_metric_(n, n) {
+  explicit dense_e_point(int n) : ps_point(n),
+				  inv_e_metric_(n, n),
+				  inv_e_metric_llt_matrixU_(n, n) {
     inv_e_metric_.setIdentity();
+    inv_e_metric_llt_matrixU_.setIdentity();
   }
 
   /**
-   * Set elements of mass matrix
+   * Set inverse metric
    *
    * @param inv_e_metric initial mass matrix
    */
-  void set_metric(const Eigen::MatrixXd& inv_e_metric) {
+  void set_inv_metric(const Eigen::MatrixXd& inv_e_metric) {
     inv_e_metric_ = inv_e_metric;
+    inv_e_metric_llt_matrixU_ = inv_e_metric_.llt().matrixU();
+  }
+
+  /**
+   * Get inverse metric
+   *
+   * @return reference to the inverse metric
+   */
+  const Eigen::MatrixXd& get_inv_metric() {
+    return inv_e_metric_;
+  }
+
+  /**
+   * Get the transpose of the lower Cholesky factor
+   *  of the inverse metric
+   *
+   * @return reference to transpose of Cholesky factor
+   */
+  const Eigen::MatrixXd& get_transpose_llt_inv_metric() {
+    return inv_e_metric_llt_matrixU_;
   }
 
   /**
--- a/src/stan/mcmc/hmc/hamiltonians/diag_e_metric.hpp
+++ b/src/stan/mcmc/hmc/hamiltonians/diag_e_metric.hpp
@@ -18,7 +18,7 @@ class diag_e_metric : public base_hamiltonian<Model, diag_e_point, BaseRNG> {
       : base_hamiltonian<Model, diag_e_point, BaseRNG>(model) {}
 
   double T(diag_e_point& z) {
-    return 0.5 * z.p.dot(z.inv_e_metric_.cwiseProduct(z.p));
+    return 0.5 * z.p.dot(z.get_inv_metric().cwiseProduct(z.p));
   }
 
   double tau(diag_e_point& z) { return T(z); }
@@ -34,7 +34,7 @@ class diag_e_metric : public base_hamiltonian<Model, diag_e_point, BaseRNG> {
   }
 
   Eigen::VectorXd dtau_dp(diag_e_point& z) {
-    return z.inv_e_metric_.cwiseProduct(z.p);
+    return z.get_inv_metric().cwiseProduct(z.p);
   }
 
   Eigen::VectorXd dphi_dq(diag_e_point& z, callbacks::logger& logger) {
@@ -46,7 +46,7 @@ class diag_e_metric : public base_hamiltonian<Model, diag_e_point, BaseRNG> {
         rand_diag_gaus(rng, boost::normal_distribution<>());
 
     for (int i = 0; i < z.p.size(); ++i)
-      z.p(i) = rand_diag_gaus() / sqrt(z.inv_e_metric_(i));
+      z.p(i) = rand_diag_gaus() / sqrt(z.get_inv_metric()(i));
   }
 };
 
--- a/src/stan/mcmc/hmc/hamiltonians/diag_e_point.hpp
+++ b/src/stan/mcmc/hmc/hamiltonians/diag_e_point.hpp
@@ -11,12 +11,13 @@ namespace mcmc {
  * Euclidean manifold with diagonal metric
  */
 class diag_e_point : public ps_point {
- public:
+private:
   /**
    * Vector of diagonal elements of inverse mass matrix.
    */
   Eigen::VectorXd inv_e_metric_;
 
+public:
   /**
    * Construct a diag point in n-dimensional phase space
    * with vector of ones for diagonal elements of inverse mass matrix.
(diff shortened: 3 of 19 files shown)


src/stan/mcmc/hmc/hamiltonians/dense_e_point.hpp
```diff
@@ -11,29 +11,54 @@ namespace mcmc {
...
   }
 
   /**
-   * Set elements of mass matrix
+   * Set inverse metric
    *
    * @param inv_e_metric initial mass matrix
    */
-  void set_metric(const Eigen::MatrixXd& inv_e_metric) {
+  void set_inv_metric(const Eigen::MatrixXd& inv_e_metric) {
```
SteveBronder: I think the API is supposed to be just `set_metric` for each point no?

src/stan/mcmc/hmc/hamiltonians/dense_e_point.hpp
```diff
@@ -11,29 +11,54 @@ namespace mcmc {
...
     inv_e_metric_ = inv_e_metric;
+    inv_e_metric_llt_matrixU_ = inv_e_metric_.llt().matrixU();
+  }
+
+  /**
+   * Get inverse metric
+   *
+   * @return reference to the inverse metric
+   */
+  const Eigen::MatrixXd& get_inv_metric() {
```
SteveBronder:
```suggestion
  const Eigen::MatrixXd& get_inv_metric() const {
```

src/stan/mcmc/hmc/hamiltonians/dense_e_point.hpp
```diff
@@ -11,29 +11,54 @@ namespace mcmc {
...
+    return inv_e_metric_;
+  }
+
+  /**
+   * Get the transpose of the lower Cholesky factor
+   *  of the inverse metric
+   *
+   * @return reference to transpose of Cholesky factor
+   */
+  const Eigen::MatrixXd& get_transpose_llt_inv_metric() {
```
SteveBronder:
```suggestion
  const Eigen::MatrixXd& get_transpose_llt_inv_metric() const {
```

src/stan/mcmc/hmc/hamiltonians/diag_e_metric.hpp
```diff
@@ -46,7 +46,7 @@ class diag_e_metric : public base_hamiltonian<Model, diag_e_point, BaseRNG> {
         rand_diag_gaus(rng, boost::normal_distribution<>());
 
     for (int i = 0; i < z.p.size(); ++i)
-      z.p(i) = rand_diag_gaus() / sqrt(z.inv_e_metric_(i));
+      z.p(i) = rand_diag_gaus() / sqrt(z.get_inv_metric()(i));
   }
```
SteveBronder: If you don't mind cleaning these up while your here I think you could make this
```cpp
z.p = z.get_inv_metric().unaryExpr([&](auto&& x) { return rand_diag_gaus() / sqrt(x)});
```

src/stan/mcmc/hmc/hamiltonians/diag_e_point.hpp
```diff
@@ -32,10 +33,19 @@ class diag_e_point : public ps_point {
...
+  void set_inv_metric(const Eigen::VectorXd& inv_e_metric) {
     inv_e_metric_ = inv_e_metric;
   }
 
+  /**
+   * Get inverse metric
+   *
+   * @return reference to the inverse metric
+   */
+  const Eigen::VectorXd& get_inv_metric() {
```
SteveBronder:
```suggestion
  const Eigen::VectorXd& get_inv_metric() const {
```

src/stan/mcmc/hmc/nuts/adapt_dense_e_nuts.hpp
```diff
@@ -29,12 +29,15 @@ class adapt_dense_e_nuts : public dense_e_nuts<Model, BaseRNG>,
       this->stepsize_adaptation_.learn_stepsize(this->nom_epsilon_,
                                                 s.accept_stat());
 
-      bool update = this->covar_adaptation_.learn_covariance(
-          this->z_.inv_e_metric_, this->z_.q);
+      Eigen::MatrixXd inv_metric;
+      
+      bool update = this->covar_adaptation_.learn_covariance(inv_metric, this->z_.q);
```
SteveBronder: Why switch to an empty matrix here?

src/stan/mcmc/hmc/hamiltonians/dense_e_point.hpp
bbbales2: Well it's the inverse metric, not the metric, so I just changed the names everywhere.

src/stan/mcmc/hmc/nuts/adapt_dense_e_nuts.hpp
bbbales2: The inv_metric argument to the learn_covariance gets written to. It's like an output parameter.

betanalpha:
The current design is intentional.

The Cholesky is needed only once per transition which is a relatively small cost compared to the many gradient evaluations needed within each transition.  Saving the Cholesky decomposition introduces an additional $\mathcal{O}(N^{2})$ memory burden which is much more than the $mathcal{O}(N)$ burden everywhere else, and becomes problematic for sufficiently large models.  Without any explicit profiling demonstrating that the Cholesky is a substantial const for typical models I don't see any strong motivation for the change.  I will add this point to the original issue.

At the same time the members of `ps_point` and its derivations are intentionally left public to avoid any code overhead for accessors and mutators.  Because these classes are not exposed through the algorithm API and everything is public we decided that there isn't any need for encapsulation.  If encapsulation were added then the accessors and mutators would need to be unit tested as is done elsewhere in the library.

src/stan/mcmc/hmc/static_uniform/adapt_diag_e_static_uniform.hpp
```diff
@@ -31,10 +31,15 @@ class adapt_diag_e_static_uniform
...
                                                 s.accept_stat());
 
-      bool update = this->var_adaptation_.learn_variance(this->z_.inv_e_metric_,
-                                                         this->z_.q);
+      Eigen::VectorXd inv_metric;
+
+      bool update
+          = this->var_adaptation_.learn_variance(inv_metric, this->z_.q);
       if (update) {
+        this->z_.set_inv_metric(inv_metric);
```
SteveBronder, line 39: [optional] At this point it's the last time you use `inv_metric` so you could just move it like

```suggestion
        this->z_.set_inv_metric(std::move(inv_metric));
```

src/stan/mcmc/hmc/static_uniform/adapt_diag_e_static_uniform.hpp
```diff
@@ -31,10 +31,15 @@ class adapt_diag_e_static_uniform
       this->stepsize_adaptation_.learn_stepsize(this->nom_epsilon_,
                                                 s.accept_stat());
 
-      bool update = this->var_adaptation_.learn_variance(this->z_.inv_e_metric_,
-                                                         this->z_.q);
+      Eigen::VectorXd inv_metric;
```
SteveBronder, line 34: Can you know the size of this when you make it?

SteveBronder (commented):
Couple little comments

src/stan/mcmc/hmc/hamiltonians/dense_e_point.hpp
```diff
@@ -11,29 +12,51 @@ namespace mcmc {
...
-   * Set elements of mass matrix
+   * Set inverse metric
    *
    * @param inv_e_metric initial mass matrix
    */
-  void set_metric(const Eigen::MatrixXd& inv_e_metric) {
+  void set_inv_metric(const Eigen::MatrixXd& inv_e_metric) {
     inv_e_metric_ = inv_e_metric;
+    inv_e_metric_llt_matrixU_ = inv_e_metric_.llt().matrixU();
+  }
```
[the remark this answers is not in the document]
bbbales2: Oh yeah good idea. This means if we `set_inv_metric(std::move(inv_e_metric))` that we won't need copies right?

src/stan/mcmc/hmc/static_uniform/adapt_diag_e_static_uniform.hpp
bbbales2: I could, but it gets resized in `learn_variance`. Should I still set the size when declaring it?

src/stan/mcmc/hmc/hamiltonians/dense_e_point.hpp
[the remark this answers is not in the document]
SteveBronder: Yeah that will just move it over instead of doing a copy

src/stan/mcmc/hmc/static_uniform/adapt_diag_e_static_uniform.hpp
SteveBronder: If you declare the sizes there then the resize is a no-op since it's already the correct size. But idt it's that big of a deal either way

SteveBronder:
> The Cholesky is needed only once per transition which is a relatively small cost compared to the many gradient evaluations needed within each transition. Saving the Cholesky decomposition introduces an additional $\mathcal{O}(N^{2})$ memory burden which is much more than the $mathcal{O}(N)$ burden everywhere else, and becomes problematic for sufficiently large models. Without any explicit profiling demonstrating that the Cholesky is a substantial const for typical models I don't see any strong motivation for the change. I will add this point to the original issue.

Thinking about this again, don't you pay the O(N^2) memory burden still computing the cholesky but it's just temporary data? If a program failed because of N is too large it would fail either way wouldn't it?

bbbales2:
> don't you pay the O(N^2) memory burden still

Oh yeah, the temporary would still be O(N^2) since you compute the Cholesky at one point or another. Good point. I'll add the forwarding real quick

src/stan/mcmc/hmc/hamiltonians/dense_e_point.hpp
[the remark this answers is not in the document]
bbbales2: Added

src/stan/mcmc/hmc/static_uniform/adapt_diag_e_static_uniform.hpp
bbbales2: I just left it off. It's kindof a weird thing happening here anyway (like `learn_variance` is effectively returning two things).

PR #2898, opened 2020-03-12
Add sparse matrix functions to io reader and writer

SteveBronder:
#### Submission Checklist

#### Summary

This adds `sparse_matrix_*` methods to the io reader and writer while also cleaning some things up. The cleanups consist of 

1. Adding brackets to for loops
2. Changing the typdefs to aliases with `using`
3. Adding const to methods if they did not change the internal state of the class
4. Allowing for named return value optimization (NRVO) by changing some methods to declare the class once at the beginning of the function and then using it as the return in places with multiple returns. For example the function below

```cpp
  inline vector_t vector(size_t m) {
    if (m == 0)
      return vector_t();
    return map_vector_t(&scalar_ptr_increment(m), m);
 }
```

Because we are returning either (1) an Eigen Vector or (2) an Eigen map that then needs to be implicitly converted to an Eigen vector this turns off RVO and NRVO. By doing declaring the return value once in the below the compiler is able to use named return value optimization to move that memory to the call site's object.

```cpp
  inline vector_t vector(size_t m) {
    vector_t ret_vec(m);
    if (m == 0) {
      return ret_vec;
    }
    ret_vec = map_vector_t(&this->scalar_ptr_increment(m), m);
    return ret_vec;
  }
```

#### Intended Effect

The stan compiler can now call

```cpp
Eigen::SparseMatrix<T> mat = reader.sparse_matrix_constrain(row_vec, col_vec, n, m);
// ...
writer.sparse_matrix_unconstrain(mat);
```

etc.

#### How to Verify

Tests were updated for the reader and writer

```bash
./runTests.py ./src/test/unit/io/ -f reader
./runTests.py ./src/test/unit/io/ -f writer
```

#### Side Effects

Should not effect any current stan programs

#### Documentation

Documentation was added for the sparse matrix methods.

#### Copyright and Licensing

Please list the copyright holder for the work you are submitting (this will be you or your assignee, such as a university or company): Stephen Bronder


SteveBronder:
Should we also support sparse vector? Eigen allows for cols/rows of Sparse matrices to be assigned from sparse vectors.

https://godbolt.org/z/T_q4S9

SteveBronder:
@bob-carpenter would you mind taking a look at this? If it's too much I can take out all the non sparse matrix stuff to a seperate PR

src/stan/io/reader.hpp
```diff
@@ -299,15 +312,92 @@ class reader {
...
+   * @param n Number of rows.
+   * @param m Number of columns.
+   * @return Eigen::Matrix made up of the next scalars.
+   */
+  template <typename VecR, typename VecC,
+            typename = require_all_std_vector_vt<std::is_integral, VecR, VecC>>
+  inline Eigen::SparseMatrix<T> sparse_matrix_constrain(const VecR& vec_r,
+                                                        const VecC& vec_c,
+                                                        size_t n, size_t m) {
+    return sparse_matrix(std::forward<VecR>(vec_r), std::forward<VecC>(vec_c),
```
bob-carpenter, line 376: [question]
Since they arguments are only constant references coming in, what does forward do here?

src/stan/io/reader.hpp
```diff
@@ -1018,15 +1108,15 @@ class reader {
    * dimensionality, reading from an unconstrained vector of the
    * appropriate size.
    *
-   * @param M Rows of Cholesky factor
-   * @param N Columns of Cholesky factor
+   * @param n Rows of Cholesky factor
```
bob-carpenter, line 1111: Nooooo!  Always m x n!  I'm going to stop writing this, but this should be put back to the conventional order everywhere.  If you want to be more verbose, Eigen uses `rows` and `cols`.

src/stan/io/reader.hpp
```diff
@@ -1227,116 +1317,351 @@ class reader {
...
+   * nonzero elements made up of the next scalars arranged in
+   * column-major order. Throws if any scalar value is less than lb.
+   *
+   * @tparam TL the type of the lower bound.
+   * @tparam VecR Vector with an integral type
+   * @tparam VecC Vector with an integral type
+   * @param lb The lower bound to check all scalar values against
+   * @param vec_r rows corresponding the nonzero entries.
+   * @param vec_c columns corresponding the nonzero entries.
+   * @param n Number of rows.
```
bob-carpenter, line 1365: OK.  Couldn't help myself.  m x n.

src/stan/io/reader.hpp
```diff
@@ -1227,116 +1317,351 @@ class reader {
...
+                                          const VecC& vec_c, size_t n,
+                                          size_t m) {
+    stan::math::check_matching_sizes("sparse_matrix_lb", "vec_r", vec_r,
+                                     "vec_c", vec_c);
+    Eigen::SparseMatrix<T> ret_mat(n, m);
+    ret_mat.reserve(vec_r.size());
+    if (m == 0 || n == 0) {
+      return ret_mat;
+    }
+    using triplet_type = Eigen::Triplet<T>;
```
bob-carpenter, line 1381: This looks cut-and-paste from the non-lb case.   Is there a way to write this once and plug in the reader like scalar_lb as another argument or would that lose efficiency?  that is, something like:

```
return construct_sparse_matrix(m, n, vec_r, vec_c, [](){ return scalar_lb(lb); });
```

An alternative that might not be much less efficient is to read the right number of scalar_lb into a vector, then use that in a call to a generic construction method for a sparse matrix from an integer vector, integer vector, and real vector).  Then this code would look like:

```
return to_sparse_matrix(m, n, vec_r, vec_c, vector_lb(lb, m * n));
```

or something like that, where you call the vector method with a lower bound that presumabl already exists somewhere.

src/stan/io/reader.hpp
```diff
@@ -1227,116 +1317,351 @@ class reader {
...
+   * @param vec_c columns corresponding the nonzero entries.
+   * @param n Number of rows.
+   * @param m Number of columns.
+   * @return Eigen::SparseMatrix made up of the next scalars.
+   */
+  template <typename TL, typename VecR, typename VecC,
+            typename = require_all_std_vector_vt<std::is_integral, VecR, VecC>>
+  inline sparse_matrix_t sparse_matrix_lb_constrain(const TL lb,
+                                                    const VecR& vec_r,
+                                                    const VecC& vec_c, size_t n,
```
bob-carpenter, line 1410: Same comments about m and n order and code duplication.

src/stan/io/reader.hpp
```diff
@@ -1227,116 +1317,351 @@ class reader {
...
 
   template <typename TU>
-  inline row_vector_t row_vector_ub_constrain(const TU ub, size_t m, T &lp) {
+  inline row_vector_t row_vector_ub_constrain(const TU ub, size_t m, T& lp) {
     row_vector_t v(m);
-    for (size_t i = 0; i < m; ++i)
+    for (size_t i = 0; i < m; ++i) {
       v(i) = scalar_ub_constrain(ub, lp);
+    }
     return v;
```
bob-carpenter, line 1517: [very optional]
I really want to go higher order and just turn all these into generic comprehensions.  Not sure it's a good idea, though.  Something like:

```
template <typename M, class F>
comprehend_matrix(int m, int n, const F& f) {
  M x(m, n);
  for (int i = 0; i < x.size(); ++i) x(i) = f();
  return M;
}
```

src/stan/io/reader.hpp
```diff
@@ -1227,116 +1317,351 @@ class reader {
...
   }
 
+  /**
+   * Return a sparse matrix of the specified dimensionality and number of
+   * nonzero elements made up of the next scalars arranged in
+   * column-major order. Throws if any scalar value is less than ub.
+   *
+   * @tparam TL the type of the upper bound.
+   * @tparam VecR Vector with an integral type
```
bob-carpenter, line 1560: [question]
When might `VecR` and `VecC` be different types?  Don't we always control the arguments to this?

src/stan/io/reader.hpp
```diff
@@ -1227,116 +1317,351 @@ class reader {
...
+    }
     return v;
   }
 
+  /**
+   * Return a sparse matrix of the specified dimensionality and number of
+   * nonzero elements made up of the next scalars arranged in
+   * column-major order. Throws if any scalar value is less than ub.
+   *
+   * @tparam TL the type of the upper bound.
```
bob-carpenter, line 1559: I strongly prefer `T`, but if there's going to be more informative letters attached, they need to match.  So presumably this should be `TU` for upper bound?

src/stan/io/reader.hpp
```diff
@@ -1380,42 +1705,161 @@ class reader {
...
+   * @return Eigen::SparseMatrix made up of the next scalars.
+   */
+  template <typename TL, typename TU, typename VecR, typename VecC,
+            typename = require_all_std_vector_vt<std::is_integral, VecR, VecC>>
+  inline sparse_matrix_t sparse_matrix_lub_constrain(const TL lb, const TU ub,
+                                                     const VecR& vec_r,
+                                                     const VecC& vec_c,
+                                                     size_t n, size_t m,
+                                                     T& lp) {
+    stan::math::check_matching_sizes("sparse_matrix_lub_constrain", "vec_r",
```
bob-carpenter, line 1847: I really want to get rid of all this cut-and-paste code with some kind of simplifying abstraction.

src/stan/io/writer.hpp
```diff
@@ -148,7 +147,7 @@ class writer {
    * @param y Bounded value.
    * @throw std::runtime_error if y is not between the lower and upper bounds
    */
-  void scalar_lub_unconstrain(double lb, double ub, T &y) {
+  void scalar_lub_unconstrain(double lb, double ub, const T &y) {
```
bob-carpenter, line 150: Thanks for tidying up the `const` correctness!

src/stan/io/writer.hpp
```diff
@@ -279,81 +282,148 @@ class writer {
...
-    for (idx_t j = 0; j < y.cols(); ++j)
-      for (idx_t i = 0; i < y.rows(); ++i)
+    for (idx_t j = 0; j < y.cols(); ++j) {
+      for (idx_t i = 0; i < y.rows(); ++i) {
         scalar_lb_unconstrain(lb, y(i, j));
+      }
+    }
+  }
+
+  void sparse_matrix_lb_unconstrain(double lb, sparse_matrix_t &y) {
```
bob-carpenter, line 326: Missing doc.

src/stan/io/writer.hpp
```diff
@@ -279,81 +282,148 @@ class writer {
...
+    for (int k = 0; k < y.outerSize(); ++k) {
+      for (sparse_matrix_inner_iter_t it(y, k); it; ++it) {
+        scalar_lb_unconstrain(lb, it.value());
+      }
+    }
   }
 
   void vector_ub_unconstrain(double ub, vector_t &y) {
```
bob-carpenter, line 334: Lots of these seem to be missing doc.

src/stan/io/writer.hpp
```diff
@@ -279,81 +282,148 @@ class writer {
...
-    for (idx_t j = 0; j < y.cols(); ++j)
-      for (idx_t i = 0; i < y.rows(); ++i)
+    for (idx_t j = 0; j < y.cols(); ++j) {
+      for (idx_t i = 0; i < y.rows(); ++i) {
         scalar_ub_unconstrain(ub, y(i, j));
+      }
+    }
+  }
+
+  void sparse_matrix_ub_unconstrain(double ub, sparse_matrix_t &y) {
```
bob-carpenter, line 355: Documentation.

Also, is there no way to get a single iterator in the right order withouthaving to double stride like this through the ouer and inner bits?

Again, I want to see all this code cut-and-paste encapsulated in something more general if possible.

src/stan/io/writer.hpp
```diff
@@ -279,81 +282,148 @@ class writer {
...
-    for (idx_t j = 0; j < y.cols(); ++j)
-      for (idx_t i = 0; i < y.rows(); ++i)
+    for (idx_t j = 0; j < y.cols(); ++j) {
+      for (idx_t i = 0; i < y.rows(); ++i) {
         scalar_lub_unconstrain(lb, ub, y(i, j));
+      }
+    }
+  }
+
+  void sparse_matrix_lub_unconstrain(double lb, double ub, sparse_matrix_t &y) {
```
bob-carpenter, line 384: Doc here and for all these new functions.  And for the old functions if you don't mind, but that's not a requirement for approval.

src/test/unit/io/reader_test.cpp
```diff
@@ -1,6 +1,29 @@
 #include <gtest/gtest.h>
 #include <stan/io/reader.hpp>
-
+#include <tuple>
+
+namespace stan {
+namespace test {
+auto get_sparse_matrix_reader_pars() {
+  int row_size = (rand() % 20) + 1;
```
bob-carpenter, line 8: I would prefer to see the tests iterate over some well; chose row and column sizes that exercise the boundaries, like

```
std::vector<int> idxs{0, 1, 2, 5, 10};
for (row_size : idxs)
  for (col_size : idxs) 
    ... test ...
```

src/test/unit/io/reader_test.cpp
```diff
@@ -1245,6 +1285,63 @@ TEST(io_reader, matrix_lb_constrain_lp) {
...
+  auto max_row = std::get<5>(sparse_pars);
+  auto max_col = std::get<6>(sparse_pars);
+  stan::io::reader<double> reader(theta, theta_i);
+  double lb = -1.5;
+  Eigen::SparseMatrix<double> y = reader.sparse_matrix_lb(
+      lb, theta_rows, theta_cols, max_row + 1, max_col + 1);
+  for (int i = 0; i < nonzero_size; i++) {
+    EXPECT_FLOAT_EQ(theta[i], y.coeffRef(theta_rows[i], theta_cols[i]));
+  }
+}
```
bob-carpenter, line 1304: Again, it seems like there might be some abstraction possible here to avoid all this cut-and-paste code.  Especially all this stuff with magic numbers in gets.

In tests, we usually don't need to worry about efficiency, so the constructors, etc. can be abstracted.

SteveBronder:
I'm closing this as sparse matrices will now be static matrices so this code will have to change

PR #2899, opened 2020-03-12
std:: int overloads test model

rok-cesnovar:
#### Submission Checklist

- [x ] Declare copyright holder and open-source license: see below

#### Summary

Adding a model that will test fun(int) and fun(real) for all functions that mirror the std:: cmath functions. 

This is added to test the changes in https://github.com/stan-dev/math/pull/1765

#### Copyright and Licensing

### DIFF
--- a/src/test/test-models/good/int_overloads.stan
+++ b/src/test/test-models/good/int_overloads.stan
@@ -27,8 +27,10 @@ transformed data {
   transformed_data_real = log10(d_int);
   transformed_data_real = log10(d_real);
 
-  transformed_data_real = pow(d_int);
-  transformed_data_real = pow(d_real);
+  transformed_data_real = pow(d_int, d_int);
+  transformed_data_real = pow(d_real, d_int);
+  transformed_data_real = pow(d_int, d_real);
+  transformed_data_real = pow(d_real, d_real);
   transformed_data_real = sqrt(d_int);
   transformed_data_real = sqrt(d_real);
   transformed_data_real = cbrt(d_int);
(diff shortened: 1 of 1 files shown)


src/test/test-models/good/int_overloads.stan
```diff
@@ -0,0 +1,194 @@
...
+  transformed_data_real = log(d_int);
+  transformed_data_real = log(d_real);
+  transformed_data_real = log1p(d_int);
+  transformed_data_real = log1p(d_real);
+  transformed_data_real = log2(d_int);
+  transformed_data_real = log2(d_real);
+  transformed_data_real = log10(d_int);
+  transformed_data_real = log10(d_real);
+
+  transformed_data_real = pow(d_int);
```
bob-carpenter: How does this compile?  `pow` requires two arguments, so we need all permutations.

src/test/test-models/good/int_overloads.stan
```diff
@@ -0,0 +1,194 @@
...
+  real y_p;
+}
+model {
+  y ~ normal(0,1);
+}
+
+transformed parameters {
+  real transformed_param_real;
+
+  transformed_param_real = abs(d_int);
```
bob-carpenter: These function applications are all already tested in the data section.  These should all be removed and replaced with ones with autodiff variable arguments.  In the case of `pow` and `atan2`, ones that do it in all combinations with `double` and `real`).

mcol:
Also I think that `sqrt`, `tan` and `tanh` appear twice in each block.

src/test/test-models/good/int_overloads.stan
rok-cesnovar: Ah, I was too qucik there.

src/test/test-models/good/int_overloads.stan
rok-cesnovar: Added.

src/test/test-models/good/int_overloads.stan
rok-cesnovar: Fixed.

rok-cesnovar:
Thanks @bob-carpenter and @mcol 

I was a bit sloppy, this should now be cleaned up.

PR #2918, opened 2020-05-08
Bugfix/ 2817 build logistic test

### LINKED ISSUE #2817: Model tests not working with g++
#### Summary:

The rundown is here: https://discourse.mc-stan.org/t/running-stan-dev-stan-tests/10940

Basically:

```
git clone --recursive https://github.com/stan-dev/stan.git
cd stan
./runTests.py -j8 src/test
```

Fails by default with g++ in Ubuntu with errors that look like:

```
/gtest_1.8.1/include -I lib/stan_math/lib/gtest_1.8.1 -I lib/stan_math/lib/gtest_1.8.1/include -I lib/stan_math/lib/gtest_1.8.1 -I lib/stan_math/lib/gtest_1.8.1/include -I lib/stan_math/lib/gtest_1.8.1 -I lib/stan_math/lib/gtest_1.8.1/include -I lib/stan_math/lib/gtest_1.8.1 -I lib/stan_math/lib/gtest_1.8.1/include -I lib/stan_math/lib/gtest_1.8.1 -I lib/stan_math/lib/gtest_1.8.1/include -I lib/stan_math/lib/gtest_1.8.1          -DBOOST_DISABLE_ASSERTS           -c -o test/stan/lang/grammars/statement_2_grammar_inst.o src/stan/lang/grammars/statement_2_grammar_inst.cpp
src/test/performance/logistic_test.cpp:47:10: fatal error: test/test-models/performance/logistic.hpp: No such file or directory
 #include <test/test-models/performance/logistic.hpp>
          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

This is because the various .stan example models (including test/test-models/performance/logistic.hpp) are not being built.

@rok-cesnovar figured out in the thread that the tests run in clang (so this is probably a gcc something or another) and that by running:

```
make test/integration/compile_models_test
```

beforehand things work. Given that, I think this comment also needs updated: https://github.com/stan-dev/stan/blob/develop/make/tests#L91

#### Current Version:
v2.20.0

rok-cesnovar:
#### Submission Checklist

#### Summary

Fixes #2817 by specifying a make rule for test/performance/logistic_test.o

#### Copyright and Licensing

### DIFF
--- a/make/tests
+++ b/make/tests
@@ -208,3 +208,10 @@ test/integration/compile_models_test$(EXE): $(patsubst src/%.stan,%.hpp-test,$(s
 # "src/test/test-models/good-standalone-functions/*.stanfuncs"
 ##
 test/integration/compile_standalone_functions_test$(EXE): $(patsubst src/%.stanfuncs,%.hpp-test,$(shell find src/test/test-models/good-standalone-functions -type f -name '*.stanfuncs'))
+
+##
+# Same trick as above for models, but for the performance model
+##
+src/unit/lang/parser_generator_test$(EXE): $(patsubst src/%.stan,%.hpp,$(shell find src/test/test-models/good/parser-generator -type f -name '*.stan'))
+
+test/performance/logistic_test.o: $(patsubst src/%.stan,%.hpp-test,$(shell find src/test/test-models/performance -type f -name '*.stan'))
--- a/make/tests
+++ b/make/tests
@@ -202,16 +202,10 @@ test/unit/lang/parser_generator_test$(EXE): $(patsubst src/%.stan,%.hpp,$(shell 
 # the compiler using the -fsyntax-only flag
 ##
 test/integration/compile_models_test$(EXE): $(patsubst src/%.stan,%.hpp-test,$(shell find src/test/test-models/good -type f -name '*.stan'))
+test/performance/logistic_test$(EXE): test/integration/compile_models_test$(EXE)
 
 ##
 # Same trick as above for models, but for standalone functions, using
 # "src/test/test-models/good-standalone-functions/*.stanfuncs"
 ##
 test/integration/compile_standalone_functions_test$(EXE): $(patsubst src/%.stanfuncs,%.hpp-test,$(shell find src/test/test-models/good-standalone-functions -type f -name '*.stanfuncs'))
-
-##
-# Same trick as above for models, but for the performance model
-##
-src/unit/lang/parser_generator_test$(EXE): $(patsubst src/%.stan,%.hpp,$(shell find src/test/test-models/good/parser-generator -type f -name '*.stan'))
-
-test/performance/logistic_test.o: $(patsubst src/%.stan,%.hpp-test,$(shell find src/test/test-models/performance -type f -name '*.stan'))


bbbales2:
Is there a chance there's some sort of race condition here?

I verified the bug and the pulled in the fix, but got this error:

```
src/test/unit/mcmc/hmc/integrators/expl_leapfrog_test.cpp:6:10: fatal error: test/test-models/good/mcmc/hmc/integrators/command.hpp: No such file or directory
 #include <test/test-models/good/mcmc/hmc/integrators/command.hpp>
          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

After running this command:
```
./runTests.py -j8 src/test
```

I then ran:
```
./runTests.py src/test
```

And after a looooong time I got a different error:
```
  first_run[7]
    Which is: 1.11536
beta.1: index 7
src/test/performance/logistic_test.cpp:123: Failure
Expected equality of these values:
  -0.51189202
  first_run[8]
    Which is: -0.226193
beta.2: index 8
[  FAILED  ] performance.values_from_tagged_version (0 ms)
[ RUN      ] performance.values_same_run_to_run
```

But I'm assuming without much reason that I can ignore that error.

Now I cleaned everything out and am just running with one core and it's been compiling for a while without error.

rok-cesnovar:
> Is there a chance there's some sort of race condition here

Could be that the dependencies are not setup correctly. Let me check.

bbbales2:
Cool the single core build is still going fine (I think the multicore build would have failed by now).

bbbales2:
I know you didn't ask but I checked with the latest push and it still is not working.

rok-cesnovar:
Oh yeah, still failing :/. Them makefiles are winning.

mitzimorris:
makefiles for those unit tests which include stan models are extremely
problematic, always have been.

in summer of 2018 Daniel Lee did a makefile refactor and added some make
dependencies magic that got rid of the need to putting explicit
instructions into the tests makefiles for each test that called a model.
I think Rok might have pruned out a pass that checks dependencies somewhere
- it speeds up the make process, because calculating dependencies is
expensive.  I'd be OK with adding explicit dependencies into the makefiles
for tests given that make goes faster overall.
Rok, does any of this make sense?  happy to take a look,
cheers,
Mitzi

syclik:
so... the problem isn't really with make / makefile, even though it seems
that way. Make is doing the right thing here and including the dependencies
correctly. The issue is that different versions of compilers generate
dependencies differently! Last I checked, even with the same version of g++
on Mac and Windows, the dependencies that are generated using the compiler
is different. And specifically, it's different for missing header files.
This is important because the logistic test header file doesn't exist when
we call it.

We have the rules in place to generate a .hpp from a .stan. And if the .hpp
were included in the dependency file generated from the compiler, we would
just generate that automatically. If there's still some sed in the
makefile, it deals with this problem a bit.

One way we could fix this is to include the .hpp file in git repo. Then
we'd need to add a rule to that .hpp that it needs to have a dependency on
all of stanc so if stanc changes, that should be rebuilt. If we kept the
latest version of the file checked in, we'd avoid this problem completely.

mitzimorris:
> One way we could fix this is to include the .hpp file in git repo. Then we'd need to add a rule to that .hpp that it needs to have a dependency on all of stanc so if stanc changes, that should be rebuilt. If we kept the latest version of the file checked in, we'd avoid this problem completely.

that seems like a hack, and not what you'd want to test changes to the compiler, no?

syclik:
If the dependencies are set up properly, the file will be regenerated with
every change to the compiler, even it it results in the same code. If the
generated code changes, that would have to be checked in by a developer.

If we wanted to be a little too clever and hacky, we could commit a blank
file with a comment about what it is. Make sure the time stamp is before
the latest change to the compiler. And force git to ignore changes to the
file. But that requires using git in a way it wasn’t designed.

PR #2922, opened 2020-05-17
Replace clock_t with std::chrono

### LINKED ISSUE #2912: What steps are needed for multiple chains?
#### Summary:

Thought it would be a good idea to talk about what steps we would need to run multiple chains in a single stan program (either serially or within parallel). @yizhang-yiz and @bbbales2 have the [mpi branch](https://github.com/stan-dev/stan/compare/mpi_warmup_auto) up, reading over their implementation they've done quite a lot!  I'd like to just sketch out what the bare minimum is for calling K chains in serial/parallel.

#### Description:

At the cmdstan level say we added options like 

```
... n_chains=4 parallel=4
```
What would we need to allow that?

It looks like a lot of it would go into [`generate_transitions`](https://github.com/stan-dev/stan/blob/develop/src/stan/services/util/generate_transitions.hpp) where

1. We would pass it the number of chains
2. We would need an atomic `util::mcmc_writer` and other callbacks
3. Assuming writer, logger, and callbacks are threadsafe, wrap the loop for the iters in generate transitions in a parallel executor (either parallel_for in tbb or whatever we want to call for MPI)

For something like campfire and the mpi version would it work to have parallism all at the level of `generate_transitions`? Could the function that calls `generate_transitions` run for N transitions on K chains, then do campfire stuff over the window and keep running and then call `generate_transitions` again? If we write it general enough we could have an option that's something like `execution_policy` that could be either MPI for clusters and `tbb` for threads.

#### Current Version:
v2.23.0

### LINKED ISSUE #2921: Measure time using std::chrono::steady_clock not clock().
#### Description:

We should move to using `std::chrono::steady_clock` that allows us to return the "wall clock time passed" for and algorithm, not the CPU time used. Our time reports with STAN_THREADS and STAN_OPENCL are broken and should not be used for any comparisons.

With threaded execution it reports the sum of execution times for all threads for the threaded parts. Which is obviously not what we are interested in. This change wont have an effect to what is reported for non-threaded or non-GPU.

#### Current Version:
v2.23.0

rok-cesnovar:
#### Submission Checklist

#### Summary

Fixes #2921 by replacing clock_t with std::chrono::steady_clock

#### Copyright and Licensing

### DIFF
--- a/src/stan/services/sample/fixed_param.hpp
+++ b/src/stan/services/sample/fixed_param.hpp
@@ -65,15 +65,13 @@ int fixed_param(Model& model, const stan::io::var_context& init,
   writer.write_sample_names(s, sampler, model);
   writer.write_diagnostic_names(s, sampler, model);
 
-  clock_t start = clock();
-
+  auto start = std::chrono::steady_clock::now();
   util::generate_transitions(sampler, num_samples, 0, num_samples, num_thin,
                              refresh, true, false, writer, s, model, rng,
                              interrupt, logger);
-  clock_t end = clock();
-
-  double sampleDeltaT = static_cast<double>(end - start) / CLOCKS_PER_SEC;
-  writer.write_timing(0.0, sampleDeltaT);
+  auto end = std::chrono::steady_clock::now();
+  std::chrono::duration<double> sample_delta_t = end-start;
+  writer.write_timing(0.0, sample_delta_t.count());
 
   return error_codes::OK;
 }
--- a/src/stan/services/sample/fixed_param.hpp
+++ b/src/stan/services/sample/fixed_param.hpp
@@ -72,7 +72,7 @@ int fixed_param(Model& model, const stan::io::var_context& init,
                              interrupt, logger);
   auto end = std::chrono::steady_clock::now();
   double sample_delta_t = std::chrono::duration_cast<std::chrono::milliseconds>(end - start).count()/1000.0;
-  writer.write_timing(0.0, sample_delta_t.count());
+  writer.write_timing(0.0, sample_delta_t);
 
   return error_codes::OK;
 }
--- a/src/stan/services/sample/fixed_param.hpp
+++ b/src/stan/services/sample/fixed_param.hpp
@@ -71,7 +71,7 @@ int fixed_param(Model& model, const stan::io::var_context& init,
                              refresh, true, false, writer, s, model, rng,
                              interrupt, logger);
   auto end = std::chrono::steady_clock::now();
-  std::chrono::duration<double> sample_delta_t = end - start;
+  double sample_delta_t = std::chrono::duration_cast<std::chrono::milliseconds>(end - start).count()/1000.0;
   writer.write_timing(0.0, sample_delta_t.count());
 
   return error_codes::OK;
--- a/src/stan/services/sample/fixed_param.hpp
+++ b/src/stan/services/sample/fixed_param.hpp
@@ -71,8 +71,10 @@ int fixed_param(Model& model, const stan::io::var_context& init,
                              refresh, true, false, writer, s, model, rng,
                              interrupt, logger);
   auto end = std::chrono::steady_clock::now();
-  double sample_delta_t = std::chrono::duration_cast
-    <std::chrono::milliseconds>(end - start).count()/1000.0;
+  double sample_delta_t
+      = std::chrono::duration_cast<std::chrono::milliseconds>(end - start)
+            .count()
+        / 1000.0;
   writer.write_timing(0.0, sample_delta_t);
 
   return error_codes::OK;
--- a/src/stan/services/sample/fixed_param.hpp
+++ b/src/stan/services/sample/fixed_param.hpp
@@ -11,6 +11,7 @@
 #include <stan/services/util/generate_transitions.hpp>
 #include <stan/services/util/create_rng.hpp>
 #include <stan/services/util/initialize.hpp>
+#include <chrono>
 #include <vector>
 
 namespace stan {
--- a/src/stan/services/sample/fixed_param.hpp
+++ b/src/stan/services/sample/fixed_param.hpp
@@ -70,7 +70,7 @@ int fixed_param(Model& model, const stan::io::var_context& init,
                              refresh, true, false, writer, s, model, rng,
                              interrupt, logger);
   auto end = std::chrono::steady_clock::now();
-  std::chrono::duration<double> sample_delta_t = end-start;
+  std::chrono::duration<double> sample_delta_t = end - start;
   writer.write_timing(0.0, sample_delta_t.count());
 
   return error_codes::OK;
--- a/src/stan/services/sample/fixed_param.hpp
+++ b/src/stan/services/sample/fixed_param.hpp
@@ -71,7 +71,8 @@ int fixed_param(Model& model, const stan::io::var_context& init,
                              refresh, true, false, writer, s, model, rng,
                              interrupt, logger);
   auto end = std::chrono::steady_clock::now();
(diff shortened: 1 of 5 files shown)


rok-cesnovar:
Running the redcard reduce_sum case study:  

##### develop

threads on - 1 thread - works fine:
```
 Elapsed Time: 72.3359 seconds (Warm-up)
               75.1681 seconds (Sampling)
               147.504 seconds (Total)


real    2m28,622s
user    2m27,747s
sys     0m0,029s
```

2 threads - broken:
```
 Elapsed Time: 69.8285 seconds (Warm-up)
               73.3394 seconds (Sampling)
               143.168 seconds (Total)


real    1m12,405s
user    2m23,323s
sys     0m0,120s
```

4 threads - broken:
```
 Elapsed Time: 73.4332 seconds (Warm-up)
               82.3326 seconds (Sampling)
               155.766 seconds (Total)


real    0m39,372s
user    2m35,813s
sys     0m0,248s
```

##### This PR

no threading:
```
 Elapsed Time: 69.3594 seconds (Warm-up)
               75.091 seconds (Sampling)
               144.45 seconds (Total)


real    2m24,741s
user    2m24,297s
sys     0m0,056s
```

1 thread:
```
 Elapsed Time: 74.651 seconds (Warm-up)
               75.4217 seconds (Sampling)
               150.073 seconds (Total)


real    2m30,399s
user    2m30,004s
sys     0m0,112s
```

2 threads:
```
 Elapsed Time: 41.023 seconds (Warm-up)
               39.0111 seconds (Sampling)
               80.0341 seconds (Total)


real    1m20,201s
user    2m38,677s
sys     0m0,156s
```

4 threads:
```
 Elapsed Time: 18.5378 seconds (Warm-up)
               17.8096 seconds (Sampling)
               36.3474 seconds (Total)


real    0m36,455s
user    2m22,777s
sys     0m0,168s
```

wds15:
This is great..I can review tomorrow.

src/stan/services/sample/fixed_param.hpp
```diff
@@ -65,15 +66,13 @@ int fixed_param(Model& model, const stan::io::var_context& init,
...
+  auto start = std::chrono::steady_clock::now();
   util::generate_transitions(sampler, num_samples, 0, num_samples, num_thin,
                              refresh, true, false, writer, s, model, rng,
                              interrupt, logger);
-  clock_t end = clock();
-
-  double sampleDeltaT = static_cast<double>(end - start) / CLOCKS_PER_SEC;
-  writer.write_timing(0.0, sampleDeltaT);
+  auto end = std::chrono::steady_clock::now();
+  std::chrono::duration<double> sample_delta_t = end - start;
```
wds15, line 74: The way you coded it is along the lines of the example on en.cppreference.com. I still do not see in the definition of the standard that one "tick" is equal to one second. Could we used instead this (if it works):

`std::chrono::seconds sample_delta_t = end - start;`

This way we make sure that its indeed seconds which we report.

See https://en.cppreference.com/w/cpp/chrono/duration and there "helper types".

Although that could report time as full seconds only... maybe test that.

Another option is to make use of the "period" member type which is part of the duration class. The period member type defines the number of seconds per tick. Maybe this is over-cautious, but I don't see in the standard that one tick is equal to one second on all systems.

rok-cesnovar: It says

It consists of a count of ticks of type Rep and a tick period, where the tick period is a compile-time rational constant representing the number of seconds from one tick to the next.

I will check if there is a way to explicitly mention seconds

wds15: I know... but the duration you use is just the number of ticks elapsed. You do not know that their base unit are seconds. So what is 1 tick is not defined by the standard, but system specific in my understanding of the standard.

If we use `std::chrono::seconds` then it will be guaranteed that we count in seconds.

rok-cesnovar: `std::chrono::seconds` is actually just `std::chrono::duration<int64_t, std::ratio<1,1>>`.

So the only other way of doing this would be:
```cpp
auto a = std::chrono::steady_clock::now();
// ... work ...
auto b = std::chrono::steady_clock::now();
double diff = std::chrono::duration_cast<std::chrono::milliseconds>(b-a).count()/1000.0;
```
But that feels cumbersome. But if you prefer that I will change it.

If we do `std::chrono::duration_cast<std::chrono::seconds>` that returns just the integer part of the number of seconds.

This is the implementation of duration.
```
 /// duration
    template<typename _Rep, typename _Period>
      struct duration
      {
	typedef _Rep						rep;
	typedef _Period 					period;
```
and `_Rep` is `double` in our case. And period is 1/(ticks per sec).

wds15: I feared that we would loose below second accuracy if we go with the seconds thing.

Let me have one more read of the C++ ref and then we will likely proceed with what you have implemented. It appears to me that the C++ ref docs are bad in this case unfortunately.

wds15:
Bump... just a reminder to myself...

src/stan/services/sample/fixed_param.hpp
wds15: I went through the C++ docs once more and googled around. The issue is still that a "tick" is never defined to be equal "one second". This is left as an implementation detail. One tick is probably commonly defined as one second, but the standard does not define it.

I think you suggestion with milliseconds is best if want to maintain sub-second accuracy. You can probably write it in an easier way, since the duration_cast should happen implicitly. So maybe try the less cumbersome

`double diff = std::chrono::milliseconds(b-a).count()/1000.0;`

I would expect that to work as well, but just give it a try.

This is the best we can do from what I see to get well defined time units. It's a shortcoming of the C++ standard here, I am afraid.

Thanks for bearing with me here.

rok-cesnovar: I still think that text here does say that: https://en.cppreference.com/w/cpp/chrono/duration

Quote:
It consists of a count of ticks of type Rep and a tick period, **where the tick period is a compile-time rational constant representing the number of seconds from one tick to the next**.

In the old time.h way, that means that period is basically something close to `1/CLOCKS_PER_SEC`. So the fraction of a second that represent one tick. Its probably written in a bizarre way.

I will try to see if I can do something close to. 
`double diff = std::chrono::milliseconds(b-a).count()/1000.0;`

I do have to note that this only helps us in "code readability", as std::chrono::milliseconds only applies the std::millli [ratio](https://en.cppreference.com/w/cpp/numeric/ratio/ratio) to the duration at compile time. So you are still counting on duration being returned in seconds.

> Thanks for bearing with me here.

I have no problem being cautious.

wds15: > It consists of a count of ticks of type Rep and a tick period, where the tick period is a compile-time rational constant representing the number of seconds from one tick to the next.

Correct. So what you could do as an alternative is to take from an instance of `std::chrono::duration<double>` the `_Preiod` definition and multiply the number which we get with `.count()` by the ratio defined by `_Period`. It's an implementation detail that `_Period` is unity (meaning it is equal to one second), but that equality ( 1 tick = 1 second ) is not spelled out anywhere.

The solution with milliseconds is much cleaner from my perspective and it results in well defined behaviour... much better!

rok-cesnovar:
@wds15 I replaced all uses with 

```cpp
auto a = std::chrono::steady_clock::now();
// ... work ...
auto b = std::chrono::steady_clock::now();
double diff = std::chrono::duration_cast<std::chrono::milliseconds>(b-a).count()/1000.0;
```
Its the only way I got it to work.

wds15:
Fine for me...why is our  test pipeline unhappy?

Thanks for  doing it this way.

rok-cesnovar:
Ah, I must have deleted a few semicolons when trying to get under the line limit. Will fix in the morning and ask for your review.

Thanks!

PR #2946, opened 2020-07-30
[NFC] base_hmc const accessor to _z

kedartal:
#### Summary

Add a const accessor to `_z` in `src/stan/mcmc/hmc/base_hmc.hpp`

#### Intended Effect

Allow access in const contexts, useful for code integration with Stan as a library.

#### How to Verify

N/A

#### Side Effects

Shouldn't be any.

#### Documentation

N/A

#### Copyright and Licensing

K2DQ, LLC

### DIFF
--- a/src/stan/mcmc/hmc/base_hmc.hpp
+++ b/src/stan/mcmc/hmc/base_hmc.hpp
@@ -15,6 +15,15 @@
 namespace stan {
 namespace mcmc {
 
+/**
+ * Base class for Hamiltonian samplers.
+ *
+ * @tparam Model The type of the Stan model.
+ * @tparam Hamiltonian The type of Hamiltonians over the (unconstrained)
+ * parameter space.
+ * @tparam Integrator The type of integrator (e.g. leapfrog).
+ * @tparam BaseRNG The type of random number generator.
+ */
 template <class Model, template <class, class> class Hamiltonian,
           template <class> class Integrator, class BaseRNG>
 class base_hmc : public base_mcmc {
@@ -132,7 +141,21 @@ class base_hmc : public base_mcmc {
     this->z_.ps_point::operator=(z_init);
   }
 
+  /**
+   * Gets the current point in the (unconstrained) parameter space.
+   *
+   * @return The current point in the (unconstrained) parameter space.
+   */
   typename Hamiltonian<Model, BaseRNG>::PointType& z() { return z_; }
+
+  /**
+   * Gets the current point in the (unconstrained) parameters space.
+   *
+   * @return The current point in the (unconstrained) parameters space.
+   */
+  const typename Hamiltonian<Model, BaseRNG>::PointType& z() const noexcept {
+    return z_;
+  }
 
   virtual void set_nominal_stepsize(double e) {
     if (e > 0)
--- a/src/test/unit/mcmc/hmc/base_hmc_test.cpp
+++ b/src/test/unit/mcmc/hmc/base_hmc_test.cpp
@@ -46,6 +46,13 @@ TEST(McmcBaseHMC, point_construction) {
   EXPECT_EQ(static_cast<int>(q.size()), sampler.z().g.size());
 }
 
+TEST(McmcBaseHMC, point_access_from_const_hmc) {
+  rng_t base_rng(0);
+  const stan::mcmc::mock_hmc sampler(stan::mcmc::mock_model(2), base_rng);
+  EXPECT_EQ(2, sampler.z().q.size());
+  EXPECT_EQ(2, sampler.z().g.size());
+}
+
 TEST(McmcBaseHMC, seed) {
   rng_t base_rng(0);
 


src/stan/mcmc/hmc/base_hmc.hpp
```diff
@@ -134,6 +134,10 @@ class base_hmc : public base_mcmc {
 
   typename Hamiltonian<Model, BaseRNG>::PointType& z() { return z_; }
 
+  const typename Hamiltonian<Model, BaseRNG>::PointType& z() const noexcept {
```
bob-carpenter, line 137: 1. Needs documentation.  This is very simple here.  There may not be doc elsewhere, but we're requiring it for new additions and appreciate it getting backfilled where it's missing.  

2. Needs a test.  This can be really simple, but we just want to make sure we have tests in place to avoid future regressions.

bob-carpenter:
Also, thanks for flagging.  I had to go look up "NFC"---it's for a non-functional change.  We don't have any conventions for that yet.  We use "WIP", but I also hadn't seen "REF" for a refactor.

I'm trying to understand the concept from here:

https://docs.civicrm.org/dev/en/latest/tools/git/#nfc

By their definition, only the most trivial changes like whitespace seem to count as NFCs.    

The utility of the tag is that it signals "easy to review", which this one certainly was.

kedartal:
@bob-carpenter:

Thanks! I'll add the documentation and a unit test shortly. 

The way I was taught to use it, NFC is meant to indicate that there's no intension to change any semantics, for any "reasonable" user. More formally, it indicates that the correctness guarantees of the system before and after the change are practically equivalent (including any new capabilities, so by definition new functionality cannot be NFC). This means that the change has narrower potential implications, since it couldn't break anything subtle: if the code compiles, the change would be transparent to everyone "reasonable". In our case, if someone did rely on `base_hmc::_z` to **not** have a const accessor, e.g. in order to select some different overloaded method, well, I'll have to revisit my notion of "reasonable" :)
Compared to the link you shared, this isn't as much about the review as it's about whether other folks may want to know this change happened, after being reviewed and accepted. Since NFC changes shouldn't impact any other code, release notes, "reasonable" expectations, ... they can be safely ignored when looking at the delta between two commits, so good to flag them as such in the commit log message.

src/stan/mcmc/hmc/base_hmc.hpp
kedartal: Done.

kedartal:
Jenkins failed compiling the TBB memory allocation injection library, most likely unrelated to the PR.
```
g++ -Werror  -c -MMD -DDO_ITT_NOTIFY -O2 -DUSE_PTHREAD -m64 -mrtm  -Wall -Wextra   -Wno-parentheses -Wno-sized-deallocation -fPIC -flifetime-dse=1 -D__TBBMALLOC_BUILD=1 -I../tbb_2019_U8/src -I../tbb_2019_U8/src/rml/include -I../tbb_2019_U8/include -I../tbb_2019_U8/src/tbbmalloc -I../tbb_2019_U8/src/tbbmalloc ../tbb_2019_U8/src/tbbmalloc/proxy.cpp
../tbb_2019_U8/src/tbbmalloc/proxy.cpp:300:7: error: ‘void* __libc_valloc(size_t)’ specifies less restrictive attributes than its target ‘void* valloc(size_t)’: ‘alloc_size’, ‘malloc’, ‘nothrow’ [-Werror=missing-attributes]
  300 | void *__libc_valloc(size_t size) __attribute__ ((alias ("valloc")));
      |       ^~~~~~~~~~~~~
../tbb_2019_U8/src/tbbmalloc/proxy.cpp:231:14: note: ‘void* __libc_valloc(size_t)’ target declared here
  231 | void *PREFIX(valloc)(ZONE_ARG size_t size) __THROW
      |              ^~~~~~
../tbb_2019_U8/src/tbbmalloc/proxy.cpp:166:22: note: in definition of macro ‘PREFIX’
  166 | #define PREFIX(name) name
      |                      ^~~~
../tbb_2019_U8/src/tbbmalloc/proxy.cpp:299:7: error: ‘void* __libc_pvalloc(size_t)’ specifies less restrictive attributes than its target ‘void* pvalloc(size_t)’: ‘malloc’, ‘nothrow’ [-Werror=missing-attributes]
  299 | void *__libc_pvalloc(size_t size) __attribute__ ((alias ("pvalloc")));
      |       ^~~~~~~~~~~~~~
../tbb_2019_U8/src/tbbmalloc/proxy.cpp:261:7: note: ‘void* __libc_pvalloc(size_t)’ target declared here
  261 | void *pvalloc(size_t size) __THROW
      |       ^~~~~~~
../tbb_2019_U8/src/tbbmalloc/proxy.cpp:298:7: error: ‘void* __libc_memalign(size_t, size_t)’ specifies less restrictive attributes than its target ‘void* memalign(size_t, size_t)’: ‘alloc_size’, ‘malloc’, ‘nothrow’ [-Werror=missing-attributes]
  298 | void *__libc_memalign(size_t alignment, size_t size) __attribute__ ((alias ("memalign")));
      |       ^~~~~~~~~~~~~~~
../tbb_2019_U8/src/tbbmalloc/proxy.cpp:225:14: note: ‘void* __libc_memalign(size_t, size_t)’ target declared here
  225 | void *PREFIX(memalign)(ZONE_ARG size_t alignment, size_t size) __THROW
      |              ^~~~~~~~
../tbb_2019_U8/src/tbbmalloc/proxy.cpp:166:22: note: in definition of macro ‘PREFIX’
  166 | #define PREFIX(name) name
      |                      ^~~~
../tbb_2019_U8/src/tbbmalloc/proxy.cpp:297:7: error: ‘void* __libc_calloc(size_t, size_t)’ specifies less restrictive attributes than its target ‘void* calloc(size_t, size_t)’: ‘alloc_size’, ‘leaf’, ‘malloc’, ‘nothrow’ [-Werror=missing-attributes]
  297 | void *__libc_calloc(size_t num, size_t size) __attribute__ ((alias ("calloc")));
      |       ^~~~~~~~~~~~~
../tbb_2019_U8/src/tbbmalloc/proxy.cpp:205:14: note: ‘void* __libc_calloc(size_t, size_t)’ target declared here
  205 | void *PREFIX(calloc)(ZONE_ARG size_t num, size_t size) __THROW
      |              ^~~~~~
../tbb_2019_U8/src/tbbmalloc/proxy.cpp:166:22: note: in definition of macro ‘PREFIX’
  166 | #define PREFIX(name) name
      |                      ^~~~
../tbb_2019_U8/src/tbbmalloc/proxy.cpp:296:7: error: ‘void* __libc_malloc(size_t)’ specifies less restrictive attributes than its target ‘void* malloc(size_t)’: ‘alloc_size’, ‘leaf’, ‘malloc’, ‘nothrow’ [-Werror=missing-attributes]
  296 | void *__libc_malloc(size_t size) __attribute__ ((alias ("malloc")));
      |       ^~~~~~~~~~~~~
../tbb_2019_U8/src/tbbmalloc/proxy.cpp:200:14: note: ‘void* __libc_malloc(size_t)’ target declared here
  200 | void *PREFIX(malloc)(ZONE_ARG size_t size) __THROW
      |              ^~~~~~
../tbb_2019_U8/src/tbbmalloc/proxy.cpp:166:22: note: in definition of macro ‘PREFIX’
  166 | #define PREFIX(name) name
      |                      ^~~~
../tbb_2019_U8/src/tbbmalloc/proxy.cpp:149:18: error: ‘void* __TBB_malloc_proxy(size_t)’ specifies less restrictive attributes than its target ‘void* malloc(size_t)’: ‘alloc_size’, ‘leaf’, ‘malloc’, ‘nothrow’ [-Werror=missing-attributes]
  149 | extern "C" void *__TBB_malloc_proxy(size_t) __attribute__ ((alias ("malloc")));
      |                  ^~~~~~~~~~~~~~~~~~
../tbb_2019_U8/src/tbbmalloc/proxy.cpp:200:14: note: ‘void* __TBB_malloc_proxy(size_t)’ target declared here
  200 | void *PREFIX(malloc)(ZONE_ARG size_t size) __THROW
      |              ^~~~~~
../tbb_2019_U8/src/tbbmalloc/proxy.cpp:166:22: note: in definition of macro ‘PREFIX’
  166 | #define PREFIX(name) name
      |                      ^~~~
cc1plus: all warnings being treated as errors
make[1]: *** [/home/jenkins-slave/jenkins-slave-files/workspace/Stan_PR-2946/lib/stan_math/lib/tbb_2019_U8/build/Makefile.tbbmalloc:62: proxy.o] Error 1
make[1]: Leaving directory '/home/jenkins-slave/jenkins-slave-files/workspace/Stan_PR-2946/lib/stan_math/lib/tbb'
make: *** [lib/stan_math/make/libraries:149: lib/stan_math/lib/tbb/tbbmalloc.def] Error 2
```

rok-cesnovar:
@kedartal The error you are seeing is because we have tests set to fail on compiler warnings. One of the test machines was upgraded recently and uses the latest compiler which warns for something with the TBB library.  So definitely not the fault of this PR. We just need to merge https://github.com/stan-dev/math/pull/1993 

Sorry for the trouble.

kedartal:
@rok-cesnovar upgrading is cool :)
While we're at it, happy to create PRs for upgrading TBB (2020.3) and Boost (1.73.0) -- I've been using Stan compiled against those for a while now, with no issues -- let me know?

rok-cesnovar:
> happy to create PRs for upgrading TBB (2020.3) and Boost (1.73.0)

Thanks for offering the help! Much appreciated. 

There is no official policy on when to upgrade but have been doing the following recently: 

For Boost we tend to wait for the upgrade of the Boost Headers package on CRAN (Boost source on CRAN). Its currently still at 1.72. rstan relies on that version and we want to be in sync, so rstan can rely on Math working with the CRAN version of Boost.

For TBB we haven't upgraded recently because rstan uses the TBB version that is part of RcppParallel that is also still behind (I think at 2018 U1). But our use of TBB isn't really exotic so no huge urge to upgrade at the moment.

I will take you up on your offer once we decide do upgrade, if you will still be up for it :) Thanks.

src/stan/mcmc/hmc/base_hmc.hpp
```diff
@@ -132,8 +132,18 @@ class base_hmc : public base_mcmc {
     this->z_.ps_point::operator=(z_init);
   }
 
+  /**
+   * The mass matrix.
```
bob-carpenter, line 136: Sorry.  I should've been more specific.  

This should follow our genreal standards for Doxygen doc and ideally also use full sentences in sentence contexts.

So this should be:

```
/**
 * Return the representation of a point in phase space.
 *
 * @tparam Model a Stan model
 * @tparam BaseRNG base random number generator type
 * @return point in phase space
 */
```

`z_` here is the triple of `q` (position), `p` (velocity), and `g` (I don't know what that is).   I believe `g` may be the metric we're using here, since the function `PointType` has a `write_metric` method.  We've bee moving to "metric" from "mass matrix" to keep everything consistent.

That way the template parameters and return get documented.  I

kedartal: The class template parameters (`Hamiltonian`, `Model`, `BaseRNG`) should probably be documented on the class rather than on its methods, so I'm adding those comments there.
I'm not sure `base_hmc` is the right place to explain the nature of a *point* in the context of HMC, it should go in `ps_point` and `{unit|diag|dense}_e_point`. I think we could reference the [NUTS paper](https://arxiv.org/pdf/1111.4246.pdf) or some other documentation and make sure we use the same nomenclature for symbol names -- the term *phase space* does not appear in that paper so I used *(unconstrained) parameter space*, but of course happy to change to anything else.
Would this work?

```cpp
/**
 * Base class for Hamiltonian samplers.
 *
 * @tparam Model The type of the Stan model.
 * @tparam Hamiltonian The type of Hamiltonians over the (unconstrained)
 * parameter space.
 * @tparam Integrator The type of integrator (e.g. leapfrog).
 * @tparam BaseRNG The type of random number generator.
 */
 template <class Model, template <class, class> class Hamiltonian,
          template <class> class Integrator, class BaseRNG>
class base_hmc : public base_mcmc {

 ...

  /**
   * Gets the current point in the (unconstrained) parameter space.
   *
   * @return The current point in the (unconstrained) parameter space.
   */
  typename Hamiltonian<Model, BaseRNG>::PointType& z() { return z_; }

  /**
   * Gets the current point in the (unconstrained) parameters space.
   *
   * @return The current point in the (unconstrained) parameters space.
   */
  const typename Hamiltonian<Model, BaseRNG>::PointType& z() const noexcept {
    return z_;
  }

...
```

I believe `g` is the precomputed gradient for the Hamiltonian, see the `dphi_dq` methods in the `{unit|diag|dense}_e_metric` classes, and the comment in the `base_hamiltonian` class `// phi = 0.5 * log | Lambda (q) | + V(q)`

bob-carpenter:
> The class template parameters (Hamiltonian, Model, BaseRNG) should probably be documented on the class rather than on its methods, so I'm adding those comments there.

Thanks---that was my bad.  The doc should definitely go on the class, which makes it technically out of scope for this PR (though we never turn down better doc).

> term phase space does not appear in that paper so I used (unconstrained) parameter space, but of course happy to change to anything else.

I'm no longer sure what paper we're talking about.  Check out Radford Neal's introduction to HMC, for example, or any paper on Hamiltonian integration in the literature.  Phase space is the position of the unconstrained parameter (referred to as `q` in the Hamiltonian literature) and the velocity (`dq/dt`, referred to as `p` in the Hamiltonian literature).  Plus there's a `g` involved that I'm guessing is the metric.   The doc can't refer to it as a point in parameter space as that's only part of the object.  

> precomputed gradient for the Hamiltonian

The Hamiltonian `H(q, p)` itself has a gradient w.r.t. `q` and `p`, but that's a function, not a fixed vector, as it varies with respect to position `(q, p)` in phase space.  It could be a gradient function, though.  It'd be weird to put that on a point, but I don't understand the factoring of the code.

The NUTS paper is not a good reference.  [Michael Betancourt's conceptual intro to HMC](https://arxiv.org/pdf/1701.02434.pdf) on arXiv is much better as a gentle intro to what's going on.  The discussion of phase space starts on p. 22.  [Radford Neal's intro paper](https://arxiv.org/pdf/1206.1901.pdf) is also good and a bit simpler mathematically;  the plots in Figure 1 are all in phase space for a 1D problem (where phase space is 2D), but he doesn't name the space of `(q, p)` as the phase space.  See the [Wikipedia on phase space](https://en.wikipedia.org/wiki/Phase_space) for more generality.  Usually one starts ODE solvers with a harmonic oscillator where the phase space trajectories are simple orbits;  this is also what you get with negative log normal density as potential energy and a standard physical quadratic kinetic energy (quadratic in velocity `p`, i.e., `p' * inv(M) * p` where `M` is the mass matrix).

bob-carpenter:
I forgot to answer the direct question.

This:

```
 /**
   * Gets the current point in the (unconstrained) parameter space.
   *
   * @return The current point in the (unconstrained) parameter space.
   */
```

is wrong because there's more in the return than just the point `q` in unconstrained parameter space.    It returns an object of type `PointType`, which is going to resolve to something like this,

https://github.com/stan-dev/stan/blob/develop/src/stan/mcmc/hmc/hamiltonians/ps_point.hpp

where the doc is "Point in a generic phase space".  

You don't need to say "unconstrained" here because that's all a feature of the model---the HMC stuff is neutral with respect to how it gets an unconstrained desnity.

In order to cut out this back and forth, I can take over the doc, or you can just replace all the specifics with "point".  The only problem with that is that the user then needs to track through all the typedefs to figure out what is in a point.   It gets defined in

https://github.com/stan-dev/stan/blob/ed5a2ae124acdd015c27e1a60bdfe20fc420e8ce/src/stan/mcmc/hmc/hamiltonians/base_hamiltonian.hpp

The lack of doc for stuff like this is why we're requiring it now.  It's a challenge to disentangle from reading the code if you don't already know how it works.

kedartal:
@bob-carpenter :

I'd suggest we pick one or a few of those papers "officially", so we could start refactoring (mostly renaming) and documenting stuff with explicit references to the papers, using the same terminology as the papers. I really don't think we should expect people to piece together such understanding from fragments of inline code comments; rather, let's make familiarity with the selected one or a few papers a prerequisite to tweaking the relevant parts of Stan's code, and make sure it's as easy as possible by aligning symbol names. Slightly bigger in scope than this PR though :)

More concretely, a *point* in the Stan codebase does include everything you mentioned, and `z_` is indeed such a *point*; I think that when you say *point* you have in mind just the location (without momentum, metric, ...), that's just not how Stan's code is currently structured. I'm happy to change the wording to anything you feel appropriate, of course, but I do think it would be misleading if we don't use the term *point* or if we use it in a different sense than its meaning throughout the rest of the library.
I find that using *unconstrained* is a good reminder that the phase space is not the same as the constrained one defined by the model -- naturally, when integrating with Stan, there are lots of other pieces of code that do actually work in the constrained space. Using *phase space* may serve the same purpose, it would be great if we could select and reference a paper that uses it in the same way (other than [the NUTS paper](https://arxiv.org/pdf/1111.4246.pdf) since that paper does not use this term).

bob-carpenter:
For a baseline paper, we should ask @betanalpha.   He's the one who wrote most of the library.

I agree about the point terminology---it's just not a point in the unconstrained space---it's a point in phase space which contains a point in the unconstrained parameter space plus a velocity variable.  But I went ahead and merged this anyway because it's not worth the effort to get the doc right when there's only one function documented in the entire library. 

Thanks for adding template parameter doc.  I'd personally like to make all that grammatical, meaning simple noun phrases don't get capitalized or periods a the end, etc., and I wrote a doc style guide to that effect, but I'm having trouble enforcing it.  Check out the "function doc" section in https://github.com/stan-dev/stan/wiki/Coding-Style-and-Idioms

betanalpha:
Bob I wish that you wouldn't merge code HMC code without checking with me
first.  In this case the internals of the sampler are not technically part
of the API, which is why we never added accessors and mutators and instead
just left the member variables public for internal use. All of the public
routes are in defined in the services.  Exposing sampler instances properly
would require much more than (const) accessors, especially with the
variety of templating combinatorics.

Regarding notation ps_point is a point in phase space as well as any
relevant local information about the manifold and target density function.
For the base class that is just the point in configuration space q, the
momenta p, and the gradient, g; note that the momenta are not "velocity
variables" -- velocities live in the tangent space, momenta lives in the
cotangent space.  Derivations to Euclidean and more general Riemannian
configuration spaces also include local metric information, if necessary.

bob-carpenter:
> Bob I wish that you wouldn't merge code HMC code without checking with me first.

Feel free to chime in on the PRs while they're in progress---it's not like this discussion was happening in secret.  I didn't want to keep blocking a trivial change because of doc wording when the underlying return type wasn't even documented.   It'd be great if you could finish up the doc and clarify this doc.

> ... we never added accessors and mutators and instead just left the member variables public for internal use. 

This PR did not add any new accessors.  This mutable accessor was already public:

```
typename Hamiltonian<Model, BaseRNG>::PointType& z() { return z_; }
```

This This PR just adds a `const`-correct version of the very same accessor:

```
const typename Hamiltonian<Model, BaseRNG>::PointType& z() const noexcept {
    return z_;
}
```

kedartal:
@betanalpha 
Restricting users of the library to those APIs does not work for all use cases -- from trivial examples like hardcoding `boost::ecuyer1988` and `std::vector` along with its default allocator, to the less trivial distributed long-running processes with multiple parallel inferences over a large sets of models.
Saying that external code should not access anything outside the `stan::services` namespace makes Stan much less relevant for larger-scale applications. I don't think that's a good position for Stan, [Hyrum's Law](https://www.hyrumslaw.com/) aside.

betanalpha:
> \Feel free to chime in on the PRs while they're in progress---it's not like this discussion was happening in secret. I didn't want to keep blocking a trivial change because of doc wording when the underlying return type wasn't even documented. It'd be great if you could finish up the doc and clarify this doc.

I’d love to have chimed in but I’ve been without a laptop for the past two weeks 
with my old one dying and a having to order an emergency replacement all while 
rushing to find intermediate replacements for my ongoing courses.  It’s helpful to
at least tag relevant people instead of assuming that they’re actively not paying 
attention.

> Restricting users of the library to those APIs does not work for all use cases -- from trivial examples like hardcoding boost::ecuyer1988 and std::vector along with its default allocator, to the less trivial distributed long-running processes with multiple parallel inferences over a large sets of models.
> Saying that external code should not access anything outside the stan::services namespace makes Stan much less relevant for larger-scale applications. I don't think that's a good position for Stan, Hyrum's Law <https://www.hyrumslaw.com/> aside.

The Stan project does not and cannot maintain support for all use cases, which 
is why the API boundaries are so important.  Changing those boundaries is very 
much open for public discussion, but discussion should proceed any actual PRs.

bob-carpenter:
> ... API boundaries are so important.  Changing those boundaries is very much open for public discussion, but discussion should proceed any actual PRs.

I agree.  We have limited resources.  

In this case, the PR only added a const-correct version of an existing non-const accessor.  That's not changing the API boundary in terms of what a client can change or access.

kedartal:
@betanalpha Sorry about your laptop, if you ever need an office in Brooklyn to work from...
I don't think it's about the direction of the project as much as about a policy for accepting commits. @betanalpha says that code motivated by "other" use cases should not be admitted, I disagree. Keeping the project focused on the family clients (RStan, PyStan etc) and their typical use cases makes perfect sense, and has nothing to do with rejecting PRs like this one. Happy to move the discussion somewhere else, or drop it :)

PR #2955, opened 2020-09-09
allow reader to accept eigen vectors and use eigen map to access inputs

### LINKED ISSUE #2954: Use Eigen Map in `reader` instead of std vectors
#### Summary:

[`reader`](https://github.com/stan-dev/stan/blob/develop/src/stan/io/reader.hpp#L37) currently holds it's data as a reference to an `std::vector<T>`. idt this would matter for anything but v large problems, but I think we could be a bit more efficient by switching those to Eigen::Maps since Eigen maps can own the data of vectors but not vice versa.

For instance when call [`stan::model::gradient`](https://github.com/stan-dev/stan/blob/develop/src/stan/model/gradient.hpp) in [`update_potential_gradients`](https://github.com/stan-dev/stan/blob/develop/src/stan/mcmc/hmc/hamiltonians/base_hamiltonian.hpp#L64) we end up making an extra copy here when we call the model's log_prob with an eigen matrix.

```cpp
  template <bool propto__, bool jacobian__, typename T_>
  inline T_ log_prob(Eigen::Matrix<T_, Eigen::Dynamic, 1>& params_r,
                     std::ostream* pstream = nullptr) const {
    std::vector<T_> vec_params_r;
    vec_params_r.reserve(params_r.size());
    for (int i = 0; i < params_r.size(); ++i)
      vec_params_r.push_back(params_r(i));
    std::vector<int> vec_params_i;
    return log_prob<propto__, jacobian__, T_>(vec_params_r, vec_params_i,
                                              pstream);
  }
```

We can just have the compiler copy/paste the body of the `std::vector`'s log_prob signature so we avoid that copy

#### Current Version:
v2.24.0

SteveBronder:
#### Submission Checklist

#### Summary

Fixes #2954 so that `reader` can accept an eigen vector as well as an `std::vector`. This also has a few cleanups of some functions we can now call some vectorized versions. 

#### Intended Effect

See notes in #2954 

#### How to Verify

```
./runTests.py ./src/test/unit/io/reader_test
```

#### Copyright and Licensing


SteveBronder:
just fyi uses some stuff from [#1844](https://github.com/stan-dev/math/pull/1844) in the math library so it needs to wait for those

src/stan/io/reader.hpp
```diff
@@ -34,33 +34,35 @@ namespace io {
 template <typename T>
 class reader {
  private:
-  std::vector<T> &data_r_;
-  std::vector<int> &data_i_;
+  Eigen::Map<Eigen::Matrix<T, -1, 1>, Eigen::Aligned8, Eigen::Stride<0, 0>>
```
bbbales2, line 37: Why require alignment here?

bbbales2 (changes requested):
So `reader` takes as input a `std::vector<double>` and `std::vector<int>` and allows different shaped variables to get read out (vectors, row vectors, scalars, matrices, etc.).

The difference now is `reader` doesn't copy-on-return? Any chance it can return `map<const Eigen::MatrixXd>` types? We certainly don't want the underlying data changing.

src/stan/io/reader.hpp
bbbales2: I guess I don't know if Aligned8 is restrictive or not. Also the stride requirement?

SteveBronder: Umm, yeah I think we can remove both of these

bbbales2:
@SteveBronder is there any returning const-Maps or whatever to avoid modifying the memory of the underlying `std::vector` arguments passed in to the reader?

SteveBronder:
Yeah I think that's a good idea I'll update this

bbbales2:
@SteveBronder merge conflicts with the other reader changes

SteveBronder:
:+1:

bbbales2 (approved):

bbbales2:
:+1:

PR #2960, opened 2020-10-07
Add slices for var matrices

SteveBronder:
#### Submission Checklist

#### Summary

Adds specializations for the indexing for `var<mat>` and general Eigen matrices. There were a lot of times that eigen matrices were being subset using the default multi-index overload. But we can specialize a handful of these for Eigen matrices to get back more efficient slices. This will also work nicely when we can start returning expressions as sub setting can continue passing around expressions.

This PR looks very big, though all the changes are essentially in `rvalue.hpp` and the tests for `rvalue.hpp`. There's also a bit of code duplication in some places because we don't have a type trait to detect whether we have a var matrix or matrix of dynamic types vs. a var vector or vector.  If this is too much to review I'm happy to break this up into smaller PRs over the different indexing types. I could also break this up to do the indexing for `var<matrix>` in one PR and the indexing for `matrix<var>` in another

#### Intended Effect

This should make vector style indexing more efficient

#### How to Verify

Tests were added in the indexing folder for the rvalue specializations. Additional tests were added for checking that reverse min_max indices are correct

```
./runTests.py -j32 ./src/test/unit/model/indexing/
```

#### Documentation

Docs added for the `var<mat>` and `mat<var>` `rvalue()` specializations

#### Copyright and Licensing

### DIFF
--- a/src/stan/model/indexing/rvalue.hpp
+++ b/src/stan/model/indexing/rvalue.hpp
@@ -37,7 +37,7 @@ inline T rvalue(T&& c, const nil_index_list& /*idx*/, const char* /*name*/ = "",
  * Return the result of indexing a type without taking a subset. Mostly used as
  * an intermediary rvalue function when doing multiple subsets.
  *
- * Types:  type[,] : type
+ * Types:  type[omni] : type
  *
  * @tparam T Any type.
  * @param[in] an object.
@@ -228,7 +228,7 @@ inline auto rvalue(const EigMat& a,
                    const cons_index_list<index_uni, nil_index_list>& idx,
                    const char* name = "ANON", int depth = 0) {
   math::check_range("matrix[uni] indexing", name, a.rows(), idx.head_.n_);
-  return a.row(idx.head_.n_ - 1);
+  return a.row(idx.head_.n_ - 1).eval();
 }
 
 /**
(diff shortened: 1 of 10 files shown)


src/stan/model/indexing/rvalue.hpp
```diff
@@ -27,115 +28,186 @@ namespace model {
...
+ * @param[in] idx Index consisting of one omni-index.
  * @param[in] name String form of expression being evaluated.
  * @param[in] depth Depth of indexing dimension.
- * @return Result of indexing vector.
+ * @return Result of indexing matrix.
  */
 template <typename T>
-inline T rvalue(const Eigen::Matrix<T, Eigen::Dynamic, 1>& v,
-                const cons_index_list<index_uni, nil_index_list>& idx,
+inline T rvalue(T&& a, const cons_index_list<index_omni, nil_index_list>& idx,
```
SteveBronder, line 50: One thing I want to call out here, I actually don't want to have to do the Type&& perfect forwarding syntax here, but if I do it here and then have const & everywhere else there's ambiguity errors with the overloads and the ones that have a template parameter in the `cons_index_list<>` like `cons_index_list<L, nil_index_list>`

wds15:
This is really cool and should make quite a few programs faster.

SteveBronder:
Yeah I think so. I've always wondered what the choking points for vectorized code was and I think this is one of them.

wds15:
Well, we coded so far around vectorized stuff by using for loops whenever we need sub expressions from vectors or matrices. Much better if we do not need to do that in the future!

SteveBronder:
Closing since this has been moved to smaller PRs

Issue #2188, opened 2016-12-24
Stop sampling on errors that aren't domain_errors


syclik:
#### Summary:
Stan treats errors in bounds the same as domain errors. We want the sampler to stop when there are `invalid_argument` or `out_of_range` type of errors.

#### Description:
Errors like `invalid_argument` should stop the sampler. They are issues with the way the program was written and won't get better by rejecting the current iteration.

#### Reproducible Steps:
```
parameters {
   real<lower = 0, upper = 1> theta[1];
}
model {
  theta[2] ~ normal(0, 1);
}
```

#### Current Output:
From CmdStan:
```
...
Informational Message: The current Metropolis proposal is about to be rejected because of the following issue:
Exception thrown at line 5: []: accessing element out of range. index 2 out of range; expecting index to be between 1 and 1; index position = 1theta
If this warning occurs sporadically, such as for highly constrained variable types like covariance matrices, then the sampler is fine,
but if this warning occurs often then your model may be either severely ill-conditioned or misspecified.

Rejecting initial value:
  Error evaluating the log probability at the initial value.
...
Initialization between (-2, 2) failed after 100 attempts. 
 Try specifying initial values, reducing ranges of constrained values, or reparameterizing the model.
```

The rejection happens 100 times.

#### Expected Output:
It should quit after the first exception. It's not something that's going to be fixed after 100 iterations.

#### Additional Information:
We already have different exceptions worked through the math library.

#### Current Version:
v2.13.1

PR #2976, opened 2020-11-18
var<matrix> slicing and assignment

SteveBronder:
#### Submission Checklist

#### Summary

(Note: This PR is waiting on [#2188](https://github.com/stan-dev/math/pull/2188) in the math repo but I wanted to open it up for two main Qs)

This adds `rvalue()` and `assign()` specializations for `var<matrix>` types. Everything is pretty cool beans besides a few things

1. About 2/3rds of this PR is unit testing code @t4c1 would you want me to break this up into two PRs so that's easier to manage or is it easier for you to look at `assign()` and `rvalue()` in the same PR?

2. `assign(var<mat>, multi_index, var<mat>)` is terribly expensive. Since we don't have any guarantees that each item in a multi index assignment won't be assigned to more than once we need to do a de-duping of the multi-index to make sure only the last item from the assigned from matrix associated with a particular index in the multi-index is used. This is rough because it makes assignment O(n^2).

3. None of these `assign()` functions handle `assign(var<mat>, index, MatrixXd);`. Is that going to have to be supported as well? I think the easiest thing would be to promote the matrix of doubles to a `var<matrix>` in the compiler (or here in a specialization) and then do the assignment.

#### Intended Effect

Support assignment and slicing of `var<matrix>` types in the stan compiler's generated C++

#### How to Verify

Tests' have been added for all the same cases we currently tests for the standard `rvalue()` and `assign()` and can be run with

```
./runTests.py src/test/unit/model/indexing/rvalue_varmat_test.cpp src/test/unit/model/indexing/lvalue_varmat_test.cpp
```

#### Copyright and Licensing

### DIFF
--- a/src/stan/model/indexing/lvalue.hpp
+++ b/src/stan/model/indexing/lvalue.hpp
@@ -130,7 +130,8 @@ inline void assign(Vec1&& x,
  * the indexed size.
  */
 template <typename Vec1, typename Vec2,
-          require_all_eigen_vector_t<Vec1, Vec2>* = nullptr>
+          require_all_vector_t<Vec1, Vec2>* = nullptr,
+          require_all_not_std_vector_t<Vec1, Vec2>* = nullptr>
 inline void assign(Vec1&& x,
                    const cons_index_list<index_min_max, nil_index_list>& idxs,
                    const Vec2& y, const char* name = "ANON", int depth = 0) {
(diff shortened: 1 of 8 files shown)


t4c1:
This thing is huge. If you can split it up in a way that makes sense, I would prefer 4+ PRs.

> 2.

mat<var> assignment does not do anything special for duplicates. Than you also don`t need to.

> 3.

That needs to be supported. Doing it with a separate conversion is quite inefficient.

PR #2977, opened 2020-11-18
adds var matrix assignment

SteveBronder:
#### Submission Checklist

#### Summary

This adds `assign()` specializations for `var<matrix>` types. Everything is pretty cool beans besides a few things

1. The only file with new functions is `lvalue_varmat.hpp`. The normal `lvalue.hpp` file has only been changed in templates, docs, and a little helper function for reusing the current `assign()` methods for `var<matrix>` types. So there's only about 500 lines of meaningful code additions here while the rest is for testing. The file `lvalue_varmat.hpp` only holds specializations for when the index type is a `uni_index` or `multi_index` since those ones require special fiddling.

2. `assign(var<mat>, multi_index, var<mat>)` is terribly expensive. Since we don't have any guarantees that each item in a multi index assignment won't be assigned to more than once we need to do a de-duping of the multi-index to make sure only the last item from the assigned from matrix associated with a particular index in the multi-index is used. This is rough because it makes assignment O(n^2). For example the following is legal stan code

```stan
vector[3] A = {1, 2, 3};
vector[6] B = {1, 2, 3, 4, 5, 6};
A[{1, 1, 2, 2, 3, 3}] = B;
```

In normal stan code a `Matrix<var, -1, 1>` is going to just loop over that duplicated `multi_index` and only the values of B associated with the last duplicate index will actually be used (in the case above that's cells 1, 3, 5 for B).

3. None of these `assign()` functions handle `assign(var<mat>, index, MatrixXd);`. I think we should do that in a separate PR

#### Intended Effect

Support assignment of `var<matrix>` types in the stan compiler's generated C++

#### How to Verify

Tests' have been added for all the same cases we currently tests for the standard `assign()` and can be run with

```
./runTests.py src/test/unit/model/indexing/lvalue_varmat_test.cpp
```

#### Documentation

Docs updated for each overload

#### Copyright and Licensing


SteveBronder:
@t4c1 I managed to cut the size of the tests in about half by writing some extra utilities. I was also able to remove the specializations for `min_max_index` with another little utility function so there's just specializations for `multi_index` and `uni_index` stuff

SteveBronder:
Closing this to open up smaller ones

SteveBronder:
Shoot I closed the wrong pr

SteveBronder:
I think for assignment to a `var<matrix>` from a `matrix<double>` we should do it in a seperate PR. @t4c1 there's a few ways I can think of doing it

1. Specializations for each `assign()`
2. convert the double matrix with `to_var_value()` then do the assign
3. An `operator=` or something like it for `var<mat>` that takes in an Eigen matrix of doubles and then does value rewind thing. We could also have a method ilke `a.update_vals(b)` so that we don't accidentally do the value rewind when someone uses `operator=`

If you have any other schemes I'm def open to them!

t4c1:
I think 2. will be slow compared to other options. I don't understand what do you gain by doing 3., so my vote is for 1.

src/test/unit/model/indexing/lvalue_varmat_test.cpp
```diff
@@ -0,0 +1,760 @@
...
+  check_vector_adjs(check_i, x, "lhs");
+  EXPECT_MATRIX_EQ(y.adj(), Vec::Ones(5));
+}
+
+TEST_F(VarAssign, omni_vec) { test_omni_vec<Eigen::VectorXd>(); }
+
+TEST_F(VarAssign, omni_rowvec) { test_omni_vec<Eigen::RowVectorXd>(); }
+
+template <typename Vec>
+void test_positive_minmax_varvector() {
```
t4c1, line 216: In what way is this test different than `test_minmax_vec`?

src/test/unit/model/indexing/lvalue_varmat_test.cpp
```diff
@@ -0,0 +1,760 @@
...
+
+TEST_F(VarAssign, uni_uni_std_vecvec) {
+  test_uni_uni_vec_eigvec<Eigen::VectorXd>();
+}
+
+TEST_F(VarAssign, uni_uni_std_vecrowvec) {
+  test_uni_uni_vec_eigvec<Eigen::RowVectorXd>();
+}
+
+// Uni Assigns
```
t4c1, line 334: Is that comment supposed to tell me anything?

src/test/unit/model/indexing/lvalue_varmat_test.cpp
```diff
@@ -0,0 +1,760 @@
...
+  auto y = generate_linear_var_vector<Eigen::VectorXd>(5, 10);
+  assign(x, index_list(index_omni(), index_uni(1)), y);
+  EXPECT_MATRIX_EQ(y.val(), x.val().col(0));
+  sum(x).grad();
+  EXPECT_MATRIX_EQ(x.val(), x_val);
+  auto check_i = [](int i) { return true; };
+  auto check_j = [](int j) { return j == 0; };
+  check_matrix_adjs(check_i, check_j, x, "lhs", 0);
+  EXPECT_MATRIX_EQ(y.adj(), Eigen::VectorXd::Ones(5));
+}
```
t4c1, line 760: I like the organization of tests. However I think some index combinations are missing. For example matrix with min and omni index. All allowed index combinations should be tested.

t4c1 (changes requested):

src/stan/model/indexing/lvalue.hpp
```diff
@@ -129,7 +130,8 @@ inline void assign(Vec1&& x,
  * the indexed size.
  */
 template <typename Vec1, typename Vec2,
-          require_all_eigen_vector_t<Vec1, Vec2>* = nullptr>
+          require_all_vector_t<Vec1, Vec2>* = nullptr,
+          require_all_not_std_vector_t<Vec1, Vec2>* = nullptr>
```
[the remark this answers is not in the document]
SteveBronder: `require_all_vector_t` is for any vector (Eigen row and col vectors, `var<row/col vector>`, and `std::vector<>`). We need the `not_std_vector_t<>` part here because we don't have anything in the library for just eigen vector or var vector

src/stan/model/indexing/lvalue_varmat.hpp
```diff
@@ -0,0 +1,396 @@
...
+inline void assign(Vec1&& x,
+                   const cons_index_list<index_multi, nil_index_list>& idxs,
+                   const Vec2& y, const char* name = "ANON", int depth = 0) {
+  stan::math::check_size_match("vector[multi] assign", "left hand side",
+                               idxs.head_.ns_.size(), name, y.size());
+  const auto x_size = x.size();
+  const auto assign_size = idxs.head_.ns_.size();
+  arena_t<std::vector<int>> x_idx;
+  arena_t<std::vector<int>> y_idx;
+  // Need to remove duplicates for cases like {2, 3, 2, 2}
```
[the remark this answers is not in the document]
SteveBronder: When I was working through this, if we have something like 

```stan
vector[3] x;
vector[6] y;
x[{1, 2, 3, 1, 2, 3}] = y;
```

Since the `Matrix<var>` effectively just assigns a pointer this is effectivly doing a de-dup and taking the last value assigned like

```
x = y[{4, 5, 6}];
```

If we don't de-dup in `var<mat>` then the y indices `{1, 2, 3}` would differ between the `mat<var>` and `var<mat>` version, where in the `mat<var>` version it's as if they were never touched but in the `var<mat>` version we would end up propagate up those adjoints. Does that make sense? I'd love to be full of baloney on this one but I think we need to do this de-duping scheme here

I've been trying to think of smarter ways to do this but I keep getting stuck at how to get out the proper `y` values

src/stan/model/indexing/lvalue_varmat.hpp
```diff
@@ -0,0 +1,396 @@
...
+  for (int i = assign_size - 1; i >= 0; --i) {
+    if (!internal::check_duplicate(x_idx, idxs.head_.ns_[i] - 1)) {
+      stan::math::check_range("vector[multi] assign", name, x_size,
+                              idxs.head_.ns_[i]);
+      y_idx.push_back(i);
```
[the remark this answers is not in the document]
SteveBronder: I'm not sure I'm following what you mean here, I just need `y_idx` for the duplicate index thing

src/stan/model/indexing/lvalue_varmat.hpp
```diff
@@ -0,0 +1,396 @@
...
+  }
+  for (Eigen::Index i = 0; i < x_idx.size(); ++i) {
+    x.vi_->val_.coeffRef(x_idx[i]) = y.vi_->val_.coeff(y_idx[i]);
+  }
+  stan::math::reverse_pass_callback([x, y, x_idx, y_idx, prev_vals]() mutable {
+    for (Eigen::Index i = 0; i < x_idx.size(); ++i) {
+      x.vi_->val_.coeffRef(x_idx[i]) = prev_vals.coeffRef(i);
+      prev_vals.coeffRef(i) = x.adj().coeffRef(x_idx[i]);
+      x.adj().coeffRef(x_idx[i]) = 0.0;
+      y.adj().coeffRef(y_idx[i]) += prev_vals.coeffRef(i);
```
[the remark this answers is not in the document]
SteveBronder: Only for the `uni_index` thought right? If we have something like 
```cpp
assign(x, index_multi({1, 2, 3}, x)
```

We would run into the aliasing issue wouldn't we?

src/stan/model/indexing/lvalue_varmat.hpp
```diff
@@ -0,0 +1,396 @@
...
+  stan::math::check_size_match("matrix[multi,multi] assign", "left hand side",
+                               assign_rows, name, y.rows());
+  arena_t<std::vector<int>> x_idx;
+  arena_t<std::vector<int>> y_idx;
+  x_idx.reserve(assign_rows);
+  y_idx.reserve(assign_rows);
+  // Need to remove duplicates for cases like {2, 3, 2, 2}
+  for (int i = assign_rows - 1; i >= 0; --i) {
+    if (!internal::check_duplicate(x_idx, idxs.head_.ns_[i] - 1)) {
+      y_idx.push_back(i);
```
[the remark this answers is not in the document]
SteveBronder: I'm not following why we can remove `y_idx`?

src/test/unit/util.hpp
```diff
@@ -59,6 +59,36 @@ void match_csv_columns(const Eigen::MatrixXd& samples,
...
 }
 
+/**
+ * Tests for  exact elementwise equality of the input matrices
+ * with the EXPECT_EQ macro from GTest.
+ *
+ * @param A first input matrix to compare
+ * @param B second input matrix to compare
+ */
+#ifndef EXPECT_MATRIX_EQ
```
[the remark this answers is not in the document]
SteveBronder: We tend to keep the tests for stan math separate so I'd rather define it here. idt anywhere else in the testing for the stan repo do we pull out anything from the math testing repo. This is just so that in the future if we ever merge the two libraries the macro is not defined twice

src/test/unit/model/indexing/lvalue_varmat_test.cpp
SteveBronder: Good catch I'll delete `test_minmax_vec`

src/test/unit/model/indexing/lvalue_varmat_test.cpp
SteveBronder: Nah not really

src/test/unit/model/indexing/lvalue_varmat_test.cpp
SteveBronder: Ooof that's a lot. What if here I did all single index and each one being on the right side of the double index? Doing all the combinations manually would be a lot of testing

SteveBronder: I added the above special case tests. It's not 100% converage but I think it captures a lot

src/stan/model/indexing/lvalue_varmat.hpp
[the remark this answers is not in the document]
t4c1: I mean you can completely remove the variable `y_idx` and replace all uses `y_idx[i]` with `i`.

src/test/unit/model/indexing/lvalue_varmat_test.cpp
t4c1: But we do support all index combinations. That means not everything we support is tested. I don't like that. 

Although writing 63 tests by hand really is a lot. Maybe it is possible to do something smart using templates or macros? For example instead of writing expected values by hand, you could write code for converting every index type into equivalent omni index. Than only omni omni combination needs to be tested by hand and all others can be compared to omni omni.

src/stan/model/indexing/access_helpers.hpp
```diff
@@ -0,0 +1,54 @@
...
+
+template <typename T, require_var_matrix_t<T>* = nullptr>
+inline auto colwise_reverse(T&& x) {
+  return std::forward<T>(x).colwise_reverse();
+}
+
+template <typename T, require_eigen_t<T>* = nullptr>
+inline auto colwise_reverse(T&& x) {
+  return std::forward<T>(x).colwise().reverse();
+}
```
t4c1, line 31: These also do not need forwarding.

src/stan/model/indexing/lvalue_varmat.hpp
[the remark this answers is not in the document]
t4c1: Oh right, aliasing. Which your code also does not handle. To handle it, you need to split this into two separate loops - one reading from adjoints and another one writing. And add tests for aliasing too.

src/stan/model/indexing/lvalue_varmat.hpp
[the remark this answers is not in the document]
t4c1: I see the issue. But that quadratic deduplication will hurt too much on large matrices. Deduplication can be done in linear time if you do duplicate checks on `std::unordered_set` instead of `std::vector`. To improve memory access pattern you should build that set in the same loop you are storing indices and do checks on it in the reverse pass loop.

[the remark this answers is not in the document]
SteveBronder: Do you mean something like this?

```
  std::unordered_set<int, std::hash<int>, std::equal_to<int>, stan::math::arena_allocator<int>> x_set;
  arena_t<std::vector<int>> x_idx;
  arena_t<std::vector<int>> y_idx;
  x_set.reserve(assign_size);
  x_idx.reserve(assign_size);
  y_idx.reserve(assign_size);
  // Need to remove duplicates for cases like {2, 3, 2, 2}
  for (int i = assign_size - 1; i >= 0; --i) {
    if (x_set.insert(idxs.head_.ns_[i]).second) {
      stan::math::check_range("vector[multi] assign", name, x_size,
                              idxs.head_.ns_[i]);
      y_idx.push_back(i);
      x_idx.push_back(idxs.head_.ns_[i] - 1);
    }
  }
```

else we could store them in a regular set and not have to have the extra mem (I think?)

src/stan/model/indexing/lvalue_varmat.hpp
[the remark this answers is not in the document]
SteveBronder: What about 

```stan
x[{1, 2, 2, 3}] = y
```

If we don't store the y index associated with the deduped x then we'll end up iterating over y[{4, 3, 2}] where we want to iterate over `y[{4, 3, 1}]`

src/stan/model/indexing/lvalue_varmat.hpp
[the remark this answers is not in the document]
t4c1: Yeah I meant like this. Or maybe we should instead just document it that duplicating indices on left hand side of assignment is not supported.

src/stan/model/indexing/lvalue_varmat.hpp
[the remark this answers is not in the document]
t4c1: I see. How about my implementation:
```
template <typename Vec1, typename Vec2,
          require_all_var_vector_t<Vec1, Vec2>* = nullptr>
inline void assign(Vec1&& x,
                   const cons_index_list<index_multi, nil_index_list>& idxs,
                   const Vec2& y, const char* name = "ANON", int depth = 0) {
  stan::math::check_size_match("vector[multi] assign", "left hand side",
                               idxs.head_.ns_.size(), name, y.size());
  const auto x_size = x.size();
  const auto assign_size = idxs.head_.ns_.size();
  arena_t<std::vector<int>> x_idx(assign_size);
  arena_t<Eigen::Matrix<double, -1, 1>> prev_vals(assign_size);
  std::unordered_set<int> x_set();
  x_set.reserve(assign_size);
  for (int i = assign_size - 1; i >= 0; --i) {
    if (likely(x_set.insert(idxs.head_.ns_[i]).second)) {
      stan::math::check_range("vector[multi] assign", name, x_size,
                              idxs.head_.ns_[i]);
      x_idx[i] = idxs.head_.ns_[i] - 1;
      prev_vals.coeffRef(i) = x.vi_->val_.coeffRef(x_idx[i]);
      x.vi_->val_.coeffRef(x_idx[i]) = y.vi_->val_.coeff(i);
	}
	else{
	  x_idx[i] = -1;
	}
  }
  stan::math::reverse_pass_callback([x, y, x_idx, prev_vals]() mutable {
    for (Eigen::Index i = 0; i < x_idx.size(); ++i) {
	  if(likely(x_idx[i]! = -1)){
        x.vi_->val_.coeffRef(x_idx[i]) = prev_vals.coeffRef(i);
        prev_vals.coeffRef(i) = x.adj().coeffRef(x_idx[i]);
        x.adj().coeffRef(x_idx[i]) = 0.0;
        y.adj().coeffRef(i) += prev_vals.coeffRef(i);
	  }
    }
  });
}
```
This avoids `y_idx` and the use of `x_set` in reverse pass at expense of additional if in reverse pass. I believe this would be faster.

src/stan/model/indexing/lvalue_varmat.hpp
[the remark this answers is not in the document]
SteveBronder: I think we should do the unordered_set thing for now. I'll make an issue in stanc3 a `unique_index` type that would just be an `std::vector` but checks at compilation that each element is unique. Then we can have a specialization to handle unique index that doesn't need to do the de-dup

src/stan/model/indexing/lvalue_varmat.hpp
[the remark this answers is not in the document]
SteveBronder: Ah! I like that

src/stan/model/indexing/lvalue_varmat.hpp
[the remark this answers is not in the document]
SteveBronder: I'm getting confused. For something like

``cpp
  auto x = generate_linear_var_vector<Vec>(5, 1);
  vector<int> ns{1, 4, 1};
  assign(x, index_list(index_multi(ns)), x.segment(0, 3));
  stan::math::sum(x).grad();
  std::cout << "\nx adj post-grad: \n" << x.adj() << "\n";
```

Should the adjoint for `x` be 

```
x adj post-grad: 
0
2
2
0
1
```

if we did a `matrix<var>` here we would be assigning cells `{1, 2}` to `{0, 3}`. So for that sum those vari pointers are used twice each (for `{0 -> 1}` and `{2 -> 3}`). So to go in reverse here for `var<mat>` the cells `{1, 2}` should have a adjoints of 2 since they were used twice and cells `{0, 3}` would have adjoints of zero. 

Is that right? If so then the current form does handle the aliasing. If not then some help here would be appreciated haha

File: src/test/unit/model/indexing/lvalue_varmat_test.cpp (hunk 3 above)
Comment (SteveBronder) [reply src/test/unit/model/indexing/lvalue_varmat_test.cpp:979]: But for the double index tests we are doing the specialization for the index on the right hand side and the specialization for the index as if it was just grabbing the rows. So while the way I have it now doesn't do all the cases doing the single index specialization and double index specialization does check each path. I'm not really understanding what edge cases we would capture by writing out each one individually

Comment (t4c1) [reply src/test/unit/model/indexing/lvalue_varmat_test.cpp:998]: Well for example an "edge case" of someone optimizing by adding a separate overload for a specific combination of indices. Also that way you test that all combinations actually can be instantiated.

File: src/stan/model/indexing/access_helpers.hpp [hunk 11]
```diff
@@ -12,42 +12,24 @@ namespace internal {
 // Internal helpers so we can reuse min_max index assign for Eigen/var<Eigen>
 template <typename T, require_var_matrix_t<T>* = nullptr>
 inline auto rowwise_reverse(T&& x) {
```
t4c1, line 14:
```suggestion
inline auto rowwise_reverse(const T& x) {
```

src/stan/model/indexing/lvalue_varmat.hpp
```diff
@@ -306,37 +309,53 @@ inline void assign(
...
-      if (!internal::check_duplicate(x_idx, idxs.head_.ns_[i] - 1,
-                                     idxs.tail_.head_.ns_[j] - 1)) {
-        y_idx.push_back(std::array<int, 2>{i, j});
-        stan::math::check_range("matrix[multi, multi] assign row", name,
-                                x.rows(), idxs.head_.ns_[i]);
-        stan::math::check_range("matrix[multi, multi] assign col", name,
-                                x.cols(), idxs.tail_.head_.ns_[j]);
-        x_idx.push_back(std::array<int, 2>{idxs.head_.ns_[i] - 1,
-                                           idxs.tail_.head_.ns_[j] - 1});
+    if (likely(x_set.insert(idxs.tail_.head_.ns_[j]).second)) {
```
t4c1, line 332: For multi multi overload you need to deduplicate both indices at the same time - by creating a set of pairs instead of two sets. This code would also "deduplicate" cases like `{2,2,3}, {2,3,3}`.

t4c1 (changes requested):

src/stan/model/indexing/lvalue_varmat.hpp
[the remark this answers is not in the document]
t4c1: In this example there is no aliasing - no value is both read and written in the assignment.

You can instead try this:
```
  auto x = generate_linear_var_vector<Vec>(5, 1);
  vector<int> ns{2, 3, 4};
  std::cout << "\nx val pre-assign: \n" << x.val() << "\n";
  assign(x, index_list(index_multi(ns)), x.segment(0, 3));
  std::cout << "\nx val post-assign: \n" << x.val() << "\n";
  stan::math::sum(x).grad();
  std::cout << "\nx adj post-grad: \n" << x.adj() << "\n";
```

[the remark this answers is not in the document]
t4c1: Sorry, I posted the code without trying to run it. The example abova also does not show the issue. This on, however, does:
```
   auto x = generate_linear_var_vector<Eigen::VectorXd>(5, 1);
   vector<int> ns{1, 2, 3};
   std::cout << "\nx val pre-assign: \n" << x.val() << "\n";
   assign(x, index_list(index_multi(ns)), x.segment(1, 3));
   std::cout << "\nx val post-assign: \n" << x.val() << "\n";
   stan::math::sum(x).grad();
   std::cout << "\nx adj post-grad: \n" << x.adj() << "\n";
```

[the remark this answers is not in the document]
SteveBronder: Ooof yes thanks! I updated all the multi indexing assignment functions to use separate loops for reading and writing on the forward pass and a separate loop for writing to the assigned from matrix on the reverse pass. I think this is the right scheme. 

Also I sort of gave up on being clever for when multi-index is on the right hand side of a double index assign and just evaluated y before looping over it. I started writing out the code to handle the aliasing manually like the rest of the `assign()` overloads but tbh it felt like the answer was more complicated/expensive relative to just evaluating y once before doing the column assignments.

src/stan/model/indexing/lvalue_varmat.hpp
SteveBronder: Yeah so this one took me a while to figure out but this is actually safe. You can checkout the `multi_multi_matrix` test here to see that I test against "deduping" against a set of pairs.

The important thing to note is that for something like `{1, 2, 3}, {4, 5, 6}` you can read it as "for column 4 assign to cells 1, 2, 3. For column 5 assign to cells 1, 2, 3, for column six...". So the rows assignment indices are reused for each column. That let's us just deduplicate the row assignment vector once

src/stan/model/indexing/lvalue_varmat.hpp
```diff
@@ -0,0 +1,458 @@
...
+      x.vi_->val_.coeffRef(x_idx[i]) = y_vals.coeff(i);
+    }
+  }
+
+  stan::math::reverse_pass_callback([x, y, x_idx, prev_vals]() mutable {
+    for (Eigen::Index i = 0; i < x_idx.size(); ++i) {
+      if (likely(x_idx[i] != -1)) {
+        x.vi_->val_.coeffRef(x_idx[i]) = prev_vals.coeffRef(i);
+        prev_vals.coeffRef(i) = x.adj().coeffRef(x_idx[i]);
+        x.adj().coeffRef(x_idx[i]) = 0.0;
```
t4c1, line 132: This line is writing adjoint, so it should be in the second loop.

src/stan/model/indexing/lvalue_varmat.hpp
t4c1: Right, Stan has a bit different understanding of 2D indexing than I do. This is fine.

t4c1 (changes requested):
The code now mostly looks good. However I would still like to see all index combinations tested.

src/stan/model/indexing/lvalue_varmat.hpp
SteveBronder: I think it needs to be in the first loop to handle values that overlap (unless I'm getting the pattern wrong). Because if we write to x and y in the same loop we are effectivly writing to X in both and I think the aliasing issue comes from that.

For like multi_vec

```
template <typename Vec>
void test_multi_alias_vec() {
  using stan::math::sum;
  using stan::math::var_value;
  using stan::model::test::check_vector_adjs;
  using stan::model::test::generate_linear_var_vector;
  auto x = generate_linear_var_vector<Eigen::VectorXd>(5, 1);
  Eigen::VectorXd x_val = x.val();
  vector<int> ns{1, 1, 2, 3};
  std::cout <<  "\nx val pre assign: \n" << x.val() << "\n";
  assign(x, index_list(index_multi(ns)), x.segment(1, 4));
  EXPECT_MATRIX_EQ(x.val().segment(0, 3), x_val.segment(2, 3));
  EXPECT_MATRIX_EQ(x.val().segment(3, 2), x_val.segment(3, 2));
  std::cout <<  "\nx val post assign: \n" << x.val() << "\n";
  stan::math::sum(x).grad();
  std::cout <<  "\nx adj: \n" << x.adj() << "\n";
  EXPECT_MATRIX_EQ(x.val(), x_val);
  Eigen::VectorXd exp_adj(5);
  exp_adj << 0, 0, 1, 2, 2;
  EXPECT_MATRIX_EQ(x.adj(), exp_adj);
}
```

We have 

```
x val pre assign: 
1
2
3
4
5

x val post assign: 
3
4
5
4
5
```

So if this was an `Eigen::Matrix<var>` it would be like cells {2, 3, 4} were put in cells {0, 1, 2}. So when we do `grad()` we should get back

```
x adj: 
0
0
1
2
2
```

Since the "var" in cells {0, 1} are overwritten their values would be zero, cell 2 overwites cell 0 so it's value should be one, and cells {3, 4} are used twice so their adjoints should be 2. But if we move 

```
x.adj().coeffRef(x_idx[i]) = 0.0;
```

To the second loop we get back

```
x adj: 
0
0
0
2
2
```

SteveBronder:
I think once we get rvalue in we can use rvalue to test all the combinations against assign. If I wrote some test suite to test all these combinations that's in effect what I would be doing anyway

src/stan/model/indexing/lvalue_varmat.hpp
t4c1: If this is setting adjoints to zero in the first loop in can set some element to zero that is due to aliasing later read in the same loop. But you are right too. I believe we need to add a third loop just for setting adjoints to zero.

SteveBronder: oof yeah I think you are right.

SteveBronder: Actually idt we need a third loop.

`x_idx` in the below is going to only access unique values so it will look something like `{-1, 3, -1, 2, 1, 0}` or something of that ilk. 

```cpp
    for (Eigen::Index i = 0; i < x_idx.size(); ++i) {
      if (likely(x_idx[i] != -1)) {
        x.vi_->val_.coeffRef(x_idx[i]) = prev_vals.coeffRef(i);
        prev_vals.coeffRef(i) = x.adj().coeffRef(x_idx[i]);
        x.adj().coeffRef(x_idx[i]) = 0.0;
      }
    }
```

t4c1: You are right, I got confused here. This is fine.

t4c1:
The code now looks ok. I don't like the idea of testing this just by comparing with rvalue, although that would be better than nothing. Unless you also intend to test all possible rvalue signatures in some other way?

SteveBronder:
The main thing for me is that I don't really know how to meaningfully write a test suite to test all of these combinations that won't end up just being the equivalent of rewriting assign or rvalue. I think what I'd like to do is use rvalue to test assign and assign to test rvalue. Then at least if we can guarantee a level of consistency between the two

t4c1:
How about my suggestion in https://github.com/stan-dev/stan/pull/2977#discussion_r533234711 to compare all indices with multi index? Do you think that is unfeasable for some reason?

SteveBronder:
It's not that it's unfeasable it's that I think it would take me a lot of time and I'm not seeing what the benefit of 100% path coverage is here. These tests do have 100% code coverage for `assign()` (I just added all the bounds and range checks to the tests) so I'm having a hard time seeing how full path coverage is going to find a new bug. You said in a previous comment you would be worried about someone writing a new specialization that is not tested here, but for that PR why wouldn't we just have the person write a new test?

Maybe I'm missing something here? Can you think of an example path we don't test right now that would fail? If there is one then I'd be pretty easily convinced that we do need 100% path coverage

SteveBronder:
(also ignore the comment in that commit that's for `assign()`)

t4c1:
> Can you think of an example path we don't test right now that would fail?

Yes. Lets say you forgot to handle and test a certain obscure combination of index types. There are too many for me and probably for anyone to keep all in their head. A month after release someone uses it in model and gets a 10MB error from gcc about not finding correct overload.

SteveBronder:
Alright I'll have to eat my shoe here and apologize for being obstinate. I thought I'd have to do some crazy tuple thing to do all of these but I managed to get something up that's pretty clean by just checking against multi-index. The only thing is that it doesn't handle repeated multi-indices like `{1, 2, 2, 3}` because that changes the adjoints of the input to 0 in one row or col. We have specific tests for that in other places so I think that's fine

t4c1 (approved):
Looks great!

SteveBronder:
Ty! @syclik are my permissions not able to let me merge to the stan repo? @t4c1 would you mind clicking the merge button on this?

PR #2979, opened 2020-11-23
allow reader to read out var<matrix> types

SteveBronder:
#### Submission Checklist

#### Summary

This adds functions in reader that allow it to output a `var<matrix>`

#### Intended Effect

This will be used by the stan compiler when reading out `var<matrix>` types in log_prob, write_array, etc

#### How to Verify

Tests were added to check that io works for the normal path, zero sized path, and when the reader has a `double` type (which for `var<matrix>` function we need to return back a `matrix<double>`

```
./runTests.py ./src/test/unit/io/reader_test
```

#### Side Effects

This currently only supports standard io and not lower bound, upper bound, correlation matrices, etc. type vars. We need a few things in stan math before we can support all those

#### Documentation

Docs added for each new function

#### Copyright and Licensing

### DIFF
--- a/src/stan/io/reader.hpp
+++ b/src/stan/io/reader.hpp
@@ -55,16 +55,19 @@ class reader {
 
  public:
   using matrix_t = Eigen::Matrix<T, Eigen::Dynamic, Eigen::Dynamic>;
-  using vector_t =  Eigen::Matrix<T, Eigen::Dynamic, 1>;
+  using vector_t = Eigen::Matrix<T, Eigen::Dynamic, 1>;
   using row_vector_t = Eigen::Matrix<T, 1, Eigen::Dynamic>;
 
   using map_matrix_t = Eigen::Map<matrix_t>;
   using map_vector_t = Eigen::Map<vector_t>;
   using map_row_vector_t = Eigen::Map<row_vector_t>;
 
-  using var_matrix_t = stan::math::var_value<Eigen::Matrix<double, Eigen::Dynamic, Eigen::Dynamic>>;
-  using var_vector_t =  stan::math::var_value<Eigen::Matrix<double, Eigen::Dynamic, 1>>;
-  using var_row_vector_t = stan::math::var_value<Eigen::Matrix<double, 1, Eigen::Dynamic>>;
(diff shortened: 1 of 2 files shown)


src/stan/io/reader.hpp
```diff
@@ -186,6 +193,33 @@ class reader {
...
+
+  /**
+   * Return a `var_value` with inner type column vector with specified
+   * dimensionality made up of the next scalars.
+   *
+   * @param m Number of rows in the vector to read.
+   * @return Column vector made up of the next scalars.
+   */
+  template <typename T_ = T, require_st_var<T_> * = nullptr>
+  inline var_vector_t var_vector(size_t m) {
```
bbbales2, line 205: Why is this a new function and not just overloading `vector(size_t m)`?

src/test/unit/io/reader_test.cpp
```diff
@@ -1429,3 +1429,93 @@ TEST(IoReader, UnitVectorThrows) {
   EXPECT_THROW(reader.unit_vector_constrain(x), std::invalid_argument);
   EXPECT_THROW(reader.unit_vector_constrain(x, lp), std::invalid_argument);
 }
+
+TEST(IoReader, var_vector) {
```
bbbales2, line 1433: Since there's the `to_var_value` here it's probably worth adding this to all the var tests:

```
vec_x.adj() = Eigen::VectorXd::Random(vec_x.size());
stan::math::grad();
for(size_t i = 0;  i < vec_x.size(); ++i) {
  EXPECT_FLOAT_EQ(vec_x.adj()(i), theta[i].adj());
}
```

I know the other tests don't do it -- just looks like something that would be easy to miss if someone comes through in the future and easy to test.

bbbales2 (changes requested):
One question in the code and a question on the tests. Basically looks good!

src/test/unit/io/reader_test.cpp
bbbales2: This we talked about and I'm fine not having a test

bbbales2 (approved):

PR #2970, opened 2020-10-31
Split large .stan test models

rok-cesnovar:
This splits large test files in two, to fix a part of https://github.com/stan-dev/math/issues/2178

This was done with a script, so if anyone feels its easier to review the script, the script is here: https://gist.github.com/rok-cesnovar/70db1bfc69233b2a334208d4f34f7808

#### Copyright and Licensing

Please list the copyright holder for the work you are submitting (this will be you or your assignee, such as a university or company):
Rok Češnovar, Uni. of Ljubljana

By submitting this pull request, the copyright holder is agreeing to license the submitted work under the following licenses:
- Code: BSD 3-clause (https://opensource.org/licenses/BSD-3-Clause)
- Documentation: CC-BY 4.0 (https://creativecommons.org/licenses/by/4.0/)

PR #2983, opened 2020-12-04
Windows integration tests optimization

rok-cesnovar:
#### Summary

If Windows integration tests run on the gelman-group-win2 machine, they seem to run for 7h to 16h (https://jenkins.mc-stan.org/blue/organizations/jenkins/Stan/detail/develop/746/pipeline/166, https://jenkins.mc-stan.org/blue/organizations/jenkins/Stan/detail/develop/738/pipeline/166).

This PR will first attempt to check if O=0 could be used (it was switched to O=1 due to assembler file size issues for huge test models). This might not be needed after https://github.com/stan-dev/stan/pull/2970

If this will not help, we will limit Windows integration tests to AWS Windows agents.

#### Copyright and Licensing

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -20,9 +20,9 @@ def runTests(String testPath, Boolean separateMakeStep=true) {
 def runTestsWin(String testPath, Boolean separateMakeStep=true) {
     withEnv(['PATH+TBB=./lib/stan_math/lib/tbb']) {
        if (separateMakeStep) {
-           bat "runTests.py -j${env.PARALLEL} ${testPath} --make-only"
+           bat "runTests.py -j8 ${testPath} --make-only"
        }
-       try { bat "runTests.py -j${env.PARALLEL} ${testPath}" }
+       try { bat "runTests.py -j8 ${testPath}" }
        finally { junit 'test/**/*.xml' }
     }
 }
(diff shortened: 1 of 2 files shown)


SteveBronder:
I kind of worry that not testing at O3 for windows could reveal some weirdness in the future. Is it okay to just use AWS windows agents? Or only set O0 when compiling on the gelman windows comp

rok-cesnovar:
We have never tested with anything above O=0 in integration tests. Until 6 months ago we only ran them on Linux.

We also test Mac/Windows integration on merges to develop now. Windows requires at least O=1, because otherwise the assembler file is too big ( regardless of the machine it runs on) because we have some really large models (1500+ lines of calling distributions with all possible signatures).

This is an exploratory PR to see if we could speedup windows testing a bit. First tried if O=0 could work after we split the largest test models. The answer for that is no. I am now trying out if limiting the parallel setting would help. If that does not help I will limit these tests to AWS. I think a test run that runs for 15h to a day is not what we want to run on all merges.

As for O=3 tests, I think its not feasible to compile 750+ models with that setting. Maybe a subset yes, but I cant imagine how long it would run for the whole and we need to be reasonable.

SteveBronder:
Oh my apologies I assumed we were already testing those at O3

PR #2988, opened 2020-12-11
remove const from reader inner maps

SteveBronder:
#### Submission Checklist

#### Summary

This is just a little fix, reader always makes copies of it's output in a stan model, but having these be const maps stops them from being assigned directly assigned to non-const maps in the model if we ever want to do that in the future.

#### Copyright and Licensing

### DIFF
--- a/src/stan/io/reader.hpp
+++ b/src/stan/io/reader.hpp
@@ -39,16 +39,16 @@ class reader {
   size_t pos_{0};
   size_t int_pos_{0};
 
-  inline T& scalar_ptr() { return data_r_.coeffRef(pos_); }
+  inline T &scalar_ptr() { return data_r_.coeffRef(pos_); }
 
-  inline T& scalar_ptr_increment(size_t m) {
+  inline T &scalar_ptr_increment(size_t m) {
     pos_ += m;
     return data_r_.coeffRef(pos_ - m);
   }
 
-  inline int& int_ptr() { return data_i_.coeffRef(int_pos_); }
+  inline int &int_ptr() { return data_i_.coeffRef(int_pos_); }
 
-  inline int& int_ptr_increment(size_t m) {
+  inline int &int_ptr_increment(size_t m) {
     int_pos_ += m;
     return data_i_.coeffRef(int_pos_ - m);
   }
--- a/src/stan/io/reader.hpp
+++ b/src/stan/io/reader.hpp
@@ -34,21 +34,21 @@ namespace io {
 template <typename T>
 class reader {
  private:
-  Eigen::Map<const Eigen::Matrix<T, -1, 1>> data_r_;
-  Eigen::Map<const Eigen::Matrix<int, -1, 1>> data_i_;
+  Eigen::Map<Eigen::Matrix<T, -1, 1>> data_r_;
+  Eigen::Map<Eigen::Matrix<int, -1, 1>> data_i_;
   size_t pos_{0};
   size_t int_pos_{0};
 
-  inline const T &scalar_ptr() { return data_r_.coeffRef(pos_); }
+  inline T& scalar_ptr() { return data_r_.coeffRef(pos_); }
 
-  inline const T &scalar_ptr_increment(size_t m) {
+  inline T& scalar_ptr_increment(size_t m) {
     pos_ += m;
     return data_r_.coeffRef(pos_ - m);
   }
 
-  inline const int &int_ptr() { return data_i_.coeffRef(int_pos_); }
+  inline int& int_ptr() { return data_i_.coeffRef(int_pos_); }
 
-  inline const int &int_ptr_increment(size_t m) {
+  inline int& int_ptr_increment(size_t m) {
     int_pos_ += m;
     return data_i_.coeffRef(int_pos_ - m);
   }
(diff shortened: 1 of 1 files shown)


bbbales2:
Is this blocking a new feature? Or did we merge the `const` Maps and now something is failing upstream (in which case we need to add upstream tests)?

The easiest escape hatch is returning non-map types by copy for now (and make an issue to turn them back into maps when we have more time).

It seems like a footgun to not const these things.

SteveBronder:
> It seems like a footgun to not const these things.

Actually the opposite is true I think it's a footgun to make these things const. If they are vars then we will end up updating the inner vari adjoints

SteveBronder:
>>  If they are vars then we will end up updating the inner vari adjoints

> Yeah but we make that assumption for all the vars in Stan.

Yep! There's just no reason for it to be const here like in the rest of Stan. I can't think of another place in the library we have a const var (it doesn't really make sense because the adjoints will be updated).

> But it also looks like stuff is passed into the generated log_prob functions without a const, so I guess we aren't const correct here to begin with.

> But if we toss const-correctness cause it's not really a thing, then how do we motivate this still? This seems exactly like what we should not allow:


There's no const-correctness issues here because they are not created as const, they are never put through a `const_cast<>`, and when we propogate the adjoints the underlying pointers here will be manipulated so while the pointer itself is const the underlying object is not.

confused why this wouldn't be allowed? I have an example case below where this would actually be quite nice

>>  having these be const maps stops them from being assigned directly assigned to non-const maps in the model if we ever want to do that in the future

> I think we are sorta protected from this behavior in algorithms cause how log_prob is used in Stan now:

Assigning a map to a transformed parameter could be bad (since they can be assigned to), but parameters are never assigned to so we could actually have them be a map. We know that all signatures in stan math treat the incoming object as a const& so inputs are never manipulated. And when an `arena_t<>` get's an Eigen map it doesn't make a copy of the data and just takes the pointer to the passed in map. So having parameters as a map actually saves a good bit of memory allocations. Think of like `multiply(multiply(a, b), b)` where b is a 10,000 x 10,000 matrix of vars. Ballpark thats about 10GB just for the pointers of `b` in one pass. Then for the second mutiply we have to store that `b` again. But if `b` was a map then no copy would happen here

bbbales2:
> I can't think of another place in the library we have a const var

`const Eigen::Matrix<var, R, C>` treats the matrix object like a const so you can't assign to it:

Like here: https://godbolt.org/z/8sbPbn

So all our `require_eigen_t<T>` + `const T&` stuff fits in

> but parameters are never assigned to so we could actually have them be a map.

And they are const maps now with the reader changes. This is just making them not const maps.

Is this breaking something somewhere? Or stopping an implementation of something?

SteveBronder:
>> I can't think of another place in the library we have a const var

> const Eigen::Matrix<var, R, C> treats the matrix object like a const so you can't assign to it:

> Like here: https://godbolt.org/z/8sbPbn

> So all our require_eigen_t<T> + const T& stuff fits in

So that's a bit different, when we use const& in signatures we are saying that we want to treat the object passed in as if it was const. Check out [Where it Gets Messy - in Parameter Passing](http://duramecho.com/ComputerInformation/WhyHowCppConst.html) from the link. 

The other issue is copy ellision, if we return a const type assigning to a non-const type (which is [what happens](https://gist.github.com/SteveBronder/a3dbe2ea7c8a00c6275ec72c6829a17d#file-glmer-hpp-L641) in the generated stan code) the compiler can't elide this and has to make a hard copy since we are moving from const to non-const 

>>  but parameters are never assigned to so we could actually have them be a map.

> And they are const maps now with the reader changes. This is just making them not const maps.

Before the last change they were also non-const. They get passed from reader as a map but they actually get assigned in the stan model to an `Eigen::Matrix`. So right now a copy happens to a non-const `Eigen::Matrix`.

> Is this breaking something somewhere? Or stopping an implementation of something?

See my reply one above. I have a case up at the bottom where we can avoid allocating about 20GB of data

Another thing, right now for parameters what we do is

```
Eigen::Matrix<T__, -1, 1> b;
b = Eigen::Matrix<double, -1, 1>(Kc);
stan::math::initialize_fill(b, DUMMY_VAR__);

current_statement__ = 1;
b = in__.var_vector(Kc);
```

The `initialize_fill` is there so that transformed parameters that have subsets assigned to them dont have empty cells.  But parameters are never assigned to, so this could be changed to

```cpp
Eigen::Matrix<T__, -1, 1> b = in__.var_vector(Kc);
```

Now all we do is

```cpp
// or auto
Eigen::Map<Eigen::Matrix<T__, -1, 1>> b = in__.var_vector(Kc);
```

Then we can just pass parameters around as maps and since the underlying functions that use `arena_t<>` just see a map they won't actually allocate anything for the reverse pass.

bbbales2:
> which is what happens in the generated stan code

```cpp
Eigen::Matrix<T__, -1, 1> b;
b = Eigen::Matrix<double, -1, 1>(Kc);
stan::math::initialize_fill(b, DUMMY_VAR__);

current_statement__ = 1;
b = in__.var_vector(Kc);
```

Oh okay, that makes sense. And then the copy still happens because the thing on the left is a matrix even if we make this non-const.

If we change to:

```cpp
auto b = in__.var_vector(Kc);
```

instead of the fills and stuff, this solves it right? And everything can stay const and we aren't writing or copying things unnecessarily?

> Then we can just pass parameters around as maps and since the underlying functions that use arena_t<> just see a map they won't actually allocate anything for the reverse pass.

Yeah I see what you're saying with the `arena_t` thing. It makes sense that we try to take advantage of that at some point. This needs it's own pull and discussion though.

SteveBronder:
> If we change to:

> `auto b = in__.var_vector(Kc);`

> instead of the fills and stuff, this solves it right? And everything can stay const and we aren't writing or copying things unnecessarily?

No idt so, I'm pretty sure it would be the same as trying to update a non-const object through a const pointer which is not allowed.

https://godbolt.org/z/WEovoh

bbbales2:
Hmm sorry coming back to this after a while hope I'm picking up in the right place.

I think

```cpp
auto b = in__.var_vector(Kc);
```

should work with the current code.

`b` will be an `Eigen::Map<const Eigen::Matrix<T__, -1, 1>>` or something.

That is different than:

```cpp
Eigen::Map<const Eigen::Matrix<T__, -1, 1>> b(...);
b = in__.var_vector(Kc);
```

where I agree that is a bad assignment.

Here's a godbolt showing the first thing working: https://godbolt.org/z/shGfdq

bbbales2:
This pull is defunct now with the deserializer on the way?

SteveBronder:
Yeah let's close this

PR #2373, opened 2017-08-04
Fixes #2364 - add C++11 support by including it from math's makefiles.

seantalts:
#### Submisison Checklist

- [x] Run unit tests: `./runTests.py src/test/unit`
- [x] Declare copyright holder and open-source license: see below

#### Copyright and Licensing

Please list the copyright holder for the work you are submitting (this will be you or your assignee, such as a university or company): Columbia University

By submitting this pull request, the copyright holder is agreeing to license the submitted work under the following licenses:
- Code: BSD 3-clause (https://opensource.org/licenses/BSD-3-Clause)
- Documentation: CC-BY 4.0 (https://creativecommons.org/licenses/by/4.0/)

PR #3012, opened 2021-02-23
Revert "Feature/bounds no inf"

SteveBronder:
This reverts the infinite bounds checks PR as we rewrote them in Stan to support the bounds checking behavior for pos/neg infinity. This branch needs to do the dual merge thing to have this and the math branch [#2373](https://github.com/stan-dev/math/pull/2373) at the same time


bbbales2 (approved):
Looks good to me (so whenever the Math is ready this can go)

Issue #3005, opened 2021-02-10
[FR] add bool template parameter to *_constrains in


SteveBronder:
#### Summary:

In the Stan compiler stuff can be written like

```cpp
      local_scalar_t__ sd_1 = in__.scalar();
      current_statement__ = 5;
      if (jacobian__) {
        current_statement__ = 5;
        sd_1 = stan::math::lb_constrain(sd_1, 0, lp__);
      } else {
        current_statement__ = 5;
        sd_1 = stan::math::lb_constrain(sd_1, 0);
      }
```

But `jacobian__` is known at compile time so we can have a `reader.*_lb_constrain<>(...)` that takes in a bool template parameter so the above can just become 

```cpp
      local_scalar_t__ sd_1 = in__.scalar_lb_constrain<jacobian__>(0, lp__);
```

for matrices things like that will matter a good bit

#### Current Version:
v2.26.0

SteveBronder:
@rok-cesnovar I wrote this up over [here](https://github.com/stan-dev/stan/compare/feature/reader-constrain-template). But if you writing the `rep_array<>()` thing should these also have the same return type as the first tempalte parameter thing so you can reuse that scheme in the compiler? Haven't written tests for this yet so pretty easy for me to flip it around and do that

rok-cesnovar:
Yes, that would be great! I am working on adding rep matrix atm , for both matrix_cl and varmat.

PR #3013, opened 2021-02-26
Add deserializer class for reading constraints easier

SteveBronder:
#### Submission Checklist

#### Summary

This does the FR for #3005 by making a `deserializer` class that handles constraints for the types and bounds used in the Stan compiler. It's a bit of a monster but handles pretty much all the parameter sizes we have in Stan. The idea here is that instead of the compiler generating code for paremeters like 

```cpp
      local_scalar_t__ sd_1 = in__.scalar();
      current_statement__ = 5;
      if (jacobian__) {
        current_statement__ = 5;
        sd_1 = stan::math::lb_constrain(sd_1, 0, lp__);
      } else {
        current_statement__ = 5;
        sd_1 = stan::math::lb_constrain(sd_1, 0);
      }
```

It can instead just use

```cpp
      local_scalar_t__ sd_1 = in__.read_lb<local_scalar_t__, Jacobian__>(lp__, 0);
```

This works for eigen matrices and arrays containing arrays so something like the following is perfectly legal

```cpp
      using big_vec = std::vector<std::vector<var_value<Eigen::Matrix<double, -1, -1>>>>;
      big_vec sd_1 = in__.read_lb<big_vec, Jacobian__>(0, lp__, 3, 3, 3, 3);
```

For general reading the function is `read<RETURN_TYPE>(dims)` where `RETURN_TYPE` is the type requested back and dims are the dimensions of the type. All of the constrain functions have the form

```cpp
read_{constrain}<{return_type}, Jacobian>({constrain_args}, lp, {dims})
```

Where 
- `{constrain}` is the kind of the constraint like `lb,ub,lub` etc, 
- `{return_type}` is the requested return type, 
- `Jacobian` is whether to account for the Jacobian calculation
- {constrain_args} is the arguments passed to the actual constrain function
-  lp is the the log prob that will be incremented if `Jacobian` is true
-  {dims} are the dimensions of the return type

This also adds `read()` methods for complex scalars, vectors, and matrices.

#### Intended Effect

This should simplify some of the compiler implementation for things like the new matrix type while also making things more performant by utilizing vectorized versions of the constrain functions when they are available.

#### How to Verify

About 2/3rds of this PR are testing, where we run all the standard tests from `reader`, then also check `std::vector` return types and `var_value<Eigen>` returns for all matrix tests. You can run the tests with

```
./runTests.py ./src/test/unit/io/deserializer_test.cpp \
 ./src/test/unit/io/deserializer_varmat_test.cpp \
 ./src/test/unit/io/deserializer_stdvector_test.cpp
```

#### Side Effects

There's two things I still need to add here, tests for the complex stuff and tests for matrices with the `offset_and_multiplier` constraints. I may need to make a Stan math PR for offset multiplier 

#### Copyright and Licensing


SteveBronder:
@rok-cesnovar do you know what's up with the github check here? It's just giving back

```
Run echo "CXX=clang++-6.0" >> make/local
No matching tests found.
exit now (03/05/21 07:02:52 UTC)
```

bbbales2 (commented):
Partial review to get comments on read-and-check-constraints versions of functions

src/stan/io/deserializer.hpp
```diff
@@ -0,0 +1,1978 @@
...
+      []() STAN_COLD_PATH {
+        throw std::runtime_error("no more integers to read");
+      }();
+    }
+  }
+
+  template <typename S, typename K>
+  using conditional_var_val_t
+      = std::conditional_t<is_var_matrix<S>::value && is_var<T>::value,
+                           return_var_matrix_t<K, S, K>, K>;
```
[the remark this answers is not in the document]
rok-cesnovar: Seems similar to https://github.com/stan-dev/math/blob/develop/stan/math/rev/meta/conditional_var_value.hpp

src/stan/io/deserializer.hpp
```diff
@@ -0,0 +1,1978 @@
...
+  template <typename Ret, typename LB, typename... Sizes,
+            require_not_std_vector_t<Ret>* = nullptr>
+  auto read_lb(const LB& lb, Sizes... sizes) {
+    auto ret = this->read<Ret>(sizes...);
+    using stan::math::check_greater_or_equal;
+    using stan::math::value_of;
+    check_greater_or_equal("io deserializer", "Lower Bound", value_of(ret),
+                           value_of(lb));
+    return ret;
+  }
```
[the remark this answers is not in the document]
bbbales2: Oh I see things like this exist in the old reader: https://github.com/stan-dev/stan/blob/develop/src/stan/io/reader.hpp#L602

Do we need these any more? I don't think they're used as-in.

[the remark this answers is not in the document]
bbbales2: I looked quickly at a stanc2 generated .hpp file and I don't think these are used there as well.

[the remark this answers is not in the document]
SteveBronder: So right now the stanc3 compiler doesn't use these but it will when it uses this function. Right now it just calls the constrain check directly as a loop, but by doing it here we can use the vectorized stuff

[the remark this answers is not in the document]
bbbales2: So is this like transform inits? Or write array or something? I'm just not sure where the constrained parameters exist as a flattened vector.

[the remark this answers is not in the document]
bbbales2: (Like, the constrained parameters exist as a flattened vector, to be read out into multiple variables)

rok-cesnovar:
> @rok-cesnovar do you know what's up with the github check here? It's just giving back

yeah, this is unrelated. Happens on develop too. Will looki into it.

src/stan/io/deserializer.hpp
```diff
@@ -0,0 +1,1978 @@
...
+   * @throw std::runtime_error if x is not positive
+   */
+  template <typename Ret, typename... Sizes>
+  auto read_pos(Sizes... sizes) {
+    auto ret = read<Ret>(sizes...);
+    using stan::math::check_positive;
+    using stan::math::value_of;
+    check_positive("deserializer", "Positive Constrained", value_of(ret));
+    return ret;
+  }
```
bbbales2, line 729: I don't see an explicit positive constrained variable in the Stan reference manual. Maybe this is a leftover type from earlier that we don't need to re-implement?

src/stan/io/deserializer.hpp
```diff
@@ -0,0 +1,1978 @@
...
+   */
+  template <typename Ret, typename... Sizes>
+  auto read_prob(Sizes... sizes) {
+    auto ret = read<Ret>(sizes...);
+    using stan::math::check_bounded;
+    using stan::math::value_of;
+    check_bounded<decltype(value_of(ret)), double, double>(
+        "deserializer", "Constrained probability", value_of(ret), 0, 1);
+    return ret;
+  }
```
bbbales2, line 1000: I think this constraint also isn't used at the language level.

src/stan/io/deserializer.hpp
```diff
@@ -0,0 +1,1978 @@
...
+            require_not_matrix_t<Ret>* = nullptr,
+            require_not_std_vector_t<Ret>* = nullptr>
+  auto read_corr(Sizes... sizes) {
+    auto ret = read<Ret>(sizes...);
+    using stan::math::check_bounded;
+    using stan::math::value_of;
+    check_bounded<decltype(value_of(ret)), double, double>(
+        "deserializer", "Correlation value", value_of(ret), -1, 1);
+    return ret;
+  }
```
bbbales2, line 1078: I think the corr constrain only works for matrices. I don't think there is a vector corr in the language, so this can be removed.

src/stan/io/deserializer.hpp
```diff
@@ -0,0 +1,1978 @@
...
+            require_not_std_vector_t<Ret>* = nullptr,
+            require_not_matrix_t<Ret>* = nullptr>
+  auto read_corr(LP& lp, Sizes... sizes) {
+    using stan::math::corr_constrain;
+    if (Jacobian) {
+      return corr_constrain(this->read<Ret>(sizes...), lp);
+    } else {
+      return corr_constrain(this->read<Ret>(sizes...));
+    }
+  }
```
bbbales2, line 1144: Similarly this can be removed if there is no corr constraint for vectors in the language (and I don't think there is)

src/stan/io/deserializer.hpp
```diff
@@ -0,0 +1,1978 @@
...
+    if (unlikely(k == 0)) {
+      []() STAN_COLD_PATH {
+        std::string msg = "deserializer: unit vectors cannot be size 0.";
+        throw std::invalid_argument(msg);
+      }();
+    }
+    auto ret = read<Ret>(k);
+    using stan::math::check_unit_vector;
+    using stan::math::value_of;
+    check_unit_vector("deserializer", "Unit Vector", value_of(ret));
```
bbbales2, line 1228: `check_unit_vector` has a non-zero size check in it ([here](https://github.com/stan-dev/math/blob/develop/stan/math/prim/err/check_unit_vector.hpp)). Since `read<Ret>(0)` is okay, we don't need the check before the read.

src/stan/io/deserializer.hpp
```diff
@@ -0,0 +1,1978 @@
...
+    if (unlikely(k == 0)) {
+      []() STAN_COLD_PATH {
+        std::string msg = "deserializer: simplex vectors cannot be size 0.";
+        throw std::invalid_argument(msg);
+      }();
+    }
+    auto ret = read<Ret>(k);
+    using stan::math::check_simplex;
+    using stan::math::value_of;
+    check_simplex("deserializer", "Simplex", value_of(ret));
```
bbbales2, line 1340: `check_simplex` checks for zero sizes ([here](https://github.com/stan-dev/math/blob/develop/stan/math/prim/err/check_simplex.hpp)) so the check above the read is not necessary

bbbales2 (changes requested):
Here's the rest of the review on the code -- after the code gets sorted out I'll do the tests.

src/stan/io/deserializer.hpp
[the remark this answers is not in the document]
bbbales2: It may be be easier to point to the old code that we're replacing than explaining it with words. I made a gist [here](https://gist.github.com/bbbales2/0053578d33035ab25560dd5e2b5a6fb1) with generated code. Which part needs the read + check constrains? It is for this model:

```stan
data {
  vector<lower = 0.0>[3] x_d;
}

transformed data {
  vector<lower = 0.0>[3] x_tf = x_d;
}

parameters {
  vector<lower = 0.0>[3] x_p;
}

transformed parameters {
  vector<lower = 0.0>[3] x_tp = x_p;
}

generated quantities {
  vector<lower = 0.0>[3] x_gq = x_p;
}
```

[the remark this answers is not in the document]
bbbales2: @SteveBronder you hear a voice from your computer

"Steve", it says, "answer meeeee"

It is your old trusty friend, stan-dev/stan #3013, discussion r588771274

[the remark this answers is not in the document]
SteveBronder: lol apologies, yeah actually I fleshed your example out a bit more and I think your right we don't need these anymore?

```stan
// deserializer not used
/*
data {
  real<lower=0> x_scal;
  vector<lower = 0.0>[3] x_v;
  matrix<lower = 0.0>[3, 3] x_m;
  matrix<lower = 0.0>[3, 3] x_vm[3];
}

transformed data {
  real<lower=0> x_tscal = x_scal;
  vector<lower = 0.0>[3] x_tv = x_v;
  matrix<lower = 0.0>[3, 3] x_tm = x_m;
  matrix<lower = 0.0>[3, 3] x_tvm[3] = x_vm;
}
*/
parameters {
  real<lower=0> x_param_scal;
  vector<lower = 0.0>[3] x_param_v;
  matrix<lower = 0.0>[3, 3] x_param_m;
  matrix<lower = 0.0>[3, 3] x_param_vm[3];
  cholesky_factor_corr[3] x_param_corr;
  cholesky_factor_corr[3] x_param_vcorr[3];
  matrix<offset = x_param_scal>[3, 3] x_opvm[3];
}

transformed parameters {
  real<lower=0> x_t_param_scal = x_param_scal;
  vector<lower = 0.0>[3] x_t_param_v = x_param_v;
  matrix<lower = 0.0>[3, 3] x_t_param_m = x_param_m;
  matrix<lower = 0.0>[3, 3] x_t_param_vm[3] = x_param_vm;
  cholesky_factor_corr[3] x_t_param_corr = x_param_corr;
  cholesky_factor_corr[3] x_t_param_vcorr[3] = x_param_vcorr;
  matrix<offset = x_param_scal>[3, 3] x_topvm[3];
}

generated quantities {
  vector<lower = 0.0>[3] x_gq = x_t_param_v;
}
```

Looking at the [gist](https://gist.github.com/SteveBronder/3e326e14db5b93586a40f0ba37b17d32) yeah I think your 100% right that we can get rid of these.

Tagging @rybern just to sanity check, is there anywhere that we currently call `reader` (which I think is just `log_prob` and `write_array`) where we need to check that an object satisfies a given constrain but doesn't perform the transform? idt so and that let's me cut out a ton of code here

src/stan/io/deserializer.hpp
```diff
@@ -0,0 +1,1978 @@
...
+    }
+  }
+
+  /**
+   * Return an Eigen row vector of size `m`.
+   * @tparam Size an integral type
+   * @tparam Ret The type to return.
+   * @param m Size of row vector.
+   */
+  template <typename Ret, typename Size,
```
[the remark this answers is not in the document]
SteveBronder: I think this was just an over-optimization on my part

src/stan/io/deserializer.hpp
```diff
@@ -0,0 +1,1978 @@
...
+            require_eigen_col_vector_t<Ret>* = nullptr,
+            require_vt_complex<Ret>* = nullptr>
+  auto read(Size m) {
+    if (unlikely(m == 0)) {
+      return Ret(map_vector_t(nullptr, m));
+    } else {
+      Ret ret(m);
+      for (Eigen::Index i = 0; i < m; ++i) {
+        ret.coeffRef(i) = std::complex<T>{data_r_.coeffRef(pos_r_++),
+                                          data_r_.coeffRef(pos_r_++)};
```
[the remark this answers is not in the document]
SteveBronder: Yes that's a good idea

src/stan/io/deserializer.hpp
```diff
@@ -0,0 +1,1978 @@
...
+   * @tparam Size an integral type
+   * @tparam Ret The type to return.
+   * @param m Size of column vector.
+   */
+  template <typename Ret, typename Size,
+            require_eigen_col_vector_t<Ret>* = nullptr,
+            require_not_vt_complex<Ret>* = nullptr>
+  auto read(Size m) {
+    if (unlikely(m == 0)) {
+      return map_vector_t(nullptr, m);
```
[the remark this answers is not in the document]
SteveBronder: Yeah this is how it was done before.

Kind of a side note to this, @bob-carpenter is there a math reason we can't have size zero simplex vectors but can have size zero vectors?

src/stan/io/deserializer.hpp
```diff
@@ -0,0 +1,1978 @@
...
+   */
+  inline T& scalar_ptr_increment(size_t m) {
+    pos_r_ += m;
+    return data_r_.coeffRef(pos_r_ - m);
+  }
+
+  /**
+   * Return pointer to current integer.
+   */
+  inline int& int_ptr() { return data_i_.coeffRef(pos_i_); }
```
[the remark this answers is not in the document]
SteveBronder: I wonder if I can just get rid of all the int things or whether we should keep them? @bob-carpenter do you know the reason these member functions existed in `reader`? Was it for discrete parameters or something?

src/stan/io/deserializer.hpp
```diff
@@ -0,0 +1,1978 @@
...
+ * must support the transforming operations, such as <code>exp(x)</code>
+ * for positive-bounded variables.  It must also support equality and
+ * inequality tests with <code>double</code> values.
+ *
+ * @tparam T Basic scalar type.
+ */
+template <typename T>
+class deserializer {
+ private:
+  Eigen::Map<Eigen::Matrix<T, -1, 1>> data_r_;    // map of reals.
```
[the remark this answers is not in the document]
SteveBronder: How about just `map_r_`?

src/stan/io/deserializer.hpp
```diff
@@ -0,0 +1,1978 @@
...
+  Eigen::Map<Eigen::Matrix<int, -1, 1>> data_i_;  // map of integers.
```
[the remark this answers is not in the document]
SteveBronder: idk, I do want to remove the int stuff but tbh I'm not sure if someone somewhere uses them or whether they still serve a purpose so I think I'd rather just leave them for now. Once we know why they exist we can talk about removing them.

src/stan/io/deserializer.hpp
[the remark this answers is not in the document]
SteveBronder: Also @rybern just wanted to sanity check, the deserializer thing here just takes code like

```cpp
      std::vector<Eigen::Matrix<local_scalar_t__, -1, -1>> x_opvm;
      x_opvm = std::vector<Eigen::Matrix<local_scalar_t__, -1, -1>>(
          3, Eigen::Matrix<local_scalar_t__, -1, -1>(3, 3));
      stan::math::fill(x_opvm, DUMMY_VAR__);

      current_statement__ = 7;
      for (int sym1__ = 1; sym1__ <= 3; ++sym1__) {
        current_statement__ = 7;
        assign(x_opvm, cons_list(index_uni(sym1__), nil_index_list()),
               in__.matrix(3, 3), "assigning variable x_opvm");
      }
      current_statement__ = 7;
      for (int sym1__ = 1; sym1__ <= 3; ++sym1__) {
        current_statement__ = 7;
        for (int sym2__ = 1; sym2__ <= 3; ++sym2__) {
          current_statement__ = 7;
          for (int sym3__ = 1; sym3__ <= 3; ++sym3__) {
            current_statement__ = 7;
            if (jacobian__) {
              current_statement__ = 7;
              assign(
                  x_opvm,
                  cons_list(index_uni(sym1__),
                            cons_list(index_uni(sym2__),
                                      cons_list(index_uni(sym3__),
                                                nil_index_list()))),
                  stan::math::offset_multiplier_constrain(
                      rvalue(x_opvm,
                             cons_list(index_uni(sym1__),
                                       cons_list(index_uni(sym2__),
                                                 cons_list(index_uni(sym3__),
                                                           nil_index_list()))),
                             "x_opvm"),
                      x_param_scal, 1, lp__),
                  "assigning variable x_opvm");
            } else {
              current_statement__ = 7;
              assign(
                  x_opvm,
                  cons_list(index_uni(sym1__),
                            cons_list(index_uni(sym2__),
                                      cons_list(index_uni(sym3__),
                                                nil_index_list()))),
                  stan::math::offset_multiplier_constrain(
                      rvalue(x_opvm,
                             cons_list(index_uni(sym1__),
                                       cons_list(index_uni(sym2__),
                                                 cons_list(index_uni(sym3__),
                                                           nil_index_list()))),
                             "x_opvm"),
                      x_param_scal, 1),
                  "assigning variable x_opvm");
            }
          }
        }
      }
```

and let's the compiler generate the following instead

```cpp
// just using an alias for readability
using std_vec_mat = std::vector<Eigen::Matrix<double, -1, -1>>;
std_vec_mat x_opvm = in__.read_offset_multiplier<std_vec_mat, Jacobian>(x_param_scal, 1, lp__, 3, 3, 3);
```

Is that cool for what you are looking to do as well? For tuples you would call

```cpp
using a_tuple_t = std::tuple<local_scalar_t__, std_vec_mat>;
 a_tuple_t foo = std::make_tuple(in__.read_lb<local_scalar_t, Jacobian>(lb_val, lp),
   in__.read_offset_multiplier<std_vec_mat, Jacobian>(offset, mult, lp, 3, 3, 3))
```

src/stan/io/deserializer.hpp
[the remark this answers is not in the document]
SteveBronder: Actually I don't like how this looks sometimes so I'd rather just leave it as is

src/stan/io/deserializer.hpp
SteveBronder: Oh? I mean maybe this is an easy type to add to the language? It seems reasonable and we have all the stuff in math for it

src/stan/io/deserializer.hpp
SteveBronder: Yeah I also couldn't find these in the stan manual? But I do know how to add these the stanc3 so if we wanted them idt it would be that bad to add

src/stan/io/deserializer.hpp
SteveBronder: Is there a corr scalar type even?

SteveBronder: Also this one is for just scalars since the requires on it stop return types of std vectors or matrics/vectors

SteveBronder:
Removing those check read functions got rid of like 2000+ likes of code. Still v big but more managable. Also found that I need to add unit vector and simplex constrain/lp tests and tests for cholesky corr constrain/lp

src/stan/io/deserializer.hpp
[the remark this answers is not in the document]
rybern: Yup should work!

src/stan/io/deserializer.hpp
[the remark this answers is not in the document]
bbbales2: I think it's that simplex vectors have the sum to 1 constraint, and a length 0 thing doesn't sum to one.

src/stan/io/deserializer.hpp
bbbales2: Yeah we can definitely clean this one up too. There is no scalar corr. I imagine maybe this was once used for partial correlations or something (constrained [-1, 1]) but upper/lower handle that now.

src/stan/io/deserializer.hpp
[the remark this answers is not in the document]
bbbales2: Cool beans

src/stan/io/deserializer.hpp
bbbales2: I'm gonna remove this. It's just easy enough to say in the language:

```
real<lower = 0.0> x;
```

src/stan/io/deserializer.hpp
bbbales2: Removing this also cause it's just `real<lower = 0.0, upper = 1.0> p;`

bbbales2:
Working through the unit tests now

src/stan/io/deserializer.hpp
SteveBronder: Lol oh yeah true true

src/stan/io/deserializer.hpp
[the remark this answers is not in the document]
rybern: @SteveBronder I actually missed your first question - 95% sure there's nowhere we do reads and not constraints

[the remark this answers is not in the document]
bbbales2: If it turns out we made a mistake cutting the checks then we can add them back in another pull without too much trouble. Code exists here in the history weeee

src/stan/io/deserializer.hpp
```diff
@@ -486,237 +436,6 @@ class deserializer {
...
-   * @param vecsize The size of the return vector.
-   * @param sizes Pack of integrals to use to construct the return's type.
-   * bounds.
-   */
-  template <typename Ret, bool Jacobian, typename Offset, typename Mult,
-            typename LP, typename... Sizes,
-            require_std_vector_t<Ret>* = nullptr>
-  inline auto read_offset_multiplier(const Offset& offset,
-                                     const Mult& multiplier, LP& lp,
-                                     const size_t vecsize, Sizes... sizes) {
```
SteveBronder: Why did this one get deleted?

bbbales2: Math handles all the `std::vector` things. Same as with lb/ub/lub constrain

SteveBronder: Oh duh because we have them in the math library

src/test/unit/io/deserializer_test.cpp
```diff
@@ -0,0 +1,899 @@
...
+
+  // should go back to working
+  EXPECT_NO_THROW(deserializer.read<int>());
+  EXPECT_THROW(deserializer.read<int>(), std::runtime_error);
+
+  // should keep throwing
+  EXPECT_THROW(deserializer.read<double>(), std::runtime_error);
+  EXPECT_THROW(deserializer.read<size_t>(), std::runtime_error);
+
+  EXPECT_THROW(deserializer.read<std::complex<double>>(), std::runtime_error);
```
bbbales2, line 325: @SteveBronder these were the only problems I found working through this. We throw an error if we read a real and don't have memory for it, so I think we should do similarly for the vectors and stuff. Can you add these in?

src/test/unit/io/deserializer_test.cpp
```diff
@@ -0,0 +1,899 @@
...
+  std::vector<int> theta_i;
+  std::vector<double> theta;
+  theta.push_back(3.0);
+  theta.push_back(-1.0);
+  theta.push_back(-2.0);
+  theta.push_back(0.0);
+  stan::io::deserializer<double> deserializer(theta, theta_i);
+  double lp = 0;
+  Eigen::VectorXd reference
+      = stan::math::unit_vector_constrain(stan::math::to_vector(theta));
```
bbbales2, line 539: I changed a lot of the tests to just check against Math library functions. Like at this point we're just testing the deserializer interface and we're assuming these functions are good, so I think it's fair. Makes it a lot easier to code and check jacobians and stuff.

src/test/unit/io/deserializer_test.cpp
```diff
@@ -0,0 +1,899 @@
+#include <stan/io/deserializer.hpp>
+// expect_near_rel comes from lib/stan_math
+#include <test/unit/math/expect_near_rel.hpp>
+#include <gtest/gtest.h>
+
+TEST(deserializer_scalar, read) {
```
bbbales2, line 6: I think the only major tests missing are if we took this file and replaced all the `double` s with `stan::math::var`. That will be the main way this code is used so probably worth it.

src/test/unit/io/deserializer_test.cpp
```diff
@@ -0,0 +1,899 @@
...
+          Eigen::Matrix<std::complex<double>, Eigen::Dynamic, Eigen::Dynamic>>(
+          2, 2)),
+      std::runtime_error);
+  EXPECT_THROW(deserializer.read<std::vector<Eigen::MatrixXd>>(2, 3, 2),
+               std::runtime_error);
+}
+
+// lb
+
+TEST(deserializer_scalar, read_lb_constrain) {
```
bbbales2, line 353: For lb/ub/lub/multiplier_offset I went pretty sparse with the tests. I didn't check that all combinations of vector/`std::vector` etc. etc. worked. I'm assuming these functions work in Math and we're just trying to test the deserializer interface to them.

bbbales2 (changes requested):
I messed with the tests a ton. I think there should be some checks to make sure that there is enough memory left when deserializing non-scalar types (there's a comment detailing this). Otherwise this looks great.

src/test/unit/io/deserializer_test.cpp
bbbales2: I'm gonna do this real quick actually.

src/stan/io/deserializer.hpp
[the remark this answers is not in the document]
bbbales2: @SteveBronder Oh yeah also this -- similar names but different function. We should rename this to just something kinda different so they don't get confused.

src/test/unit/io/deserializer_test.cpp
bbbales2: I added them in `deserializer_var_test.cpp`. Now that I copy-pasted them there I'm not sure how much value they add. Maybe just needed a couple `read` tests to make sure that things compiled rather than the full gamut of read_lb, read_ub, read_corr_matrix, etc.

src/test/unit/io/deserializer_test.cpp
bbbales2: I added checks for this using `check_r_capacity`.

bbbales2:
Tests are waiting on https://github.com/stan-dev/stanc3/pull/845

bbbales2:
Aaah, k. I'm pretty happy with the code here. Take a look through it and if it's all good we can get it merged in tomorrow hopefully.

I'm gonna get started tidying the varmat compatibility test script (so it's easy to generate a list of compatible functions and it's not a script hidden on my computer).

SteveBronder:
Sounds good!

Do you think it's better to make the script or to go through and do the doc thing I mentioned in the math issue? I'm fine either way

bbbales2:
I think both. Docs will be good just to have them.

The script is testing that the compiling works so it should be more thorough and automatic -- similar to expressions testing.

bbbales2:
Script is mostly written -- so hopefully won't be long turnaround time on this. There's also a temporary list over here: https://github.com/stan-dev/cmdstan/issues/980#issuecomment-777478394 so that any compiler-writing can start based on that.

SteveBronder:
Pinging @rybern FYI this should get merged tmrw so the compiler impl can start using it

SteveBronder:
@bbbales2 good to go?

rok-cesnovar:
Can you push a commit here to bump jenkins?

bbbales2:
@SteveBronder yeah I'm happy with what is here. You good with it? Jump pushed a commit to get Jenkins to go again.

bbbales2 (approved):

SteveBronder:
Yep!

rybern:
Congrats!!

Would a matching interface for the _free variables make sense? In stanc3, we generate the _constrain and _free calls in the same way, so this PR will let me simplify the _constrain calls, but with a new _free interface I can get rid of the old code altogether. (Same thing with checks to a lesser extent)

SteveBronder:
Ty! Yes so `transform_inits()` I've been thinking about `transform_inits()` for the free functions and one good idea and one okay idea

So the `var_context` in the signature here is a bit of trouble because that class is an abstract base class (it only has undefined virtual functions with no other members). So the actual class going into that function is one of it's children like `stan::io::array_context` etc. who does have those functions defined. One rule with virtual methods is that they can't have templates so we can't do the exact same trick we did with `deserializer`.

```cpp
  inline void transform_inits(
      const stan::io::var_context& context,
      Eigen::Matrix<double, Eigen::Dynamic, 1>& params_r,
      std::ostream* pstream = nullptr) const final {
//...
}
```

But in Stan, `transform_inits()` is only used in [`initialize.hpp`](https://github.com/stan-dev/stan/blob/develop/src/stan/services/util/initialize.hpp#L101) and [`standalone_gqs`](https://github.com/stan-dev/stan/blob/develop/src/stan/services/sample/standalone_gqs.hpp#L113). 

Looking at both of those functions that use `transform_inits()` we know that they are trying to supply the model with the parameters names, *an input array*, the parameter dimesions, and an *output array*. But we know the model already knows the 
dimensions and who cares about names!

So I think we could just use a serializer and deserializer!

What we could do is add a `transform_inits()` functions with the signature

```cpp
transform_inits(Eigen::VectorXd& input_r, Eigen::VectorXd& output_r, std::ostream* pstream__ = nullptr) {
    stan::io::deserializer deserializer(input_r);
    // Made stuff calling the read<>() functions.
    Eigen::MatrixXd X = deserializer.read<Eigen::MatrixXd>(N, M);
    // Write to serializer
    stan::io::serializer serializer(output_r);
    serializer.serialize_free_lb(X, lb);
```

Thinking about it more could we just do the free's in deserializer and write

```cpp
transform_inits(Eigen::VectorXd& input_r, Eigen::VectorXd& output_r, std::ostream* pstream__ = nullptr) {
    stan::io::deserializer deserializer(input_r);
    stan::io::serializer serializer(output_r);
    // Make stuff calling the read_free_lb<>() functions.
    serializer.serialize(deserializer.read_free_lb<Eigen::MatrixXd>(lb, N, M));
```

@bbbales2 do you have any opinions on whether the free functions should be in the serializer or deserializer? I think it would be nice to keep the serializer simple and just put all this in the serializer

I have to double check that this works, but if so does that work for @rybern? My only concern is that we have to worry about parameter order, though last time I looked at the contexts it just takes the vector you give it and uses that data to create sub arrays of those dimensions internally so I think this would be fine. Tagging @betanalpha as I know he has had opinions on this in the past

SteveBronder:
> The things missing on this are like the array forms of simplex_free, corr_matrix_free, etc. Maybe we shoulda just put all the array stuff in Math. None of that would be hard to write or test. The array forms for the functions we haven't written are straightforward (ub/lb/lub/offset_multiplier were the hard ones).

Yeah we could either do these in the deserializer (since it's just calling `simplex_free` on each submatrix in the array) or do them in math. I'm fine with either way

> I'm hoping that "add" is actually a "replace"? Cause I assume we'd be deleting the old-style transform_inits.

Yeah so I was thinking of a triage'd style thing where we

1. Add the things for deserializer/serializer along with the virtual functions for the new `transform_inits()` in `model_base/model_base_crtp` (Since those virtual functions are not called yet it's fine if they are not defined in the model)
2. Ryan writes the stuff to do the deserializer/serializer form of transform inits in the model class
3. We change the stan code to use the new pattern in initialize and standalone generated quantities hpp files

Unless we want to do one big PR in Stan that does (1) and (3), then merge them at the same time as Ryan updates the stanc3 code

> Touching the var_contexts scares me a bit cause I don't know what downstream code exists that would need to be updated.

> At worst we could just serialize to a local variable and then push those values onto the var_context. This part of the code shouldn't be performance sensitive.

Yeah I agree, like looking around it seems like this is fine but we should probs reach out to folks we wrote this stuff to confirm what we are talking about here is going to work smoothly

bbbales2:
Hmm well yeah since you mention it all the array free stuff can just go here in a `serializer` class and we don't need to mess with Math. That's one pull instead of two.

Since you have a picture of serializer in your head, you want to go ahead and kick that off? I remain inclined to ditch all things integer here, fwiw.

SteveBronder:
Yeah sure I'm trying to put together some docs for the compiler today and I'm going to give up on that in like 20 mins or so and I'll ge started on this

SteveBronder:
It won't let me click the merge button but Ben feel free to merge. @rok-cesnovar do I need some privileges I don't have? Tagging @syclik as well

rok-cesnovar:
You have write privileges in this repo so this is definitely weird.

rybern:
Sorry for the delay!

 @SteveBronder I think what you described works great for my purposes (I probably haven't grokked the whole story though)

@bbbales2 
> The upside here would be the cleanup right?

I do think it'll simplify of the hairier bits of OCaml, because the logic would be easier and match the new constraints, and it'll generalize better for tuples. Also, maybe we won't have to deal with \_free__ variables across unrelated parts of the compiler.

It might have a little performance gain too, right?

Just to make things concrete for my benefit, here's @nhuurre's example:
```
parameters {
  array[10] simplex[3] param;
}
```
For constraints we can now generate:
```
param = in__.read_simplex<T, jacobian__>(lp__, 10, 3);
```
For `transform_inits`, we're currently generating this big boy:
```
      std::vector<Eigen::Matrix<double, -1, 1>> param_free__;
      param_free__ = std::vector<Eigen::Matrix<double, -1, 1>>(10, Eigen::Matrix<double, -1, 1>(
        (3 - 1)));
      stan::math::fill(param_free__, std::numeric_limits<double>::quiet_NaN());
      
      current_statement__ = 1;
      for (int sym1__ = 1; sym1__ <= 10; ++sym1__) {
        current_statement__ = 1;
        assign(param_free__, cons_list(index_uni(sym1__), nil_index_list()),
          stan::math::simplex_free(param[(sym1__ - 1)]),
          "assigning variable param_free__");}
      for (int sym1__ = 1; sym1__ <= 10; ++sym1__) {
        for (int sym2__ = 1; sym2__ <= (3 - 1); ++sym2__) {
          vars__.emplace_back(param_free__[(sym1__ - 1)][(sym2__ - 1)]);}}
```
I'm imagining that it's possible to do something more direct along the lines of `vars__.emplace_back<..>(.., param, 10, 3)` and that that might be more efficient.

bbbales2:
[This](https://github.com/stan-dev/stan/pull/3018) has the code to do something like:

```
std::vector<Eigen::VectorXd> param_free__ = deserializer.simplex_free(param);
```

I think Steve's working on a replacement for the `vars__.emplace_back` bit.

rybern:
Awesome!

SteveBronder:
Yeah Ryan you may actually be able to write something like

```cpp
stan::io::deserializer deserializer(input_vec__)
stan::io::serializer serializer(vars__);
serialize.write(deserializer.free_lb<std::vector<std::vector<Matrix>>>(lb, lp, {dims...})
```

Is it just those free functions called in transform inits or does something else happen?

rybern:
@SteveBronder sorry Steve I missed that you asked a question here

> Is it just those free functions called in transform inits or does something else happen?

I think in transform_inits it's just doing the free stuff, @seantalts could you confirm?

PR #3018, opened 2021-03-11
Added prototype free functions to deserializer

bbbales2:
#### Submission Checklist

#### Summary

I added an example of how unit vector free would work

#### Copyright and Licensing


bbbales2:
@SteveBronder I think these should be copy-pasted into the serializer class. They aren't making use of anything in deserializer.

Otherwise I think all the frees are here and the new bits are tested.

src/stan/io/deserializer.hpp
```diff
@@ -944,6 +944,301 @@ class deserializer {
...
+   * @tparam S Type of bounded input
+   * @tparam L Type of lower bound
+   * @tparam U Type of upper bound
+   * @param y Upper bounded input
+   * @param lb Lower bound
+   * @param ub Upper bound
+   * @return Unconstrained variable
+   */
+  template <typename S, typename L, typename U>
+  inline auto free_lub(const S& y, const L& lb, const U& ub) {
```
SteveBronder, line 988: [optional] @rybern is it okay that the name pattern here is

```
free_{constrain_name}(object, {constrain_params});
```

Or would it be easier for you if we had the params come first like in the constrain functions?

Also @bbbales2 should we name these `read_free_{constrain_name}` and change the constrain functions to `read_constrain_{constrain_name}`? That would just help keep the API more consistent

src/test/unit/io/deserializer_free_test.cpp
```diff
@@ -0,0 +1,459 @@
...
+
+  std::vector<int> theta_i;
+  std::vector<double> theta;
+  stan::io::deserializer<double> deserializer(theta, theta_i);
+
+  std::vector<Eigen::VectorXd> std_vec = {vec1, vec2},
+                               std_vec_bad = {vec1, vec_bad};
+
+  std::vector<std::vector<Eigen::VectorXd>> std_std_vec
+      = {{vec1, vec2}, {vec2, vec1}};
```
SteveBronder, line 62: Would have saved us a lot of code if I had remembered that std::vector versions just call the underlying and could have done them all in one go. Very nice!

SteveBronder (approved):
Looks good! Quick q for @rybern on the API pattern here and we should be good to go.

src/stan/io/deserializer.hpp
rybern: That shouldn't make it any easier or harder for me, up to you.

This looks awesome!

SteveBronder (changes requested):
Actually Ben I goofed here approving this. The signature should match the read functions like 

```cpp
template <typename Ret, require_not_std_vector_t<Ret>* = nullptr>
  inline auto free_ordered(dims...) {
```

Our goal is to replace the code here


https://gist.github.com/SteveBronder/283a8cc24787383232d26590e207ba0c#file-original_model-hpp-L1076


```cpp
      double x_param_scal;
      x_param_scal = std::numeric_limits<double>::quiet_NaN();
      x_param_scal = context__.vals_r("x_param_scal")[(1 - 1)];
      double x_param_scal_free__;
      x_param_scal_free__ = std::numeric_limits<double>::quiet_NaN();
      x_param_scal_free__ = stan::math::lb_free(x_param_scal, 0);
```

With code like

```cpp
double x_param_scal_free__ = deserializer.read_free_lb<double>(0);
```

bbbales2:
@SteveBronder it was way easier to write tests with the serializer in place, so this will wait on that pull to go in.

There are a couple more changes I still need to add but will get them tomorrow or something:
1. Write a unit test for the changes [here](https://github.com/stan-dev/stan/pull/3018/files#diff-b0565fd88eecb37a8e58aecee30e057366d11107a61cb19f2eba15fb3e0e46efR542) (I may have broken other tests or something I'm not sure)
2. Rename read_lb and stuff to read_constrain_lb etc.

SteveBronder:
Aight the serializer pr should be good to go

bbbales2:
@SteveBronder This should be good to go once #3019 is in

bbbales2:
@serban-nicusor-toptal Looks like Jenkins is throwing out of space errors:

```
make/tests:93: recipe for target 'src/stan/io/dump.hpp-test' failed
make: *** [src/stan/io/dump.hpp-test] Error 1
make: *** Waiting for unfinished jobs....
test/dummy.cpp:1:22: fatal error: error writing to /tmp/ccAA3ozy.s: No space left on device
    1 | int main() {return 0;}
```

SteveBronder:
@rybern fyi we are going with the signatures `read()` for basic reads, `read_constrain_*` for reading and constraining variables, and `read_free_*` for reading and free'ing variables

bbbales2:
@SteveBronder this is green

SteveBronder:
Let me give it one quick read through though should be good

rybern:
@SteveBronder excellent

src/stan/io/deserializer.hpp
```diff
@@ -935,12 +946,335 @@ class deserializer {
...
+  template <typename Ret, typename... Sizes,
+            require_std_vector_t<Ret>* = nullptr>
+  inline auto read_free_simplex(size_t vecsize, Sizes... sizes) {
+    std::decay_t<Ret> ret;
+    ret.reserve(vecsize);
+    for (size_t i = 0; i < vecsize; ++i) {
+      ret.emplace_back(read_free_simplex<value_type_t<Ret>>(sizes...));
+    }
+    return ret;
+  }
```
SteveBronder, line 1087: Does this need a

```cpp
  template <typename Ret, require_not_std_vector_t<Ret>* = nullptr>
  inline auto read_free_simplex(size_t size) {
```

Or nah because you can't have a std::vector simplex? I'm pretty sure now we don't need it

SteveBronder (approved):
Cool and good!

SteveBronder:
It won't let me click the button for some reason but this is good to go!

goodcoolgood

src/stan/io/deserializer.hpp
bbbales2: I added it anyway for symmetry with everything else lol. Pretty sure we don't need it too cause the number of size arguments will define which thing gets used.

PR #3028, opened 2021-03-21
Parallel Run Adaptive Sampler

SteveBronder:
#### Submission Checklist

#### Summary

This is a WIP to figure out some of the odds and ends of the parallel version of adaptive sampling. The main issues 

1. The `sample_writers` and `diagnostic_writers` come in as a vector of streams where the current impl comes in as a writer for each stream. Is that fine? At the cmdstan level I suppose we can just check that if `n_chain > 1` we can spin up writers to `output_{chain_#}`?

2. wrt RNGs, do we need to have a special PRNG or is it safe to just have multiple RNGs? Should we use a [different RNG](https://www.boost.org/doc/libs/1_66_0/doc/html/boost_random/reference.html#boost_random.reference.generators) with a longer cycle and set each seed to user given initial + `2^chain_id` or something?

3. For what @wds15 brought up [here](https://github.com/stan-dev/cmdstan/pull/987#issuecomment-803323921) should the `isolate()` thing happen before we put this PR in or after? I think as long as it's not exposed to users yet it's fine to work on this as long as we do that before we make this available via cmdstan.

Till we sort out 1-3 I just brought in the adaptive_sampler from [`cmdstan#987`](https://github.com/stan-dev/cmdstan/pull/987), once we have those 3 sorted we will have a version that just blandly runs each chain in parallel, but it will be very easy to make new methods like auto warmup across chains since we'd just have some `run_auto_warmup_adaptive_sampler()` with `parallel_for`s for the warmup and sampling where the warmup is just something like 

```cpp
for (size_t i = 0; i < max_warmup; i += n_warmup_checkpoints) {
  for (int iter = i; iter < n_warmup_checkpoints; ++iter) {
      tbb::parallel_for(
      tbb::blocked_range<size_t>(0, n_chain, 1),
       // cool stuff!
      );
      // smart warmup stuff here
 }
}
```

#### Intended Effect

Run Chains in Parallel

#### Copyright and Licensing

### DIFF
--- a/src/stan/callbacks/file_stream_writer.hpp
+++ b/src/stan/callbacks/file_stream_writer.hpp
@@ -23,15 +23,15 @@ class file_stream_writer final : public writer {
    * @param[in] comment_prefix string to stream before each comment line.
    *  Default is "".
    */
-   explicit file_stream_writer(std::unique_ptr<std::ostream>&& output,
-                         const std::string& comment_prefix = "")
+  explicit file_stream_writer(std::unique_ptr<std::ostream>&& output,
+                              const std::string& comment_prefix = "")
       : output_(std::move(output)), comment_prefix_(comment_prefix) {}
-
 
   file_stream_writer();
   file_stream_writer(file_stream_writer& other) = delete;
-  file_stream_writer(file_stream_writer&& other) :
-   output_(std::move(other.output_)), comment_prefix_(std::move(other.comment_prefix_)) {}
+  file_stream_writer(file_stream_writer&& other)
+      : output_(std::move(other.output_)),
+        comment_prefix_(std::move(other.comment_prefix_)) {}
   /**
    * Virtual destructor
    */
@@ -51,9 +51,7 @@ class file_stream_writer final : public writer {
   /**
    * Get the underlying stream
    */
-  auto& get_stream() {
-    return *output_;
-  }
+  auto& get_stream() { return *output_; }
 
   /**
    * Writes a set of values in csv format followed by a newline.
@@ -113,7 +111,7 @@ class file_stream_writer final : public writer {
   }
 };
 
-}
-}
+}  // namespace callbacks
+}  // namespace stan
 
 #endif
--- a/src/stan/callbacks/file_stream_writer.hpp
+++ b/src/stan/callbacks/file_stream_writer.hpp
@@ -1,5 +1,5 @@
-#ifndef STAN_CALLBACKS_STREAM_WRITER_HPP
-#define STAN_CALLBACKS_STREAM_WRITER_HPP
+#ifndef STAN_CALLBACKS_FILE_STREAM_WRITER_HPP
+#define STAN_CALLBACKS_FILE_STREAM_WRITER_HPP
 
 #include <stan/callbacks/writer.hpp>
 #include <ostream>
@@ -10,27 +10,32 @@ namespace stan {
 namespace callbacks {
 
 /**
- * <code>stream_writer</code> is an implementation
- * of <code>writer</code> that writes to a stream.
+ * <code>file_stream_writer</code> is an implementation
+ * of <code>writer</code> that writes to a file.
  */
-class stream_writer final : public writer {
+class file_stream_writer final : public writer {
  public:
   /**
-   * Constructs a stream writer with an output stream
+   * Constructs a file stream writer with an output stream
    * and an optional prefix for comments.
    *
-   * @param[in, out] output stream to write
-   * @param[in] comment_prefix string to stream before
-   *   each comment line. Default is "".
+   * @param[in, out] A unique pointer to a type inheriting from `std::ostream`
+   * @param[in] comment_prefix string to stream before each comment line.
+   *  Default is "".
    */
-  explicit stream_writer(std::ostream& output,
+   explicit file_stream_writer(std::unique_ptr<std::ostream>&& output,
                          const std::string& comment_prefix = "")
-      : output_(output), comment_prefix_(comment_prefix) {}
+      : output_(std::move(output)), comment_prefix_(comment_prefix) {}
 
+
+  file_stream_writer();
+  file_stream_writer(file_stream_writer& other) = delete;
+  file_stream_writer(file_stream_writer&& other) :
+   output_(std::move(other.output_)), comment_prefix_(std::move(other.comment_prefix_)) {}
   /**
    * Virtual destructor
    */
-  virtual ~stream_writer() {}
+  virtual ~file_stream_writer() {}
 
   /**
    * Writes a set of names on a single line in csv format followed
@@ -42,6 +47,12 @@ class stream_writer final : public writer {
    */
   void operator()(const std::vector<std::string>& names) {
     write_vector(names);
+  }
+  /**
+   * Get the underlying stream
+   */
+  auto& get_stream() {
+    return *output_;
   }
 
   /**
@@ -57,7 +68,7 @@ class stream_writer final : public writer {
   /**
    * Writes the comment_prefix to the stream followed by a newline.
    */
-  void operator()() { output_ << comment_prefix_ << std::endl; }
+  void operator()() { *output_ << comment_prefix_ << std::endl; }
 
   /**
    * Writes the comment_prefix then the message followed by a newline.
@@ -65,14 +76,14 @@ class stream_writer final : public writer {
    * @param[in] message A string
    */
   void operator()(const std::string& message) {
-    output_ << comment_prefix_ << message << std::endl;
+    *output_ << comment_prefix_ << message << std::endl;
   }
 
  private:
   /**
    * Output stream
    */
-  std::ostream& output_;
+  std::unique_ptr<std::ostream> output_;
 
   /**
    * Comment prefix to use when printing comments: strings and blank lines
@@ -97,11 +108,12 @@ class stream_writer final : public writer {
 
     for (typename std::vector<T>::const_iterator it = v.begin(); it != last;
          ++it)
-      output_ << *it << ",";
-    output_ << v.back() << std::endl;
+      *output_ << *it << ",";
+    *output_ << v.back() << std::endl;
   }
 };
 
-}  // namespace callbacks
-}  // namespace stan
+}
+}
+
 #endif
--- a/src/stan/mcmc/hmc/hamiltonians/softabs_metric.hpp
+++ b/src/stan/mcmc/hmc/hamiltonians/softabs_metric.hpp
@@ -174,26 +174,26 @@ class softabs_metric : public base_hamiltonian<Model, softabs_point, BaseRNG> {
 
   // Threshold below which a power series
   // approximation of the softabs function is used
-  static double lower_softabs_thresh;
+  static constexpr double lower_softabs_thresh = 1e-4;
 
   // Threshold above which an asymptotic
   // approximation of the softabs function is used
-  static double upper_softabs_thresh;
+  static constexpr double upper_softabs_thresh = 18;
 
   // Threshold below which an exact derivative is
   // used in the Jacobian calculation instead of
   // finite differencing
-  static double jacobian_thresh;
+  static constexpr double jacobian_thresh = 1e-10;
 };
 
 template <class Model, class BaseRNG>
-double softabs_metric<Model, BaseRNG>::lower_softabs_thresh = 1e-4;
+constexpr double softabs_metric<Model, BaseRNG>::lower_softabs_thresh;
 
 template <class Model, class BaseRNG>
-double softabs_metric<Model, BaseRNG>::upper_softabs_thresh = 18;
+constexpr double softabs_metric<Model, BaseRNG>::upper_softabs_thresh;
 
 template <class Model, class BaseRNG>
-double softabs_metric<Model, BaseRNG>::jacobian_thresh = 1e-10;
+constexpr double softabs_metric<Model, BaseRNG>::jacobian_thresh;
 }  // namespace mcmc
 }  // namespace stan
 #endif
--- a/src/stan/services/util/generate_transitions.hpp
+++ b/src/stan/services/util/generate_transitions.hpp
@@ -44,7 +44,7 @@ void generate_transitions(stan::mcmc::base_mcmc& sampler, int num_iterations,
                           util::mcmc_writer& mcmc_writer,
                           stan::mcmc::sample& init_s, Model& model,
                           RNG& base_rng, callbacks::interrupt& callback,
-                          callbacks::logger& logger) {
+                          callbacks::logger& logger, size_t n_chain = 0) {
   for (int m = 0; m < num_iterations; ++m) {
     callback();
 
@@ -52,6 +52,9 @@ void generate_transitions(stan::mcmc::base_mcmc& sampler, int num_iterations,
         && (start + m + 1 == finish || m == 0 || (m + 1) % refresh == 0)) {
       int it_print_width = std::ceil(std::log10(static_cast<double>(finish)));
       std::stringstream message;
+      if (n_chain > 0) {
+        message << "Chain [" << (n_chain + 1) << "]";
+      }
       message << "Iteration: ";
       message << std::setw(it_print_width) << m + 1 + start << " / " << finish;
       message << " [" << std::setw(3)
--- a/src/stan/services/util/generate_transitions.hpp
+++ b/src/stan/services/util/generate_transitions.hpp
@@ -51,7 +51,8 @@ void generate_transitions(stan::mcmc::base_mcmc& sampler, int num_iterations,
     if (refresh > 0
         && (start + m + 1 == finish || m == 0 || (m + 1) % refresh == 0)) {
       int it_print_width = std::ceil(std::log10(static_cast<double>(finish)));
-      std::stringstream message;
+      std::string mes;
+      std::stringstream message(mes, std::ios_base::in);
       if (n_chain > 0) {
         message << "Chain [" << (n_chain + 1) << "]";
       }
--- a/src/stan/services/util/generate_transitions.hpp
+++ b/src/stan/services/util/generate_transitions.hpp
@@ -51,8 +51,7 @@ void generate_transitions(stan::mcmc::base_mcmc& sampler, int num_iterations,
     if (refresh > 0
         && (start + m + 1 == finish || m == 0 || (m + 1) % refresh == 0)) {
       int it_print_width = std::ceil(std::log10(static_cast<double>(finish)));
-      std::string mes;
-      std::stringstream message(mes, std::ios_base::in);
+      std::stringstream message;
       if (n_chain > 0) {
         message << "Chain [" << (n_chain + 1) << "]";
       }
--- a/src/stan/services/util/run_adaptive_sampler.hpp
+++ b/src/stan/services/util/run_adaptive_sampler.hpp
@@ -124,10 +124,11 @@ void run_adaptive_sampler(std::vector<Sampler>& samplers, Model& model,
     samples.emplace_back(cont_params, 0, 0);
   }
   std::vector<double> warm_delta_v(n_chain, 0);
+  std::vector<double> sample_delta_v(n_chain, 0);
   tbb::parallel_for(
       tbb::blocked_range<size_t>(0, n_chain, 1),
       [num_warmup, num_samples, num_thin, refresh, save_warmup, &samples,
-       &warm_delta_v, &writers, &samplers, &model, &cont_vectors, &rngs,
+       &warm_delta_v, &sample_delta_v, &writers, &samplers, &model, &cont_vectors, &rngs,
        &interrupt, &logger, &sample_writers,
        &diagnostic_writers](const tbb::blocked_range<size_t>& r) {
         for (size_t i = r.begin(); i != r.end(); ++i) {
@@ -161,15 +162,18 @@ void run_adaptive_sampler(std::vector<Sampler>& samplers, Model& model,
                                      refresh, true, false, writer, samp, model,
                                      rngs[i], interrupt, logger, i);
           auto end_sample = std::chrono::steady_clock::now();
-          double sample_delta_t
+          sample_delta_v[i]
               = std::chrono::duration_cast<std::chrono::milliseconds>(
                     end_sample - start_sample)
                     .count()
                 / 1000.0;
-          writer.write_timing(warm_delta_v[i], sample_delta_t);
         }
       },
       tbb::simple_partitioner());
+    for (int i = 0; i < n_chain; ++i) {
+      writers[i].write_timing(warm_delta_v[i], sample_delta_v[i]);
+    }
+
 }
 
 }  // namespace util
--- a/src/stan/services/util/run_adaptive_sampler.hpp
+++ b/src/stan/services/util/run_adaptive_sampler.hpp
@@ -100,18 +100,18 @@ void run_adaptive_sampler(std::vector<Sampler>& samplers, Model& model,
                           std::vector<SampT>& sample_writers,
                           std::vector<DiagnoseT>& diagnostic_writers,
                           size_t n_chain) {
-  std::vector<services::util::mcmc_writer> writers;
-  writers.reserve(n_chain);
+  if (n_chain == 0) {
+    run_adaptive_sampler(samplers[0], model, cont_vectors[0], num_warmup, num_samples,
+     num_thin, refresh, save_warmup, rngs[0], interrupt, logger, sample_writers[0],
+     diagnostic_writers[0]);
+  }
   std::vector<stan::mcmc::sample> samples;
   samples.reserve(n_chain);
-
   for (int i = 0; i < n_chain; ++i) {
-    auto&& sample_writer = sample_writers[i];
-    auto&& diagnostic_writer = diagnostic_writers[i];
     auto&& sampler = samplers[i];
-    Eigen::Map<Eigen::VectorXd> cont_params(cont_vectors[i].data(),
-                                            cont_vectors[i].size());
     sampler.engage_adaptation();
+    Eigen::Map<Eigen::VectorXd> cont_params(cont_vectors[i].data(),
+                                           cont_vectors[i].size());
     try {
       sampler.z().q = cont_params;
       sampler.init_stepsize(logger);
@@ -120,19 +120,17 @@ void run_adaptive_sampler(std::vector<Sampler>& samplers, Model& model,
       logger.info(e.what());
       return;
     }
-    writers.emplace_back(sample_writer, diagnostic_writer, logger);
     samples.emplace_back(cont_params, 0, 0);
   }
-  std::vector<double> warm_delta_v(n_chain, 0);
-  std::vector<double> sample_delta_v(n_chain, 0);
   tbb::parallel_for(
       tbb::blocked_range<size_t>(0, n_chain, 1),
       [num_warmup, num_samples, num_thin, refresh, save_warmup, &samples,
-       &warm_delta_v, &sample_delta_v, &writers, &samplers, &model,
-       &cont_vectors, &rngs, &interrupt, &logger, &sample_writers,
+       &samplers, &model,
+       &rngs, &interrupt, &logger, &sample_writers,
        &diagnostic_writers](const tbb::blocked_range<size_t>& r) {
         for (size_t i = r.begin(); i != r.end(); ++i) {
-          auto&& writer = writers[i];
+          auto&& sample_writer = sample_writers[i];
+          auto writer = services::util::mcmc_writer(sample_writer, diagnostic_writers[i], logger);
           auto&& sampler = samplers[i];
           auto&& samp = samples[i];
 
@@ -140,39 +138,36 @@ void run_adaptive_sampler(std::vector<Sampler>& samplers, Model& model,
           writer.write_sample_names(samp, sampler, model);
           writer.write_diagnostic_names(samp, sampler, model);
 
-          auto start_warm = std::chrono::steady_clock::now();
+          const auto start_warm = std::chrono::steady_clock::now();
           util::generate_transitions(sampler, num_warmup, 0,
                                      num_warmup + num_samples, num_thin,
                                      refresh, save_warmup, true, writer, samp,
                                      model, rngs[i], interrupt, logger, i);
-          auto end_warm = std::chrono::steady_clock::now();
-          warm_delta_v[i]
+          const auto end_warm = std::chrono::steady_clock::now();
+          auto warm_delta
               = std::chrono::duration_cast<std::chrono::milliseconds>(
                     end_warm - start_warm)
                     .count()
                 / 1000.0;
           sampler.disengage_adaptation();
-          auto&& sample_writer = sample_writers[i];
           writer.write_adapt_finish(sampler);
           sampler.write_sampler_state(sample_writer);
 
-          auto start_sample = std::chrono::steady_clock::now();
+          const auto start_sample = std::chrono::steady_clock::now();
           util::generate_transitions(sampler, num_samples, num_warmup,
                                      num_warmup + num_samples, num_thin,
                                      refresh, true, false, writer, samp, model,
                                      rngs[i], interrupt, logger, i);
-          auto end_sample = std::chrono::steady_clock::now();
-          sample_delta_v[i]
+          const auto end_sample = std::chrono::steady_clock::now();
+          auto sample_delta
               = std::chrono::duration_cast<std::chrono::milliseconds>(
                     end_sample - start_sample)
                     .count()
                 / 1000.0;
+          writer.write_timing(warm_delta, sample_delta);
         }
       },
       tbb::simple_partitioner());
-  for (int i = 0; i < n_chain; ++i) {
-    writers[i].write_timing(warm_delta_v[i], sample_delta_v[i]);
-  }
 }
 
 }  // namespace util
--- a/src/stan/services/util/run_adaptive_sampler.hpp
+++ b/src/stan/services/util/run_adaptive_sampler.hpp
@@ -34,7 +34,7 @@ namespace util {
  * @param[in,out] sample_writer writer for draws
  * @param[in,out] diagnostic_writer writer for diagnostic information
  */
-template <class Sampler, class Model, class RNG>
+template <typename Sampler, typename Model, typename RNG>
 void run_adaptive_sampler(Sampler& sampler, Model& model,
                           std::vector<double>& cont_vector, int num_warmup,
                           int num_samples, int num_thin, int refresh,
@@ -87,6 +87,91 @@ void run_adaptive_sampler(Sampler& sampler, Model& model,
                           / 1000.0;
   writer.write_timing(warm_delta_t, sample_delta_t);
 }
+
+template <typename Sampler, typename Model, typename RNG, typename SampleWriter,
+          typename DiagnosticWriter>
+void run_adaptive_sampler(Sampler& sampler, Model& model,
+                          std::vector<std::vector<double>>& cont_vector,
+                          int num_warmup, int num_samples, int num_thin,
+                          int refresh, bool save_warmup, RNG& rng,
+                          callbacks::interrupt& interrupt,
+                          callbacks::logger& logger,
+                          std::vector<SampleWriter>& sample_writer,
+                          std::vector<DiagnosticWriter>& diagnostic_writer,
+                          size_t n_chains = 0) {
+  std::vector<services::util::mcmc_writer> writers;
+  writers.reserve(n_chain);
+  std::vector<stan::mcmc::sample> samples;
+  samples.reserve(n_chain);
+
+  for (int i = 0; i < n_chain; ++i) {
+    auto&& sample_writer = sample_writers[i];
+    auto&& diagnostic_writer = diagnostic_writers[i];
+    auto&& sampler = samplers[i];
+    Eigen::Map<Eigen::VectorXd> cont_params(cont_vectors[i].data(),
+                                            cont_vectors[i].size());
+    sampler.engage_adaptation();
+    try {
+      sampler.z().q = cont_params;
+      sampler.init_stepsize(logger);
+    } catch (const std::exception& e) {
+      logger.info("Exception initializing step size.");
+      logger.info(e.what());
+      return;
+    }
+    writers.emplace_back(sample_writer, diagnostic_writer, logger);
+    samples.emplace_back(cont_params, 0, 0);
+  }
+  std::vector<double> warm_delta_v(n_chain, 0);
+  tbb::parallel_for(
+      tbb::blocked_range<size_t>(0, n_chain, 1),
+      [num_warmup, num_samples, num_thin, refresh, save_warmup, &samples,
+       &warm_delta_v, &writers, &samplers, &model, &cont_vectors, &rngs,
+       &interrupt, &logger, &sample_writers,
+       &diagnostic_writers](const tbb::blocked_range<size_t>& r) {
+        for (size_t i = r.begin(); i != r.end(); ++i) {
+          auto&& writer = writers[i];
+          auto&& sampler = samplers[i];
+          auto&& samp = samples[i];
+
+          // Headers
+          writer.write_sample_names(samp, sampler, model);
+          writer.write_diagnostic_names(samp, sampler, model);
+
+          auto start_warm = std::chrono::steady_clock::now();
+          util::generate_transitions(sampler, num_warmup, 0,
+                                     num_warmup + num_samples, num_thin,
+                                     refresh, save_warmup, true, writer, samp,
+                                     model, rngs[i], interrupt, logger, i);
+          auto end_warm = std::chrono::steady_clock::now();
+          warm_delta_v[i]
+              = std::chrono::duration_cast<std::chrono::milliseconds>(
+                    end_warm - start_warm)
+                    .count()
+                / 1000.0;
+          sampler.disengage_adaptation();
+          auto&& sample_writer = sample_writers[i];
+          auto&& sampler = samplers[i];
+          writer.write_adapt_finish(sampler);
+          sampler.write_sampler_state(sample_writer);
+
+          auto start_sample = std::chrono::steady_clock::now();
+          util::generate_transitions(sampler, num_samples, num_warmup,
+                                     num_warmup + num_samples, num_thin,
+                                     refresh, true, false, writer, samp, model,
+                                     rngs[i], interrupt, logger, i);
+          auto end_sample = std::chrono::steady_clock::now();
+          double sample_delta_t
+              = std::chrono::duration_cast<std::chrono::milliseconds>(
+                    end_sample - start_sample)
+                    .count()
+                / 1000.0;
+          writer.write_timing(warm_delta_v[i], sample_delta_t);
+        }
+      },
+      tbb::simple_partitioner());
+}
+
 }  // namespace util
 }  // namespace services
 }  // namespace stan
--- a/src/stan/services/util/run_adaptive_sampler.hpp
+++ b/src/stan/services/util/run_adaptive_sampler.hpp
@@ -128,8 +128,8 @@ void run_adaptive_sampler(std::vector<Sampler>& samplers, Model& model,
   tbb::parallel_for(
       tbb::blocked_range<size_t>(0, n_chain, 1),
       [num_warmup, num_samples, num_thin, refresh, save_warmup, &samples,
-       &warm_delta_v, &sample_delta_v, &writers, &samplers, &model, &cont_vectors, &rngs,
-       &interrupt, &logger, &sample_writers,
+       &warm_delta_v, &sample_delta_v, &writers, &samplers, &model,
+       &cont_vectors, &rngs, &interrupt, &logger, &sample_writers,
        &diagnostic_writers](const tbb::blocked_range<size_t>& r) {
         for (size_t i = r.begin(); i != r.end(); ++i) {
           auto&& writer = writers[i];
@@ -170,10 +170,9 @@ void run_adaptive_sampler(std::vector<Sampler>& samplers, Model& model,
         }
       },
       tbb::simple_partitioner());
-    for (int i = 0; i < n_chain; ++i) {
-      writers[i].write_timing(warm_delta_v[i], sample_delta_v[i]);
-    }
-
+  for (int i = 0; i < n_chain; ++i) {
+    writers[i].write_timing(warm_delta_v[i], sample_delta_v[i]);
+  }
 }
 
 }  // namespace util
--- a/src/stan/services/util/run_adaptive_sampler.hpp
+++ b/src/stan/services/util/run_adaptive_sampler.hpp
@@ -108,8 +108,8 @@ void run_adaptive_sampler(std::vector<Sampler>& samplers, Model& model,
   }
   tbb::parallel_for(
       tbb::blocked_range<size_t>(0, n_chain, 1),
-      [num_warmup, num_samples, num_thin, refresh, save_warmup,
-       &samplers, &model, &rngs, &interrupt, &logger, &sample_writers, &cont_vectors,
+      [num_warmup, num_samples, num_thin, refresh, save_warmup, &samplers,
+       &model, &rngs, &interrupt, &logger, &sample_writers, &cont_vectors,
        &diagnostic_writers](const tbb::blocked_range<size_t>& r) {
         for (size_t i = r.begin(); i != r.end(); ++i) {
           auto&& sampler = samplers[i];
--- a/src/stan/services/util/run_adaptive_sampler.hpp
+++ b/src/stan/services/util/run_adaptive_sampler.hpp
@@ -101,9 +101,10 @@ void run_adaptive_sampler(std::vector<Sampler>& samplers, Model& model,
                           std::vector<DiagnoseT>& diagnostic_writers,
                           size_t n_chain) {
   if (n_chain == 0) {
-    run_adaptive_sampler(samplers[0], model, cont_vectors[0], num_warmup, num_samples,
-     num_thin, refresh, save_warmup, rngs[0], interrupt, logger, sample_writers[0],
-     diagnostic_writers[0]);
+    run_adaptive_sampler(samplers[0], model, cont_vectors[0], num_warmup,
+                         num_samples, num_thin, refresh, save_warmup, rngs[0],
+                         interrupt, logger, sample_writers[0],
+                         diagnostic_writers[0]);
   }
   std::vector<stan::mcmc::sample> samples;
   samples.reserve(n_chain);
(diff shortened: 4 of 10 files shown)


SteveBronder:
Tagging parties interested @wds15 @yizhang-yiz @mitzimorris @bbbales2 once we have the API sorted out this should be decently easy peasy

wds15:
The isolation thing is easy to implement. Let‘s just do an issue for that and fix it. That‘s anyway good to have to make things safe whenever Stan chains run in parallel environments (within a single arena).

wds15:
One point for consideration is the thread allocation / requesting. So far we have (by design of one chain per process) a by-chain thread allocation controlled by the environment variable `STAN_NUM_THREADS`. The way the current PR is structured is such that there won't anymore be a control over how many threads are being used per chain. I would expect that to be a good thing and it will leave resource allocation to the Intel TBB scheduler, which is what the thing is build for. If we stay with such a design, then we can leave the setup of the size of the used tbb arena within cmdstan and we do not need to deal with setting up an arena within the services. To me that makes sense. @SteveBronder , @rok-cesnovar & others ? Note that the meaning of `STAN_NUM_THREADS` will change (depending on how you see it) to become "total number of threads to be used for all chains" if we don't do anything. The best solution I think is to add to cmdstan a program option which controls the number of threads being used. All of these are then used by the `n_chains` requested from the services.

SteveBronder:
Yeah @wds15 I think what you are saying makes sense. What we could do is finally add the `parallel` option to cmdstan so users would call something like 

```
./my_model sample data file=./my_data.json chains n_chains=8 parallel n_threads=16
```

Making sure I read your comment right, I think we are both on the same page that the user should just tell stan the total amount of available threads and on the backend we can sort out how to use those. I think Stan's preference would always be to use a thread for each chain since the overall chain is the more expensive thing to run. Then any other available threads can be used in the model

bbbales2:
> we do not need to deal with setting up an arena within the services. To me that makes sense. @SteveBronder , @rok-cesnovar & others ?

Other than segfaults, I guess the only thing I worry about here is I expect work done outside of reduce_sum to scale more efficiently than work in reduce_sum (because reduce_sum has the shared argument copy overhead). If there's lots of outside-reduce_sum work and lots of inside-reduce_sum work, I'd hope that the outside work gets priority cause the inside work scales worse.

I assume that would show up quickly with experiments though so probably just easier to see if it happens than worry about it too much ahead of time.

> The sample_writers and diagnostic_writers come in as a vector of streams where the current impl comes in as a writer for each stream. Is that fine?

So there's a separate output file for each chain? Question there is what would the interface look like.

It seems awkward to do:

```
output file="file1.csv,file2.csv,..."
```

The other option would be do some sort of basename or automatic name-rewriting. So if the user specified:

```
output file="myfile.csv"
```

They would get myfile.csv.0, myfile.csv.1, etc. or something.

My guess is we should just have a meeting with @mitzimorris and @rok-cesnovar and talk that one through. I like the name re-writing which looks like what is being done in the prototyping. There's probably also some extra output that should be written in the not-quite-csv output headers.

It looks like this is something that can remain compatible with the single chain interface which is neat.

> or is it safe to just have multiple RNGs?

Seems simpler to think about to have multiple RNGs unless they aren't threadsafe or something. Like, if you wanted to reproduce a calculation and you only wanted to run one thread, then no need to reproduce in parallel.

If seed isn't specified then I assume picking a seed however we currently do is the way to go. We'd probably need to be able to specify the seed as a list or something now. There are probably other arguments that may make sense to be optionally-lists? Inits? I'm not sure. That's a question to coordinate with cmdstanr and cmdstanpy I guess.

> For what @wds15 brought up here should the isolate() thing happen before we put this PR in or after?

Sounds like it shouldn't negatively affect behavior so yeah def if it makes this easier.

What is the file_stream_writer here doing? I searched for it but it didn't show up in the run_adaptive thing for me. Could have missed something.

Looks like the basic coding is under control. I am happy to review, help write the tests, or help with the cmdstanr/cmdstanpy stuff for this.

wds15:
> If there's lots of outside-reduce_sum work and lots of inside-reduce_sum work, I'd hope that the outside work gets priority cause the inside work scales worse.

The work anyway needs to get done.

... the only risk there is that the TBB scheduler slices the nested reduce_sum work into more slices than what is healthy if given more cores, but I don‘t think that this is the case.

All that said... let‘s just do a basic benchmark, sure.


> Seems simpler to think about to have multiple RNGs unless they aren't threadsafe or something.

The rng topic needs to be handled careful. It should not make a difference how many threads are being used as to what the numerical result is. The numerical result should always be the same ideally.

SteveBronder:
>> we do not need to deal with setting up an arena within the services. To me that makes sense. @SteveBronder , @rok-cesnovar & others ?

> Other than segfaults, I guess the only thing I worry about here is I expect work done outside of reduce_sum to scale more efficiently than work in reduce_sum (because reduce_sum has the shared argument copy overhead). If there's lots of outside-reduce_sum work and lots of inside-reduce_sum work, I'd hope that the outside work gets priority cause the inside work scales worse.

> I assume that would show up quickly with experiments though so probably just easier to see if it happens than worry about it too much ahead of time.

Yeah I think we would notice if we are on a bad beat here

>> The sample_writers and diagnostic_writers come in as a vector of streams where the current impl comes in as a writer for each stream. Is that fine?

> So there's a separate output file for each chain? Question there is what would the interface look like.

> The other option would be do some sort of basename or automatic name-rewriting. So if the user specified:

I like this option
> output file="myfile.csv"

> They would get myfile.csv.0, myfile.csv.1, etc. or something.

Though I think I'd just prefer `myfile_0.csv` just so things that import csvs know it's a csv still

> My guess is we should just have a meeting with @mitzimorris and @rok-cesnovar and talk that one through. I like the name re-writing which looks like what is being done in the prototyping. There's probably also some extra output that should be written in the not-quite-csv output headers.

Agree!

> It looks like this is something that can remain compatible with the single chain interface which is neat.

Yep! 

> Seems simpler to think about to have multiple RNGs unless they aren't threadsafe or something. Like, if you wanted to reproduce a calculation and you only wanted to run one thread, then no need to reproduce in parallel.

So the PRNG thing, I think we will be fine if we set the seed to be just `user_seed + chain_num` for each chain. I might be overreacting but I've read in the past that in parallel systems you can have issues where for certain PRNGs you can end up with overlap of the stream of random numbers (since they are pseudo RNGs).  From [here](https://www.boost.org/doc/libs/1_75_0/doc/html/boost_random/reference.html#boost_random.reference.generators) it seems the length of the ecuyer1988 prng is about `2^61`. It would be nice to not worry about this but it's just an area I'm uncertain about. Tagging @betanalpha as he might know more about whether we should worry about this or not

> What is the file_stream_writer here doing? I searched for it but it didn't show up in the run_adaptive thing for me. Could have missed something.

So that `file_stream_writer` holds unique pointers to each of the file streams. I had some issues using `stream_writer` where somewhere a reference was being dereferenced or something was falling out of scope (`stream_writer` just holds it's `ostream` by refrerence) so I thought the easiest solution was just to wrap that ostream into an `std::unique_ptr`. Then we also get promises that no-one ever copies that thing.

> Looks like the basic coding is under control. I am happy to review, help write the tests, or help with the cmdstanr/cmdstanpy stuff for this.

Much appreciated! I think at this point I still need to figure out some patterns etc. that we can probs hash out over a call. But besides that I think the testing etc. I can handle for now and can holler otherwise

> The rng topic needs to be handled careful. It should not make a difference how many threads are being used as to what the numerical result is. The numerical result should always be the same ideally.

We can def have reproducibility no sweat. The main issue is just making sure we have good random numbers (imagine spinning up 8 chains but then each chain has correlated random numbers :-x)


I got a test up and running, but running the address sanitizer over it I'm seeing leaks happening after the test finishes inside of `AutodiffStackSingleton`. Full output below, but it seems like it's not destroying the auto diff stack at the end of the program. But do we even want it to delete that? It just get's tossed once the program is done. Should we do something that makes sure that once the global `ad_tape_observer` destructor is called at the end of the program those ChainableStacks are also removed? Or is this some other issue?

https://gist.github.com/SteveBronder/7c2b2c371d4ce805fbd301fca32c088f

rok-cesnovar:
I agree that the user should just set the max number of cores to use (threads to spawn) and let us deal with how that is used.

rok-cesnovar:
I am a bit confused here. The name of the PR is "Parallel Adaptive Sampler". Does that mean the first step of multi chain in service will already include the campfire-communication-between-chains-in-warmup things? I think step 1 should be just multi chains in services. I am probably overthinking the title :)

SteveBronder:
No sorry tbc this is just running `n` adaptation samplers in parallel with no communication. This does the "simpler" thing of multiple chains in services

SteveBronder:
Hmm, running 3 chains with two threads also gives some nasty stuff 

https://gist.github.com/SteveBronder/b414e37cb081ba2be7cb5da512557c7c

SteveBronder:
Actually above may be something about https://github.com/stan-dev/math/issues/2428

SteveBronder:
Okay more confusion, if I compile that test with 

```
STAN_THREADS=true
TBB_LIBRARIES=tbb tbbmalloc tbbmalloc_proxy
```

Then run the test with 

```
STAN_NUM_THREADS=2 ./runTests.py -j30 ./src/test/unit/services/util/run_adaptive_sampler_parallel_test.cpp
```

I no longer get  the vptr undefined error and just get the error about the end of the application not cleaning up memory.

wds15:
Can u use an external up to date version of the one tbb? That should be possible with our makefiles and hopefully resolves the cleanup issue.

SteveBronder:
I'll try that later today or tomorrow. But even if that solves it folks will still not be able to use this because of how RcppParallel is on the old version right? Tbh I wonder if we could just submit `OneTBB` on Cran that just gives C++ access to a an updated version of OneTBB.

SteveBronder:
Ugh so I'd like to just bump up if we can, then if that's possible we can talk to the RcppParallel folks about whether they should update, whether we should make a new library, etc. But it looks like the new oneTBB must use cmake so we will have that as a depedency as well. It feels onerous to ask users to supply their own version of tbb, though I'm not sure what's normal here. I have a branch of math below with the new tbb version, but need to sort out how to hook it up to our makefile stuff. Maybe it's just time to use cmake?

https://github.com/stan-dev/math/compare/update/onetbb

SteveBronder:
Pinging @hsbadr as I feel like he might have a better idea of what route to take here

hsbadr:
@SteveBronder `RcppParallel` does support external TBB library using `TBB_LIB` and `TBB_INC` environment variables (https://github.com/RcppCore/RcppParallel#intel-tbb), just like Stan's `math` (https://github.com/stan-dev/math#intel-tbb). You need to (force) reinstall `RcppParallel` after correctly setting the environment variables.

So, we may remove the TBB source code from `math` (make it an external dependency by default). This will ultimately fix the conflicting TBB headers (with system `libtbb-dev`, `RcppParallel`'s TBB, ... etc) where Stan can explicitly declare the supported versions of TBB that can also be adopted by `RcppParallel`.

We could use external `oneTBB` binaries (from https://github.com/oneapi-src/oneTBB/releases), ship `oneTBB` binaries in `math` to avoid `cmake` for now, update the internal TBB source code and the build system in https://github.com/stan-dev/math/tree/update/onetbb, or create a new `oneTBB` R package. Then, we can give instructions to reinstall `RcppParallel` with `TBB_LIB` and `TBB_INC` set. I haven't tested external `oneTBB` on non-Unix targets yet.

If/when needed, we can maintain code compatibility via `TBB_INTERFACE_VERSION` (from `oneapi/tbb/version.h` in `oneTBB`, or `tbb/tbb_stddef.h` in the old versions/interface of TBB) which is currently controlled by `TBB_INTERFACE_NEW` since we have only two conflicting (old and new) TBB interfaces.

I haven't reviewed the code in this PR yet, but I'll take a look ASAP.

wds15:
> So, we may remove the TBB source code from math (make it an external dependency by default). This will ultimately fix the conflicting TBB headers (with system libtbb-dev, RcppParallel's TBB, ... etc) where Stan can explicitly declare the supported versions of TBB that can also be adopted by RcppParallel.

So far we have always kept all required libraries together which are required to get Stan running. I don't think we would want to move away from that.

@SteveBronder Maybe things are not so desperate. Either we can work around the issue or maybe this problem is not really an issue as things anyway get cleaned up. Let's see.

To correct what I said earlier: If we start 4 chains with 4 threads, then any nested reduce_sum should not be given priority. We then still want the 4 chains to run in parallel instead of running 1 chain at a time with reduce_sum using 4 cores simultaneously. In order to test this I suggest that we code up a version of this thing where the different chain are seeded with independent RNGs, but those run in sync. This way we fire up 4 times the same work and we should get their results in basically the same time. For the final release it should not be possible to fire up 4 exactly equal chains, of course.

t4c1:
> We'd probably need to be able to specify the seed as a list or something now.

We can use single seed if we are smart about it. If we initialize one PRNG with user provided seed and use that to generate seeds for PRNGs in threads. Obviously this kind of initialization must use a different PRNG than threads.

> wrt RNGs, do we need to have a special PRNG or is it safe to just have multiple RNGs? Should we use a different RNG with a longer cycle and set each seed to user given initial + 2^chain_id or something?

> I might be overreacting but I've read in the past that in parallel systems you can have issues where for certain PRNGs you can end up with overlap of the stream of random numbers (since they are pseudo RNGs). From here it seems the length of the ecuyer1988 prng is about 2^61.

For usage on a single CPU period of 2^61 seems adequate. For example if we run 200 threads and generate 10^7 random numbers per thread that gives up probability that any streams overlap of 10^-6. With more threads (thinking about GPUs or clusters) we should use a PRNG with longer period.

I already did some testing of PRNGs in parallel (with focus on GPUs, but it should be applicable here): https://link.springer.com/article/10.1007/s11227-019-02756-2

rok-cesnovar:
- re RNGs: 

   Is starting 4 chains in TBB any different than starting 4 chains in separate processes, which is what the interfaces do now? Based on my understanding it is not. The services layer takes in the seed and the chain id and then multiplies the RNG with a discard stride 2^50 (see https://github.com/stan-dev/stan/blob/develop/src/stan/services/util/create_rng.hpp#L27). So either we have been doing this wrong all along and we need to fix this or we just use this.

- re downstream libraries and supplying them: 

  I agree with @wds15, we should not move away from supplying the base libraries on our own. We have to know most of our users are not C++ developers or even programmers and they do not want to deal with package managers, paths and whatnot. Controlling the versions of downstream libraries also vastly simplifies our testing. We could move them out of git and just supply them in releases, but I doubt that makes a huge difference.

  We should enable (or at least make it somewhat easy) users to plug in their own downstream library installation (like @hsbadr did for TBB) if they want to, but at that point they are on their own in terms of if things will work as we do not guarantee (and cant) it will work.

  Its fine if we dont supply any libraries that are only targeting power-users or advanced users, but the libraries that are needed for the basic stuff we should supply.

Q for @hsbadr : What is the last version of TBB/oneTBB that we can just plug into Math (so just replace sources and expect things to work with our current makefiles). Maybe we should move to that and help RcppParallel move to that as well? So we have a bit of extra time to figure out what to do next.

t4c1:
> Is starting 4 chains in TBB any different than starting 4 chains in separate processes, which is what the interfaces do now?

No.

> The services layer takes in the seed and the chain id and then multiplies the RNG with a discard stride 2^50 (see https://github.com/stan-dev/stan/blob/develop/src/stan/services/util/create_rng.hpp#L27). So either we have been doing this wrong all along and we need to fix this or we just use this.

Not every PRNG supports this. Since we are using one that does, it is fine.

wds15:
> > Is starting 4 chains in TBB any different than starting 4 chains in separate processes, which is what the interfaces do now?
> 
> No.
> 
> > The services layer takes in the seed and the chain id and then multiplies the RNG with a discard stride 2^50 (see https://github.com/stan-dev/stan/blob/develop/src/stan/services/util/create_rng.hpp#L27). So either we have been doing this wrong all along and we need to fix this or we just use this.
> 
> Not every PRNG supports this. Since we are using one that does, it is fine.

The thing that is new for the services is that we must deal with multiple instances of the rng. So far we only had a single rng and now we have one per chain being (at the same time as it's not any more happening in different cmdstan runs).

rok-cesnovar:
Good point, though I guess each TBB task will call its own https://github.com/stan-dev/stan/blob/develop/src/stan/services/sample/hmc_nuts_dense_e.hpp#L59 and thus the task will have its thread-local rng, the rng variable is created there. Or is the plan to make all the services like hmc_nuts_dense_e run the chains?

wds15:
As I recall, the stan service so far got their rng instance passed into from the client code (cmdstan). Within Stan services we only passed references around. Maybe my memory is wrong on that, but all this needs some consideration.

rok-cesnovar:
The services only get the random seed as an `unsigned int`. The random seed is either user-set or chosen based on time.

And then the rng's are created in 2 places: the service used (like the example linked above) and the model's constructor.
The model's constructor has no chain ID info, which means that if you start 4 chains with the same seed, the chains will also have the exact same rng. Which is the correct behaviour as we want to run multiple chains with the exact same data.

Just paste the below model in https://rok-cesnovar.github.io/stanc3js-demo/ and check the constructor.
```stan
transformed data {
  int N = 10;
  vector[N] y;
  for (n in 1:N)
    y[n] = normal_rng(0, 1);
  
}
```

CmdStan creates no rngs on its own. But yes, we have to be careful with this.

wds15:
Cool parser web-site that is!

hsbadr:
> So far we have always kept all required libraries together which are required to get Stan running. I don't think we would want to move away from that.

I see. Upgrading to `oneTBB` is worth it then. Since the current version of TBB in `math` isn't compatible with the CRAN version of `RcppParallel`, there's no need to freeze it. We can instruct the users to reinstall `RcppParallel` with our TBB library. This would require that the TBB binaries built by Stan be available via `TBB_INC` and `TBB_LIB` environment variables. 

> Q for @hsbadr : What is the last version of TBB/oneTBB that we can just plug into Math (so just replace sources and expect things to work with our current makefiles). Maybe we should move to that and help RcppParallel move to that as well? So we have a bit of extra time to figure out what to do next.

So far, any version of `oneTBB` (up to `oneTBB 2021.1.1`) can be plugged into `math` with `TBB_INTERFACE_NEW=true`. The only difference in the latest versions is that the headers have been moved to `oneapi/tbb/*` from `tbb/*`, but they're being redirected anyway ([`tbb/*`](https://github.com/oneapi-src/oneTBB/tree/master/include/tbb): `#include "../oneapi/tbb/*"`; so, no need to change anything).

wds15:
@hsbadr So R users are forced to install oneTBB to escape the one supplied by `RcppParallel`? If yes, then it would be really good to help the `RcppParallel` folks to update as well to oneTBB. Otherwise that will be a major burden for R users and add to the installation headaches users must go through. A solution not requiring installing additional things would be best.

SteveBronder:
> So far we have always kept all required libraries together which are required to get Stan running. I don't think we would want to move away from that.

Yes I agree with this. I don't really like the idea of making users supply their own tbb. 

> We can use single seed if we are smart about it. If we initialize one PRNG with user provided seed and use that to generate seeds for PRNGs in threads. Obviously this kind of initialization must use a different PRNG than threads.

Do you mean like we need to use one PRNG like ecuyer1988 to initialize then like taus88 for the actual thread PRNGS?

> For usage on a single CPU period of 2^61 seems adequate. For example if we run 200 threads and generate 10^7 random numbers per thread that gives up probability that any streams overlap of 10^-6. With more threads (thinking about GPUs or clusters) we should use a PRNG with longer period.

This sounds like we don't have to worry about it and I'm very cool with not having to worry about this

> So either we have been doing this wrong all along and we need to fix this or we just use this.

Yes let's do it the way we do things now and if this is an issue we notice we can do something fancier later

> So far, any version of oneTBB (up to oneTBB 2021.1.1) can be plugged into math with TBB_INTERFACE_NEW=true. The only difference in the latest versions is that the headers have been moved to oneapi/tbb/* from tbb/*, but they're being redirected anyway (tbb/*: #include "../oneapi/tbb/*"; so, no need to change anything).

I think my Q is how do we update the tbb in Stan math. I'm guessing we can't just move the tbb folder from the new versions into `tbb_2019_U8` to use the makefiles there.

SteveBronder:
@alashworth made the Stan monorepo using cmake. I wonder if we could use a lot of that stuff to use cmake throughout the project? We could for a bit have both cmake and the makefiles, then slowly add cmake to each repo. Then once we have cmake up to cmdstan we remove the make stuff

wds15:
> Do you mean like we need to use one PRNG like ecuyer1988 to initialize then like taus88 for the actual thread PRNGS?

seed and chain id determine the initial state of the PRNG ecuyer which we use. In this parallel world, I would create as many instances of the ecuyer as we are asked to run chains. Again, the seed and chain id determine the starting state of the chain specific PRNG. In the parallel world we should keep one instance of the ecuyer per instance around.

betanalpha:
Regarding the pseudo-random number generator issue:  there are two approaches depending on the nature of the pseudo-random number generator.

If the generator admits efficient jump-ahead/seekability then it can be strided, staring with an initial state generated from the given seed then jumping ahead enough states that the stream appears uncorrelated with the initial stream.  Choosing the stride is a bit of an art — it has to be smaller than the generator period divided by the number of desired streams, but even that might not guarantee sufficiently uncorrelated streams.  Usually if the stride is much longer than the number of states needed then the behavior will be okay, but the results are empirical.

In some cases the generator can also produce multiple streams that are independent by construction by using updates that are “orthogonal” in some sense.  Here each stream uses the same initial seed/state but the different configurations of the sequential update that ensure distinct streams.

The latter is faster and has better guarantees than the former but the overhead will be trivial for Stan so there’s no real difference.  For more see Sections 2.5 and 4.3 of https://www.pcg-random.org/pdf/hmc-cs-2014-0905.pdf <https://www.pcg-random.org/pdf/hmc-cs-2014-0905.pdf>; http://www.iro.umontreal.ca/~lecuyer/myftp/papers/streams00.pdf <http://www.iro.umontreal.ca/~lecuyer/myftp/papers/streams00.pdf> also has useful information although it a bit more dense. The L’Ecuyer generator supports striding and that was the intended strategy for handling multiple chains (and also why it was chosen over the twister, if I remember correctly).  It should also work here provided that the chain_id naming is consistent.

One issue with the model model prng generators is that they control both the transformed data and the generated quantities blocks.  While there are cases where a common generator for the transformed data makes sense there are fewer were a common generator for generated quantities makes sense.  Indeed I would rather have different generators across the model replications than a common one — it’s safer, more appropriate to simulation calibration methods, and if anyone does want common data it’s easy enough to split the process up into two steps — generating common data with the fixed_param sampler and then passing that to the chains.

hsbadr:
> I think my Q is how do we update the tbb in Stan math. I'm guessing we can't just move the tbb folder from the new versions into `tbb_2019_U8` to use the makefiles there.

The current TBB build rules in `math` won't work with `oneTBB`, and we don't have to remove `tbb_2019_U8` if we'd like to support backward compatibility (switching the sources based on `TBB_INTERFACE_NEW`). It may not be optimal but we could add rules to download and copy the `oneTBB` binaries to `$(TBB_BIN)` skipping the build process until `math` supports `cmake` or we backport `oneTBB` build rules in `math/make`.

SteveBronder:
> The current TBB build rules in math won't work with oneTBB, and we don't have to remove tbb_2019_U8 if we'd like to support backward compatibility (switching the sources based on TBB_INTERFACE_NEW). It may not be optimal but we could add rules to download and copy the oneTBB binaries to $(TBB_BIN) skipping the build process until math supports cmake or we backport oneTBB build rules in math/make.

Oooof I just spent like all day trying to figure out how to get stan math setup with cmake and it was brutal. It seems like cmake extremely does not like when a package builds all of it's dependencies which tbh is pretty annoying. I can try one more crack tmrw

alashworth:
> > The current TBB build rules in math won't work with oneTBB, and we don't have to remove tbb_2019_U8 if we'd like to support backward compatibility (switching the sources based on TBB_INTERFACE_NEW). It may not be optimal but we could add rules to download and copy the oneTBB binaries to $(TBB_BIN) skipping the build process until math supports cmake or we backport oneTBB build rules in math/make.
> 
> Oooof I just spent like all day trying to figure out how to get stan math setup with cmake and it was brutal. It seems like cmake extremely does not like when a package builds all of it's dependencies which tbh is pretty annoying. I can try one more crack tmrw

CMake is not a package manager. Some projects get away with using it as one, but past a certain complexity it just becomes a huge headache, just like it would be if you tried to solve the problem in autoconf. In the Stan monorepo, I used Conan to handle dependencies, and in the past there I have also discussed using Bazel because of the upcoming transition the OcaML stan compiler (currently it uses Nix which unfortunately has nonexistent Windows support). It might be worth your time to check Conan out.

SteveBronder:
> CMake is not a package manager. Some projects get away with using it as one, but past a certain complexity it just becomes a huge headache, just like it would be if you tried to solve the problem in autoconf. In the Stan monorepo, I used Conan to handle dependencies, and in the past there I have also discussed using Bazel because of the upcoming transition to the OcaML-based stan compiler (currently it is built using Nix which unfortunately has nonexistent Windows support). It might be worth your time to check Conan out.

What I'm confused by is, what is cmake for then? Is it just for linking to headers and static / dynamic libraries? tmk I think all of our dependencies use cmake, so I don't really get why it's difficult to tell our cmake, "here is each repo we depend on, here are their cmake files, when you need to build tests for Stan Math we need the headers from X and a local dynamic library of Y."

alashworth:
CMake is a cross-platform replacement for Make and Autotools. I'm sure you're aware of how many hoops you have to jump through to get anything to work on Windows when you use Linux tooling. There's a lot of projects that just surrender and say: "better install a Linux virtual machine like WSL if you want things to work". Config, linking, filesystem paths, all of that works differently on Linux vs Windows.

If you ask me why CMake is popular, it's because a lot of C++ projects need to build with MSVC on Windows and GCC on Linux, and it transparently handles cross-platform and cross-compiler issues.

I don't quite follow your last paragraph where you're having issues with CMake - these things should be possible, but I don't recommend them. You will encounter pain when dealing with optional dependencies like MPI, and so will users if they want to use their own accelerated version of Sundials, for example, because they are not likely to want to edit the project Make/CMakefiles. That is the reason I recommend Conan as a cross-platform package manager.

syclik:
@alashworth, thanks for the comparison of CMake and Conan!

SteveBronder:
Yes @alashworth much appreciated! What your saying makes 100% sense, I think I'm just old and at trying to learn something new.

Looking at the cmake files of our depedencies it seems like what I mentioned above should be possible, but I was going about it trying to use our current downloads of the dependencies or hacking around getting particular git tags and then getting confused when it couldn't find the right cmake files. 

I have this branch up with the downloads setup using the `FetchContent` module. It's pretty neat! you just tell it the git repo and tag you want and it downloads those for you. Then if you actually need to compile and install things you can do it in the build repo (I think, everything I've tried so far is just a header library) if you have anytime to look at it that would be much appreciated! If I can get the gtest stuff up and running then I'll consider this a very good step 1 success

https://github.com/stan-dev/math/compare/trying/cmake-deps

alashworth:
> Yes @alashworth much appreciated! What your saying makes 100% sense, I think I'm just old and at trying to learn something new.
> 
> Looking at the cmake files of our depedencies it seems like what I mentioned above should be possible, but I was going about it trying to use our current downloads of the dependencies or hacking around getting particular git tags and then getting confused when it couldn't find the right cmake files.
> 
> I have this branch up with the downloads setup using the `FetchContent` module. It's pretty neat! you just tell it the git repo and tag you want and it downloads those for you. Then if you actually need to compile and install things you can do it in the build repo (I think, everything I've tried so far is just a header library) if you have anytime to look at it that would be much appreciated! If I can get the gtest stuff up and running then I'll consider this a very good step 1 success
> 
> https://github.com/stan-dev/math/compare/trying/cmake-deps

Yes, `fetch_content` is a good solution for 90% of the time. Now, I haven't looked at Stan's deps since last year, but the biggest headache then was using CMake to handle patching of Sundials and dealing with various MPI bindings on Linux. In the end you can do everything in CMake as at the end of the day you can do anything in a shell script, but it just wasn't a clean solution.

If you want to save some time I would really recommend the book from [Crascit](https://crascit.com/professional-cmake/) - it is written by a core CMake developer. Then also watch [video 1](https://www.youtube.com/watch?v=bsXLMQ6WgIk) and [video 2](https://www.youtube.com/watch?v=eC9-iRN2b04).

mitzimorris:
regarding GNU-make vs. alternatives, it has to work on all supported platforms.  Windows being the biggest headache, but not the only one.

rok-cesnovar:
Before we dive in too deep are we even sure the sanitizer error is "legit". Has it manifested in an actual bug?

wds15:
...and maybe we move to a different thread? Possibly on discourse?

SteveBronder:
Yeah I think we'd need a much bigger discussion before thinking more about this move. After messing around with it I like the `FetchContent` stuff for managing / building dependencies. You give cmake the url and tag you want and can supply patches on top of that etc. Looking through things it seems like once we wrote this stuff it would be easy to bump up versions of dependencies.

But it is kind of confusing. It's yet another thing new contributors have to learn and we need to spend time documenting how the cmake system would work, change our python scripts to run in the build directory with the new stuff, and have people who can manage the whole thing. The make stuff can be a bit wugh but we already have spent the time catching a lot of our corner cases and setting up the build dependencies properly along with scripts to update our dependencies and apply patches. I also don't know how we can do things like the jumbo tests in cmake. The version of cmake I'm using is also rather new and I'm not sure it would be available for older systems like the older rhel stuff. Without the nice FetchContent module stuff I think managing dependencies would be a lot grosser looking. It's hard to judge whether it's worth the time it would take to move over.


After messing around with it I realized, instead of moving totally to cmake why don't we just call the tbb cmake stuff from our current make script? That would make cmake a dependency but it seems pretty easy to install and the tbb supports a pretty old version available by default on linux comps. I can test whether just moving up the tbb version would make this error go away and then if so we can talk to RcppParallel about how to handle that.

> Before we dive in too deep are we even sure the sanitizer error is "legit". Has it manifested in an actual bug?

I'll look into this more, maybe it's not actually a bug. It was kind of confusing when reading the threads on this in the onetbb issue tracker.

> Yes, fetch_content is a good solution 90% of the time. It falls down under three cases: 1) can't handle binary dependencies, like hypothetical ocaml stan compiler,

@alashworth can't the binary for the new compiler be brought in using something like the example from cmake [here](https://cmake.org/cmake/help/latest/module/FetchContent.html#examples)

```cmake
include(FetchContent)
FetchContent_Declare(
  stanc3
  URL  https://stanc.url/binary_stuff
)
FetchContent_MakeAvailable(stanc3)

```

alashworth:
OK, you got me 👍, hah. You can use `fetch content` as a wrapper around curl. And it will work for simple dependency management, just like git submodule or shelling out to curl.

wds15:
@SteveBronder have a look: https://github.com/stan-dev/math/pull/2445

SteveBronder:
Noice!

So good news, updating tbb to 2020.3 works with both the new and old versions of `task_scheduler_init` and `tbb::task_arena` so we should be able to upgrade RcppParallel. Using the `TBB_INTERFACE_NEW` with the same amount of threads as chains the memory leak for in the stack allocator go away!

Bad news, if we increase the number of threads the leaks come back again :-(. I might have an answer to this, but they also happen at the end of the program. I'm inclined to believe this is not actually a "real leak" as all that happens is that the thread local stack allocator is not deleted at the end of the program. @wds15 this seems like odd behavior to me though, it implies for some reason having the exact number of threads allows it to be cleaned up while having more causes a whoopsie.

SteveBronder:
Link to updated tbb math

https://github.com/stan-dev/math/tree/feature/update-tbb2020-3

@hsbadr is it pretty reasonable to apply the windows patches for RcppParallel? If so it might be nice to bump up the version. Then in the future we can talk about how to do a hard deprecation wrt `task_scheduler_init` and newer versions of tbb

SteveBronder:
And for replication purposes I have in my make/local

```
CXX=clang++-11
CXXFLAGS+= -DTBB_INTERFACE_NEW -ggdb3 -fsanitize=address,undefined,leak -fno-omit-frame-pointer
O=g
STAN_THREADS=true
```

and am doing 

```
cd lib/stan_math
git pull
git checkout feature/update-tbb2020-3
cd  ../../
make clean-all
# cool and good
STAN_NUM_THREADS=3 ./runTests.py -j25 ./src/test/unit/services/util/run_adaptive_sampler_parallel_test.cpp
# bad :-(
STAN_NUM_THREADS=4 ./runTests.py -j25 ./src/test/unit/services/util/run_adaptive_sampler_parallel_test.cpp
```

SteveBronder:
Alright so after looking around I found some [docs](https://spec.oneapi.com/versions/latest/elements/oneTBB/source/task_scheduler/task_arena/task_scheduler_observer_cls.html#_CPPv417on_scheduler_exitb) and this [thread](https://community.intel.com/t5/Intel-oneAPI-Threading-Building/task-scheduler-observer-on-scheduler-exit-not-called/td-p/854736) which describes the behavior here and I think there's enough evidence to say that, for the version that uses the new tbb, this is a leak but not a dangerous leak

Cleanup for the ad tape in each thread normally happens from the global [`ad_type_observer`](https://github.com/stan-dev/math/blob/develop/stan/math/rev/core/init_chainablestack.hpp#L50)'s `on_scheduler_exit()` method. But in the tbb docs and the above thread it's stated that 

> A process does not wait for the worker threads to clean up, and can terminate before on_scheduler_exit is invoked.

So what's happening is that before `on_scheduler_exit()` can be called for a thread, the program is ending and the leak detector is getting confused as it sees memory that exists at end of program but is never cleaned up. Because the threads are never cleaned up it throws these leak warnings, but since the main thread is terminated there's no reason to care that the other threads are cleaned up. Since the tests run fine and the leak only happens after the tests end I feel pretty confident that this is not a true leak.

But the other memory error with our current version of the tbb about the vptr being incorrect I think is an actual bug inside of that version of the tbb. So I think if we bump up the tbb version to 2020-3 (which uses the makefiles still so no cmake) and use `TBB_INTERFACE_NEW` by default then everything here can be okie dokie! I have a PR to do that [here](https://github.com/stan-dev/math/tree/feature/update-tbb2020-3) I can open up tmrw

SteveBronder:
> One issue with the model model prng generators is that they control both the transformed data and the generated quantities blocks.  While there are cases where a common generator for the transformed data makes sense there are fewer were a common generator for generated quantities makes sense.  Indeed I would rather have different generators across the model replications than a common one — it’s safer, more appropriate to simulation calibration methods, and if anyone does want common data it’s easy enough to split the process up into two steps — generating common data with the fixed_param sampler and then passing that to the chains.

So the good news is that for the generated quantities the signature actually supplies an RNG, so for that the current impl I have with multiple PRNGs should be alrighty. Multiple PRNG for the transformed data might have to come some other time. One nice benefit here is that the single model is shared across all the processes which cuts out memory overhead, but if we want multiple PRNG for the transformed data then we'll also have to make multiple models to have different transformed data. Maybe we could have an option at the cmdstan level to support sbc which would then give each thread it's own model? Though I think we should do that another time as the stuff here so far seems like a nice layup (I need to read the stuff you and @t4c1 posted but so far I think my concerns were overblown and we should be okay).

hsbadr:
> is it pretty reasonable to apply the windows patches for RcppParallel?

I think so. Just need to test it on Windows.

> If so it might be nice to bump up the version. Then in the future we can talk about how to do a hard deprecation wrt `task_scheduler_init` and newer versions of tbb

That sounds like a great transition to the new TBB interface, with `2020.3`.

SteveBronder:
Awesome! Just put up a pr with the new tbb version in math

hsbadr:
> Awesome! Just put up a pr with the new tbb version in math

Great! It'd be optimal to update TBB in `RcppParallel` to `2020.3` as well.

src/stan/mcmc/hmc/hamiltonians/softabs_metric.hpp
```diff
@@ -174,26 +174,26 @@ class softabs_metric : public base_hamiltonian<Model, softabs_point, BaseRNG> {
...
   // Threshold above which an asymptotic
   // approximation of the softabs function is used
-  static double upper_softabs_thresh;
+  static constexpr double upper_softabs_thresh = 18;
 
   // Threshold below which an exact derivative is
   // used in the Jacobian calculation instead of
   // finite differencing
-  static double jacobian_thresh;
+  static constexpr double jacobian_thresh = 1e-10;
```
SteveBronder, line 186: fyi I just did a ctrl+f for any objects that were static in the samplers and made them constexpr so we don't get a hiccup in the future

SteveBronder:
Alright I think this is ready for review! I'm going to open up another PR on top of this one soon that has the signature for nuts with a diagonal metric

bbbales2:
> another PR on top

Is this a cmdstan PR or is there something else in stan that needs changed?

SteveBronder:
This is just a Stan PR, once #3033 goes in we can do the cmdstan stuff.

For review no, but it would be a good idea to wait to merge until the tbb pr is in

bbbales2:
> This is just a Stan PR, once #3033 goes in

Okay so this is a services function, and then we'll do a samplers pull, and then we'll move on to cmdstan?

SteveBronder:
Yep!  #3033 is the sampler pull

src/stan/services/util/generate_transitions.hpp
```diff
@@ -44,14 +44,17 @@ void generate_transitions(stan::mcmc::base_mcmc& sampler, int num_iterations,
...
+                          callbacks::logger& logger, size_t n_chain = 0) {
   for (int m = 0; m < num_iterations; ++m) {
     callback();
 
     if (refresh > 0
         && (start + m + 1 == finish || m == 0 || (m + 1) % refresh == 0)) {
       int it_print_width = std::ceil(std::log10(static_cast<double>(finish)));
       std::stringstream message;
+      if (n_chain > 0) {
+        message << "Chain [" << (n_chain + 1) << "]";
```
bbbales2, line 56: This makes sense cause I think cmdstanr/cmdstanpy append similar things, but let's check with the format.

src/stan/callbacks/file_stream_writer.hpp
```diff
@@ -0,0 +1,117 @@
...
+   * @param[in, out] A unique pointer to a type inheriting from `std::ostream`
+   * @param[in] comment_prefix string to stream before each comment line.
+   *  Default is "".
+   */
+  explicit file_stream_writer(std::unique_ptr<std::ostream>&& output,
+                              const std::string& comment_prefix = "")
+      : output_(std::move(output)), comment_prefix_(comment_prefix) {}
+
+  file_stream_writer();
+  file_stream_writer(file_stream_writer& other) = delete;
```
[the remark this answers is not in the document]
bbbales2: Oh okay I looked back at what you previously wrote:

```
So that file_stream_writer holds unique pointers to each of the file streams. I had some
issues using stream_writer where somewhere a reference was being dereferenced or
something was falling out of scope (stream_writer just holds it's ostream by refrerence)
so I thought the easiest solution was just to wrap that ostream into an std::unique_ptr.
Then we also get promises that no-one ever copies that thing.
```

It seems cleaner to replace stream_writer with something that behaves better. Presumably the other use cases of stream_writer can still work.

src/stan/services/util/run_adaptive_sampler.hpp
```diff
@@ -87,6 +88,83 @@ void run_adaptive_sampler(Sampler& sampler, Model& model,
...
+          typename DiagnoseT>
+void run_adaptive_sampler(std::vector<Sampler>& samplers, Model& model,
+                          std::vector<std::vector<double>>& cont_vectors,
+                          int num_warmup, int num_samples, int num_thin,
+                          int refresh, bool save_warmup, std::vector<RNG>& rngs,
+                          callbacks::interrupt& interrupt,
+                          callbacks::logger& logger,
+                          std::vector<SampT>& sample_writers,
+                          std::vector<DiagnoseT>& diagnostic_writers,
+                          size_t n_chain) {
```
bbbales2, line 102: Alright so the model, interrupt, and loggers passed into this all need to be threadsafe.

Is every function in the model threadsafe (I guess other than constructor, destructor)? And I would check this by inspecting a compiled model.hpp?

I see the test interrupt and logger have been made threadsafe.

It seems like as part of this pull there'd be a bunch of other interrupt/logger non-test code making things threadsafe. Is that not necessary for some reason? Or is it already threadsafe?

src/stan/services/util/generate_transitions.hpp
```diff
@@ -44,14 +44,17 @@ void generate_transitions(stan::mcmc::base_mcmc& sampler, int num_iterations,
...
+      }
```
bbbales2, line 57: This makes sense but we can run it by interfaces.

src/stan/callbacks/file_stream_writer.hpp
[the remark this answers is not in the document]
bbbales2: Oh I see your previous comment on this:

```
So that file_stream_writer holds unique pointers to each of the file streams. I had some
issues using stream_writer where somewhere a reference was being dereferenced or
something was falling out of scope (stream_writer just holds it's ostream by refrerence)
so I thought the easiest solution was just to wrap that ostream into an std::unique_ptr.
Then we also get promises that no-one ever copies that thing.
```

Can we clean up the stream_writer implementation though instead of adding another thing that does something very similar?

src/stan/services/util/run_adaptive_sampler.hpp
bbbales2, line 102: So the model, interrupt, and logger bits need to be threadsafe.

Is every function in the model (other than constructors/destructors) thread safe? I think this threading stuff will mess up the profiles, so we gotta figure out what to do there.

For the interrupts and loggers, I see that the implementations in the test had to change but shouldn't there be changes in the main source too? Or were these already threadsafe or something?

I see that interrupt is for the interfaces to capture keystrokes. Are the interrupts signals generated at the interfaces and transmitted in here -- or are they signals generated in the samplers and transmitted back to the interfaces?

src/stan/services/util/generate_transitions.hpp
bbbales2, line 56: Makes sense to me but we should check with interfaces on this.

bbbales2 (changes requested):
Left some comments. I'm gonna go read the sampler PR. Gotta piece more of this together.

bbbales2:
Computer crashed mid-review here. I came back and comments were gone, so I added them again. It seems like in the end some of the comments got doubled. Sorry for the noise. I cleaned up duplicates.

src/stan/callbacks/file_stream_writer.hpp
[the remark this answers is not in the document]
SteveBronder: I'd like to replace stream writer, but I was having trouble with backwards compatibility. I think once cmdstan uses the `file_stream_writer` we can remove `stream_writer`

src/stan/services/util/run_adaptive_sampler.hpp
SteveBronder: So the logger goes to `std::cout` which is threadsafe (see [here](https://en.cppreference.com/w/cpp/io/cout))

>> Unless sync_with_stdio(false) has been issued, it is safe to concurrently access these objects from multiple threads for both formatted and unformatted output. 

The interrupts, at least in cmdstan, are actually just the base impl `stan::callbacks::interrupt` which doesn't actually do anything. I'm looking at the [R interrupt](https://rdrr.io/github/HenrikBengtsson/RNativeAPI/man/R_CheckUserInterrupt.html) and it seems like that would be threadsafe. 

> Is every function in the model (other than constructors/destructors) thread safe?

Yep! I've looked the model class over. The model is essentially immutable after construction so it's safe to pass around.

>  I think this threading stuff will mess up the profiles, so we gotta figure out what to do there.

Oooh hm idk how this will effect profiling. @rok-cesnovar looking at the profiling gen code

```
      {
        profile<local_scalar_t__> profile__("testerrr", const_cast<profile_map&>(profiles__));
        
        current_statement__ = 3;
        assign(transf_testp, stan::math::exp(testp),
          "assigning variable transf_testp");
      }
```

I think we might need a mutex in the profile constructor here

```cpp
  profile(std::string name, profile_map& profiles)
      : key_({name, std::this_thread::get_id()}) {
    {
      std::lock_guard<std::mutex> map_lock(profile_mutex);
     // idt search is thread safe
      profile_map::iterator p = profiles.find(key_);
      if (p == profiles.end()) {
        // idt insertion is thread safe either
        profiles[key_] = profile_info();
    }
    // rest business as usual
    }
```

> I see that interrupt is for the interfaces to capture keystrokes. Are the interrupts signals generated at the interfaces and transmitted in here -- or are they signals generated in the samplers and transmitted back to the interfaces?

They are called in [`generate_transitions`](https://github.com/stan-dev/stan/blob/develop/src/stan/services/util/generate_transitions.hpp#L49) which is for the keystroke thing for R / Python / higher level systems that need to tell the C++ to stop. It's not used in cmdstan since you can just call ctrl+c on a binary to tell the system to stop whatever program is running atm.

src/stan/services/util/generate_transitions.hpp
SteveBronder: Yeah anyone that is depending on the current behavior shouldn't be effected by this so I think it's fine

src/stan/services/util/run_adaptive_sampler.hpp
```diff
@@ -87,6 +88,83 @@ void run_adaptive_sampler(Sampler& sampler, Model& model,
...
+                         num_samples, num_thin, refresh, save_warmup, rngs[0],
+                         interrupt, logger, sample_writers[0],
+                         diagnostic_writers[0]);
+  }
+  tbb::parallel_for(
+      tbb::blocked_range<size_t>(0, n_chain, 1),
+      [num_warmup, num_samples, num_thin, refresh, save_warmup, &samplers,
+       &model, &rngs, &interrupt, &logger, &sample_writers, &cont_vectors,
+       &diagnostic_writers](const tbb::blocked_range<size_t>& r) {
+        for (size_t i = r.begin(); i != r.end(); ++i) {
```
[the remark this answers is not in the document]
SteveBronder: So, I didn't realize this until I started working on #3033 lol. Yeah I think that would work? Let me try to do that in #3033. Though I will say I kind of would like to keep this code only as an example of how to do `parallel_for()` at this level which is what the auto warmup size thing will need to do. For the auto warmup stuff you mostly just need to take this code, break it up into two parallel for loops, then add another loop around the warmup `parallel_for()` to do the warmup in chunk sizes.

src/stan/callbacks/file_stream_writer.hpp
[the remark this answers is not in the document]
SteveBronder: Yeah that's the main issue

src/stan/services/util/run_adaptive_sampler.hpp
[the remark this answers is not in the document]
SteveBronder: Alright so just ran this locally and it gives back good answers. If we are going to cut this code out then what I can actually do it remove the changes to run_adaptive_sampler in #3033 and just do the simpler parallel loop

src/stan/services/util/run_adaptive_sampler.hpp
wds15: Is it an option to run the profiler in only one thread or avoid the need for a mutex by using multiple model copies? We do not want a mutex unless we have to.

bbbales2: I think we want separate profiles for each thread if we want profiling to act like it currently does (a separate profile for each chain). Is Stan data actually copied into the model? Or is it references? I guess if it's references then we could actually have lots of models? But maybe that is undesirable.

rok-cesnovar: Do we need a mutex? The thread id is a key in the storage map and each thread is thus guaranteed to always access only its "own" profiles. `std::map` is supposed to be fine in that case: https://stackoverflow.com/questions/15067160/stdmap-thread-safety and other resources if you google "std::map thread safety".

We could switch to using tbb::concurent_hash_map though I think that is unecessary.

SteveBronder: The profiler itself runs in one thread, the problem is that in the above constructor the `profile_map` is a global `std::map` and idt there's any reasonable way to get around that. 

imo idt a mutex is that bad here. I think we can actually put it in the `if (p == profiles.end())` since `find` should be threadsafe since it's searching for a key that also has the thread_id in it. But for insertion with `profiles[key_]` we do not want to be adding keys simultaneously as we could end up in a situation where the map reallocs memory


One solution here is to prefill the profile_map keys. Since we know each key is decided by a unique name and thread combination no two threads can cause a race condition because they can write to different parts of the map. So we only need to worry about the map reallocing for new keys. for an example model like  


```stan
model {
  profile("ayyy") {
    theta ~ beta(1,1);  // uniform prior on interval 0,1
  }
  profile("yooo") {
    y ~ bernoulli(theta);    
  }
}
```

We can have a function in the model like 

```
void init_profile_map(){
   const auto max_threads = get_max_threads();
   for (auto&& id : std::array<const char*, 2>{"ayyy", "yooo")) {
     for (int i = 0; i < max_threads; ++i) {
       profile_map[{name, std::this_thread::get_id()}] = empty_profile_info();
     }
   } 
}
```

Or something like that to fill up the keys before we do any other insertions. Then we don't need to worry about data reallocation since all the keys are there.

But still I think a mutex here is not that bad. It's short lived and only causes locks whenever we are doing profiling (which imo is a usage pattern that is not for performance anyway). I think what I would like is to just put a mutex in the if for now and then later we can do the lock free pattern

SteveBronder: Oh shoot I typed that out before Rok posted his reply. Yeah from there it looks like it is safe to insert since each key is distinct for each thread. I'll look into this more to make sure it's fine. If I can't sort out anything conclusive then I think moving to a ` tbb::concurent_hash_map` is a reasonable solution

wds15: Global variables are problematic for obvious reasons. The TBB data structures should be a good alternative... How often is this code being run over? I mean, is it performance critical?

rok-cesnovar: Insert to a map happens the first time a profile is registered. So once in a lifetime of a model. After that its just read access to get the reference to profile info. And those are separated for separate threads, no race conditions there.

If locking the insert in a mutex is something that would make us sleep more calmly, thats not a big price to pay. It does seem its not needed.

SteveBronder: Yeah idt it's needed. Let's not worry about this for now and we should notice it with a pretty simple test once the cmdstan stuff is up

SteveBronder:
fyi I'm going to close this PR in lieu of #3033 because of @bbbales2 comment [here](https://github.com/stan-dev/stan/pull/3028#discussion_r601448388) I realized we actually don't need to do parallelism at this level and can just wrap `run_adaptive_sampler` in a `parallel_for`. Much simpler!

PR #3030, opened 2021-03-22
Check for overflow in metric adaptation

### LINKED ISSUE #3029: `base_hmc.init_stepsize()` loops forever if stepsize is NaN
#### Summary:
Run this model with `num_warmup=2000`
```stan
parameters {
  real x;
}
model {
  target += -log_sum_exp(0, -x);
}
```
It hangs after ~1900 warmup samples and never terminates.
The problem is that after the first ~800 samples metric overflows to infinity. After that stepsize becomes `NaN` (and the sampler is completely stuck). When the last adaptation window finishes the sampler updates the metric and calls `base_hmc.init_stepsize()` but that method does not handle `NaN` correctly and causes an infinite loop.

#### Current Output:
Process hangs.

#### Expected Output:
Error that says the posterior is improper.

#### Current Version:
v2.26.1

nhuurre:
#### Submission Checklist

#### Summary

Fix #3029 
Check that metric does not overflow during adaptation and ensure that `base_hmc.init_stepsize()` errors properly if `stepsize=NaN`.

#### Intended Effect

More robust detection of improper posteriors.

#### Copyright and Licensing

Copyright holder: Niko Huurre

### DIFF
--- a/src/stan/mcmc/covar_adaptation.hpp
+++ b/src/stan/mcmc/covar_adaptation.hpp
@@ -28,6 +28,13 @@ class covar_adaptation : public windowed_adaptation {
               + 1e-3 * (5.0 / (n + 5.0))
                     * Eigen::MatrixXd::Identity(covar.rows(), covar.cols());
 
+      if (!covar.allFinite())
+        throw std::runtime_error(
+            "Numerical overflow in metric adaptation. "
+            "Posterior is too wide for Stan to estimate it. "
+            "This problem may indicate that the posterior is improper. "
+            "Please check your model.");
+
       estimator_.restart();
 
       ++adapt_window_counter_;
--- a/src/stan/mcmc/covar_adaptation.hpp
+++ b/src/stan/mcmc/covar_adaptation.hpp
@@ -31,9 +31,10 @@ class covar_adaptation : public windowed_adaptation {
       if (!covar.allFinite())
         throw std::runtime_error(
             "Numerical overflow in metric adaptation. "
-            "Posterior is too wide for Stan to estimate it. "
-            "This problem may indicate that the posterior is improper. "
-            "Please check your model.");
+            "This occurs when the sampler encounters extreme values on the "
+            "unconstrained space; this may happen when the posterior density "
+            "function is too wide or improper. "
+            "There may be problems with your model specification.");
 
       estimator_.restart();
 
--- a/src/stan/mcmc/hmc/base_hmc.hpp
+++ b/src/stan/mcmc/hmc/base_hmc.hpp
@@ -127,7 +127,7 @@ class base_hmc : public base_mcmc {
         this->nom_epsilon_ = direction == 1 ? 2.0 * this->nom_epsilon_
                                             : 0.5 * this->nom_epsilon_;
 
-      if (this->nom_epsilon_ > 1e7)
+      if (!(this->nom_epsilon_ < 1e7))
         throw std::runtime_error(
             "Posterior is improper. "
             "Please check your model.");
@@ -139,6 +139,13 @@ class base_hmc : public base_mcmc {
     }
 
     this->z_.ps_point::operator=(z_init);
+  }
+
+  void metric_check(bool ok) {
+    if (!ok)
+      throw std::runtime_error(
+          "Posterior is improper. "
+          "Please check your model.");
   }
 
   /**
--- a/src/stan/mcmc/hmc/base_hmc.hpp
+++ b/src/stan/mcmc/hmc/base_hmc.hpp
@@ -82,7 +82,8 @@ class base_hmc : public base_mcmc {
     ps_point z_init(this->z_);
 
     // Skip initialization for extreme step sizes
-    if (this->nom_epsilon_ == 0 || this->nom_epsilon_ > 1e7)
+    if (this->nom_epsilon_ == 0 || this->nom_epsilon_ > 1e7
+        || std::isnan(this->nom_epsilon_))
       return;
 
     this->hamiltonian_.sample_p(this->z_, this->rand_int_);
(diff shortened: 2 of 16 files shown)


src/stan/mcmc/hmc/base_hmc.hpp
```diff
@@ -127,7 +127,7 @@ class base_hmc : public base_mcmc {
         this->nom_epsilon_ = direction == 1 ? 2.0 * this->nom_epsilon_
                                             : 0.5 * this->nom_epsilon_;
 
-      if (this->nom_epsilon_ > 1e7)
+      if (!(this->nom_epsilon_ < 1e7))
```
betanalpha, line 130: This inequality change is incorrect.  The current warning is meant to capture cases where the gradient vanishes (which occurs when the posterior density is flat, which will be encountered with probability > 0 only if the posterior is improper in at least some neighborhoods), _any_ stepwise will satisfy the initialization condition, and the heuristic initialization will result in an extremely large step size.

src/stan/mcmc/hmc/base_hmc.hpp
```diff
@@ -141,6 +141,13 @@ class base_hmc : public base_mcmc {
     this->z_.ps_point::operator=(z_init);
   }
 
+  void metric_check(bool ok) {
```
betanalpha, line 144: The error message is incorrect here.  Because the metric adaptation is convex the metric elements can become NaN only if the unconstrained states are NaN.  That in turn can happen from underflow or overflow in the gradient evaluation or even divergent numerical trajectories, which are both distinct  from the conditions of an improper posterior density where the gradient vanishes identically.

src/stan/mcmc/hmc/nuts/adapt_dense_e_nuts.hpp
```diff
@@ -33,6 +33,7 @@ class adapt_dense_e_nuts : public dense_e_nuts<Model, BaseRNG>,
           this->z_.inv_e_metric_, this->z_.q);
 
       if (update) {
+        this->metric_check(this->z_.inv_e_metric_.allFinite());
```
betanalpha, line 36: The name here is misleading -- `check_metric` should be changed to something like `check_metric_condition` or just accept the inverse metric by reference and check allFinite itself.

src/stan/mcmc/hmc/base_hmc.hpp
nhuurre: I suppose it would make more sense to instead change the check that bails out at the start of the function
https://github.com/stan-dev/stan/blob/37ad8c36b83299d04fd9da824d324fa6517fc6e0/src/stan/mcmc/hmc/base_hmc.hpp#L84-L86

src/stan/mcmc/hmc/base_hmc.hpp
nhuurre: In the example model given on the issue the _metric_ does not become NaN but _infinite_ which makes `accept_stat` and `stepsize` into `NaN`, and that happens at finite value of unconstrained state. (The square of a sufficiently large finite value is infinite.)
It's true that technically this can happen without the posterior being improper (and improper posterior does not require the gradient to vanish) but I figured that if the distribution is so wide that floating-point arithmetic fails then we might as well pretend it reaches all the way out to infinity.

src/stan/mcmc/hmc/nuts/adapt_dense_e_nuts.hpp
nhuurre: True enough. I wasn't sure what to call it--I just wanted to put error message in one place. Most logical would be to accept the metric by reference but `diag_e` and `dense_e` are `Eigen::VectorXd` and `Eigen::MatrixXd` and I'm not sure you can mix those...I guess it could be a virtual method on `ps_point`?

src/stan/mcmc/hmc/base_hmc.hpp
betanalpha: The extreme step size occurs _during_ the initialization, hence the check immediately afterwards.

src/stan/mcmc/hmc/base_hmc.hpp
betanalpha: Yes, I am treating NaN/inf as the same for diagnostic purposes.

The problem with "improper" is that is can never be fully verified empirically.  The original warning was designed for flat posterior densities functions because of the propensity of early users to use uniform prior densities with nonidentified likelihood functions, where the flat posterior density identified the flat ledge that (was assumed to) roll all the way to infinity.   Most common improper prior densities are asymptotically flat and eventually encounter the same behavior.

The example here is different because the target density is worse than flat.  But we have no way of verifying if the problem will continue to infinity (and actually be improper) because the floating point falls apart and we don't have an interpretation like the above to motivate a more theoretical extrapolation.

We can avoid the ambiguity entirely by just being precise about the warning -- if the inverse metric elements are NaN then report that they NaN'd and note that should be caused by NaN states, and if they're infinite report that they're infinite and suggest that is probably caused by extremely large states.

src/stan/mcmc/hmc/nuts/adapt_dense_e_nuts.hpp
betanalpha: That would be the most natural for the current OOP design.

betanalpha (changes requested):

src/stan/mcmc/covar_adaptation.hpp
```diff
@@ -28,6 +28,13 @@ class covar_adaptation : public windowed_adaptation {
               + 1e-3 * (5.0 / (n + 5.0))
                     * Eigen::MatrixXd::Identity(covar.rows(), covar.cols());
 
+      if (!covar.allFinite())
+        throw std::runtime_error(
+            "Numerical overflow in metric adaptation. "
+            "Posterior is too wide for Stan to estimate it. "
```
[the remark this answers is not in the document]
nhuurre: Done.

betanalpha (approved):

betanalpha:
Thanks!

PR #3033, opened 2021-03-24
Parallel Adaptive Nuts

SteveBronder:
#### Submission Checklist

#### Summary

Okay so this PR has the cmdstan facing version of `hmc_nuts_diag_e_adapt()` (and I plan to also do dense in this PR). The signature differs from the original `hmc_nuts_diag_e_adapt()` in that the `init`, `init_inv_metric`, `sample_writer`, and `diagnostic_writer` are std::vector<>`'s of what the normal signature is.

My suggestion is that for cmdstan we should have all of these be std::vectors when they are created. Then for the algorithms that are not parallel yet we set `vec_argument[0]` as the input argument for the algorithm. That will let us get diag/dense matric adaptive nuts into cmdstan and then start adding parallel versions of all the other algorithms. Does that work for everyone?

#### Intended Effect

Allow multiple chains to be invoked for adaptive nuts with diag/dense metrics

#### How to Verify

Test added for parallel diag e adapt

```
STAN_NUM_THREADS=4 ./runTests.py ./src/test/unit/services/sample/hmc_nuts_diag_e_adapt_parallel_test.cpp
```

#### Documentation

Still need to add docs

#### Copyright and Licensing


src/stan/services/util/create_unit_e_dense_inv_metric.hpp
```diff
@@ -17,17 +17,12 @@ namespace util {
...
+                       "", "", "inv_metric <- structure(c(", dims);
   std::stringstream txt;
-  txt << "inv_metric <- structure(c(";
-  for (size_t i = 0; i < num_elements; i++) {
-    txt << inv_metric(i);
-    if (i < num_elements - 1)
-      txt << ", ";
-  }
-  txt << "),.Dim=c(" << num_params << ", " << num_params << "))";
+  txt << Eigen::MatrixXd::Identity(num_params, num_params).format(RFmt);
```
bbbales2, line 25: These changes seem unrelated. Nothing wrong with quality of life changes, but it does make it harder to review these pulls cause I gotta check these side things.

src/stan/services/sample/hmc_nuts_diag_e_adapt.hpp
```diff
@@ -53,20 +53,20 @@ namespace sample {
...
  * @return error_codes::OK if successful
  */
-template <class Model>
+template <typename Model, typename InitContext, typename InitMetricContext,
+          typename SampleWriter, typename DiagnosticWriter>
 int hmc_nuts_diag_e_adapt(
-    Model& model, const stan::io::var_context& init,
-    const stan::io::var_context& init_inv_metric, unsigned int random_seed,
+    Model& model, const InitContext& init,
+    const InitMetricContext& init_inv_metric, unsigned int random_seed,
```
bbbales2, line 60: Are the types changing here? I don't see an advantage to templating this. Now the template hides even the base polymorphic type.

src/stan/services/sample/hmc_nuts_diag_e_adapt.hpp
```diff
@@ -133,27 +133,120 @@ int hmc_nuts_diag_e_adapt(
...
+        samplers[i].get_stepsize_adaptation().set_gamma(gamma);
+        samplers[i].get_stepsize_adaptation().set_kappa(kappa);
+        samplers[i].get_stepsize_adaptation().set_t0(t0);
+        samplers[i].set_window_params(num_warmup, init_buffer, term_buffer,
+                                      window, logger);
+      }
+    } catch (const std::domain_error& e) {
+      return error_codes::CONFIG;
+    }
+    util::run_adaptive_sampler(samplers, model, cont_vectors, num_warmup,
```
bbbales2, line 210: Similarly to the other code, I don't like how the 1 chain and N chain implementations are separate. It seems like they could be the same code and the 1 chain thing would just have 1 chain and that's fine.

bbbales2 (changes requested):
Comments. I don't like having the 1 chain/n-chain code paths be separate.

src/stan/services/util/create_unit_e_dense_inv_metric.hpp
SteveBronder: Yeah sorry about that, was just looking through the function we use here and saw this one could be cleaned up a little bit. I'll avoid this in the future

src/stan/services/sample/hmc_nuts_diag_e_adapt.hpp
SteveBronder: Ah they are! Notice that now when an initial inverse metric is not given we actually return back an `stan::io::dump` (or vector of). This is nice for the compiler because now it knows the real type coming in so no devirtualization happens a lot easier.

src/stan/services/sample/hmc_nuts_diag_e_adapt.hpp
SteveBronder: We have two right now mostly because of cmdstan. Once we add the parallel version to cmdstan we can remove 1 chain version

SteveBronder:
@bbbales2 I think this PR should have everything in it that we need to do the cmdstan stuff. Once we think the `hmc_nuts_diag_e_adapt` version looks good I can pretty easily update the other samplers and add tests for them

SteveBronder:
Also I updated the cmdstan branch for the parallel prototype in this branch

https://github.com/stan-dev/cmdstan/compare/proto/multi-chain

bbbales2 (dismissed):
Comments. Implementing this with overloads seems pretty simple and I like it.

Don't update the math submodule.

---- logging

I think rstan and cmdstan are wrapping stdout/stderr for their logging. That should be threadsafe, but I think we might need to add little synchronizers to make it so that each thread gets to print a full line before another thread does. Compared to the autodiff, I don't think this sorta I/O sync would be expensive.

Edit, what I'm scared about is output that looks like:
```
[chain 1] Iteration 0 / 10[chain 2] Iteration 100/1000 yada yada000 yada yada

[chain3]...
```

instead of:
```
[chain 1] Iteration 0 / 10000 yada yada
[chain 2] Iteration 100/1000 yada yada
[chain3]...
```

httpstan uses [this](https://github.com/stan-dev/httpstan/blob/main/httpstan/socket_logger.hpp) for its logger, and would need it to be threadsafe to work. (Edit: not that we need it now, since httpstan will just use the non-threading overload for now, was just investigating what would have to port)

---- interrupts

I don't think httpstan uses an interrupt. Cmdstan doesn't either. Appears to be exclusively a rstan thing, and that goes through R_checkUserInterrupt. This pdf mentions something about threadsafe versions of stdout and the checkUserInterrupt thing: https://cran.r-project.org/web/packages/RcppThread/vignettes/RcppThread-vignette.pdf . Anyway we could add our own thread-safeness around these things too.

---- profiler

I think the way we use the profiler is okay even if the profiler doesn't have chain information for each of the threads (which is different than it is now). I was thinking originally we might want to change that, but it still seems fine to say "these are the profile counters for each thread in a Stan program". Now a single Stan program has multiple chains -- but I think the counters remain useful in the same way they were previously.

---- file_stream_writer

If I get a chance I'm gonna investigate that file_stream_writer. That still bothers me adding a second think that's doing the same thing but maybe it's the only way.

src/stan/services/sample/hmc_nuts_diag_e_adapt.hpp
```diff
@@ -133,27 +154,133 @@ int hmc_nuts_diag_e_adapt(
...
-
   return hmc_nuts_diag_e_adapt(
-      model, init, unit_e_metric, random_seed, chain, init_radius, num_warmup,
+      model, init, dmp, random_seed, chain, init_radius, num_warmup,
       num_samples, num_thin, save_warmup, refresh, stepsize, stepsize_jitter,
       max_depth, delta, gamma, kappa, t0, init_buffer, term_buffer, window,
       interrupt, logger, init_writer, sample_writer, diagnostic_writer);
 }
 
+template <class Model, typename InitContext, typename InitInvContext,
```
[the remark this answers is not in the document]
bbbales2: I guess it's worth saying here model, interrupt, and logger should be thread-safe and saying that the `std::vector`s of things should all be the same length (equal to n_chains) and that they should be different.

src/stan/services/sample/hmc_nuts_diag_e_adapt.hpp
```diff
@@ -19,6 +19,26 @@ namespace stan {
 namespace services {
 namespace sample {
 
+namespace internal {
+/**
+ * Specializtion to get a const reference to the underlying value in a
+ *  shared_ptr.
+ */
+template <typename Context>
+inline const auto& get_context(const std::shared_ptr<Context>& x) {
```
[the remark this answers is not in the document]
SteveBronder: No so this is a weird thing I'm sorting out about cmdstan. Since cmdstan actually wraps the `var_context` for inits inside of a `shared_ptr<>` we need to access the underlying pointer. Because in cmdstan we are going to pass a `std::vector<shared_ptr<var_context>>` and the current version just passes the underlying of the shared pointer. Though I might be able to get rid of this so I'll check

src/stan/services/sample/hmc_nuts_diag_e_adapt.hpp
SteveBronder: Is a `var_context` more informative? It's just an abstract base class so you still need to go look at the callee to see what's happening when this objects member functions are called. What if I included in the docs that this is either going to be a `stan::io::dump` or `stan::io::json_context` etc.?

src/stan/services/sample/hmc_nuts_diag_e_adapt.hpp
```diff
@@ -133,27 +154,133 @@ int hmc_nuts_diag_e_adapt(
...
+int hmc_nuts_diag_e_adapt(
+    Model& model, const std::vector<InitContext>& init,
+    const std::vector<InitInvContext>& init_inv_metric,
+    unsigned int random_seed, unsigned int chain, double init_radius,
+    int num_warmup, int num_samples, int num_thin, bool save_warmup,
+    int refresh, double stepsize, double stepsize_jitter, int max_depth,
+    double delta, double gamma, double kappa, double t0,
+    unsigned int init_buffer, unsigned int term_buffer, unsigned int window,
+    callbacks::interrupt& interrupt, callbacks::logger& logger,
+    callbacks::writer& init_writer, std::vector<SampleWriter>& sample_writer,
```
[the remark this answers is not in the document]
SteveBronder: Good catch, in cmdstan this is just a `stan::io::writer` (which is an empty base class so everything it does is a no-op). If higher level stuff uses these to write to individual files then I'll make these into vectors as well

src/stan/services/sample/hmc_nuts_diag_e_adapt.hpp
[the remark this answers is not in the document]
SteveBronder: Yeah I'm going to add docs for all these this weekend. At first I was like meh but I thought about it and there's enough differences here to write docs

src/stan/services/sample/hmc_nuts_diag_e_adapt.hpp
```diff
@@ -133,27 +154,133 @@ int hmc_nuts_diag_e_adapt(
...
     callbacks::logger& logger, callbacks::writer& init_writer,
-    callbacks::writer& sample_writer, callbacks::writer& diagnostic_writer) {
+    SampleWriter& sample_writer, DiagnosticWriter& diagnostic_writer) {
   stan::io::dump dmp
       = util::create_unit_e_diag_inv_metric(model.num_params_r());
-  stan::io::var_context& unit_e_metric = dmp;
-
   return hmc_nuts_diag_e_adapt(
-      model, init, unit_e_metric, random_seed, chain, init_radius, num_warmup,
+      model, init, dmp, random_seed, chain, init_radius, num_warmup,
```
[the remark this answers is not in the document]
SteveBronder: Yep!

src/stan/services/sample/hmc_nuts_diag_e_adapt.hpp
bbbales2: > Is a var_context more informative

Well it's still easier to search for (you can find that it's an abstract base class, and then you can find what inherits from it and whatnot) and this is how everything is kinda done in the rest of services.

I think the reason we'd go templating here is if we were doing something we couldn't just solve with the existing polymorphism, or it's just easier to code the templated thing.

If that's the case, let's do templating. If that's not the case, let's do templating it if you really want to. Not the biggest deal but I prefer the existing stuff.

SteveBronder:
@bbbales2 I think we should just do the nuts dense and diag adapt algorithms in parallel right now because I think we'll have a lot to do on the cmdstan side

bbbales2:
I think this removes the get_underlying thing: https://github.com/stan-dev/stan/compare/feature/parallel-nuts...review/parallel-nuts-no-get

I want to investigate the loggers a little more tomorrow morning. I want to make sure that transformed data is getting handled the same as everything else cause it happens in a different place than other things. I think that stringstream trick for handling multithreading you found is good though.

> we'll have a lot to do on the cmdstan side

In terms of figuring out the arguments and stuff? I saw the var contexts thing got fancier today.

SteveBronder:
https://github.com/stan-dev/stan/compare/feature/parallel-nuts...review/parallel-nuts-no-get#diff-b6a2187bc6862ee3a2771cb7a1efc813f93c7bbc4d8c10f60aabb88882a8b621R243

That's fine as long as we specify that the class in the `std::vector<>` is a pointer or smart pointer. I like get underlying since it can be overwritten in case packages that use the services want to access the object in another way (like just passing a vector of initializations that are not pointers.) But idt this is a big deal so I'm fine with restricting it to vectors holding pointers. Looking at `Rcpp::Xptr<>` I think yours will work fine

https://github.com/RcppCore/Rcpp/blob/master/inst/include/Rcpp/XPtr.h#L152

bbbales2:
> That's fine as long as we specify that the class in the std::vector<> is a pointer or smart pointer.

Yeah that does seem bad.

Because of how we read things in cmdstan, polymorphism is convenient, and because polymorphism, we're doing vectors of pointers, and because of that, we're making the interface require pointers. So yeah maybe get underlying is better.

Would it clean this all up if the json reader and the .dat reader returned the same type here: https://github.com/stan-dev/cmdstan/blob/develop/src/cmdstan/command.hpp#L82 ? Then we're not juggling shared_ptrs everywhere just so polymorphism works.

bbbales2:
I looked at this more. I think we could make two functions in cmdstan without much trouble (given the existing code):

```
array_var_context build_array_var_context_from_json(std::string filename);
array_var_context build_array_var_context_from_rdata(std::string filename);
```

Then the shared_ptr stuff can go away since we don't depend on polymorphism because we get the same type out regardless of whether we read json or rdata, and then there's no need for funny business inside stan-dev/stan (no dereferencing pointers or getting the underlying).

If you think this makes sense I can put together a pull pointing at proto/multi-chain that does this.

SteveBronder:
Srry for delayed response, spent an irresponsible amount of time trying to fix the adjoint ODE solver test thing

> Then the shared_ptr stuff can go away since we don't depend on polymorphism because we get the same type out regardless of whether we read json or rdata, and then there's no need for funny business inside stan-dev/stan (no dereferencing pointers or getting the underlying).

idk if I'm following. Why should the shared ptr thing go away? For instance that lets us handle the case where a user passes one init file and we can just share it across all the chains. Why do you want to move to an array var context?

Also idt I understand why de-referencing a pointer is funny business, it's a smart pointer and I want to see what's inside of it. If we were passing around raw pointers then yeah that might be smelly but I think that's totally fine in a smart pointer. I think we should leave `get_underlying()` because it allows for both actual objects (which it just returns), the standard smart pointers, and can be overwritten pretty easily for whatever fancy smart pointer thing upstream folks have.

SteveBronder:
@wds15 do you have time to review this? I have a branch of cmdstan [here](https://github.com/stan-dev/cmdstan/pull/987) that has everything hooked up if you want to try it out (the interface for that uses `parallel` as an argument with sub arguments but that's going to change in the actual PR).

One thing to note, in the above cmdstan version it throws if we try to run multiple chains without `STAN_THREADS` defined. If stan threads is not defined should we just use a simple for loop instead of the `parallell_for`? It seems like tbb is making a copy of a writer somewhere that is getting lost.

wds15:
This is ready and there is a chance to get it into 2.27?

I can try, but can’t promise…

wds15:
> It seems like tbb is making a copy of a writer somewhere that is getting lost.

Not sure I understand this comment... but for case of `STAN_THREADS` not being defined why can't we just continue to use `parallel_for`? The TBB would need to be initialized to use just one core, but then it should behave just like a single core program and there is not need to worry about anything.

SteveBronder:
We don't need it for the next release, I don't see any reason to rush here. Let me dig into what's going up downstream, maybe I'm doing something silly

wds15 (changes requested):
I need to look further into this. My main comment so far is about simplification. Why not get rid of the 1-core versions which we can reroute to the multi-core ones?

wds15:
Bump. Is this PR on a break?

SteveBronder:
Oh sorry! Yeah I'm going to look at it again on Monday. I have to sort out a weird segfault I was getting in cmdstan when `STAN_THREADS` is not defined

src/stan/services/sample/hmc_nuts_dense_e_adapt.hpp
```diff
@@ -156,6 +155,219 @@ int hmc_nuts_dense_e_adapt(
...
+        },
+        tbb::simple_partitioner());
+#else
+  for (size_t i = 0; i != num_chains; ++i) {
+    util::run_adaptive_sampler(
+        samplers[i], model, cont_vectors[i], num_warmup, num_samples,
+        num_thin, refresh, save_warmup, rngs[i], interrupt, logger,
+        sample_writer[i], diagnostic_writer[i], i + 1, num_chains);
+  }
+#endif
```
SteveBronder, line 285: @wds15 I was looking at the code for `reduce_sum()` and was getting a bit confused. Do I need to have code like this for when `STAN_THREADS` is defined or not? Going through `tbb::parallel_for()` without `STAN_THREADS` caused the updated version of cmdstan (below) to segfault.

src/stan/services/sample/hmc_nuts_dense_e_adapt.hpp
```diff
@@ -156,6 +155,211 @@ int hmc_nuts_dense_e_adapt(
...
+    unsigned int random_seed, unsigned int init_chain_id, double init_radius,
+    int num_warmup, int num_samples, int num_thin, bool save_warmup,
+    int refresh, double stepsize, double stepsize_jitter, int max_depth,
+    double delta, double gamma, double kappa, double t0,
+    unsigned int init_buffer, unsigned int term_buffer, unsigned int window,
+    callbacks::interrupt& interrupt, callbacks::logger& logger,
+    std::vector<InitWriter>& init_writer,
+    std::vector<SampleWriter>& sample_writer,
+    std::vector<DiagnosticWriter>& diagnostic_writer) {
+  if (num_chains == 1) {
```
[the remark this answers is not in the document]
SteveBronder: I don't have a strong opinion about this, but tbh I'd like to just make sure we are 100% cool with backwards compatibility by always deferring to the original API version if `chains=1`

src/stan/services/util/run_adaptive_sampler.hpp
```diff
@@ -34,15 +35,16 @@ namespace util {
...
 void run_adaptive_sampler(Sampler& sampler, Model& model,
                           std::vector<double>& cont_vector, int num_warmup,
                           int num_samples, int num_thin, int refresh,
                           bool save_warmup, RNG& rng,
                           callbacks::interrupt& interrupt,
                           callbacks::logger& logger,
                           callbacks::writer& sample_writer,
-                          callbacks::writer& diagnostic_writer) {
+                          callbacks::writer& diagnostic_writer,
+                          size_t chain_id = 0, size_t n_chain = 1) {
```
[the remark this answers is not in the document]
SteveBronder: Ooof, should be 1!

src/stan/services/sample/hmc_nuts_dense_e_adapt.hpp
wds15: Actually this type of code should not be necessary. What I think happens is that the TBB is spawning happily multiple threads, but it should not. So I would suggest to patch

https://github.com/stan-dev/math/blob/develop/stan/math/prim/core/init_threadpool_tbb.hpp

such that whenever `STAN_THREADS` is not set, then the TBB thread pool will be set to use only 1 thread. Then the above should be fine without the ifdef. Can you try that? That is patch the init function for the case `STAN_THREADS` is not set to always just use 1 thread and then disable the ifdef above.

wds15:
I need to a few days to dive into this...

src/stan/services/sample/hmc_nuts_dense_e_adapt.hpp
SteveBronder: Yep this was the issue! Just put in a PR in math to fix this

wds15: Cool. So we can get rid of the ifdef once this fix in math is in. Great.

src/stan/services/sample/hmc_nuts_dense_e_adapt.hpp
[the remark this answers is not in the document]
wds15: Na... let's just ensure we have good tests for that and instead make the old code call the new one with chains=1.

One issue you have when doing that is that you get references to inits and other things passed into the function. These then need to be wrapped into a size 1 vector of refs. Since you went with templates one can this use `std::ref` to make this work if I am not mistaken. I'd really like to avoid code duplication which is a source for errors in the future.

wds15:
Hmm... I am not sure that I see tests which confirm that we get the striding behavior as outlined in the design doc. So we want to ensure that the sampler will do the same if we do 4 chains with chain id 1...4 and just one thread each (old interface) and when doing chain 1..4 with 4 threads at once. Maybe also test that things match up when you do this 1..2 with 2 threads and again 3..4 with 2 threads.

Maybe we can test that the sampler arrived at the same state for lp on a test model? I'd also test that the rng is in the same sate - either explicitly or implicitly through something tied to the sampler which must be based on that.

I'd prefer to have some more tests in this direction in then we get rid of the chains == 1 code by deferring to the multi-chain  methods with chains set to 1.

SteveBronder:
> Hmm... I am not sure that I see tests which confirm that we get the striding behavior as outlined in the design doc. So we want to ensure that the sampler will do the same if we do 4 chains with chain id 1...4 and just one thread each (old interface) and when doing chain 1..4 with 4 threads at once. Maybe also test that things match up when you do this 1..2 with 2 threads and again 3..4 with 2 threads.

> Maybe we can test that the sampler arrived at the same state for lp on a test model? I'd also test that the rng is in the same sate - either explicitly or implicitly through something tied to the sampler which must be based on that.

> I'd prefer to have some more tests in this direction in then we get rid of the chains == 1 code by deferring to the multi-chain methods with chains set to 1.

I agree tests for the sampler states with multi threads vs multiple programs makes sense. I'll have to think about how to do this / find a simple model for the tests. I think we could do something for implicitly checking the RNG by doing something with generated quantities

SteveBronder:
Alrighty! @wds15 check out the test in `src/test/unit/services/sample/hmc_nuts_diag_e_adapt_parallel_match_test.cpp` I think I got everything lined up where I check if the multi chain diagonal adaptation samples matches the single chain diagonal adaptation samples for the `rosenbrock` model. The samples match exactly so I think that implicitly means the rng has to be getting the states right. 

If that all makes sense then I'll do the same thing for dense adaptation (and then I think we are good!)

src/stan/services/util/run_adaptive_sampler.hpp
```diff
@@ -34,15 +35,16 @@ namespace util {
...
 void run_adaptive_sampler(Sampler& sampler, Model& model,
                           std::vector<double>& cont_vector, int num_warmup,
                           int num_samples, int num_thin, int refresh,
                           bool save_warmup, RNG& rng,
                           callbacks::interrupt& interrupt,
                           callbacks::logger& logger,
                           callbacks::writer& sample_writer,
-                          callbacks::writer& diagnostic_writer) {
+                          callbacks::writer& diagnostic_writer,
+                          size_t chain_id = 1, size_t n_chain = 1) {
```
wds15, line 47: these new arguments need doc, please.

src/test/unit/services/sample/hmc_nuts_dense_e_adapt_parallel_test.cpp
```diff
@@ -0,0 +1,175 @@
...
+#include <gtest/gtest.h>
+#include <stan/io/empty_var_context.hpp>
+#include <test/test-models/good/optimization/rosenbrock.hpp>
+#include <test/unit/services/instrumented_callbacks.hpp>
+#include <iostream>
+
+auto&& blah = stan::math::init_threadpool_tbb();
+
+static constexpr size_t num_chains = 4;
+class ServicesSampleHmcNutsDenseEAdaptPar : public testing::Test {
```
wds15, line 11: and here two comments explaining what the test does would be great.

src/stan/services/util/create_rng.hpp
```diff
@@ -19,14 +19,17 @@ namespace util {
  * duplicated.
  *
  * @param[in] seed the random seed
- * @param[in] chain the chain id
+ * @param[in] init_chain_id the chain id
+ * @param[in] chain_num For multi-chain, the ch
```
wds15, line 23: doc got cut?

src/stan/services/sample/hmc_nuts_diag_e_adapt.hpp
```diff
@@ -143,17 +147,219 @@ int hmc_nuts_diag_e_adapt(
...
+        tbb::blocked_range<size_t>(0, num_chains, 1),
+        [num_warmup, num_samples, num_thin, refresh, save_warmup, num_chains,
+         &samplers, &model, &rngs, &interrupt, &logger, &sample_writer,
+         &cont_vectors,
+         &diagnostic_writer](const tbb::blocked_range<size_t>& r) {
+          for (size_t i = r.begin(); i != r.end(); ++i) {
+            util::run_adaptive_sampler(
+                samplers[i], model, cont_vectors[i], num_warmup, num_samples,
+                num_thin, refresh, save_warmup, rngs[i], interrupt, logger,
+                sample_writer[i], diagnostic_writer[i], i + 1, num_chains);
```
wds15, line 274: the chain-id given to `run_adaptive_sampler` is `i+1`... which labels from 1 to num_chains... but shouldn't we make this `init_chain_id + i`? I do think so. In case you like 1...num_chains  numbering then the least we need to do is to aling the argument name for the downstream functions. That is right now called `chain_id`... which it is not if we don't change it as suggested.

I'd rather like to get the actual chain id's being displayed.

wds15 (changes requested):
We need to sort out init_chain_id passing around and some more doc is needed for the new arguments for chain id management. Looking good. Thanks for the test!

src/stan/services/util/create_rng.hpp
SteveBronder: Nope! Names of the arguments changed

src/stan/services/sample/hmc_nuts_diag_e_adapt.hpp
SteveBronder: I think what we want is `init_chain_id + i + 1` right? for `n_chain = 1` `generate_transitions` gives the output we have now, but for `n_chain > 1` I think we want the initial chain id + what chain we are on. So for `init_chain_id = 5` and `n_chain = 4` we would get back 

```bash
Chain [5] Iteration: ... 
Chain [6] Iteration: ... 
Chain [7] Iteration: ... 
Chain [8] Iteration: ... 
```

I'd prefer not changing the output for the default/current stuff. I'm sure someone somewhere is logging and grepping that and I'd rather not cause a headache somewhere

SteveBronder:
Ty for the review! If you agree with the printing out of the chain id above then should be good!

wds15:
The ball is in your court, right?

SteveBronder:
Lol oh I was waiting on you, is there something I'm the review I missed?

src/stan/services/sample/hmc_nuts_diag_e_adapt.hpp
wds15: I think it's `init_chain_id + i`, because `init_chain_id` is the first chain id you want to see.

I am ok with leaving the old code around and have slightly different outputs for multi chain vs single chain runs.

This point still needs update in the code, right?

wds15:
Sorry.. no, you did not miss things... I missed your update. The last thing to sort out is the chain id output labeling. See above.

Once we got that, then I need to check that pre-compilation of the services is still all doing its job. Quick to do by just timing the bernoulli example compile time.

Then we are good from my view.

src/stan/services/sample/hmc_nuts_diag_e_adapt.hpp
SteveBronder: > I think it's init_chain_id + i, because init_chain_id is the first chain id you want to see.

Do we want to index from 0 or 1 for the printing? Looking at the code for `generate_transitions()` if we did `init_chain_id + i` for 4 chains and init value of 0 (the default for cmdstan) we will have 

```
Chain [0] Iteration: ... 
Chain [1] Iteration: ... 
Chain [2] Iteration: ... 
Chain [3] Iteration: ... 
```

> This point still needs update in the code, right?

If we are indexing from 1 then the code I have right now is right, but if want to index from zero then I need to remove the `+ 1`

wds15: If we index from 0 right now, then we should display it like that, I think. I would find it very confusing to have the printed id to be different than the actual id of the chain.

@mitzimorris what do you think?

SteveBronder: Sorry, by

> If we index from 0 right now, then we should display it like that, I think

what indexing are you referring to? The initial id? That is a value of zero. idk if there is a current precedent for this. But if we are going to do the chain outputs as `{output_name}_{chain_id}.csv` starting from zero then I do think that it makes sense to index from 0 here. Does that make sense? I think what I'm saying is that if the output csv's names start from 0 we should have these start from 0 as well. Though we could change those to start at 1. I'm fine either way, though I kind of like indexing from 1 since the Stan language indexes from 1.

wds15: OH... good point. So yeah, the outputs should be aligned with the chain id is my view here. It is unfortunate that cmdstan allows chain ids to start with 0 and even the default is 0. My proposal is to align chain id with whatever number we show on screen or put into file names. I would find it extremely confusing to see chain id 2 being sampled in the initial print and then the chain is being referred to as chain 3??

In order to remedy the situation we can consider to change the default chain id in cmdstan to be 1 instead of 0 (while still allowing a chain id of 0).

SteveBronder: Yep I'm cool with that! Just made the change to `init_chain_id + i`. We can talk about changing the default init id to 0 in the cmdstan branch

SteveBronder:
Alrighty @wds15 I also updated https://github.com/stan-dev/cmdstan/pull/987 with this branch + the change for the init being set to 1.

I moved the `chains` argument to be under `sample` as for now that's the only one that uses it. I think the only other one that would use it would be `variational` and we can make `chains` an argument for variational once we have a version of VB that uses multiple chains. So calling is now like 

```
examples/diamonds/diamonds sample num_samples=150 num_warmup=150 chains=8 \
 data file=examples/diamonds/diamonds2.json threads=8
```

wds15:
So one of my last tests is to compare the compile time with 2.27.0 is:

```
[21:35:41][sebi@sebastians-macbook-pro-1:~/work/cmdstan-2.27.0]$ time make examples/bernoulli/bernoulli

--- Translating Stan model to C++ code ---
bin/stanc  --o=examples/bernoulli/bernoulli.hpp examples/bernoulli/bernoulli.stan

--- Compiling, linking C++ code ---
clang++ -DSTAN_THREADS -std=c++1y -Wno-unknown-warning-option -Wno-tautological-compare -Wno-sign-compare -D_REENTRANT -Wno-ignored-attributes      -I stan/lib/stan_math/lib/tbb_2020.3/include   -O3 -I src -I stan/src -I lib/rapidjson_1.1.0/ -I lib/CLI11-1.9.1/ -I stan/lib/stan_math/ -I stan/lib/stan_math/lib/eigen_3.3.9 -I stan/lib/stan_math/lib/boost_1.75.0 -I stan/lib/stan_math/lib/sundials_5.7.0/include    -DBOOST_DISABLE_ASSERTS         -c -include-pch stan/src/stan/model/model_header.hpp.gch -x c++ -o examples/bernoulli/bernoulli.o examples/bernoulli/bernoulli.hpp
clang++ -DSTAN_THREADS -std=c++1y -Wno-unknown-warning-option -Wno-tautological-compare -Wno-sign-compare -D_REENTRANT -Wno-ignored-attributes      -I stan/lib/stan_math/lib/tbb_2020.3/include   -O3 -I src -I stan/src -I lib/rapidjson_1.1.0/ -I lib/CLI11-1.9.1/ -I stan/lib/stan_math/ -I stan/lib/stan_math/lib/eigen_3.3.9 -I stan/lib/stan_math/lib/boost_1.75.0 -I stan/lib/stan_math/lib/sundials_5.7.0/include    -DBOOST_DISABLE_ASSERTS               -Wl,-L,"/Users/sebi/work/cmdstan-2.27.0/stan/lib/stan_math/lib/tbb" -Wl,-rpath,"/Users/sebi/work/cmdstan-2.27.0/stan/lib/stan_math/lib/tbb"      examples/bernoulli/bernoulli.o src/cmdstan/main.o        -Wl,-L,"/Users/sebi/work/cmdstan-2.27.0/stan/lib/stan_math/lib/tbb" -Wl,-rpath,"/Users/sebi/work/cmdstan-2.27.0/stan/lib/stan_math/lib/tbb"   stan/lib/stan_math/lib/sundials_5.7.0/lib/libsundials_nvecserial.a stan/lib/stan_math/lib/sundials_5.7.0/lib/libsundials_cvodes.a stan/lib/stan_math/lib/sundials_5.7.0/lib/libsundials_idas.a stan/lib/stan_math/lib/sundials_5.7.0/lib/libsundials_kinsol.a  stan/lib/stan_math/lib/tbb/libtbb.dylib stan/lib/stan_math/lib/tbb/libtbbmalloc.dylib stan/lib/stan_math/lib/tbb/libtbbmalloc_proxy.dylib -o examples/bernoulli/bernoulli
rm -f examples/bernoulli/bernoulli.o

real	0m6.449s
user	0m5.979s
sys	0m0.449s
```

and with this branch I got

```
--- Translating Stan model to C++ code ---
bin/stanc  --o=examples/bernoulli/bernoulli-2.hpp examples/bernoulli/bernoulli-2.stan

--- Compiling, linking C++ code ---
clang++ -std=c++1y -Wno-unknown-warning-option -Wno-tautological-compare -Wno-sign-compare -D_REENTRANT -Wno-ignored-attributes     -DSTAN_THREADS -I stan/lib/stan_math/lib/tbb_2020.3/include   -O3 -I src -I stan/src -I lib/rapidjson_1.1.0/ -I lib/CLI11-1.9.1/ -I stan/lib/stan_math/ -I stan/lib/stan_math/lib/eigen_3.3.9 -I stan/lib/stan_math/lib/boost_1.75.0 -I stan/lib/stan_math/lib/sundials_5.7.0/include    -DBOOST_DISABLE_ASSERTS         -c -include-pch stan/src/stan/model/model_header_threads.hpp.gch -x c++ -o examples/bernoulli/bernoulli-2.o examples/bernoulli/bernoulli-2.hpp
clang++ -std=c++1y -Wno-unknown-warning-option -Wno-tautological-compare -Wno-sign-compare -D_REENTRANT -Wno-ignored-attributes     -DSTAN_THREADS -I stan/lib/stan_math/lib/tbb_2020.3/include   -O3 -I src -I stan/src -I lib/rapidjson_1.1.0/ -I lib/CLI11-1.9.1/ -I stan/lib/stan_math/ -I stan/lib/stan_math/lib/eigen_3.3.9 -I stan/lib/stan_math/lib/boost_1.75.0 -I stan/lib/stan_math/lib/sundials_5.7.0/include    -DBOOST_DISABLE_ASSERTS               -Wl,-L,"/Users/sebi/work/cmdstan/stan/lib/stan_math/lib/tbb" -Wl,-rpath,"/Users/sebi/work/cmdstan/stan/lib/stan_math/lib/tbb"      examples/bernoulli/bernoulli-2.o src/cmdstan/main_threads.o        -Wl,-L,"/Users/sebi/work/cmdstan/stan/lib/stan_math/lib/tbb" -Wl,-rpath,"/Users/sebi/work/cmdstan/stan/lib/stan_math/lib/tbb"   stan/lib/stan_math/lib/sundials_5.7.0/lib/libsundials_nvecserial.a stan/lib/stan_math/lib/sundials_5.7.0/lib/libsundials_cvodes.a stan/lib/stan_math/lib/sundials_5.7.0/lib/libsundials_idas.a stan/lib/stan_math/lib/sundials_5.7.0/lib/libsundials_kinsol.a  stan/lib/stan_math/lib/tbb/libtbb.dylib stan/lib/stan_math/lib/tbb/libtbbmalloc.dylib stan/lib/stan_math/lib/tbb/libtbbmalloc_proxy.dylib -o examples/bernoulli/bernoulli-2
rm -f examples/bernoulli/bernoulli-2.o

real	0m6.941s
user	0m6.432s
sys	0m0.458s
```

So possibly this branch does slow down things a tiny bit wrt to compilation (upon the second time to be clear), but I'd assume this is within noise.

SteveBronder:
Yeah imo I'd expect possibly a tiny slowdown in compilation, but overall that seems to be fine.

You think this is ready to merge?

wds15:
From my memory this is good now. Let me do one last round over this.

This is my first bigger PR review for Stan, so that's why I take a bit more time. I'd guess we are good.

src/stan/services/sample/hmc_nuts_dense_e_adapt.hpp
```diff
@@ -156,6 +155,211 @@ int hmc_nuts_dense_e_adapt(
...
+ * @tparam InitWriter A type derived from `stan::callbacks::writer`
+ * @param[in] model Input model to test (with data already instantiated)
+ * @param[in] init An std vector of init var contexts for initialization of each
+ chain.
+ * @param[in] init_inv_metric An std vector of var contexts exposing an initial
+ diagonal inverse Euclidean metric for each chain (must be positive definite)
+ * @param[in] random_seed random seed for the random number generator
+ * @param[in] init_chain_id first chain id. The pseudo random number generator
+ will advance by for each chain by an integer sequence from `init_chain_id` to
+ `num_chains`
```
wds15, line 178: the last chain id won't be `num_chains`, but `init_chain_id+num_chains-1`

src/stan/services/util/create_rng.hpp
wds15: But the doc "For multi-chain, the ch" is not meaningful to me. What about 

```
@param[in] init_chain_id start of chain ids
 @param[in] chain_num chain id offset such that chain_id is init_chain_id+chain_num
  ```

Also... reading the comment here suggest to me that we should actually not allow for a chain id of 0, right??? (I mean the comment to the function)

**Review by wds15 — COMMENTED** [review]
Sorry, did not get to finish the review. Leaving these comments for now on the doc.

File: src/stan/services/sample/hmc_nuts_dense_e_adapt.hpp [hunk 16]
```diff
@@ -156,6 +155,211 @@ int hmc_nuts_dense_e_adapt(
...
+    std::vector<SampleWriter>& sample_writer,
+    std::vector<DiagnosticWriter>& diagnostic_writer) {
+  if (num_chains == 1) {
+    return hmc_nuts_dense_e_adapt(
+        model, *init[0], *init_inv_metric[0], random_seed, init_chain_id,
+        init_radius, num_warmup, num_samples, num_thin, save_warmup, refresh,
+        stepsize, stepsize_jitter, max_depth, delta, gamma, kappa, t0,
+        init_buffer, term_buffer, window, interrupt, logger, init_writer[0],
+        sample_writer[0], diagnostic_writer[0]);
+  } else {
```
wds15, line 228: The else branch is not needed. A simple `if` for the num_chains=1 case is sufficient (same for the other functions).

src/stan/services/sample/hmc_nuts_dense_e_adapt.hpp
```diff
@@ -156,6 +155,211 @@ int hmc_nuts_dense_e_adapt(
...
+    std::vector<DiagnosticWriter>& diagnostic_writer) {
+  if (num_chains == 1) {
+    return hmc_nuts_dense_e_adapt(
+        model, *init[0], random_seed, init_chain_id, init_radius, num_warmup,
+        num_samples, num_thin, save_warmup, refresh, stepsize, stepsize_jitter,
+        max_depth, delta, gamma, kappa, t0, init_buffer, term_buffer, window,
+        interrupt, logger, init_writer[0], sample_writer[0],
+        diagnostic_writer[0]);
+  } else {
+    std::vector<std::unique_ptr<stan::io::dump>> unit_e_metrics;
```
wds15, line 348: else is not needed.

src/stan/services/sample/hmc_nuts_diag_e_adapt.hpp
```diff
@@ -143,17 +147,220 @@ int hmc_nuts_diag_e_adapt(
...
+ * @tparam InitWriter A type derived from `stan::callbacks::writer`
+ * @param[in] model Input model to test (with data already instantiated)
+ * @param[in] init An std vector of init var contexts for initialization of each
+ chain.
+ * @param[in] init_inv_metric An std vector of var contexts exposing an initial
+ diagonal inverse Euclidean metric for each chain (must be positive definite)
+ * @param[in] random_seed random seed for the random number generator
+ * @param[in] init_chain_id first chain id. The pseudo random number generator
+ will advance by for each chain by an integer sequence from `init_chain_id` to
+ `num_chains`
```
wds15, line 179: doc needs correction as before. The range is `init_chain_id` to `init_chain_id+num_chains-1`

src/stan/services/sample/hmc_nuts_diag_e_adapt.hpp
```diff
@@ -143,17 +147,220 @@ int hmc_nuts_diag_e_adapt(
...
+ * @param[in] term_buffer width of final fast adaptation interval
+ * @param[in] window initial width of slow adaptation interval
+ * @param[in,out] interrupt Callback for interrupts
+ * @param[in,out] logger Logger for messages
+ * @param[in,out] init_writer std vector of Writer callbacks for unconstrained
+ inits of each chain.
+ * @param[in,out] sample_writer std vector of Writers for draws of each chain.
+ * @param[in,out] diagnostic_writer std vector of Writers for diagnostic
+ information of each chain.
+ * @param[in] num_chains The number of chains to run in parallel. `init`,
```
wds15, line 203: `num_chains` doc string should appear after Model so that it matches the order of the arguments.

src/stan/services/sample/hmc_nuts_diag_e_adapt.hpp
```diff
@@ -143,17 +147,220 @@ int hmc_nuts_diag_e_adapt(
...
+    std::vector<DiagnosticWriter>& diagnostic_writer) {
+  if (num_chains == 1) {
+    return hmc_nuts_diag_e_adapt(
+        model, *init[0], *init_inv_metric[0], random_seed, init_chain_id,
+        init_radius, num_warmup, num_samples, num_thin, save_warmup, refresh,
+        stepsize, stepsize_jitter, max_depth, delta, gamma, kappa, t0,
+        init_buffer, term_buffer, window, interrupt, logger, init_writer[0],
+        sample_writer[0], diagnostic_writer[0]);
+  } else {
+    using sample_t = stan::mcmc::adapt_diag_e_nuts<Model, boost::ecuyer1988>;
```
wds15, line 230: else not needed

src/stan/services/util/create_rng.hpp
wds15: BTW, do we actually need to have `create_rng` take a `init_chain_id` and a `num_chains`? Why don't we leave the function as is and just pre-compute the chain_id's in the parallel sample functions? That seems simpler to me.

src/stan/services/sample/hmc_nuts_diag_e_adapt.hpp
```diff
@@ -143,17 +147,220 @@ int hmc_nuts_diag_e_adapt(
...
+    std::vector<DiagnosticWriter>& diagnostic_writer) {
+  if (num_chains == 1) {
+    return hmc_nuts_diag_e_adapt(
+        model, *init[0], random_seed, init_chain_id, init_radius, num_warmup,
+        num_samples, num_thin, save_warmup, refresh, stepsize, stepsize_jitter,
+        max_depth, delta, gamma, kappa, t0, init_buffer, term_buffer, window,
+        interrupt, logger, init_writer[0], sample_writer[0],
+        diagnostic_writer[0]);
+  } else {
+    std::vector<std::unique_ptr<stan::io::dump>> unit_e_metrics;
```
wds15, line 349: no else here needed

src/stan/services/util/run_adaptive_sampler.hpp
```diff
@@ -66,7 +71,7 @@ void run_adaptive_sampler(Sampler& sampler, Model& model,
   auto start_warm = std::chrono::steady_clock::now();
   util::generate_transitions(sampler, num_warmup, 0, num_warmup + num_samples,
                              num_thin, refresh, save_warmup, true, writer, s,
-                             model, rng, interrupt, logger);
+                             model, rng, interrupt, logger, chain_id, n_chain);
```
wds15, line 74: Sorry for the late catch...but why is it called `n_chain` here and not `num_chain`? `num_chain` looks more consistent with all the other variable conventions...would need to be addressed in the entire PR.

wds15 (changes requested):
Small laundry items, not more. Then we are good. Will look at the cmdstan num_thread thing now.

src/stan/services/util/create_rng.hpp
SteveBronder: No we do not need to change this good find! I'll switch it back

wds15:
It's on me again?

SteveBronder:
Yep go for it!

SteveBronder:
@wds15 bump!

wds15:
Monday…sorry…

SteveBronder:
All good!

wds15 (approved):

wds15:
LGTM!

PR #3037, opened 2021-04-12
Skip metric regularization if adapt window=0

nhuurre:
#### Submission Checklist

#### Summary

This came up in https://github.com/stan-dev/stan/pull/3027#issuecomment-814223714

CmdStan accepts `window=0` as a valid argument even though it means the adaptation cannot possibly estimate the inverse metric. The intended effect appears to be adapt only stepsize and leave the metric at its initial value, either the default identity or a user specified value. That works _but only if_ you also set `init_buffer=0`. If `init_buffer>0` the sampler tries to update the metric anyway and messes it up.

The metric update code is
https://github.com/stan-dev/stan/blob/35a15c8c206a09415318fbdfc0b324993929fd08/src/stan/mcmc/var_adaptation.hpp#L24-L28
The Welford estimator's `sample_variance()` is a no-op if there are not enough samples so that part's fine.
The problem is the subsequent regularization which in case of zero samples just discards the (unchanged) metric.
The fix is to simply skip the regularization when `sample_variance()` didn't do anything.

#### Intended Effect

Setting `window=0` always keeps the initial metric.

#### How to Verify

Run the `examples/bernoulli` model with `adapt init_buffer=10 window=0` and look for the mass matrix in output.csv.
On develop it says
```
# Diagonal elements of inverse mass matrix:
# 0.001
```
On this branch
```
# Diagonal elements of inverse mass matrix:
# 1
```

#### Copyright and Licensing

Please list the copyright holder for the work you are submitting (this will be you or your assignee, such as a university or company): Niko Huurre

### DIFF
--- a/src/stan/mcmc/covar_adaptation.hpp
+++ b/src/stan/mcmc/covar_adaptation.hpp
@@ -23,10 +23,12 @@ class covar_adaptation : public windowed_adaptation {
 
       estimator_.sample_covariance(covar);
 
-      double n = static_cast<double>(estimator_.num_samples());
-      covar = (n / (n + 5.0)) * covar
-              + 1e-3 * (5.0 / (n + 5.0))
-                    * Eigen::MatrixXd::Identity(covar.rows(), covar.cols());
+      if (estimator_.num_samples() > 1) {
+        double n = static_cast<double>(estimator_.num_samples());
+        covar = (n / (n + 5.0)) * covar
+                + 1e-3 * (5.0 / (n + 5.0))
+                      * Eigen::MatrixXd::Identity(covar.rows(), covar.cols());
+      }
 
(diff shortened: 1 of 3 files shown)


betanalpha:
Although I largely agree with the update technically inverse metric updating is skipped for both adapt_window = 0 _and_ adapt_window = 1.  I'll change the pull name before merging.

betanalpha (changes requested):

betanalpha:
If changes are made according to https://github.com/stan-dev/stan/pull/3027#issuecomment-820638658 then no warning message would be needed for the base_window = 0 case, in which case messages wouldn't have to be passed, but if inverse metric updating is also disabled for base_window = 1 then I think we still need the message.

PR #3040, opened 2021-04-25
Refactor Integration tests

rok-cesnovar:
This PR will remove the extremely large test models from the regular Integration test models batch. 
The models being removed:
- `distributions/univariate/continuous/exp_mod_normal`
- `distributions/univariate/continuous/pareto_type_2`
- `distributions/univariate/continuous/skew_normal`
- `distributions/univariate/continuous/student_t`
- `distributions/univariate/continuous/wiener`

They are being removed from being used in every PR because they represent 70%+ of execution time for these tests and do not add a lot of test coverage, if any.
These models are simple but huge. They test that all different signatures compile. And because these distributions have many arguments
the models are huge. They thus also consume a lot of RAM which limits the parallel setting we can use (we quickly run out of swap, as 1500 line so lpdf calls can consume up to 10GB of RAM).

The almost exact same test (testing that signatures compile) is already being tested in Math distribution tests.
These will only be tested on request (via Jenkins) and in a separate weekly test for CmdStan we are preparing.

This is not ready for review yet, as I need to do some testing prior. Will tag once ready.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -192,62 +192,62 @@ pipeline {
                 }
             }
         }
-        stage('Unit tests') {
-            when {
-                expression {
-                    !skipRemainingStages
-                }
-            }
-            parallel {
-                stage('Windows Headers & Unit') {
-                    agent { label 'windows' }
-                    when {
-                        expression {
-                            ( env.BRANCH_NAME == "develop" ||
-                            env.BRANCH_NAME == "master" ||
-                            params.run_tests_all_os ) &&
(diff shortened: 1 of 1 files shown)


SteveBronder:
> This is not ready for review yet, as I need to do some testing prior. Will tag once ready.

:+1: ping me when this is ready

rok-cesnovar:
Should be ready now.

SteveBronder (approved):
lcbtm!

PR #3048, opened 2021-06-28
Opencl indexing

t4c1:
#### Submission Checklist

#### Summary

Adds indexing for kernel generator (OpenCL) types. This was made to match indexing on CPU as closely as possible, although there still are significant differences.

`matrix_cl` and kernel generator expressions do not store compile time information about their shape. So we can not make different overloads for `matrix`, `vector`, `row_vector` and `array` types. The way this is made now everything is assumed to be 2d matrix. This will work for `matrix`, `vector` and 1-dimensional `array` types. For `row_vector`s we will need some help from the stanc compiler. `row_vector[index]` will need to be replaced by `row_vector[index_omni, index]`. Similarly indexing `vector` and `row_vector` types with `index_uni` will need to be changed. `vector[index_uni(i)]` -> `vector[index_uni(i), index_uni(1)]`, `row_vector[index_uni(i)]` -> `row_vector[index_uni(1), index_uni(i)]`.

For `index_multi` we need to have the array of indices on the OpenCL device. To keep things simple I decided to achieve that by replacing `index_multi` with `matrix_cl<int>` containing indices in a single column. This way the OpenCL version of the `index_multi` can be made by `to_matrix_cl(index_multi.ns_)`.

This also fixes a bug in `rvalue_index_size` and `rvalue_at` when called with `index_min_max` with reverse order (`.max_` being smaller than `.min_`). Some tests that were passing on wrong behavior were also fixed.

#### Intended Effect

Indexing OpenCL types is possible.

#### How to Verify

Run the new tests.

#### Documentation

In doxygen. 

#### Copyright and Licensing

Please list the copyright holder for the work you are submitting (this will be you or your assignee, such as a university or company): Tadej Ciglarič


src/stan/model/indexing/rvalue_cl.hpp
```diff
@@ -0,0 +1,395 @@
...
+}
+
+/**
+ * Index a prim kernel generator expression with two indices.
+ *
+ * @tparam Expr type of the exoression
+ * @tparam RowIndex type of row index
+ * @tparam ColIndex type of column index
+ * @param expr a prim kernel generator expression to index
+ * @param name string form of expression being evaluated
```
SteveBronder, line 122:
```suggestion
 * @param name string of name of rvalue being indexed if named, otherwise an empty string
```

src/stan/model/indexing/rvalue_cl.hpp
```diff
@@ -0,0 +1,395 @@
...
+template <typename Expr, typename RowIndex,
+          require_all_kernel_expressions_and_none_scalar_t<Expr>* = nullptr>
+inline auto rvalue(Expr&& expr, const char* name, const RowIndex& row_index) {
+  auto checks = internal::index_check(row_index, name, expr.rows());
+  try {
+    math::index_apply<std::tuple_size<decltype(checks)>::value>(
+        [&checks](auto... Is) {
+          math::results(std::get<Is>(checks).first...)
+              = math::expressions(std::get<Is>(checks).second...);
+        });
```
SteveBronder, line 106: Is there a reason to use `index_apply` over `apply` here? i.e.

```suggestion
    math::apply([](const auto& check) {
          math::results(check.first...) = math::expressions(check.second...);
    }, checks);
```

It looks like you use this pattern a lot, would be good to add an internal function that just does this ala

```cpp
template <typename T>
void index_check_cl(const T& checks) {
    math::apply([](const auto& check) {
          math::results(check.first...) = math::expressions(check.second...);
    }, checks);
}
```

src/stan/model/indexing/rvalue_cl.hpp
```diff
@@ -0,0 +1,395 @@
...
+
+  return math::indexing(expr,
+                        internal::cl_row_index(row_index, expr.rows(), name),
+                        internal::cl_col_index(col_index, expr.cols(), name));
+}
+
+/**
+ * Index a prim kernel generator expression with two single indices.
+ *
+ * @tparam Expr type of the exoression
```
SteveBronder, line 157: do a find/replace for `exoression`

```suggestion
 * @tparam Expr type of the expression
```

SteveBronder (commented):
Just a few quick comments for what I had time to look at today. I'll give this a more thorough review tmrw!

src/stan/model/indexing/rvalue_cl.hpp
t4c1: Fixed

src/stan/model/indexing/rvalue_cl.hpp
t4c1: Actually this is way more complicated than it needs to be. Fixed.

src/stan/model/indexing/rvalue_cl.hpp
t4c1: Fixed.

SteveBronder:
> For `row_vectors` we will need some help from the stanc compiler. `row_vector[index]` will need to be replaced by `row_vector[index_omni, index]`. Similarly indexing `vector` and `row_vector` types with `index_uni` will need to be changed. `vector[index_uni(i)] -> vector[index_uni(i), index_uni(1)]`, `row_vector[index_uni(i)] -> row_vector[index_uni(1), index_uni(i)]`.

It will take a bit of fancy footwork but I _think_ I can do this in the compiler. I started messing around with adding compile time size information in the compiler. I wonder if it would be easier to implement this in math if instead of allowing for general compile time sized `matrix_cl` types we just support knowing if we are looking at a matrix (-1, -1), vector (-1, 1), or row vector (1, -1). Then I think at least the scope of things needed to change shrinks a good bit (or at least the logic does).

https://github.com/stan-dev/math/tree/feature/compile-time-sizes-for-matrix_cl

src/stan/model/indexing/assign_cl.hpp
```diff
@@ -0,0 +1,369 @@
...
+  auto& lhs = lhs_holder;
+  math::arena_matrix_cl<double> prev_vals = lhs;
+  lhs = expr_rhs;
+  math::reverse_pass_callback([expr_lhs, row_index, prev_vals, name]() mutable {
+    auto&& lhs_val = rvalue(expr_lhs.val_op(), name, row_index);
+    auto&& lhs_adj = rvalue(expr_lhs.adj(), name, row_index);
+    math::results(lhs_val, lhs_adj) = math::expressions(
+        prev_vals, math::constant(0.0, lhs_adj.rows(), lhs_adj.cols()));
+  });
+}
```
SteveBronder, line 147: I'm confused about what's going on here, where is the `expr_rhs` getting assigned to the `expr_lhs`'s values? Is it because `lhs` is a ref to `lhs_holder` and `rvalue()` returns a reference to the `expr_lhs` values? So then `lhs = expr_rhs` is actually assigning to `expr_lhs`'s values?

src/stan/model/indexing/assign_cl.hpp
```diff
@@ -0,0 +1,369 @@
...
+  }
+  auto&& lhs_holder = rvalue(expr_lhs.val_op(), name, row_index);
+  auto& lhs = lhs_holder;
+  math::arena_matrix_cl<double> prev_vals = lhs;
+  lhs = expr_rhs.val();
+  math::reverse_pass_callback(
+      [expr_lhs, expr_rhs, row_index, name, prev_vals]() mutable {
+        auto&& lhs_val = rvalue(expr_lhs.val_op(), name, row_index);
+        auto&& lhs_adj_holder = rvalue(expr_lhs.adj(), name, row_index);
+        auto& lhs_adj = lhs_adj_holder;
```
SteveBronder, line 270: Why `auto&&` and then `auto&`?

src/test/unit/model/indexing/rvalue_cl_test.cpp
```diff
@@ -0,0 +1,273 @@
...
+auto from_matrix_cl_nonscalar(const int a) { return a; }
+
+template <typename T>
+void set_adjoints(T& v) {
+  for (int i = 0; i < v.rows(); i++) {
+    for (int j = 0; j < v.cols(); j++) {
+      v(i, j).adj() += i + 3;
+    }
+  }
+}
```
SteveBronder, line 45: Can you make a `util` file to put these in with the things used in the assign test as well?

src/test/unit/model/indexing/rvalue_cl_test.cpp
```diff
@@ -0,0 +1,273 @@
...
+                      = stan::math::to_matrix_cl(m_v2);
+                  auto correct = rvalue(m_v1, "", ind1, ind2);
+                  auto res = from_matrix_cl_nonscalar(rvalue(
+                      m_v_cl, "", opencl_index(ind1), opencl_index(ind2)));
+                  expect_eq(correct.val(), res.val());
+                  set_adjoints(correct);
+                  set_adjoints(res);
+                  stan::math::grad();
+                  expect_eq(m_v1.adj(), m_v2.adj());
+                  stan::math::recover_memory();
```
SteveBronder, line 171: Are you testing somewhere that the values are the same after grad() when slicing assignment happens?

SteveBronder (changes requested):
Mostly questions about design stuff and making sure the compiler and this impl will make sense. In a couple places I'd prefer using `apply()` and `for_each()` instead of `stan::math::index_apply`

src/stan/model/indexing/rvalue_cl.hpp
```diff
@@ -0,0 +1,329 @@
...
+ * @tparam Expr type of the expression
+ * @tparam RowIndex type of index
+ * @param expr a prim kernel generator expression to index
+ * @param name name of value being indexed (if named, otherwise an empty string)
+ * @param row_index index
+ * @return result of indexing
+ */
+template <typename Expr, require_rev_kernel_expression_t<Expr>* = nullptr>
+inline auto rvalue(Expr&& expr, const char* name,
+                   const math::matrix_cl<int>& row_index) {
```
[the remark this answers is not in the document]
t4c1: > Why not have rvalue() and assign() here that take in std::vector<int> types and call to_matrix_cl() here?

Because if the same index is used in every iteration we only need to copy it to GPU once. I it was done in indexing functions it would be copied every time.

src/test/unit/model/indexing/assign_cl_test.cpp
```diff
@@ -0,0 +1,413 @@
...
+template <typename T>
+T opencl_index(T i) {
+  return i;
+}
+stan::math::matrix_cl<int> opencl_index(const index_multi& i) {
+  return stan::math::to_matrix_cl(i.ns_);
+}
+
+template <typename T, stan::require_not_rev_kernel_expression_t<T>* = nullptr>
+void set_adjoints1(T& v) {
```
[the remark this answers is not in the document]
t4c1: They are just set adjoints in two different patterns, so as not to make any tests pass just because all the adjoints are the same. What would you name them?

src/test/unit/model/indexing/assign_cl_test.cpp
```diff
@@ -0,0 +1,413 @@
...
+  stan::math::matrix_cl<int> m2_i_cl(m2_i);
+  stan::math::matrix_cl<double> m_err(5, 5);
+  stan::math::matrix_cl<int> m_i_err(5, 5);
+  auto indices = std::make_tuple(
+      index_omni(), index_multi(std::vector<int>{1, 3, 2}), index_min(2),
+      index_max(3), index_min_max(2, 3), index_min_max(3, 1));
+  constexpr int N_ind = std::tuple_size<decltype(indices)>::value;
+  stan::math::index_apply<N_ind>([&](auto... Is1) {
+    std::initializer_list<int>{(
+        [&](const auto& ind1) {
```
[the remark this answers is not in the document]
t4c1: That is kind of in the test name.

src/stan/model/indexing/assign_cl.hpp
t4c1: Exactly.

src/test/unit/model/indexing/rvalue_cl_test.cpp
t4c1: I feel like we already have many util files scattered around. I always have to check what is in which one. So I don't think making another one would improve clarity.

src/test/unit/model/indexing/rvalue_cl_test.cpp
t4c1: What do you mean by slicing assignment?

src/test/unit/model/indexing/assign_cl_test.cpp
```diff
@@ -0,0 +1,413 @@
+#ifdef STAN_OPENCL
```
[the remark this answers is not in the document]
serban-nicusor-toptal: Hey @SteveBronder I think we can why not, is implementation similar to [math repository](https://github.com/stan-dev/math/blob/develop/Jenkinsfile#L232) ? I can create a PR next week with it.

src/stan/model/indexing/rvalue_cl.hpp
```diff
@@ -0,0 +1,329 @@
...
+  try {
+    cl::Event copy_event;
+    queue.enqueueReadBuffer(
+        expr_eval.buffer(), true,
+        sizeof(Val)
+            * (row_index.n_ - 1 + (col_index.n_ - 1) * expr_eval.rows()),
+        sizeof(Val), &res, &expr_eval.write_events(), &copy_event);
+    copy_event.wait();
+  } catch (const cl::Error& e) {
+    stan::math::check_opencl_error("uni uni indexing", e);
```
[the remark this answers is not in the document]
t4c1: fixed

src/test/unit/model/indexing/assign_cl_test.cpp
```diff
@@ -0,0 +1,413 @@
...
+          stan::math::matrix_cl<double> m_test_cl = m1_cl;
+          Eigen::VectorXi m_i_test = m1_i;
+          stan::math::matrix_cl<int> m_i_test_cl = m1_i_cl;
+
+          auto ind1_cl = opencl_index(ind1);
+
+          assign(m_test, rvalue(m2, "", ind1), "", ind1);
+          assign(m_test_cl, rvalue(m2_cl, "", ind1_cl), "", ind1_cl);
+          assign(m_i_test, rvalue(m2_i, "", ind1), "", ind1);
+          assign(m_i_test_cl, rvalue(m2_i_cl, "", ind1_cl), "", ind1_cl);
```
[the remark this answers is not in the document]
t4c1: done

src/test/unit/model/indexing/assign_cl_test.cpp
```diff
@@ -0,0 +1,413 @@
...
+  stan::math::matrix_cl<int> m1_i_cl(m1_i);
+  stan::math::matrix_cl<int> m2_i_cl(m2_i);
+  stan::math::matrix_cl<double> m_err(5, 5);
+  stan::math::matrix_cl<int> m_i_err(5, 5);
+  auto indices = std::make_tuple(
+      index_omni(), index_multi(std::vector<int>{1, 3, 2}), index_min(2),
+      index_max(3), index_min_max(2, 3), index_min_max(3, 1));
+  constexpr int N_ind = std::tuple_size<decltype(indices)>::value;
+  stan::math::index_apply<N_ind>([&](auto... Is1) {
+    std::initializer_list<int>{(
```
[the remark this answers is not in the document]
t4c1: yes

src/test/unit/model/indexing/assign_cl_test.cpp
```diff
@@ -0,0 +1,413 @@
...
+  auto indices = std::make_tuple(
+      index_uni(1), index_omni(), index_multi(std::vector<int>{1, 3, 2}),
+      index_min(2), index_max(3), index_min_max(2, 3), index_min_max(3, 1));
+  constexpr int N_ind = std::tuple_size<decltype(indices)>::value;
+  stan::math::index_apply<N_ind>([&](auto... Is1) {
+    std::initializer_list<int>{(
+        [&](const auto& ind1) {
+          stan::math::index_apply<N_ind>([&](auto... Is2) {
+            std::initializer_list<int>{(
+                [&](const auto& ind2) {
```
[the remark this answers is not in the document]
t4c1: done

src/stan/model/indexing/assign_cl.hpp
```diff
@@ -0,0 +1,369 @@
...
+inline void assign(ExprLhs&& expr_lhs, const ExprRhs& expr_rhs,
+                   const char* name, const RowIndex& row_index) {
+  if (std::is_same<RowIndex, index_omni>::value) {
+    stan::math::check_size_match("omni assign", "left hand side rows",
+                                 expr_lhs.rows(), name, expr_rhs.rows());
+    stan::math::check_size_match("omni assign", "left hand side columns",
+                                 expr_lhs.cols(), name, expr_rhs.cols());
+  }
+  rvalue(expr_lhs, name, row_index) = expr_rhs;
+}
```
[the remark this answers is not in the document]
t4c1: Yes, in case we have `index_omni` and two `matrix_cl`s it is worth it.

src/stan/model/indexing/assign_cl.hpp
t4c1: Actually this is what `decltype(auto)` is for. I did not think of it when I was first writing this.

src/stan/model/indexing/assign_cl.hpp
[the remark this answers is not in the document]
t4c1: For rev=rev I found it simpler to add a separate overload for the case where forwarding can be beneficial.

t4c1:
>  I wonder if it would be easier to implement this in math if instead of allowing for general compile time sized matrix_cl types we just support knowing if we are looking at a matrix (-1, -1), vector (-1, 1), or row vector (1, -1).

I think that would be quite a significant refactor, maybe even also affecting stan and cmdstan repos. Is it that much work to handle sizes in the compiler?

src/test/unit/model/indexing/rvalue_cl_test.cpp
SteveBronder: So when we have 

```stan
matrix[N, M] x = exp(y);
x[1:10, 1:10] = exp(x);
```

Or some slicing assignment like that, making sure that the values after the forward pass are correct and that they are put back after the reverse pass.

src/test/unit/model/indexing/rvalue_cl_test.cpp
SteveBronder: idt it's a big deal to have util files around, and we can just use another of the util files as well. It's mostly that the functions here and in the other file are very similar so I don't see a reason to have two sets of them.

src/stan/model/indexing/assign_cl.hpp
SteveBronder: Can you leave a note on the docs on what's going on here? It's a bit hard to read

src/test/unit/model/indexing/assign_cl_test.cpp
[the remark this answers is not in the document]
SteveBronder: Why not have one function that brings in what you set the adjoint to? Then you can just call it `set_adjoints(...)`

src/stan/model/indexing/rvalue_cl.hpp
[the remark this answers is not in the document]
SteveBronder: So there's two cases I'm thinking about

1.

```stan

data {
  int N;
  int idx[N];
}

parameters {
  vector[N] y;
}

model {
  vector[N] x = exp(y[idx]);
}
```
For that in the compiler I would just make an `idx_cl` that is moved over when the model is constructed (aka what you have written here)

```stan
model {
  int idx2[5] = {1, 2, 3, 4, 5}
  vector[5] x = exp(y[idx2]);
  vector[5] x = exp(y[{1, 2, 3, 4 ,5}]);
}
```

For these I think I'd have to do some odd things to figure out when they need wrapped in `to_matrix_cl()` since they will always be copied since they are constructed in the model block. For these cases it would be easier to just have a version of this that copies a vector over in the function.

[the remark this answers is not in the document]
t4c1: I don't understand what is odd about this case. `to_matrix_cl()` needs to be called, if you want to call an OpenCL overload of a function and you don't have the data on OpenCL device. This is the same as any other function with OpenCL overloads.

I would really prefer to keep this consistent with all other functions and not add overloads that accept mix of OpenCL and non-OpenCL arguments.

src/test/unit/model/indexing/assign_cl_test.cpp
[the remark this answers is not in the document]
t4c1: That would be possible if all adjoints were set to the same value in one call of `set_adjoints`. However, to avoid missing any errors, all adjoints need to have different values, even ones set in the same call. So we have two functions, each setting adjoints according to different pattern. This way all adjoints are different.

src/stan/model/indexing/rvalue_cl.hpp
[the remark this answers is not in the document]
rok-cesnovar: I agree with @t4c1 here. The second example is written inefficiently and we dont need to fix those cases. 

Its the same if a user has an expression with only data arguments anywhere outside trans. data. We dont fix those and we dont need to, at least not as a default mode.

src/test/unit/model/indexing/assign_cl_test.cpp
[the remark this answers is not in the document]
rok-cesnovar: Yeah, maybe we could have a few OpenCL models that we compile in the tests. Maybe a 5-10 models for starters.

src/stan/model/indexing/assign_cl.hpp
t4c1: Added some comments.

src/test/unit/model/indexing/rvalue_cl_test.cpp
t4c1: fine

src/test/unit/model/indexing/rvalue_cl_test.cpp
t4c1: Tests for assignments are in `assign_cl_test.cpp`. See for example line 92 in that file.

SteveBronder:
>> I wonder if it would be easier to implement this in math if instead of allowing for general compile time sized matrix_cl types we just support knowing if we are looking at a matrix (-1, -1), vector (-1, 1), or row vector (1, -1).

> I think that would be quite a significant refactor, maybe even also affecting stan and cmdstan repos. Is it that much work to handle sizes in the compiler?

Yeah it would be, let's continue with the scheme you proposed. If I have trouble doing this in the compiler we can think about that

src/stan/model/indexing/rvalue_cl.hpp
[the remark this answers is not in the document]
SteveBronder: Okay I think that makes sense. I'm mostly pinging these things as I'm worried about how hard the compiler impl will be. But once I start on it we can sort out whether these things are actually a difficulty or not and do whatever then. I agree this is fine as is

src/stan/model/indexing/assign_cl.hpp
```diff
@@ -0,0 +1,385 @@
...
+    stan::math::check_size_match("omni assign", "left hand side rows",
+                                 expr_lhs.rows(), name, expr_rhs.rows());
+    stan::math::check_size_match("omni assign", "left hand side columns",
+                                 expr_lhs.cols(), name, expr_rhs.cols());
+  }
+  decltype(auto) lhs = rvalue(expr_lhs.val_op(), name, row_index);
+  math::arena_matrix_cl<double> prev_vals = lhs;
+  lhs = expr_rhs.val();  // assign the values
+  math::reverse_pass_callback(
+      [expr_lhs, expr_rhs, row_index, name, prev_vals]() mutable {
```
SteveBronder, line 272: Is `RowIndex` always trivially copyable? Might be good to include that in the docs if so

src/stan/model/indexing/assign_cl.hpp
```diff
@@ -0,0 +1,385 @@
...
+  decltype(auto) lhs = rvalue(expr_lhs.val_op(), name, row_index, col_index);
+  math::arena_matrix_cl<double> prev_vals = lhs;
+  lhs = expr_rhs.val();  // assign the values
+  math::reverse_pass_callback([expr_lhs, expr_rhs, row_index, col_index, name,
+                               prev_vals]() mutable {
+    auto&& lhs_val_holder
+        = rvalue(expr_lhs.val_op(), name, row_index, col_index);
+    auto&& lhs_adj_holder = rvalue(expr_lhs.adj(), name, row_index, col_index);
+    auto& lhs_val = lhs_val_holder;
+    auto& lhs_adj = lhs_adj_holder;
```
SteveBronder, line 331: Why does there need to be `lhs_val_holder` and `lhs_val`? I usually don't like comments near code, but If there's some precise mechanics going on here it would be nice to have something written down either here or in the doc comment

src/stan/model/indexing/rvalue_cl.hpp
```diff
@@ -0,0 +1,333 @@
...
+  math::check_range("uni indexing", name, expr.cols(), col_index.n_);
+  cl::CommandQueue queue = stan::math::opencl_context.queue();
+  Val res;
+  try {
+    cl::Event copy_event;
+    queue.enqueueReadBuffer(
+        expr.val().buffer(), true,
+        sizeof(Val) * (row_index.n_ - 1 + (col_index.n_ - 1) * expr.rows()),
+        sizeof(Val), &res, &expr.val().write_events(), &copy_event);
+    copy_event.wait();
```
SteveBronder, line 247: Do these expressions need evaluated before we write to their buffer?

SteveBronder (changes requested):
Couple more review comments. For the docs I think we should have a bit more than

```
@tparam Expr type of the expression
```

Ideally it would be nice to have the docs explain the requirements

src/stan/model/indexing/assign_cl.hpp
```diff
@@ -0,0 +1,385 @@
...
+                   const RowIndex& row_index) {
+  if (std::is_same<RowIndex, index_omni>::value) {
+    stan::math::check_size_match("omni assign", "left hand side rows",
+                                 expr_lhs.rows(), name, expr_rhs.rows());
+    stan::math::check_size_match("omni assign", "left hand side columns",
+                                 expr_lhs.cols(), name, expr_rhs.cols());
+  }
+  decltype(auto) lhs = rvalue(expr_lhs.val_op(), name, row_index);
+  math::arena_matrix_cl<double> prev_vals;
+  if (std::is_rvalue_reference<ExprRhs&&>::value) {
```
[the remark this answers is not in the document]
t4c1: Well `ExprRhs&&` is the type of the argument. So I slightly prefer this to the negation.

src/stan/model/indexing/assign_cl.hpp
```diff
@@ -0,0 +1,385 @@
...
+}
+
+/**
+ * Assign one primitive kernel generator expression to a reverse mode kernel
+ * generator expression, using given indices.
+ *
+ * @tparam ExprLhs type of the expression on the left hand side of the
+ * assignment
+ * @tparam ExprRhs type of the expression on the right hand side of the
+ * assignment
```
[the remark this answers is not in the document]
t4c1: I think describing what types are valid kernel generator expressions would get a bit long to copy-paste to every argument of every function. So I will keep it short by using prim/rev/assignable kernel generator expression.

src/stan/model/indexing/assign_cl.hpp
```diff
@@ -0,0 +1,385 @@
...
+ * @tparam RowIndex type of row index
+ * @tparam ColIndex type of columnindex
```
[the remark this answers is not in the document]
t4c1: done

src/stan/model/indexing/assign_cl.hpp
```diff
@@ -0,0 +1,385 @@
...
+ * @tparam ExprLhs type of the expression on the left hand side of the
+ * assignment
+ * @tparam ExprRhs type of the expression on the right hand side of the
+ * assignment
+ * @tparam RowIndex type of index
+ * @param[in,out] expr_lhs expression on the left hand side of the assignment
+ * @param expr_rhs expression on the right hand side of the assignment
+ * @param name Name of lvalue variable
+ * @param row_index index used for indexing `expr_lhs`
+ * @throw std::out_of_range If the index is out of bounds.
```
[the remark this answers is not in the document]
t4c1: Yes, this check and throw happens within the `rvalue` call.

src/stan/model/indexing/assign_cl.hpp
t4c1: No it is not. We need to forward it into lambda. Except that lambda captures do not support forwarding, so we need a workaround by wrapping it in a tuple ... Fixed.

src/stan/model/indexing/assign_cl.hpp
t4c1: I forgot to replace `auto&&` with `decltype(auto)` in some places. Fixed.

src/stan/model/indexing/rvalue_cl.hpp
t4c1: Expressions in general do not have buffers. Only matrices do. So we need to evaluate the expression to get a matrix. Also this is reading the buffer, not writing it.

src/stan/model/indexing/assign_cl.hpp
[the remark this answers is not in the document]
SteveBronder: :+1:

src/test/unit/model/indexing/assign_cl_test.cpp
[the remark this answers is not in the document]
SteveBronder: @serban-nicusor-toptal apologies if you were waiting for my reply. By thumps up I meant yes this should be the same setup as the Stan math library

src/stan/model/indexing/assign_cl.hpp
SteveBronder: Sorry what's the logic with wrapping it in a tuple? I'm not sure I follow that

src/test/unit/model/indexing/rvalue_cl_test.cpp
```diff
@@ -0,0 +1,247 @@
...
+  stan::math::matrix_cl<double> m_cl(m);
+  Eigen::VectorXi m_i(4);
+  m_i << 1, 2, 3, 4;
+  stan::math::matrix_cl<int> m_i_cl(m_i);
+  auto indices = std::make_tuple(
+      index_omni(), index_multi(std::vector<int>{1, 2, 1, 3, 1}), index_min(2),
+      index_max(3), index_min_max(2, 3), index_min_max(3, 1));
+  constexpr int N_ind = std::tuple_size<decltype(indices)>::value;
+  stan::math::index_apply<N_ind>([&indices, &m, &m_cl, &m_i,
+                                  &m_i_cl](auto... Is1) {
```
SteveBronder, line 41: Can you swap these for for_each() as well?

SteveBronder (changes requested):
I think besides the tuple logic Q I have and changes these to `for_each()` this should be good!

SteveBronder:
Oh and also the tests for the Stan repo that turn on the OpenCL tests

SteveBronder:
Ack sorry clicked wrong button

src/stan/model/indexing/assign_cl.hpp
t4c1: So we can not use perfect forwarding, but we want to. So we pass in a tuple by value and forward the tuple element.

SteveBronder: Sorry which one is not trivially copyable? I'm looking at `index.hpp` and the only one I see is `index_multi` right? And the OpenCL assign / lvalues won't see that one because they are going to be pulled out as `matrix_cl<int>` types. So for that what if we just extended `to_arena()` for the indices that was essentially a no-op for the other indices and made a `matrix_cl<int>` in the arena when it sees `matrix_cl<int>`? The rest just have int values / default constructors / descructors, etc.

SteveBronder: Also why the need for perfect forwarding into the reverse pass callback? I mostly just care that we are sure stuff will exist and get properly cleaned up for the reverse pass

t4c1: Because `matrix_cl` is expensive to copy. Although what you suggest with to_arena should work as well and is simpler. Will change to that.

src/test/unit/model/indexing/rvalue_cl_test.cpp
t4c1: done

SteveBronder:
Alright I think this all makes sense! At this point I think all we need is the OpenCL tests on Jenkins for this stuff and it should be good to go. I'm fine with doing that either in a separate PR or this PR

serban-nicusor-toptal:
Hey @SteveBronder today I had an insane day, I'll try to do the Jenkins part tomorrow my morning.

SteveBronder:
Word no worries!

rok-cesnovar:
Do we just merge this and work on the Jenkinsfile separately? I think that makes most sense wrt to reviewing the Jenkinsfile changes and all. I can work with @serban-nicusor-toptal on that.

t4c1:
I would prefer if the tests passes on Jenkins before this is merged. Also reviewing Steves PR with varmat assign bugfixes I got an idea, how I might be able to shorten the code here a bit.

rok-cesnovar:
Ok cool, @serban-nicusor-toptal lets make a PR towards this branch then if you dont mind.

SteveBronder:
Nice that's a big cut! And yeah I agree on Jenkins tests. I'll give your changes another look today

rok-cesnovar:
Good to go?

src/test/unit/mcmc/hmc/hamiltonians/softabs_metric_test.cpp
```diff
@@ -59,6 +59,8 @@ TEST(McmcSoftAbs, sample_p) {
   EXPECT_EQ("", fatal.str());
 }
 
+#ifndef STAN_OPENCL
+
```
SteveBronder, line 63: why this change?

SteveBronder (changes requested):
Couple small fixes

src/test/unit/mcmc/hmc/hamiltonians/softabs_metric_test.cpp
rok-cesnovar: I added this one because this test also uses fvars and we dont support to_matrix_cl for fvars. The alternative is to only run the assign_cl and rvalue_cl tests.

rok-cesnovar:
I have another request. Please add 

```
#ifdef STAN_OPENCL
#include <stan/model/indexing/assign_cl.hpp>
#include <stan/model/indexing/rvalue_cl.hpp>
#endif
```

to https://github.com/stan-dev/stan/blob/develop/src/stan/model/indexing.hpp

This way its automatically included in the model.

src/stan/model/indexing/assign_cl.hpp
```diff
@@ -0,0 +1,276 @@
...
+  math::arena_matrix_cl<double> prev_vals = lhs;
+  lhs = math::value_of(expr_rhs);  // assign the values
+  math::reverse_pass_callback(
+      [expr_lhs, expr_rhs, name, prev_vals,
+       row_index_tup = std::tuple<RowIndex>(std::forward<RowIndex>(row_index)),
+       col_index_tup
+       = std::tuple<ColIndex>(std::forward<ColIndex>(col_index))]() mutable {
+        decltype(auto) lhs_val
+            = rvalue(expr_lhs.val_op(), name, std::get<0>(row_index_tup),
+                     std::get<0>(col_index_tup));
```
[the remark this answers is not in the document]
t4c1: Yes. When using `auto&&` and the function on the right hand side returns a value the variable type is rvalue reference. So it is the same as moving it into whatever function is using is. When using `decltype(auto)` in the same case the variable type is justz a value and it is passes as a lvalue reference into any functions called on it. So I am using `auto&&` wherever the variable is used in exactly one place and `decltype(auto)` when using the variable more than one time.

src/stan/model/indexing/assign_cl.hpp
```diff
@@ -0,0 +1,276 @@
...
+                                 expr_lhs.cols(), name, expr_rhs.cols());
+  }
+  decltype(auto) lhs = rvalue(expr_lhs.val_op(), name, row_index, col_index);
+  math::arena_matrix_cl<double> prev_vals = lhs;
+  lhs = math::value_of(expr_rhs);  // assign the values
+  math::reverse_pass_callback(
+      [expr_lhs, expr_rhs, name, prev_vals,
+       row_index_tup = std::tuple<RowIndex>(std::forward<RowIndex>(row_index)),
+       col_index_tup
+       = std::tuple<ColIndex>(std::forward<ColIndex>(col_index))]() mutable {
```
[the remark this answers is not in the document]
t4c1: fixed

src/stan/model/indexing/assign_cl.hpp
```diff
@@ -0,0 +1,276 @@
...
+ * @throw std::out_of_range If the index is out of bounds.
+ */
+template <typename ExprLhs, typename ScalRhs,
+          require_kernel_expression_lhs_t<ExprLhs>* = nullptr,
+          require_stan_scalar_t<ScalRhs>* = nullptr>
+inline void assign(ExprLhs&& expr_lhs, const ScalRhs& scal_rhs,
+                   const char* name, const index_uni row_index,
+                   const index_uni col_index) {
+  math::block_zero_based(expr_lhs, row_index.n_ - 1, col_index.n_ - 1, 1, 1)
+      = math::constant(scal_rhs, 1, 1);
```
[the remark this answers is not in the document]
t4c1: assignmnet does the check for us.

src/stan/model/indexing/assign_cl.hpp
```diff
@@ -0,0 +1,276 @@
...
+ * @throw std::out_of_range If the index is out of bounds.
+ */
+template <typename ExprLhs, typename ScalRhs,
+          require_rev_kernel_expression_t<ExprLhs>* = nullptr,
+          require_stan_scalar_t<ScalRhs>* = nullptr>
+inline void assign(ExprLhs&& expr_lhs, const ScalRhs& scal_rhs,
+                   const char* name, const index_uni row_index,
+                   const index_uni col_index) {
+  decltype(auto) lhs = math::block_zero_based(
+      expr_lhs.val_op(), row_index.n_ - 1, col_index.n_ - 1, 1, 1);
```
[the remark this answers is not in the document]
t4c1: sure

src/stan/model/indexing/assign_cl.hpp
```diff
@@ -0,0 +1,276 @@
...
+  math::arena_matrix_cl<double> prev_val = lhs;
+  lhs = math::constant(math::value_of(scal_rhs), 1, 1);  // assign the value
+  math::reverse_pass_callback(
+      [expr_lhs, scal_rhs, row_index, col_index, prev_val]() mutable {
+        auto&& lhs_val = math::block_zero_based(
+            expr_lhs.val_op(), row_index.n_ - 1, col_index.n_ - 1, 1, 1);
+        auto&& lhs_adj = math::block_zero_based(
+            expr_lhs.adj(), row_index.n_ - 1, col_index.n_ - 1, 1, 1);
```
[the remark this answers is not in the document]
t4c1: Yeah it probably should. This works as it is just because we dont have a separate overload of `from_matrix_cl` that would actually move rvalue inputs.

SteveBronder (approved):
Cool!

PR #3049, opened 2021-07-14
Fix: Eigen double to matrix var assignment

SteveBronder:
#### Submission Checklist

#### Summary

This fixes the `assign()` errors from https://github.com/stan-dev/stanc3/pull/898 for doing `var<Matrix> = Eigen::Matrix<double>`. Combining this with https://github.com/stan-dev/math/pull/2535 removed the errors on the sample of test models I tried out in the stanc test suite that were previously failing.

Explicit paths had to be added in the callbacks for assignment using `index_uni` and `index_multi`. For all the other functions I wrote a `assign_impl()` function that is just `A = B` for all cases besides `var<Matrix> = Eigen::Matrix<double>`. For that case we do a procedure to update the var values and then in a reverse pass callback put back the old values and reset the adjoints to zero.

#### Intended Effect

Allow assignment from `Eigen::Matrix<double>` to `var<Matrix>` in the stanc generated C++.

#### How to Verify

All the old tests for `var<Matrix>` assignment were pulled out and made into functions that test assignment from both a `var<Eigen::Matrix<double>>` and `Eigen::Matrix<double>`. These can be run with 

```
./runTests.py ./src/test/unit/model/indexing/assign_varmat_test.cpp
```

#### Side Effects

None

#### Documentation

Docs added for new `assign_impl()` function.

#### Copyright and Licensing

### DIFF
--- a/src/stan/model/indexing/assign.hpp
+++ b/src/stan/model/indexing/assign.hpp
@@ -53,7 +53,7 @@ template <
     typename T, typename U,
     require_t<std::is_assignable<std::decay_t<T>&, std::decay_t<U>>>* = nullptr>
 inline void assign(T&& x, U&& y, const char* name) {
-  x = std::forward<U>(y);
+  internal::assign_impl(x, std::forward<U>(y));
 }
 
 /**
@@ -390,7 +390,8 @@ inline void assign(Mat1&& x, Mat2&& y, const char* name, index_min_max idx) {
     stan::math::check_size_match("matrix[reverse_min_max] assign",
                                  "left hand side rows", row_size, name,
                                  y.rows());
-    internal::assign_impl(x.middleRows(idx.max_ - 1, row_size), internal::colwise_reverse(y));
+    internal::assign_impl(x.middleRows(idx.max_ - 1, row_size),
+                          internal::colwise_reverse(y));
     return;
   }
 }
@@ -433,7 +434,8 @@ inline void assign(Mat1&& x, Mat2&& y, const char* name, index_min_max row_idx,
       stan::math::check_size_match("matrix[min_max, min_max] assign",
                                    "left hand side columns", col_size, name,
                                    y.cols());
-      internal::assign_impl(x.block(row_idx.min_ - 1, col_idx.min_ - 1, row_size, col_size), y);
+      internal::assign_impl(
+          x.block(row_idx.min_ - 1, col_idx.min_ - 1, row_size, col_size), y);
       return;
     } else {
       auto row_size = row_idx.max_ - (row_idx.min_ - 1);
@@ -444,7 +446,9 @@ inline void assign(Mat1&& x, Mat2&& y, const char* name, index_min_max row_idx,
       stan::math::check_size_match("matrix[min_max, reverse_min_max] assign",
                                    "left hand side columns", col_size, name,
                                    y.cols());
-      internal::assign_impl(x.block(row_idx.min_ - 1, col_idx.max_ - 1, row_size, col_size), internal::rowwise_reverse(y));
+      internal::assign_impl(
+          x.block(row_idx.min_ - 1, col_idx.max_ - 1, row_size, col_size),
+          internal::rowwise_reverse(y));
       return;
     }
   } else {
@@ -457,7 +461,9 @@ inline void assign(Mat1&& x, Mat2&& y, const char* name, index_min_max row_idx,
       stan::math::check_size_match("matrix[reverse_min_max, min_max] assign",
                                    "left hand side columns", col_size, name,
                                    y.cols());
-      internal::assign_impl(x.block(row_idx.max_ - 1, col_idx.min_ - 1, row_size, col_size), internal::colwise_reverse(y));
+      internal::assign_impl(
+          x.block(row_idx.max_ - 1, col_idx.min_ - 1, row_size, col_size),
+          internal::colwise_reverse(y));
       return;
     } else {
       auto row_size = row_idx.min_ - (row_idx.max_ - 1);
@@ -468,7 +474,9 @@ inline void assign(Mat1&& x, Mat2&& y, const char* name, index_min_max row_idx,
       stan::math::check_size_match(
           "matrix[reverse_min_max, reverse_min_max] assign",
           "left hand side columns", col_size, name, y.cols());
-      internal::assign_impl(x.block(row_idx.max_ - 1, col_idx.max_ - 1, row_size, col_size), y.reverse());
+      internal::assign_impl(
+          x.block(row_idx.max_ - 1, col_idx.max_ - 1, row_size, col_size),
+          y.reverse());
       return;
     }
   }
--- a/src/stan/model/indexing/assign.hpp
+++ b/src/stan/model/indexing/assign.hpp
@@ -2,6 +2,7 @@
 #define STAN_MODEL_INDEXING_ASSIGN_HPP
 
 #include <stan/math/prim.hpp>
+#include <stan/model/indexing/assign_varmat.hpp>
 #include <stan/model/indexing/access_helpers.hpp>
 #include <stan/model/indexing/index.hpp>
 #include <stan/model/indexing/rvalue_at.hpp>
@@ -133,13 +134,13 @@ inline void assign(Vec1&& x, const Vec2& y, const char* name,
     const auto slice_size = idx.max_ - slice_start;
     stan::math::check_size_match("vector[min_max] assign", "left hand side",
                                  slice_size, name, y.size());
-    x.segment(slice_start, slice_size) = y;
+    internal::assign_impl(x.segment(slice_start, slice_size), y);
   } else {
     const auto slice_start = idx.max_ - 1;
     const auto slice_size = idx.min_ - slice_start;
     stan::math::check_size_match("vector[reverse_min_max] assign",
                                  "left hand side", slice_size, name, y.size());
-    x.segment(slice_start, slice_size) = y.reverse();
+    internal::assign_impl(x.segment(slice_start, slice_size), y.reverse());
   }
 }
 
@@ -165,7 +166,7 @@ inline void assign(Vec1&& x, const Vec2& y, const char* name, index_min idx) {
   stan::math::check_range("vector[min] assign", name, x.size(), idx.min_);
   stan::math::check_size_match("vector[min] assign", "left hand side",
                                x.size() - idx.min_ + 1, name, y.size());
-  x.tail(x.size() - idx.min_ + 1) = y;
+  internal::assign_impl(x.tail(x.size() - idx.min_ + 1), y);
 }
 
 /**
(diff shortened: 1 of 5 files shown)


src/stan/model/indexing/assign_varmat.hpp
```diff
@@ -96,40 +175,59 @@ inline void assign(Vec1&& x, const Vec2& y, const char* name,
...
+        }
       }
-    }
-    for (Eigen::Index i = 0; i < x_idx.size(); ++i) {
-      if (likely(x_idx[i] != -1)) {
-        y.adj().coeffRef(i) += prev_vals.coeffRef(i);
+      using stan::math::forward_as;
+      using stan::math::promote_scalar_t;
+      using stan::math::var;
+      if (!is_constant<Vec2>::value) {
```
t4c1, line 211: This if is redundant - it is within an if with the same condition.

src/stan/model/indexing/assign_varmat.hpp
```diff
@@ -147,22 +245,25 @@ inline void assign(Vec1&& x, const Vec2& y, const char* name,
...
-        x.vi_->val_.coeffRef(row_idx_val, col_idx_val) = prev_val;
-        y.adj() += x.adj().coeffRef(row_idx_val, col_idx_val);
-        x.adj().coeffRef(row_idx_val, col_idx_val) = 0.0;
-      });
+  x.vi_->val_.coeffRef(row_idx_val, col_idx_val) = stan::math::value_of(y);
+  stan::math::reverse_pass_callback([x, y, row_idx_val, col_idx_val,
+                                     prev_val]() mutable {
+    x.vi_->val_.coeffRef(row_idx_val, col_idx_val) = prev_val;
+    using stan::math::forward_as;
+    using stan::math::var;
```
t4c1, line 261: [optional] I am not a fan of using declarations for functions and types used only once.

src/stan/model/indexing/assign_varmat.hpp
```diff
@@ -353,7 +490,7 @@ inline void assign(Mat1&& x, const Mat2& y, const char* name,
       for (int i = assign_rows - 1; i >= 0; --i) {
         if (likely(x_row_idx[i] != -1)) {
           prev_vals.coeffRef(i, j) = x.vi_->val_(x_row_idx[i], x_col_idx[j]);
-          y_vals(i, j) = y.vi_->val_(i, j);
+          y_vals(i, j) = y_val(i, j);
```
t4c1, line 493: Unrelated to changes in this PR, but would not it be better to use `.coeff()` and `.coeffRef()` here?

src/stan/model/indexing/assign.hpp
```diff
@@ -2,6 +2,7 @@
 #define STAN_MODEL_INDEXING_ASSIGN_HPP
 
 #include <stan/math/prim.hpp>
+#include <stan/model/indexing/assign_varmat.hpp>
```
t4c1, line 5: I would prefer if you moved the `assign_impl` overloads that are not varmat specific into this file, so there was no need for this include.

src/test/unit/model/indexing/util.hpp
```diff
@@ -156,9 +191,11 @@ auto generate_linear_vector(Eigen::Index n, double start = 0) {
  * @param n Number of cells.
  * @param start Where the linear sequence should start from.
  */
-template <typename Vec = Eigen::Matrix<double, -1, 1>>
+template <typename Vec = Eigen::Matrix<double, -1, 1>,
+          typename RhsScalar = stan::math::var>
 auto generate_linear_var_vector(Eigen::Index n, double start = 0) {
-  using ret_t = stan::math::var_value<Vec>;
+  using ret_t = std::conditional_t<is_var<RhsScalar>::value,
+                                   stan::math::var_value<Vec>, Vec>;
```
t4c1, line 198: The name of this function is now confusing, as depending in template it does not generate a var vector. Same for the matrix one.

t4c1 (changes requested):

src/stan/model/indexing/assign_varmat.hpp
```diff
@@ -55,16 +131,19 @@ namespace model {
...
   double prev_val = x.val().coeffRef(coeff_idx);
-  x.vi_->val_.coeffRef(coeff_idx) = y.val();
+  x.vi_->val_.coeffRef(coeff_idx) = stan::math::value_of(y);
   stan::math::reverse_pass_callback([x, y, coeff_idx, prev_val]() mutable {
     x.vi_->val_.coeffRef(coeff_idx) = prev_val;
-    y.adj() += x.adj().coeffRef(coeff_idx);
+    using stan::math::forward_as;
+    using stan::math::var;
+    if (!is_constant<U>::value) {
+      forward_as<var>(y).adj() += x.adj().coeffRef(coeff_idx);
```
t4c1, line 145:
```suggestion
      math::adjoint_of(y) += x.adj().coeffRef(coeff_idx);
```
This can be simplified.

src/stan/model/indexing/assign_varmat.hpp
```diff
@@ -14,6 +14,82 @@ namespace stan {
...
+ * @param x The var matrix to assign to
+ * @param y The eigen matrix to assign from.
+ */
+template <typename Mat1, typename Mat2, require_var_matrix_t<Mat1>* = nullptr,
+          require_eigen_st<std::is_arithmetic, Mat2>* = nullptr>
+void assign_impl(Mat1&& x, Mat2&& y) {
+  auto prev_vals = stan::math::to_arena(x.val());
+  x.vi_->val_ = std::forward<Mat2>(y);
+  stan::math::reverse_pass_callback([x, prev_vals]() mutable {
+    x.vi_->val_ = prev_vals;
```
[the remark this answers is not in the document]
SteveBronder: Actually reading this I'm kind of confused. When we use `to_arena()` on an `arena_matrix<>` does it make a new matrix? Like in the below shouldn't `prev_vals` point to the same memory as `x.vi_->val_`?

```cpp
  auto prev_vals = stan::math::to_arena(x.val());
  x.vi_->val_ = std::forward<Mat2>(y);
```

In my brain it feels like this should fail but it's passing my tests for omni assignment

src/stan/model/indexing/assign_varmat.hpp
```diff
@@ -14,6 +14,82 @@ namespace stan {
...
+ * @tparam Mat1 A `var_value` with inner type derived from `EigenBase`
+ * @tparam Mat2 A type derived from `EigenBase` with an arithmetic scalar.
+ * @param x The var matrix to assign to
+ * @param y The eigen matrix to assign from.
+ */
+template <typename Mat1, typename Mat2, require_var_matrix_t<Mat1>* = nullptr,
+          require_eigen_st<std::is_arithmetic, Mat2>* = nullptr>
+void assign_impl(Mat1&& x, Mat2&& y) {
+  auto prev_vals = stan::math::to_arena(x.val());
+  x.vi_->val_ = std::forward<Mat2>(y);
```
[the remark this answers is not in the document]
SteveBronder: I'll remove the forward. I think we do not want to do pointer swapping when the rhs is map. The data in the Stan model is made as a map and I think in the scheme we are doing here we could end up accidentally overwriting parts of that map if we just swapped pointers.

src/stan/model/indexing/assign_varmat.hpp
SteveBronder: Yeah I went through and fixed these up!

src/test/unit/model/indexing/util.hpp
SteveBronder: I agree, I changed the name to just `conditionally_generate_linear_var_vector()` etc.

src/stan/model/indexing/assign_varmat.hpp
SteveBronder: Nice! Is there a reason we don't have this for the `matrix<var>` case?

src/stan/model/indexing/assign_varmat.hpp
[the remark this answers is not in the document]
t4c1: Yeah, you are right. `y` would also have to be a rvalue reference for this to be ok.

src/stan/model/indexing/assign_varmat.hpp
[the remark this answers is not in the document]
t4c1: :facepalm: I actually made `arena_matrix` smart enough to do pointer assignment if possible and forgot about that. So both of these assignments modify just pointers, not data. Nothing to change here.

src/stan/model/indexing/assign.hpp
```diff
@@ -14,6 +13,29 @@ namespace stan {
...
+    = require_not_t<stan::math::conjunction<is_var_matrix<T1>, is_eigen<T2>>>;
+/**
+ * Base case of assignment
+ * @tparam T1 Any type that's not a var matrix.
+ * @tparam T2 Any type that's not a var matrix.
+ * @param x The value to assign to
+ * @param y The value to assign from.
+ */
+template <typename T1, typename T2,
+          require_nand_var_matrix_or_eigen<T1, T2>* = nullptr>
```
t4c1, line 32: Reading nand makes my head hurt. Also I think this is wrong and should be just `require_not_var_matrix<T2>`.You can also add `require_any_t<is_var_matrix<T1>, is_eigen<T1>>` if you want, but it is not really needed.

src/stan/model/indexing/assign_varmat.hpp
```diff
@@ -205,24 +178,21 @@ inline void assign(Vec1&& x, const Vec2& y, const char* name,
...
-          if (likely(x_idx[i] != -1)) {
-            forward_as<promote_scalar_t<var, Vec2>>(y).adj().coeffRef(i)
-                += prev_vals.coeffRef(i);
-          }
+      for (Eigen::Index i = 0; i < x_idx.size(); ++i) {
+        if (likely(x_idx[i] != -1)) {
+          math::forward_as<math::promote_scalar_t<math::var, Vec2>>(y)
+              .adj()
+              .coeffRef(i)
+              += prev_vals.coeff(i);
```
t4c1, line 186: [optional] This could also be simplified with `adjoint_of` if we added `coeffRef` to its `nonexisting_adjoint` return type. I know it is another repo ...

src/stan/model/indexing/assign_varmat.hpp
t4c1: It was probably implemented before `adjoint_of`.

t4c1 (changes requested):
a few more

src/stan/model/indexing/assign.hpp
SteveBronder: Yeah I did a bad job at naming this, ended up simplifying it to `require_any_not_t<is_var_matrix<T1>, is_eigen<T2>>`. It has to be the opposite of the one specialization right?

src/stan/model/indexing/assign_varmat.hpp
SteveBronder: I agree that would be v nice, let's not do that in this PR but I'll make a math PR to do that and update it here

t4c1 (approved):

SteveBronder:
Shoot it won't let me click the merge button, can someone click it?

PR #3051, opened 2021-07-26
Revert "Opencl indexing"

SteveBronder:
Reverts stan-dev/stan#3048

We're seeing testing issues upstream in (https://github.com/stan-dev/math/pull/2535#issuecomment-886419617) so I think it's better to do the revert and revert revert PR  until we have things sorted out


bob-carpenter (approved):
Just approving as it's a reversion and tests pass.

PR #3065, opened 2021-09-29
Additional eval return

andrjohns:
Hey @hsbadr, got another rvalue return that needed to be a plain type for a reverse dependency (`idem`) to build.

Andrew


hsbadr (approved):
LGTM. Thanks @andrjohns!

PR #3082, opened 2021-11-09
bool is_empty() for not writing to files - redo of 3081

rok-cesnovar:
#### Submission Checklist

This is just a re-do of #3081 done with the master branch as base. The changes are all Steve's.

#### Copyright and Licensing

Steve Bronder
Rok Češnovar (all I did was create the branch and open the PR though)

### DIFF
--- a/src/stan/callbacks/stream_writer.hpp
+++ b/src/stan/callbacks/stream_writer.hpp
@@ -70,7 +70,10 @@ class stream_writer : public writer {
     }
   }
 
-  inline bool is_empty() const {
+  /**
+   * Check if the writer is writing to an empty stream
+   */
+  inline bool is_empty() const noexcept final {
     return empty_;
   }
 
@@ -86,7 +89,10 @@ class stream_writer : public writer {
    */
   std::string comment_prefix_;
 
-  bool empty_;
+  /**
+   * Used as check for whether output stream needs to be written to.
+   */
+  bool empty_{false};
   /**
    * Writes a set of values in csv format followed by a newline.
    *
--- a/src/stan/callbacks/stream_writer.hpp
+++ b/src/stan/callbacks/stream_writer.hpp
@@ -24,7 +24,8 @@ class stream_writer : public writer {
    *   each comment line. Default is "".
    */
   explicit stream_writer(std::ostream& output,
-                         const std::string& comment_prefix = "", bool is_empty = false)
+                         const std::string& comment_prefix = "",
+                         bool is_empty = false)
       : output_(output), comment_prefix_(comment_prefix), empty_(is_empty) {}
 
   /**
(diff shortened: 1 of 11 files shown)


SteveBronder (approved):
@rok-cesnovar I approved this (though I wrote the original code) If this looks good to you then I think we can merge

rok-cesnovar:
I have been testing this over the weekend (with a change in cmdstan). Still seeing a perf regression :/ Will more details a bit later.

SteveBronder:
Oh dang, can you send me the branch you are trying out? I think you just want to make [this](https://github.com/stan-dev/cmdstan/blob/develop/src/cmdstan/command.hpp#L497) line to 

```cpp
      diagnostic_writers.emplace_back(
          std::make_unique<std::fstream>(diagnostic_filename,
                                         std::fstream::out, "#", true),
```

That should turn off all of the diagnostics to know whether or not writing diagnostics is the issue. You can also do the same for the other writer for output. If that fixes the regression then the actual issue is writing to the stringstream before writing to disk which I think we can be looser about now that I'm thinking about it. I think at the time I assumed multiple threads could write to one `unique_stream_writer<>`, but each stream writer is associated with one output file so we may not actually need that string stream for anything besides the loggers that write to stdout and stderr

rok-cesnovar:
The model used for testing:

```stan
transformed data {
 int N = 10000;
}
parameters {
 vector[N] a;
}
model {
 a ~ normal(0,1);
}
```

Then various versions:

Between each switch, I ran make stan-update, make clean-all, make build. Also set STANC3_VERSION=v2.x.y so the correct stanc3 binary is associated.

- on 2.27.1:

 Elapsed Time: 8.892 seconds (Warm-up)
               13.219 seconds (Sampling)
               22.111 seconds (Total)

- on master (2.28.1):

 Elapsed Time: 8.296 seconds (Warm-up)
               19.1 seconds (Sampling)
               27.396 seconds (Total)

- on fix/diagnostic branch (points to this branch in stan):

 Elapsed Time: 8.752 seconds (Warm-up)
               12.481 seconds (Sampling)
               21.233 seconds (Total)

rok-cesnovar:
Link to branch: https://github.com/stan-dev/cmdstan/tree/fix_diagnostic_writer

So I guess I should have put it better, it looks better, but still not at the level of 2.27.0. Maybe that is fine and expected?

I would be fine with this amount of a perf. regression as long as we know where it comes from.

SteveBronder:
It looks like it's faster than 2.27.1? Which would be before any of these changes

rok-cesnovar:
Yeah, it is faster than 2.28.1. Maybe the rest of it is due to the use of streams and there is not much we can do besides that.

SteveBronder:
Sorry am I not reading your posted times right?

> on 2.27.1:
> Elapsed Time: 8.892 seconds (Warm-up)
> 13.219 seconds (Sampling)
> 22.111 seconds (Total)

> on fix/diagnostic branch (points to this branch in stan):
> Elapsed Time: 8.752 seconds (Warm-up)
> 12.481 seconds (Sampling)
> 21.233 seconds (Total)

Is the top one 2.27.1 and the bottom this branch?

rok-cesnovar:
Ah, sorry I think I botched something. Let me run the benchmark again. I think the last number is wrong. Will run again to avoid any confusion...

SteveBronder:
Closing this as I think I found a simpler solution in https://github.com/stan-dev/stan/pull/3087

PR #3101, opened 2022-02-16
Use is_uninitialized() instead of size() when assigning var matrix types

SteveBronder:
#### Submission Checklist

#### Summary

Fixes a bug found in 2.29. What happens is that for `assign_impl(var<Matrix> x, var<Matrix> y)` we would call `x.size()`, but the `size()` function for var matrices calls the `size()` function for the underlying `vari` like the below.

```
// Inside of the var_value<Matrix> class
Eigen::Index size() {
 return this->vi_.size();
}
```
But for uninitialized var matrices that `vi_` is just a `nullptr` so `this->vi_.size();` is undefined behavior. 

This PR fixes this by using `is_uninitialized()` instead of `size()` inside of `assign_impl`

#### Intended Effect

Fix bug found in 2.29 for assigning var matrix types

#### How to Verify

Tests added to `assign_varmat_test.cpp`, everything can be verified as still working by running 

```
./runTests.py -j4 src/test/unit/model/indexing/
```

#### Documentation

I also updated the docs for `assign_impl` here so things matched up properly

#### Copyright and Licensing


rok-cesnovar (approved):

rok-cesnovar:
There is another issue that we need to get to the bottom of, then we will release the patch Cmdstan https://github.com/stan-dev/cmdstanr/issues/623

PR #3121, opened 2022-04-15
Add assignment for tuples

SteveBronder:
#### Submission Checklist

## Summary

This adds an assign signature for tuples like `stan::assign(tuple, tuple, tuple_name)`

#### Intended Effect

As stated above

#### How to Verify

Tests for tuples of same and mixed types have been added to the tests and can be run with 

```
./runTests.py ./src/test/unit/model/indexing/assign_test.cpp 
```

#### Side Effects

There's one odd thing to call out which is that we cannot use standard assignment via the `assign` on line 59 of `assign.hpp`. So what we do instead of just have one assign for tuples with the exact same inner types and another where the inner types can differ (which would be like the lhs has a `matrix<var>` inside but the rhs is `matrix<double>`

#### Documentation

Documentation added for each of the tuple assignments.

#### Copyright and Licensing

### DIFF
--- a/src/stan/model/indexing/access_helpers.hpp
+++ b/src/stan/model/indexing/access_helpers.hpp
@@ -143,6 +143,23 @@ void assign_impl(Mat1&& x, Mat2&& y, const char* name) {
     x = stan::math::var_value<plain_type_t<Mat2>>(std::forward<Mat2>(y));
   }
 }
+
+template <typename... Types>
+struct is_tuple_impl : std::false_type {};
+template <typename... Types>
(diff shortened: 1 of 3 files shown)


SteveBronder:
@WardBrian actually I need to test this also works with `std::vector<tuple>` as well

WardBrian:
Yeah, the standard set of tests for anything with tuples should include some mix of the following:

a tuple
An array of tuples 
A tuple which contains an array
A tuple which contains a tuple
A tuple which contains an array of tuples


Usually if you have enough logic to handle those 4, the further recursions work out.

PR #3144, opened 2022-10-06
Add a #define flag to enable nested autodiff in model_base class

WardBrian:
#### Submission Checklist

#### Summary

The model base class exposes overloads to all the log_prob functions for `double` and `math::var`. It would be useful to expose `math::fvar<var>` overloads to allow higher-order autodiff from base class instances, but this does not work with all Stan models (particularly ODE models and other uses of implicit functions).

This PR introduces opt-in only overloads. This allows downstream users, who want to provide model Hessians or Hessian-vector products, to still use `stan::model::model_base` as a type. Currently, the Hessian functionality in (e.g.) `stan::model::hessian` relies on a template type which can never be filled by `stan::model::model_base`; after this PR it would be possible if the define `STAN_MODEL_AD_HESSIAN` is used.

These signatures are identical to the previously existing ones except for their scalar type. 

To make this even more concrete, in an interface like [bridgestan](https://github.com/roualdes/bridgestan), we can use this same flag to switch from using finite difference Hessians to full-AD for models which support it.

#### Intended Effect

The ability to use `model_base` instances in `stan::model::hessian` on an opt-in basis.

#### Side Effects

Hopefully none.

#### Documentation

I'm not sure where this would be best to document

#### Copyright and Licensing

Simons Foundation

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -36,3 +36,25 @@ jobs:
           echo "CXX=clang++-${{ env.clangppVersion }}" >> make/local
           echo "O=0" >> make/local
           python runTests.py -j 2 src/test/unit
+
+run-fvar-var-tests:
+    name: run stan fvar-var model tests
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-latest]
+        python-version: [3.8]
+      fail-fast: false
+    steps:
+      - name: Check out source code
+        uses: actions/checkout@v2
+        with:
+          submodules:  true
+      - name: Install clang++
+        run: sudo apt-get install clang-${{ env.clangppVersion }}++
+      - name: Run tests
+        run: |
+          echo "CXX=clang++-${{ env.clangppVersion }}" >> make/local
+          echo "CXXFLAGS+=-DSTAN_MODEL_FVAR_VAR" >> make/local
+          echo "O=0" >> make/local
+          python runTests.py -j 2 src/test/unit/model/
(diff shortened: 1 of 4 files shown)


WardBrian:
@bob-carpenter agreed it is better to use a different name. 

I’m not sure how to test this - I don’t know if our test framework lets me specify a `#define` for only some tests, but I can look. In principle, if the fvar versions did disagree, that would be a bug in Stan math or stanc3, not in these signatures

rok-cesnovar (approved):
Looks good to me. Regarding tests, I agree with Brian's statement: "In principle, if the fvar versions did disagree, that would be a bug in Stan math or stanc3, not in these signatures".

bob-carpenter:
>  I don’t know if our test framework lets me specify a #define for only some tests

I suggest asking @serban-nicusor-toptal for help.  I suspect it's already possible because we probably need it for testing all the threading and GPU stuff, which are also controlled by environment variables.

rok-cesnovar:
You can also always add another Github Action test job, specify whatever you need to and run a few tests. Like we do here https://github.com/stan-dev/math/blob/cd60e75a4e63fa5c321ed6effcf825adfbde92c0/.github/workflows/header_checks.yml#L66

WardBrian:
I've added tests to the existing file which tested the other signatures, and a new GHA which runs just the `src/test/unit/model/` folder with the `STAN_MODEL_FVAR_VAR` define

rok-cesnovar:
Great! Merge once Jenkins gives the green light.

PR #3147, opened 2022-10-19
Improve CI times

WardBrian:
#### Summary

The step "Integration Linux" is identical to two steps in stanc3's CI ("Compile tests - good" and "Compile tests - example models") but takes ~3 hours while those two combined take 65 minutes. So far as @serban-nicusor-toptal and I can tell, the only difference between the two setups is PRECOMPILED_HEADERS and the misspecified PARALLEL setting.

#### Copyright and Licensing

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -64,7 +64,7 @@ String stan_pr() {
         env.BRANCH_NAME
     }
 }
-String integration_tests_flags() { 
+String integration_tests_flags() {
     if (params.compile_all_model) {
         '--no-ignore-models '
     } else {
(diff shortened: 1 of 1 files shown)


WardBrian:
@rok-cesnovar we'd also appreciate if you had any thoughts on why the two would be so different here

WardBrian:
These changes immediately dropped that step from 3 hours down to 59 minutes, which is great. I think this is ready for review, unless @rok-cesnovar has further suggestions or any objections.

If I recall correctly, the reason we had these disabled was due to them causing intermittent failures on the Amazon EC2 instances. I think that the Flatiron machines mean we need to be less cautious (clearly the Stanc3 workflow has worked in that repo)

PR #3159, opened 2022-12-18
Allow providing external dependencies

andrjohns:
#### Submission Checklist

#### Summary

This PR follows the work in the [Math library](https://github.com/stan-dev/math/pull/2834) to allow users to provide external dependencies for Stan (instead of the bundled `rapidjson` and `stan_math`).

The PR introduces the following makeflags:

- `MATH_INC`
- `RAPIDJSON_INC`

Which can be used to provide paths to external dependencies

#### Intended Effect

Allow for increased portability and future options for packaging with [homebrew](https://github.com/stan-dev/math/issues/2723) and [linux repos](https://github.com/stan-dev/math/pull/2834#issuecomment-1344112159)

###### Homebrew

```
brew install rapidjson
```

###### Debian/Ubuntu

```
sudo apt install rapidjson-dev
```

#####

Add to `make/local`: 

###### Homebrew
```
RAPIDJSON_INC = /opt/homebrew/Cellar/rapidjson/1.1.0/include
```

###### Debian/Ubuntu
```
RAPIDJSON_INC = /usr/include/
```

#### Side Effects

N/A

#### Documentation

N/A

#### Copyright and Licensing

### DIFF
--- a/makefile
+++ b/makefile
@@ -14,8 +14,17 @@ help:
 -include $(HOME)/.config/stan/make.local  # user-defined variables
 -include make/local                       # user-defined variables
 
-MATH ?= lib/stan_math/
-RAPIDJSON ?= lib/rapidjson_1.1.0/
+ifdef MATH_INC
+  MATH ?= $(MATH_INC)
+else
+  MATH ?= lib/stan_math/
+endif
+
+ifdef RAPIDJSON_INC
+  RAPIDJSON ?= $(RAPIDJSON_INC)
+else
+  RAPIDJSON ?= lib/rapidjson_1.1.0/
+endif
 
 -include $(MATH)make/compiler_flags
 -include $(MATH)make/dependencies


makefile
```diff
@@ -14,8 +14,17 @@ help:
 -include $(HOME)/.config/stan/make.local  # user-defined variables
 -include make/local                       # user-defined variables
 
-MATH ?= lib/stan_math/
-RAPIDJSON ?= lib/rapidjson_1.1.0/
+ifdef MATH_INC
```
syclik, line 17: @andrjohns, I believe the existing code does what you want and is cleaner makefile code. 

In your make/local, if you change your example to use the existing variables, it should work. 

Does it not? If not, we can trace why it's not picking up on changes made at the make/local level. 

* could you also explain the reason for duplicating the variables?
* could you give some context to the names? (Why _INC instead of something more descriptive if we're going that way? Include is ambiguous as a term here... source include or library include?)

syclik (changes requested):
@andrjohns I made some comments, please ping me with any questions, comments, or if you want to have a discussion about this.

syclik:
Maybe what the repo needs are better examples...

makefile
andrjohns: This is for consistency with the existing approach with providing external headers for the TBB in the Math library (described in the README [here](https://mc-stan.org/math#title3)), where the `_INC` suffix denotes the path to header includes and the `_LIB` suffix denotes the path to shared libraries (i.e., dll/so/dylib).

I felt that having the `_INC` suffix for the math library and rapidjson made for a unified approach (as it is also being used in the linked Math library PR for the other dependencies)

andrjohns: Hmm but after thinking about it more, if the current makefiles are already compatible then it's unnecessary to be doing this. I'll close this PR and instead just check that the current makefiles work without issue - and open a PR with any fixes needed if not

PR #3165, opened 2023-02-18
Feature/3161 parse tuples json

mitzimorris:
#### Submission Checklist

#### Summary
Updates json parser and data handler to allow tuples, including tuples of tuples, arrays of tuples, tuples of arrays of tuples, etc.

#### Intended Effect
Allow users to supply init files of JSON data corresponding to Stan program tuple variables.

#### How to Verify
Unit tests.

#### Side Effects
Parser allows nested objects.

#### Documentation
N/A

#### Copyright and Licensing


src/test/unit/util.hpp
```diff
@@ -89,6 +89,39 @@ void match_csv_columns(const Eigen::MatrixXd& samples,
...
+  return path_separator;
+}
+
+/**
+ * Returns the path as a string with the appropriate path separator.
+ *
+ * @param file_path vector of strings representing path to the file
+ * @return the string representation of the path
+ */
+std::string paths_to_fname(const std::vector<std::string>& path) {
```
WardBrian, line 113: Are these intentionally duplicates of the ones in `test/unit/io/json/util`?

src/test/test-data/array_tuple_multi.json
```diff
@@ -0,0 +1,10 @@
+{ "x" : [ { "1" : [ { "1" : 111, "2": [ 1, 2 ] },
```
WardBrian, line 1: Should all of these test files be under the `unit/io` folder? I noticed there is a `test_csv_files` directory there, so it seems like a natural place

src/test/unit/io/json/json_data_tuples_test.cpp
```diff
@@ -0,0 +1,163 @@
...
+  test_int_var(jdata, "x.2.1", expected_vals_i, expected_dims);
+  expected_vals_r.clear();
+  expected_vals_r.push_back(6.28);
+  test_real_var(jdata, "x.2.2", expected_vals_r, expected_dims);
+  expected_vals_r.clear();
+  expected_vals_r.push_back(0.0000123);
+  test_real_var(jdata, "y", expected_vals_r, expected_dims);
+}
+
+// array[2] (array[3] real, real) x;
```
WardBrian, line 58: It seems like this comment is out of date. I like having them here (I find it makes the tests much more readable) but it will be difficult to enforce that they're correct.

src/stan/io/json/json_data_handler.hpp
```diff
@@ -81,100 +369,156 @@ class json_data_handler : public stan::json::json_handler {
...
-    reset();
-  }
-
-  void promote_to_double() {
-    if (is_int_) {
-      for (std::vector<int>::iterator it = values_i_.begin();
-           it != values_i_.end(); ++it)
-        values_r_.push_back(*it);
-      is_int_ = false;
+  void end_array() {
```
WardBrian, line 471: I'd appreciate a couple high-level comments which provide an overview of the relationships between these functions/the invariants between them

WardBrian (changes requested):
First round of review, mostly comments and questions.

The testing for this looks really robust, I appreciate all the effort there a lot.

src/test/unit/io/json/util.hpp
```diff
@@ -0,0 +1,177 @@
...
+ * @param filename path to the file
+ */
+void show_parse(const std::string &fileName) {
+  recording_handler handler;
+  std::ifstream ifs(fileName.c_str(),
+                    std::ios::in | std::ios::binary | std::ios::ate);
+  ifs.seekg(0, std::ios::beg);
+  stan::json::rapidjson_parse(ifs, handler);
+  std::cout << handler.os_.str() << std::endl << std::flush;
+}
```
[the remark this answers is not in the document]
mitzimorris: another debugging function.

src/stan/io/json/json_data_handler.hpp
```diff
@@ -7,72 +7,360 @@
...
+    // only used for debugging
+    std::stringstream ss;
+    ss << " num dims: " << dims.size() << "\tdim szs: ";
+    for (auto& x : dims)
+      ss << " " << x;
+    ss << "\tdims_acc cts: ";
+    for (auto& x : dims_acc)
+      ss << " " << x;
+    return ss.str();
+  }
```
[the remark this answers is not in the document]
mitzimorris: removed.

src/stan/io/json/json_data_handler.hpp
```diff
@@ -7,72 +7,360 @@
...
+  size_t array_start_i;  // index into values_i
+  size_t array_start_r;  // index into values_r
+  int event;
+
+  void reset_values() {
+    values_r.clear();
+    values_i.clear();
+    array_start_i = 0;
+    array_start_r = 0;
+  }
```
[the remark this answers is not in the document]
mitzimorris: adding code comments.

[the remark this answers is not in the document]
mitzimorris: the answer is yes, very intentional.  `values_r` and `values_i` are used to accumulate all values for an array element.  as soon as the handler gets the next key, it copies or appends the contents of `values_r` and `values_i` over to the corresponding variable entry in `vars_r` and `vars_i`.   once copied, as part of function `save_key_value_pair`, we clear the accumulators for reuse for the next variable definition.

src/stan/io/json/json_data_handler.hpp
```diff
@@ -7,72 +7,360 @@
...
+
+  void unexpected_error(std::string where) {
+    std::stringstream errorMsg;
+    errorMsg << "Variable " << where << " ill-formed data.";
+    throw json_error(errorMsg.str());
+  }
+
+  // for debugging only
+  void dump_state(std::string where) {
+    std::string slot_type("unknown");
```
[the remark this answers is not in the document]
mitzimorris: removed

src/test/unit/io/json/util.hpp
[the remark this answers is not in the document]
mitzimorris: removed util debug functions `show_parse` and `show_file`.

src/test/test-data/array_tuple_multi.json
mitzimorris: moved to directory `unit/io/test_json_files`

src/test/unit/io/json/json_data_tuples_test.cpp
mitzimorris: agreed.  JSON doesn't allow comments and Stan doesn't allow string variables, so no way to put them closer to the JSON data.   will remove.

src/stan/io/json/json_data_handler.hpp
```diff
@@ -81,100 +369,156 @@ class json_data_handler : public stan::json::json_handler {
...
     }
-    if (dim_idx_ > 0 && dim_last_ == dim_idx_) {
-      std::stringstream errorMsg;
-      errorMsg << "variable: " << key_ << ", error: non-scalar array value";
-      throw json_error(errorMsg.str());
+    event = meta_event::KEY;
+    reset_values();
+    key_stack.push_back(key);
+    if (slot_types_map.count(key_str()) == 0) {
+      slot_types_map[key_str()] = meta_type::SCALAR;
```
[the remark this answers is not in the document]
mitzimorris: won't the compiler take care of this?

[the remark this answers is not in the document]
mitzimorris: changing - makes code much easier to read.

src/stan/io/json/json_data_handler.hpp
```diff
@@ -81,100 +369,156 @@ class json_data_handler : public stan::json::json_handler {
...
+  }
+
+  void start_object() {
+    event = meta_event::OBJ_OPEN;
+    if (is_init())
+      return;
+    if (slot_types_map[key_str()] == meta_type::ARRAY) {
+      slot_types_map[key_str()] = meta_type::ARRAY_OF_TUPLES;
+    } else if (slot_types_map[key_str()] == meta_type::SCALAR) {
+      slot_types_map[key_str()] = meta_type::TUPLE;
```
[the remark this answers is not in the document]
mitzimorris: changed.

src/test/unit/util.hpp
mitzimorris: added twice by mistake - leaving here, removing from `test/unit/io/json/util`.

src/stan/io/json/json_data_handler.hpp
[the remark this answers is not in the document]
WardBrian: Just for completeness I checked and the answer is "probably not". GCC12 doesn't seem to at either `-O2` or `-O3`, and my example is less complicated than a stateful class: https://godbolt.org/z/8exzrG44M

I would have put at least 50% probability on it doing so, so this is interesting to me!

src/stan/io/json/json_data_handler.hpp
mitzimorris: added high-level comments to the event handlers for key as well as tuple and array open, close.
this was a very good way to identify code cruft as well as a way to tighten up processing - in particular, for single-dimensional arrays, we don't need to do anything to the vector of values, just save it.

mitzimorris:
ready for re-review.

src/test/unit/io/json/json_data_tuples_test.cpp
```diff
@@ -0,0 +1,180 @@
...
+
+#include <boost/limits.hpp>
+#include <boost/math/concepts/real_concept.hpp>
+#include <boost/math/special_functions/fpclassify.hpp>
+#include <gtest/gtest.h>
+#include <complex>
+#include <iostream>
+#include <fstream>
+
+// tuple(int, real) x - also real y;
```
WardBrian, line 18: Do we have a way of testing bad inputs?

I noticed some worse-than-before errors for things like repeated keys, e.g. if you changed this test file to

```json
{ "x": {"1": 1, "2": 6.28},
"y": 0.0000123,
"x": 0.0000123
}
```

The test still passes (e.g. redefining `x` does not error out).

Similarly, re-defining an name which previously defined an array-of-tuples yields ` error: non-rectangular array.` which isn't really the right error for this. To see this, update `array_tuple_multi.json` to also have an `"x": []` element.

src/stan/io/json/json_data_handler.hpp
```diff
@@ -7,174 +7,527 @@
...
+          std::pair<std::vector<int>, std::vector<size_t>> pair;
+          pair = make_pair(values_i, dims);
+          vars_i[key_str()] = pair;
+        } else {
+          std::pair<std::vector<double>, std::vector<size_t>> pair;
+          pair = make_pair(values_r, dims);
+          vars_r[key_str()] = pair;
+        }
+      } else {
+        if (!is_array_tuples(key_stack)) {
```
WardBrian, line 242: This seems suspect to me - why does the `redefine variable` error only occur for non-array-tuples?

src/stan/io/json/json_data_handler.hpp
```diff
@@ -7,174 +7,527 @@
...
+/** Enum which tracks handler events.
+    Used to identify first and last slot of a tuple.
+*/
+struct meta_event {
+  enum {
+    OBJ_OPEN = 0,   // {
+    OBJ_CLOSE = 1,  // }
+    KEY = 2,
+  };
+};
```
WardBrian, line 52: It seems like we're only ever actually care if we're in OBJ_OPEN when we see a key, so do we need to store the additional states? I still have a hard time following the `save_key_value_pair` logic this is tied to

WardBrian (commented):
I started reviewing before the new comments, which may have answered some questions (they look great!)

A few more style things, and then a request for testing for error messages (to whatever extent that is possible). In particular, re-definition logic seems to not work correctly right now.

src/stan/io/json/json_data_handler.hpp
```diff
@@ -7,174 +7,486 @@
...
+    }
+    int_slots_map[key_str()] = false;
+  }
+
+  /* Save non-tuple vars and innermost tuple slots to vars_i and vars_r.
+   * For arrays of tuples we need to check that new elements are consistent
+   * with previous tuple elements.
+   */
+  void save_key_value_pair() {
+    if (0 == key_stack.size())
```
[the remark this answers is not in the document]
mitzimorris: done!

src/stan/io/json/json_data_handler.hpp
```diff
@@ -7,174 +7,486 @@
...
+      return;
+    if (slot_types_map.count(key_str()) < 1)
```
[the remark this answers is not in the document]
mitzimorris: done

src/stan/io/json/json_data_handler.hpp
```diff
@@ -7,174 +7,486 @@
...
+      unexpected_error(key_str());
+    if (slot_types_map[key_str()] == meta_type::TUPLE
+        || slot_types_map[key_str()] == meta_type::ARRAY_OF_TUPLES) {
+      {
+      }
+    } else {
```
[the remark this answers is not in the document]
mitzimorris: thanks - it's much cleaner to say
```
if (slot_types_map[key] == meta_type::SCALAR
        || slot_types_map[key] == meta_type::ARRAY) {
```

src/stan/io/json/json_data_handler.hpp
```diff
@@ -7,174 +7,486 @@
...
+  }
+
+  void promote_to_double() {
+    if (int_slots_map[key_str()]) {
+      for (std::vector<int>::iterator it = values_i.begin();
+           it != values_i.end(); ++it)
+        values_r.push_back(*it);
+      array_start_r = array_start_i;
+    }
+    int_slots_map[key_str()] = false;
```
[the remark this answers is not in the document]
mitzimorris: doh!

src/stan/io/json/json_data_handler.hpp
```diff
@@ -7,174 +7,486 @@
...
+      unexpected_error(key);
+    }
+    slot_dims_map[key] = update;
+  }
+
+  void promote_to_double() {
+    if (int_slots_map[key_str()]) {
+      for (std::vector<int>::iterator it = values_i.begin();
+           it != values_i.end(); ++it)
+        values_r.push_back(*it);
```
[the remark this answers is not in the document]
mitzimorris: cleaned up.

src/test/unit/io/json/json_data_tuples_test.cpp
mitzimorris: problem is this:  the tuple results in Stan program vars "x.1" and "x.2"
the scalar var is just "x"
code needs to check that prefix is unique.   will add logic for this.

WardBrian: It should be valid to also insert “x” as an (empty) variable into the map, if that would get us the errors “for free”

src/stan/io/json/json_data_handler.hpp
mitzimorris: I think logic at line 222 is the problem

src/stan/io/json/json_data_handler.hpp
```diff
@@ -7,174 +7,525 @@
...
+      key = boost::algorithm::join(key_stack, ".");
+      unexpected_error(key);
+    }
+    slot_dims_map[key] = update;
+  }
+
+  void promote_to_double() {
+    if (int_slots_map[key_str()]) {
+      int_slots_map[key_str()] = false;
+      values_r.reserve(values_i.size());
```
WardBrian, line 201: Is values_r always empty at this point? If not, we need to `reserve(values_r.size() + values_i.size())` (reserve is for total size, not additional size)

mitzimorris: yes, `values_r` is always empty.   we're just processing one slot's worth of data here.

mitzimorris:
re-review?

mitzimorris:
I broke the unit tests for ill-formed JSON which the handler should detect and report into from file "json_data_test.cpp" into a new test file, "ill_formed_json_test.cpp", and then added more tests for redefining tuple and array of tuple vars.

src/stan/io/json/json_data_handler.hpp
```diff
@@ -237,11 +237,8 @@ class json_data_handler : public stan::json::json_handler {
           vars_r[key] = pair;
         }
       } else {
-        if (!is_array_tuples(key_stack)) {
-          std::stringstream errorMsg;
-          errorMsg << "Attempt to redefine variable: " << key << ".";
-          throw json_error(errorMsg.str());
-        }
+        if (slot_types_map[key_stack[0]] != meta_type::ARRAY_OF_TUPLES)
+          unexpected_error(key);
```
WardBrian, line 241: This change here seems to have broken tuples-of-array-of-tuples.

e.g., this Stan decl:

```stan
  tuple(array[2] real, int, array[3] tuple(real, array[4] int)) tuple_arr_tuple;
```

and this JSON

```json
{
  "tuple_arr_tuple": {
    "1": [ 5.6413957047783221, 6.86041453790074 ],
    "2": 3,
    "3": [
      { "1": 6.32749625894004, "2": [ 2, 3, 6, 4 ] },
      { "1": 3.3358848771578735, "2": [ 5, 3, 3, 2 ] },
      { "1": 2.3144782736186058, "2": [ 3, 4, 2, 5 ] }
    ]
  }
}
```


This previously worked but now raises `Variable tuple_arr_tuple.3.1 ill-formed data.`

src/test/unit/io/json/json_ill_formed_test.cpp
```diff
@@ -0,0 +1,206 @@
...
+TEST(errJson, jsonData_redefine_var_3) {
+  std::vector<std::string> json_path;
+  json_path = {
+      "src", "test", "unit", "io", "test_json_files", "redefine_vars_3.json"};
+  std::string filename = paths_to_fname(json_path);
+  std::string txt = file2str(filename);
+  test_exception(txt, "Attempt to redefine variable: x.");
+}
+
+TEST(errJson, jsonData_redefine_var_4) {
```
WardBrian, line 199: This new file is great, thanks for adding. 

Could you also add some tests for:

- Non-regular arrays, e.g.  `"x": [ {"1": 0, "2": 1}, {"1": 1}]` and `"x": [ {"1": 0}, {"1": 1, "2": 1}]`
- Changing tuple types in an array, e.g. `"x": [ {"1": 0, "2": 1}, {"1": [1.5], "2":2}]`
- More redefines: What happens if I have a tuple `"x": {"1": [1,2,3]}` and then I add a JSON entry with the name `"x.1"`? Can you also add the reversed-order tests of the redefine_vars_1 and redefine_vars_4 tests (e.g. for vars_1, the order where the tuple is second, and for vars_4, the order where the array-of-tuples is first)

WardBrian: In general I think we can't have too much testing of these edge cases, so I'm happy to generate more as well.

WardBrian (changes requested):
The latest changes seemed to have introduced a regression, which suggests a new case to test against! (tuple-of-array-of-tuple seems to not trigger the array-of-tuple logic)

And a few more concrete failure tests I'd like to see

src/stan/io/json/json_data_handler.hpp
```diff
@@ -7,174 +7,527 @@
...
+    key_stack.pop_back();
+  }
+
+  /* Process array variables
+   *  a. for array of tuples, concatenate dimensions
+   *  b. if multi-dim array, convert vector of values
+   *      from row-major order to column-major order.
+   * Update vars_i and vars_r accordingly.
+   */
+  void convert_arrays() {
```
WardBrian, line 275: I think this function is doing the wrong thing for arrays inside tuples.

Given the decl ` array[2] tuple(real, simplex[4]) t;` and json

```json
{
  "t": [
    {
      "1": 5.8215377737734162,
      "2": [
        0.014580671457001414, 0.30736191423550285, 0.2920816565582352,
        0.38597575774926052
      ]
    },
    {
      "1": 5.6235875354192935,
      "2": [
        0.26041728896763167, 0.29828428322619338, 0.22664074293149306,
        0.2146576848746819
      ]
    }
  ]
}
```

I would expect that `vals_r("t.2")` would return `{0.014580671457001414, 0.30736191423550285, 0.2920816565582352, 0.38597575774926052, 0.26041728896763167, 0.29828428322619338, 0.22664074293149306, 0.2146576848746819}`

E.g. the returned vector is the result of concatinating `t[0].2` and `t[1].2`.

The actual result currently returned is `{0.0145807, 0.260417, 0.307362, 0.298284, 0.292082, 0.226641, 0.385976, 0.214658}`

Which is the result of striping `t[0].2` and `t[1].2`

The later would be difficult for the compiler to correctly un-fold, and I don't think it makes sense to "colum-majorize" these tuple elements

WardBrian: This wasn't explicit in the design doc, I've updated the example there to clarify

mitzimorris: given that we've got the `slot_types_map`, we should be able to handle this.

but when would someone want to do this in a Stan program?  this means that a 2-D array is sometimes row-major, sometimes column-major.   do you have examples of this kinds of operations a user would want to do and why?  is it because we expect users to iterate over an array of tuples element by element?

WardBrian: >  this means that a 2-D array is sometimes row-major, sometimes column-major.

Any 2-D array declared by the user should still be column-major. But, if you have an array of tuples, and an element of those tuples is an array, it creates a "phantom" 2-D array in the `var_context` which needs to be read into tuples chunk-by-chunk.

In pseudocode, here is how the `t` above is read into memory

```
vector<tuple<double, Eigen::Vector>> t = //init to NANs;

vector<double> t_1_flat = context.vars_r("t.1");
int t_2_flat_pos = 1;
vector<double> t_2_flat = context.vars_r("t.2")
int t_2_flat_pos = 1;

for ( i = 0; i < 2 < i++) {
  tuple(double, Eigen::Vector> temp;
  std::get<0>(temp) = t_1_flat[t_1_flat_pos]; 
  t_2_flat_pos = t_2_flat_pos + 1;

  std::get<1>(temp) = t_2_flat[t_2_flat_pos: t_2_flat_pos + 4]; // IMPORTANT LINE HERE
  
  t[i] = temp;
}
```

If we treat "t.2" like any other 2D array, the "`IMPORTANT LINE HERE`" part goes from a simple indexing slice to a very expensive memory striping problem. 

You make a good point about what happens if you have a `array[2] tuple(double, array[2,3]) t2;`, where the inner-array _is_ actually 2D, so now the third dimension is the "phantom" one. Again, it is important that we have the memory such that `t2[1].2` is a contiguous chunk, and `t2[2].2` follows it. I think we can accomplish this by doing the column major conversion not at the complete end of a key, but at the end of the array literals?


Does this make sense? I'm not sure I conveyed that super clearly

WardBrian: I didn't realize it before, but this does imply that `array[2] tuple(array[3] real, array[4] int) x` actually _should_ be serialized differently than `tuple(array[2, 3] real, array[2, 4] int) x`. In the second case we'd want the column major behavior, but it doesn't really make sense in the first.

mitzimorris: great - given the behavoir you describe, this makes sense.

mitzimorris: question:  how will `write_array` serialize variable `array[2] tuple(array[3] real, array[4] int) x` and how will it serialize var `tuple(array[2, 3] real, array[2, 4] int) x`?

WardBrian: `write_array`s only contract seems to be that it writes out parameters in the same order `constrained_param_names` does. And those two would have different orders of naming. I believe these are equivalent to the in-order traversal if you think of the type as a tree.

Currently, the first generates this as a CSV header:
```
x.1:1.1,x.1:1.2,x.1:1.3,x.1:2.1,x.1:2.2,x.1:2.3,x.1:2.4,x.2:1.1,x.2:1.2,x.2:1.3,x.2:2.1,x.2:2.2,x.2:2.3,x.2:2.4
```

The second will generate:
```
x:1.1.1,x:1.2.1,x:1.1.2,x:1.2.2,x:1.1.3,x:1.2.3,x:2.1.1,x:2.2.1,x:2.1.2,x:2.2.2,x:2.1.3,x:2.2.3,x:2.1.4,x:2.2.4
```

(as per the design doc, `:` indicates a tuple index, and `.` is still used for all other indices)

Unwinding all this in cmdstanpy/r will be...fun.

mitzimorris: fun indeed - but it only needs to be done once.

src/test/unit/io/json/json_ill_formed_test.cpp
mitzimorris: > More redefines: What happens if I have a tuple "x": {"1": [1,2,3]} and then I add a JSON entry with the name "x.1"? What happens if the order is reversed and I have a json entry for "x.1" and then a tuple called "x" later?

this isn't possible - "x.1" is not a valid variable name - `{ "x.1" : 3.14 }` is invalid - or should be.

WardBrian: It’s not a valid name in Stan, but someone could type it into a json file

mitzimorris: it looks like the JSON parser isn't checking for valid Stan variable name, so
```
{
    "x.1": 42,
    "x": {"1": 1, "2": 6.28}
}
```
parses just fine.
we could either disallow Stan variable names which aren't legal - the most robust thing to do would be to parse them, but then not save them.

mitzimorris: got a handy regex for legal Stan variable name in C++?

WardBrian: I think `[a-zA-Z][a-zA-Z0-9_]*` should work. (Starts with a letter, can be alphanumeric plus underscores)

technically we also don’t allow it to end in an undescore but that is harder to check with just a regex

src/stan/io/json/json_data_handler.hpp
mitzimorris: fixed, added to unit tests.

src/test/unit/io/json/json_ill_formed_test.cpp
mitzimorris: I like the ignore option.
trying to work out how the event handler should do this gracefully.

mitzimorris: > Non-regular arrays, e.g. "x": [ {"1": 0, "2": 1}, {"1": 1}] and "x": [ {"1": 0}, {"1": 1, "2": 1}]
> Changing tuple types in an array, e.g. "x": [ {"1": 0, "2": 1}, {"1": [1.5], "2":2}]

These are tests by the unit tests `json_ill_formed.cpp` and the corresponding JSON files are in folder `src/test/unit/io/test_json_files/` names "inconsistent_array_tuples".

mitzimorris:
Added logic to ignore top-level entries with names that aren't legal Stan varnames and corresponding unit tests.
Ready for re-review.

Issue #2577, opened 2018-07-10
Clean up code: `.str().length()`


syclik:
#### Summary:
There are many instances of code that looks like:
```
if (msg.str().length() > 0)
  logger.info(msg);
```

We should either: 
- have the info() method ignore empty messages in general
- wrap the call in a function

#### Additional Information:
Originally from this PR comment: https://github.com/stan-dev/stan/pull/2570#discussion_r201151286

#### Current Version:
v2.17.1

Ishvina:
Hi! I am in an intro to swe class that is requiring me to work on my first open-source project. I was wondering if this issue is still open and if so could is this beginner-friendly?

WardBrian:
Hi @Ishvina -

There are indeed still places where this pattern appears in the Stan code, for example:
https://github.com/stan-dev/stan/blob/5621ebde258fe9926c6a36160f12e998e028670e/src/stan/services/optimize/lbfgs.hpp#L96-L97

If you'd like to follow the first suggestion and change the logger, the main implementation of the logger can be found here:
https://github.com/stan-dev/stan/blob/develop/src/stan/callbacks/stream_logger.hpp

If you know some C++ I think this is a good beginner project, but building Stan itself can be complicated, especially on Windows machines. There are some helpful guides [here](https://github.com/stan-dev/stan/wiki/Introduction-to-Stan-for-New-Developers) and [here](https://github.com/stan-dev/stan/wiki/Developer-process-overview), and if you need human support you can try the #developers channel on our [Slack](mc-stan.slack.com)

AnikaAChowdhury:
Hi @Ishvina! Are you still working on this project? If you are not, could I work on this? I am in the same class as you.

Ishvina:
@AnikaAChowdhury yea you can work on it!

mitzimorris:
wecome Ishvana and Anika!  happy to help answer your questions!

AnikaAChowdhury:
It is Intro to Software Engineering class for CS majors at UF.

AnikaAChowdhury:
Hi @WardBrian! The Slack link does not work. Could you provide me with the link again? Also, could you elaborate a little more on the issue I have to change?
Thank You!

AnikaAChowdhury:
Could anyone explain to me what is the purpose of this issue and what could be achieved by fixing this issue?

mitzimorris:
this issue is the result of a suggestion made during code review:

[...]

PR #3173, opened 2023-03-20
Standalone GQ: Output NaNs when exception is caught

WardBrian:
#### Submission Checklist

#### Summary

Closes https://github.com/stan-dev/cmdstan/issues/1007. Since https://github.com/stan-dev/stanc3/pull/1165 we have made sure that `write_array` always populates the array with NaNs, so writing it out is safe even after an exception. 

#### Intended Effect

Standalone generate quantities behavior now matches that of sampling (nans output if write_array fails) and the number of rows in the output of standalone generated quantities is now always the number of rows in the input `fitted_params`.

#### How to Verify

The example from https://github.com/stan-dev/cmdstan/issues/1007 works well

#### Copyright and Licensing

Simons Foundation

### DIFF
--- a/src/stan/services/util/gq_writer.hpp
+++ b/src/stan/services/util/gq_writer.hpp
@@ -81,7 +81,6 @@ class gq_writer {
       if (ss.str().length() > 0)
         logger_.info(ss);
       logger_.info(e.what());
-      return;
     }
     if (ss.str().length() > 0)
       logger_.info(ss);
--- a/src/stan/services/util/gq_writer.hpp
+++ b/src/stan/services/util/gq_writer.hpp
@@ -77,13 +77,13 @@ class gq_writer {
     try {
       model.write_array(rng, const_cast<std::vector<double>&>(draw), params_i,
                         values, false, true, &ss);
+      if (ss.str().length() > 0)
+        logger_.info(ss);
     } catch (const std::exception& e) {
       if (ss.str().length() > 0)
         logger_.info(ss);
       logger_.info(e.what());
     }
-    if (ss.str().length() > 0)
-      logger_.info(ss);
 
     std::vector<double> gq_values(values.begin() + num_constrained_params_,
                                   values.end());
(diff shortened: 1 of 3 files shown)


src/stan/services/util/gq_writer.hpp
```diff
@@ -81,7 +81,6 @@ class gq_writer {
       if (ss.str().length() > 0)
         logger_.info(ss);
       logger_.info(e.what());
-      return;
```
bob-carpenter, line 84: If this isn't going to return, then the following `logger_` call outside the catch duplicates the `ss` message.  The full context is this:

```cpp
    try {
      model.write_array(rng, const_cast<std::vector<double>&>(draw), params_i,
                        values, false, true, &ss);
      if (ss.str().length() > 0)
        logger_.info(ss);
      logger_.info(e.what());
    }
    if (ss.str().length() > 0)
      logger_.info(ss);
```

The conditional logging after the catch should be moved into the try block to avoid duplicating `logger_.info(ss)`.   

This is also making me think the logger itself should be ignoring empty string messages rather than duplicating this everywhere.

src/test/unit/services/util/gq_writer_test.cpp
```diff
@@ -40,6 +41,23 @@ TEST_F(ServicesUtilGQWriter, t2) {
...
+  boost::ecuyer1988 rng1 = stan::services::util::create_rng(0, 1);
+  std::vector<double> draw;
+  draw.push_back(2.345);
+  draw.push_back(6.789);
+  stan::services::util::gq_writer writer(sample_writer, logger, 2);
+  writer.write_gq_values(model, rng1, draw);
+  // model test_gq.stan generates 4 values, 3 commas
+  EXPECT_EQ(count_matches(",", sample_ss.str()), 3);
+
+  EXPECT_EQ(count_matches("nan", sample_ss.str()), 4);
```
bob-carpenter, line 62: Thanks for adding the test.

bob-carpenter (changes requested):
Looks good other than scope of a logger in one file.

src/stan/services/util/gq_writer.hpp
WardBrian: Good catch.

Ignoring empty strings has a (nearly 5 year old!) issue open: https://github.com/stan-dev/stan/issues/2577

bob-carpenter (approved):
Thanks for the update.

bob-carpenter:
Thanks, and feel free to merge when tests finish.

PR #3177, opened 2023-04-05
Remove old/unused IO code

WardBrian:
#### Submission Checklist

#### Summary

Removes unused code from `stan/io`:

- `reader.hpp` and `writer.hpp` - replaced by `deserializer.hpp` and `serializer.hpp`, no longer used
- `is_whitespace.hpp` - used in the stanc2 compiler, now no longer needed
- `util.hpp` - contained only one function, `utc_time_string`, which is never used

Updated a few tests to reflect

#### Intended Effect

Removes old code no longer needed

#### Documentation

None

#### Copyright and Licensing


WardBrian:
It seems like a bunch of tests were relying on the fact that `model_header.hpp` was including `io/dump`  (which models never depend on!) and did not include it themselves. I'm updating those tests now

rok-cesnovar (approved):
🧹

bgoodri:
I was using the `reader` in the the last section of the StanHeaders [vignette](https://cran.r-project.org/web/packages/StanHeaders/vignettes/stanmath.html). I was eventually able to figure out that I needed to use `deserializer`, but just because something is not used within Stan doesn't mean it is not used by someone writing their own C++ on top of Stan. We are trying to encourage more of that, but breaking their code does not help.

WardBrian:
> We are trying to encourage more of that

I thought we were trying to encourage [less of that](https://discourse.mc-stan.org/t/r-package-maintainers-using-stan-please-read/24131), at least in the R world?

My understanding is that Stan had not had any internal usages of this code in the 4 versions prior to when we removed it. We really have no way of knowing what other usages exist for something like this

bgoodri:
I agree it is impossible to know how stuff in Stan Math or Library may be used in external C++ projects (unless they complain when stuff breaks). But other C++ header libraries like Boost and Eigen won't just remove stuff in a point release for that reason.

PR #3179, opened 2023-04-11
Expose new base class method 'unconstrain_array' 

WardBrian:
#### Submission Checklist

#### Summary

This exposes the method added in https://github.com/stan-dev/stanc3/pull/1305 as a direct inverse function for `write_array`. `unconstrain_array` takes in a vector of parameters on the constrained scale _in the same order that `write_array` produced them_ and produces a vector of parameters on the **un**constrained scale in the same order that `write_array` and `log_prob` expect.

This allows us to write what we want to write in almost all cases where we are using `array_var_context` followed by `transform_inits` at the moment (standalone generated quantities, the `log_prob` service function, etc). This is very useful for new features like tuples, ragged arrays, etc, since it weakens the reliance on the `model.get_dims()` function which fundamentally assumes that parameters are rectangular. 

#### Intended Effect

Exposes `unconstrain_array` to the model base class. This function is already generated by the compiler.

#### How to Verify

Tests are added for the function directly and the standalone generated quantities service has been updated to use this new code. 

#### Side Effects

None

#### Copyright and Licensing

Simons Foundation


src/test/unit/model/array_functions_roundtrip_test.cpp
```diff
@@ -0,0 +1,91 @@
...
+    out.str("");
+
+    std::vector<std::string> json_path = {
+        "src", "test", "test-models", "good", "model", "parameters.inits.json"};
+    std::string filename = paths_to_fname(json_path);
+    std::ifstream in(filename);
+    inits = new stan::json::json_data(in);
+  }
+
+  ~ModelArrayFunctionsRoundtripTest() { delete inits; }
```
SteveBronder, line 21: I would do something like this so the deletion just gets handled by a `unique_ptr`

```suggestion
auto init_inits() {
  std::vector<std::string> json_path = {
    "src", "test", "test-models", "good", "model", "parameters.inits.json"};
  std::string filename = paths_to_fname(json_path);

  std::fstream stream(filename, std::fstream::in);
  return std::unique_ptr<stan::io::var_context>(new stan::json::json_data(stream));
}

class ModelArrayFunctionsRoundtripTest : public testing::Test {
 public:
  ModelArrayFunctionsRoundtripTest()
      : model(context, 0, nullptr), rng(12324232), inits(init_inits()) {}

  void TearDown() override {
    out.str("");
  }


```

SteveBronder (approved):

rok-cesnovar (approved):

PR #3191, opened 2023-05-16
Feature/3176 json

mitzimorris:
#### Submission Checklist

#### Summary

Add callback writer to output JSON data.

#### Intended Effect

Allow structured outputs in JSON format, per design-docs spec https://github.com/stan-dev/design-docs/blob/master/designs/0032-stan-output-formats.md

#### How to Verify

Unit tests

#### Side Effects

None

#### Documentation

N/A

#### Copyright and Licensing

Please list the copyright holder for the work you are submitting (this will be you or your assignee, such as a university or company):  Coumbia University


src/stan/callbacks/unique_stream_writer.hpp
```diff
@@ -66,17 +67,34 @@ class unique_stream_writer final : public writer {
...
+    if (output_ == nullptr)
+      return;
+    write_vector(state);
+  }
 
-  /**
-   * Writes the comment_prefix to the stream followed by a newline.
-   */
-  void operator()() {
+  void operator()(const std::tuple<Eigen::VectorXd, Eigen::VectorXd>& state) {
```
WardBrian, line 76: What is the purpose of this overload?

src/stan/callbacks/unique_stream_writer.hpp
```diff
@@ -28,7 +30,10 @@ class unique_stream_writer final : public writer {
    */
   explicit unique_stream_writer(std::unique_ptr<Stream>&& output,
                                 const std::string& comment_prefix = "")
-      : output_(std::move(output)), comment_prefix_(comment_prefix) {}
+      : output_(std::move(output)), comment_prefix_(comment_prefix) {
+    if (output_ == nullptr)
+      throw std::invalid_argument("output writer cannot be null");
+  }
```
WardBrian, line 36: If we're preventing the output from being `nullptr` we can delete the a lot of the `if` statements that check for this in each operator() overload, such as on line 71

src/stan/callbacks/unique_stream_writer.hpp
```diff
@@ -66,17 +67,34 @@ class unique_stream_writer final : public writer {
...
     if (output_ == nullptr)
       return;
-    *output_ << comment_prefix_ << std::endl;
+    Eigen::IOFormat CommaInitFmt(Eigen::StreamPrecision, Eigen::DontAlignCols,
+                                 ", ", "", "", "\n", "", "");
+    *output_ << std::get<0>(state).transpose().eval();
+    *output_ << std::get<1>(state).transpose().eval();
+  }
+
+  void operator()(const Eigen::MatrixXd& states) {
```
WardBrian, line 85: This overload needs a docstring (although this isn't part of `stan::io::callbacks::writer`, so maybe it isn't used?)

Is there a story behind the name `states`?

src/stan/callbacks/json_writer.hpp
```diff
@@ -0,0 +1,375 @@
...
+
+  /**
+   * Write a key-value pair to the output stream with a value of null as the
+   * value.
+   * @param key Name of the value pair
+   */
+  void write(const std::string& key) {
+    write_sep();
+    write_key(key);
+    *output_ << "\"null\" ";
```
WardBrian, line 218: Why `"null"` vs `null`?

src/stan/callbacks/json_writer.hpp
```diff
@@ -0,0 +1,375 @@
...
+    record_internal_needs_comma_ = false;
+  }
+  /**
+   * Writes "]", final token of a JSON list.
+   */
+  void end_list() {
+    *output_ << "]";
+    write_sep();
+    *output_ << "\n";
+  }
```
WardBrian, line 208: Because all of the `write` functions assume a key is present, I think the only valid purpose for these functions is to construct a list of objects, which currently does not produce valid JSON.

Test case:
```c++
TEST_F(StanInterfaceCallbacksJsonWriter, list_of_objects) {
  std::string key("key");
  const int N = 5;
  std::vector<double> x;
  for (int n = 0; n < N; ++n)
    x.push_back(n);

  writer.begin_list();
  writer.begin_record();
  EXPECT_NO_THROW(writer.write(key, x));
  writer.end_record();

  writer.begin_record();
  EXPECT_NO_THROW(writer.write(key, x));
  writer.end_record();
  writer.end_list();

  EXPECT_EQ(
      "[{\"key\" : [ 0, 1, 2, 3, 4 ]}"
      ", {\"key\" : [ 0, 1, 2, 3, 4 ]}]",
      ss.str());

}
```

Currently, the comma is missing between the two objects.

src/stan/callbacks/json_writer.hpp
```diff
@@ -0,0 +1,375 @@
...
+    write_key(key);
+    Eigen::IOFormat json_format(Eigen::StreamPrecision, Eigen::DontAlignCols,
+                                ", ", "", "", "", "[", "]");
+    *output_ << vec.format(json_format);
+  }
+
+  /**
+   * Reset state
+   */
+  void reset() {
```
WardBrian, line 366: What is the use case for this function?

src/test/unit/callbacks/json_writer_test.cpp
```diff
@@ -0,0 +1,132 @@
...
+  const int N = 5;
+  std::vector<std::string> x;
+  for (int n = 0; n < N; ++n)
+    x.push_back(boost::lexical_cast<std::string>(n));
+
+  EXPECT_NO_THROW(writer.write("key", x));
+  EXPECT_EQ("\"key\" : [ 0, 1, 2, 3, 4 ]", ss.str());
+}
+
+TEST_F(StanInterfaceCallbacksJsonWriter, write_null) {
```
WardBrian, line 124: Also worth testing `NaN` and `inf` as floats

src/stan/callbacks/json_writer.hpp
```diff
@@ -0,0 +1,375 @@
...
+      *output_ << "}";
+      record_depth_--;
+      if (record_depth_ > 0) {
+        record_needs_comma_ = true;
+      }
+      record_internal_needs_comma_ = false;
+    } else {
+      throw std::runtime_error(
+          "Attempted to end record, but there is no open record.");
+    }
```
WardBrian, line 191: We throw an error if you try to close a record which doesn't exist, which I think is great. However, we don't do any other checking that json is valid, e.g. that we are inside an object before any key is written.

I'd like to have that check such that it is a runtime error to do 

```c++
json_writer out(...);
out.write("hello", "world");
```
which would produce `hello: "world"`, while the nearest valid JSON is `{ hello: "world" }`

WardBrian (changes requested):

src/stan/callbacks/unique_stream_writer.hpp
mitzimorris: updated unique_stream_writer.hpp - the above was present in an old version of pathfinder, checked in by mistake.

src/stan/callbacks/unique_stream_writer.hpp
mitzimorris: fixed.

src/stan/callbacks/unique_stream_writer.hpp
mitzimorris: gone - same as above - early pathfinder version, checked in by mistake.

src/stan/callbacks/json_writer.hpp
mitzimorris: mistake. fixed.

src/stan/callbacks/json_writer.hpp
SteveBronder: idt there is any way to meaningfully validate a steam of json that is not also very difficult (hence why it's so hard / weird to do streaming json in rapidjson.)

Take the following case as an example where the try throws and someone wants to attempt to fix the output file

```c++
std::string write_to_file = "./path/to/my/file.json";
try {
  json_writer out(write_to_file);
  out.begin_record();
  out.write("hello", "world");
  // Other stuff happens that could throw
  out.end_record();
  // checks if a simple fix like adding double quote or a comma can produce a valid line of json
  if (attempt_to_fix_json(write_to_file)) {
    json_writer out(write_to_file);
    out.write("error",  e.what());
    out.end_record();
  }   
  // If the json is now invalid we stop, else we can keep going
  if (invalid_json(write_to_file)) {
      throw e;
  }
}
```

in the above calling `end_record()` would be a valid attempt to fix the file, but we've the added state from checking if `begin_record()` is called already would cause this to fail

mitzimorris: I think that the best thing to do is remove the `end_record` check and document that this is a streaming writer which doesn't perform validation on the JSON, because, as Steve points out, we don't know anything about the output stream we're writing to.

src/stan/callbacks/json_writer.hpp
WardBrian: @SteveBronder and I also think we can probably just remove the begin/end list functions, as they're not strictly needed for our use cases. If we need them later we can add them then and do the extra heavy lifting to make them correct

SteveBronder: Def agree on removing begin and end list idt we use them anywhere in pathfinder

src/stan/callbacks/json_writer.hpp
mitzimorris: needed if the caller is reusing the json writer and also writing to the output stream.
right now it's used by the unit tests

src/stan/callbacks/json_writer.hpp
mitzimorris: updated initial class doc comment, removed check on balanced close objects.

src/stan/callbacks/json_writer.hpp
mitzimorris: gone

mitzimorris:
OK, ready for re-review

src/stan/callbacks/unique_stream_writer.hpp
```diff
@@ -64,9 +70,31 @@ class unique_stream_writer final : public writer {
...
+   * transposed for the output.
    */
-  void operator()(const std::vector<double>& state) { write_vector(state); }
+  void operator()(const Eigen::MatrixXd& values) {
+    if (output_ == nullptr)
+      return;
+    Eigen::IOFormat CommaInitFmt(Eigen::StreamPrecision, Eigen::DontAlignCols,
+                                 ", ", "", "", "\n", "", "");
+    *output_ << values.transpose().format(CommaInitFmt);
+  }
```
WardBrian, line 97: Does this flatten the matrix into a single row of output or output the matrix as a 2-d object?

I know this PR is not primarily about the CSV writer, but this seems like an odd method to me to add, with the implicit transpose and the required contract that the number of rows is the same as the length of the header of the file.

src/stan/callbacks/json_writer.hpp
WardBrian: It's not needed in the tests, each test is run on a newly instantiated object (see the [FAQ](https://fuchsia.googlesource.com/third_party/googletest/+/HEAD/googletest/docs/faq.md#ctorvssetup)).

I also don't think we should support re-using the same writer on a new file, it just seems like a recipe for problems, and constructing these isn't particularly expensive. I think `reset` should not be part of the API of this

src/test/unit/callbacks/unique_stream_writer_test.cpp
WardBrian, line 1: It appears this test file has been completely replaced by accident

src/test/unit/callbacks/json_writer_test.cpp
```diff
@@ -0,0 +1,142 @@
...
+}
+
+TEST_F(StanInterfaceCallbacksJsonWriter, write_double_vector_nan_inf) {
+  std::string key("key");
+  std::vector<double> x;
+  x.push_back(std::numeric_limits<double>::quiet_NaN());
+  x.push_back(std::numeric_limits<double>::infinity());
+  x.push_back(-std::numeric_limits<double>::infinity());
+  writer.write(key, x);
+  EXPECT_EQ("\"key\" : [ nan, inf, -inf ]", ss.str());
```
WardBrian, line 101: The output here is not something which is readable by our own JSON parser for inputs. Tested with this model:

```stan
data {
   array[3] real key;
}
transformed data {
   print(key);
}
```

output:
```
Error in JSON parsing 
at offset 12: 
Invalid value.
```

Outputting the correct values (especially if they appear inside of an Eigen matrix or something) will be a bit tedious, but it is important if we ever want to consume the output of these files.

src/stan/callbacks/json_writer.hpp
```diff
@@ -0,0 +1,401 @@
...
+    *output_ << v.back() << " ]";
+  }
+
+  /**
+   * Writes a single value.  Corrects capitalization for inf and nans.
+   *
+   * @param[in] v value
+   */
+  void write_element(double v) {
```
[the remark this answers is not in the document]
mitzimorris: `unlikely(...)` is since C++20 - what is allowed in Stan these days?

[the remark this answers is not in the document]
WardBrian: We're only on 14 still, but we have a macro which does the same thing (and is also called `unlikely`) when it can: https://github.com/stan-dev/math/blob/develop/stan/math/prim/meta/compiler_attributes.hpp

We use it e.g. here:
https://github.com/stan-dev/stan/blob/1b126f5c8da9d374d0bee934779a6600112ed08c/src/stan/io/deserializer.hpp#L179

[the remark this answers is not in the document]
mitzimorris: using this everywhere - for double, Eigen, and complex values.

src/test/unit/callbacks/json_writer_test.cpp
mitzimorris: fixed.

mitzimorris:
ready for re-re-review

src/stan/callbacks/json_writer.hpp
```diff
@@ -0,0 +1,419 @@
...
+        write_value(v[i]);
+        *output_ << ", ";
+      }
+      write_value(v[last]);
+    }
+    *output_ << " ]";
+  }
+
+ public:
+  json_writer() : output_(nullptr) {}
```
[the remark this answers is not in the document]
WardBrian: I think that will also require adding checks for nullptr to every method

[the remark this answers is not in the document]
mitzimorris: can we extend this class with a "sink" class that overrides all the `write` functions and does nothing?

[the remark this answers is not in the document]
WardBrian: That’s essentially treating this as a base class, so sure. You still would need to have a pointer indirection or a templated function any place you wanted to be able to accept both of them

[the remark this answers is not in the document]
mitzimorris: added checks for nullptr everywhere; leaving no-arg constructor as is.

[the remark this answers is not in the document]
WardBrian: I'm fine with `if (output == nullptr)`. At the very least it is consistent with `unique_stream_writer`.

@mitzimorris mind adding a test which uses a null output and just asserts that everything does not throw?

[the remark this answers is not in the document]
mitzimorris: there's one there already - last test in file `stan/src/test/unit/callbacks/json_writer_test.cpp` (line 223 or so)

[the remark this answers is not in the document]
mitzimorris: added another one.

WardBrian (approved):

SteveBronder:
Actually @mitzimorris and @WardBrian one last thing, how should NA and infinite values get handled here? I'm unfamiliar with how json should handle these.

WardBrian:
We print them such that the JSON parser we use can read them back in. This involves using some non-standard atoms:

| Value | ""JSON"" |
|--------|--------|
| Not-a-number | `NaN` |
| Infinity | `Inf` |
| Negative infinity | `-Inf` |

mitzimorris:
OK to merge?

WardBrian:
I think so, @SteveBronder?

SteveBronder:
Yep good to merge!

PR #3202, opened 2023-06-30
Issue/3200 pathfinder diagnostic

mitzimorris:
#### Submission Checklist

#### Summary

Add hooks to json_writer to put newlines around records.

- begin_record takes optional arg "newline"; if true, will output newline before record start or key.
- end_record takes optional arg "newline"; if true, puts newline after record end, furthermore, final close record will be on its own line.

#### Intended Effect

Produce readable-ish JSON.

#### How to Verify

Unit tests at CmdStan level.

#### Side Effects

N/A

#### Documentation

N/A

#### Copyright and Licensing

### DIFF
--- a/src/stan/callbacks/json_writer.hpp
+++ b/src/stan/callbacks/json_writer.hpp
@@ -259,13 +259,14 @@ class json_writer {
 
   /**
    * Writes "{", initial token of a JSON record.
+   * @param[in] newline (optional) if true, add newline after open token.
    */
-  void begin_record(bool newline_after=false) {
+  void begin_record(bool newline=false) {
     if (output_ == nullptr)
       return;
     write_record_comma_if_needed();
     *output_ << "{";
-    if (newline_after)
+    if (newline)
       *output_ << std::endl;
     record_needs_comma_ = false;
     record_depth_++;
@@ -274,27 +275,27 @@ class json_writer {
   /**
    * Writes "\"key\" : {", initial token of a named JSON record.
    * @param[in] key The name of the record.
-   * @param[in] newline_before if true, add newline before open token.
+   * @param[in] newline (optional) if true, add newline between key, open token.
    */
-  void begin_record(const std::string& key, bool newline_before=false) {
+  void begin_record(const std::string& key, bool newline=false) {
     if (output_ == nullptr)
       return;
     write_record_comma_if_needed();
     *output_ << "\"" << key << "\":";
-    newline_before ? *output_ << std::endl : *output_ << " ";
+    newline ? *output_ << std::endl : *output_ << " ";
     *output_ <<  "{";
     record_needs_comma_ = false;
     record_depth_++;
   }
   /**
    * Writes "}", final token of a JSON record.
-   * @param[in] newline_after if true, add newline at end.
+   * @param[in] newline (optional) if true, add newline after close token.
    */
-  void end_record(bool newline_after=false) {
+  void end_record(bool newline=false) {
     if (output_ == nullptr)
       return;
     *output_ << "}";
-    if (newline_after)
+    if (newline)
       *output_ << std::endl;
       
     record_depth_--;
--- a/src/stan/callbacks/json_writer.hpp
+++ b/src/stan/callbacks/json_writer.hpp
@@ -281,7 +281,7 @@ class json_writer {
     if (output_ == nullptr)
       return;
     write_record_comma_if_needed(newline);
-    *output_ << "\"" << key << "\":{";
+    *output_ << "\"" << key << "\" : {";
     record_needs_comma_ = false;
     record_depth_++;
   }
@@ -292,15 +292,17 @@ class json_writer {
   void end_record(bool newline=false) {
     if (output_ == nullptr)
       return;
-    *output_ << "}";
-    if (newline)
-      *output_ << std::endl;
-      
     record_depth_--;
     if (record_depth_ > 0) {
       record_needs_comma_ = true;
-    }
+    } else {
+      if (newline)
+        *output_ << std::endl;
+    }      
     record_element_needs_comma_ = false;
+    *output_ << "}";
+    if (newline)
+      *output_ << std::endl;
   }
 
   /**
--- a/src/stan/callbacks/json_writer.hpp
+++ b/src/stan/callbacks/json_writer.hpp
@@ -260,11 +260,13 @@ class json_writer {
   /**
    * Writes "{", initial token of a JSON record.
    */
-  void begin_record() {
+  void begin_record(bool newline_after=false) {
     if (output_ == nullptr)
       return;
     write_record_comma_if_needed();
     *output_ << "{";
+    if (newline_after)
+      *output_ << std::endl;
     record_needs_comma_ = false;
     record_depth_++;
   }
@@ -272,23 +274,27 @@ class json_writer {
   /**
    * Writes "\"key\" : {", initial token of a named JSON record.
    * @param[in] key The name of the record.
+   * @param[in] newline_before if true, add newline before open token.
    */
-  void begin_record(const std::string& key) {
+  void begin_record(const std::string& key, bool newline_before=false) {
     if (output_ == nullptr)
       return;
     write_record_comma_if_needed();
-    *output_ << "\"" << key << "\": {";
+    *output_ << "\"" << key << "\":";
+    newline_before ? *output_ << std::endl : *output_ << " ";
+    *output_ <<  "{";
     record_needs_comma_ = false;
     record_depth_++;
   }
   /**
    * Writes "}", final token of a JSON record.
+   * @param[in] newline_after if true, add newline at end.
    */
-  void end_record(bool add_newline=false) {
+  void end_record(bool newline_after=false) {
     if (output_ == nullptr)
       return;
     *output_ << "}";
-    if (add_newline)
+    if (newline_after)
       *output_ << std::endl;
       
     record_depth_--;
--- a/src/stan/callbacks/json_writer.hpp
+++ b/src/stan/callbacks/json_writer.hpp
@@ -284,10 +284,13 @@ class json_writer {
   /**
    * Writes "}", final token of a JSON record.
    */
-  void end_record() {
+  void end_record(bool add_newline=false) {
     if (output_ == nullptr)
       return;
     *output_ << "}";
+    if (add_newline)
+      *output_ << std::endl;
+      
     record_depth_--;
     if (record_depth_ > 0) {
       record_needs_comma_ = true;
--- a/src/stan/callbacks/json_writer.hpp
+++ b/src/stan/callbacks/json_writer.hpp
@@ -43,10 +43,12 @@ class json_writer {
   /**
    * Writes a comma separator for the record's parent object if needed.
    */
-  void write_record_comma_if_needed() {
+  void write_record_comma_if_needed(bool newline=false) {
     if (record_depth_ > 0 && record_needs_comma_) {
       *output_ << ",";
     }
+    if (newline)
+      *output_ << std::endl;
   }
 
   /**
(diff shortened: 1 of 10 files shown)


mitzimorris:
is there a pathfinder unit test that checks output files?   if so, can add tests for single-path pathfinder.

WardBrian:
How ugly is it to always emit newlines after closing records? I feel like adding a flag for this is not the cleanest solution

mitzimorris:
we would have to add "write_newline" to json_writer.   also a fine way to go and much much simpler.

mitzimorris:
with this change, Pathfinder output diagnostic files look like this (CmdStan's example/bernoulli model and data):

```
{
"0" : {"iter" : 0, "unconstrained_parameters" : [ 0.66028 ], "grads" : [ -4.91188 ]}
,"1" : {"iter" : 1, "unconstrained_parameters" : [ 0.66028 ], "grads" : [ -4.91188 ], "history_size" : 1, "lbfgs_success" : true, "pathfinder_success" : true, "x_center" : [ 0.373672 ], "logDetCholHk" : -2.84129, "L_approx" : [ [ -0.0583501 ] ], "Qk" : [  ], "alpha" : [ 1 ], "full" : true, "lbfgs_note" : ""}
,"2" : {"iter" : 2, "unconstrained_parameters" : [ 0.169092 ], "grads" : [ 3.50607 ], "history_size" : 2, "lbfgs_success" : true, "pathfinder_success" : true, "x_center" : [ -1.07247 ], "logDetCholHk" : -0.519062, "L_approx" : [ [ 0.595079 ] ], "Qk" : [  ], "alpha" : [ 0.354119 ], "full" : true, "lbfgs_note" : ""}
,"3" : {"iter" : 3, "unconstrained_parameters" : [ -0.948132 ], "grads" : [ 0.351127 ], "history_size" : 3, "lbfgs_success" : true, "pathfinder_success" : true, "x_center" : [ -1.09769 ], "logDetCholHk" : -0.426745, "L_approx" : [ [ 0.65263 ] ], "Qk" : [  ], "alpha" : [ 0.425926 ], "full" : true, "lbfgs_note" : ""}
,"4" : {"iter" : 4, "unconstrained_parameters" : [ -1.07247 ], "grads" : [ 0.0591971 ], "history_size" : 4, "lbfgs_success" : true, "pathfinder_success" : true, "x_center" : [ -1.09861 ], "logDetCholHk" : -0.408829, "L_approx" : [ [ 0.664428 ] ], "Qk" : [  ], "alpha" : [ 0.441464 ], "full" : true, "lbfgs_note" : ""}
,"5" : {"iter" : 5, "unconstrained_parameters" : [ -1.09769 ], "grads" : [ 0.00208353 ], "history_size" : 5, "lbfgs_success" : true, "pathfinder_success" : true, "x_center" : [ -1.09861 ], "logDetCholHk" : -0.405582, "L_approx" : [ [ 0.666589 ] ], "Qk" : [  ], "alpha" : [ 0.444341 ], "full" : true, "lbfgs_note" : ""}
}
```

mitzimorris:
re-implemented, added function "newline" with optional argument "add_comma".

this isn't particularly pretty, but it allows the single-path pathfinder to create JSON files one iteration per line.

```
{
"0" : {"iter" : 0, "unconstrained_parameters" : [ -1.55164 ], "grads" : [ 0.901806 ]},
"1" : {"iter" : 1, "unconstrained_parameters" : [ -1.55164 ], "grads" : [ 0.901806 ], "history_size" : 1, "lbfgs_success" : true, "pathfinder_success" : true, "x_center" : [ -1.50214 ], "logDetCholHk" : -2.90239, "L_approx" : [ [ -0.0548919 ] ], "Qk" : [  ], "alpha" : [ 1 ], "full" : true, "lbfgs_note" : ""},
"2" : {"iter" : 2, "unconstrained_parameters" : [ -1.46146 ], "grads" : [ -0.741069 ], "history_size" : 2, "lbfgs_success" : true, "pathfinder_success" : true, "x_center" : [ -1.09998 ], "logDetCholHk" : -0.358948, "L_approx" : [ [ 0.698411 ] ], "Qk" : [  ], "alpha" : [ 0.487778 ], "full" : true, "lbfgs_note" : ""},
"3" : {"iter" : 3, "unconstrained_parameters" : [ -1.08508 ], "grads" : [ 0.0305572 ], "history_size" : 3, "lbfgs_success" : true, "pathfinder_success" : true, "x_center" : [ -1.09861 ], "logDetCholHk" : -0.407202, "L_approx" : [ [ 0.66551 ] ], "Qk" : [  ], "alpha" : [ 0.442904 ], "full" : true, "lbfgs_note" : ""},
"4" : {"iter" : 4, "unconstrained_parameters" : [ -1.09822 ], "grads" : [ 0.000893818 ], "history_size" : 4, "lbfgs_success" : true, "pathfinder_success" : true, "x_center" : [ -1.09861 ], "logDetCholHk" : -0.405519, "L_approx" : [ [ 0.666631 ] ], "Qk" : [  ], "alpha" : [ 0.444397 ], "full" : true, "lbfgs_note" : ""}
}
```

mitzimorris:
I realize that breaking iterations out by line is completely unneccessary.  the minimal fix for this PR is just adding the "newline" method, no fanciness w/r/t comma between records.

if the reviewers prefer,  I can simplify the PR and leave the diagnostic JSON as is, simply adding a newline at the end.

WardBrian:
I think I prefer simpler. I wouldn’t be opposed to automatically adding new lines at the end of lists/records, if that produces desirable output, but doing it case by case seems clunky

mitzimorris:
simplified.   will try to remember this for next time.

src/stan/callbacks/json_writer.hpp
```diff
@@ -287,12 +286,12 @@ class json_writer {
   void end_record() {
     if (output_ == nullptr)
       return;
-    *output_ << "}";
     record_depth_--;
     if (record_depth_ > 0) {
       record_needs_comma_ = true;
     }
     record_element_needs_comma_ = false;
+    *output_ << "}";
```
SteveBronder, line 294: Is there a reason we don't just add a newline here like 

```suggestion
    *output_ << "}\n";
```

instead of having a new function? Or do we need `std::endl`?

SteveBronder (changes requested):

src/stan/callbacks/json_writer.hpp
mitzimorris: that's another option - no objections to that either.

mitzimorris: the output files end up looking funny because the logic to insert the comma between records is invoked elsewhere,
resulting in "}\n," instead of "},\n".    I wasted an awful lot of time trying to make this pretty.  not worth it.

SteveBronder: imo I'd just like to not complicate the interface with a new function. if `},\n` is valid json then I think that's fine

mitzimorris: yes, "}\n," is valid JSON.

@WardBrian - what do you think?

WardBrian: My preference is also for simpler, so if it can be done without another function great. If the `}\n,` is aesthetically un-pleasing, it's also reasonable to store an extra bool called `should_newline` and then make the actual printing of `\n` part of the comma logic/end of file logic?

src/stan/callbacks/json_writer.hpp
```diff
@@ -46,6 +46,7 @@ class json_writer {
   void write_record_comma_if_needed() {
     if (record_depth_ > 0 && record_needs_comma_) {
       *output_ << ",";
+      record_needs_comma_ = false;
```
bob-carpenter, line 49: This can be 

```cpp
record_needs_comma_ = record_needs_comma_ && record_depth_ > 0;
if (record_needs_comma_) {
  *output_ << ",";
}
```

But then I wonder why is the variable is being set?  (I also reordered the conjunction---there is no `&&=`, alas)

src/stan/callbacks/json_writer.hpp
```diff
@@ -287,12 +286,12 @@ class json_writer {
   void end_record() {
     if (output_ == nullptr)
       return;
-    *output_ << "}";
     record_depth_--;
     if (record_depth_ > 0) {
```
bob-carpenter, line 290: This can also be an assignment.

```cpp
record_needs_comma_ = record_needs_comma_ || record_depth_ > 0;
```

src/stan/callbacks/json_writer.hpp
```diff
@@ -287,12 +286,12 @@ class json_writer {
   void end_record() {
     if (output_ == nullptr)
```
bob-carpenter, line 287: Why is this using a pointer rather than a reference?  Is the idea that you can supply a `nullptr` to turn off writing?

SteveBronder: > Is the idea that you can supply a nullptr to turn off writing?

Yes if it's not initialized then we don't want to allow writes

src/stan/callbacks/json_writer.hpp
mitzimorris: it gets set on line 293.

src/stan/io/dump.hpp
```diff
@@ -269,8 +268,8 @@ class dump_reader {
   int get_int() {
     int n = 0;
     try {
-      n = boost::lexical_cast<int>(buf_);
-    } catch (const boost::bad_lexical_cast& exc) {
+      n = std::stol(buf_);
```
SteveBronder, line 271:
```suggestion
      n = std::stoi(buf_);
```
?

imo can we back out these changes for `lexical_cast` for another PR?

src/stan/io/dump.hpp
```diff
@@ -242,8 +241,8 @@ class dump_reader {
     scan_optional_long();
     size_t d = 0;
     try {
-      d = boost::lexical_cast<size_t>(buf_);
-    } catch (const boost::bad_lexical_cast& exc) {
+      d = std::stoull(buf_);
+    } catch (const std::exception& exc) {
```
SteveBronder, line 245: If we are using the `std::sto*` functions they can only call two exceptions

```
> [std::invalid_argument](https://en.cppreference.com/w/cpp/error/invalid_argument) if no conversion could be performed.
> [std::out_of_range](https://en.cppreference.com/w/cpp/error/out_of_range) if the converted value would fall out of the range of the result type or if the underlying function (std::strtol or std::strtoll) sets [errno](https://en.cppreference.com/w/cpp/error/errno) to [ERANGE](https://en.cppreference.com/w/cpp/error/errno_macros).
```

I'd rather have try-catches for each of these

SteveBronder: https://en.cppreference.com/w/cpp/string/basic_string/stol

src/stan/callbacks/json_writer.hpp
```diff
@@ -265,8 +268,8 @@ class json_writer {
       return;
     write_record_comma_if_needed();
     *output_ << "{";
-    record_needs_comma_ = false;
     record_depth_++;
+    record_element_needs_comma_ = false;
```
SteveBronder, line 272: Why was this changed from `record_needs_comma` to `record_element_needs_comma_`?

SteveBronder (changes requested):

src/stan/callbacks/json_writer.hpp
mitzimorris: because that logic has been moved into `write_record_comma_if_needed`

mitzimorris: added more tests to make sure this works properly.

src/stan/io/dump.hpp
mitzimorris: noted - backed out changes.

src/stan/io/dump.hpp
mitzimorris: done

SteveBronder (approved):
lgtm!

PR #3227, opened 2023-09-05
Make json_writer a structured_writer, improve output formatting

WardBrian:
#### Submission Checklist

#### Summary

Follow up on #3191.

This PR:
- Improves the output of the JSON writer to be a bit more human readable (mainly by more consistently outputting newlines between key:value pairs)
- Updates both `structured_writer` and `json_writer` such that the latter can be a subclass of the former. 
- Fixed a bug where `j.begin_record("foo"); j.end_record(); j.write("anything here");` would generate invalid JSON

#### Copyright and Licensing

### DIFF
--- a/src/stan/callbacks/json_writer.hpp
+++ b/src/stan/callbacks/json_writer.hpp
@@ -103,8 +103,9 @@ class json_writer final : public structured_writer {
 
   template <typename T>
   void write_int_like(const std::string& key, T value) {
-    if (output_ == nullptr)
+    if (output_ == nullptr) {
       return;
+    }
     write_sep();
     write_key(key);
     *output_ << value;
@@ -190,8 +191,9 @@ class json_writer final : public structured_writer {
    * Writes "{", initial token of a JSON record.
    */
   void begin_record() {
-    if (output_ == nullptr)
+    if (output_ == nullptr) {
       return;
+    }
     write_sep();
     *output_ << "{";
     record_depth_++;
@@ -203,8 +205,9 @@ class json_writer final : public structured_writer {
    * @param[in] key The name of the record.
    */
   void begin_record(const std::string& key) {
-    if (output_ == nullptr)
+    if (output_ == nullptr) {
       return;
+    }
     write_sep();
     write_key(key);
     *output_ << "{";
@@ -215,8 +218,9 @@ class json_writer final : public structured_writer {
    * Writes "}", final token of a JSON record.
    */
   void end_record() {
-    if (output_ == nullptr)
+    if (output_ == nullptr) {
       return;
+    }
     record_depth_--;
     *output_ << "\n" << std::string(record_depth_ * 2, ' ') << "}";
     if (record_depth_ > 0) {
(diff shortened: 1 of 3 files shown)


src/stan/callbacks/json_writer.hpp
```diff
@@ -109,7 +97,17 @@ class json_writer {
...
+  }
+
+  template <typename T>
+  void write_int_like(const std::string& key, T value) {
+    if (output_ == nullptr)
+      return;
+    write_sep();
+    write_key(key);
+    *output_ << value;
   }
```
SteveBronder, line 111: I would just call this `write_` since it takes anything that is valid for `operator<<`

src/stan/callbacks/json_writer.hpp
```diff
@@ -410,13 +347,91 @@ class json_writer {
    * @param key Name of the value pair
    * @param values vector to write.
    */
-  template <typename T>
-  void write(const std::string& key, const std::vector<T>& values) {
+  void write(const std::string& key, const std::vector<std::string>& v) {
+    if (output_ == nullptr)
+      return;
```
SteveBronder, line 352:
```suggestion
    if (output_ == nullptr) {
      return;
    }
```

src/stan/callbacks/json_writer.hpp
```diff
@@ -410,13 +347,91 @@ class json_writer {
...
+
+    *output_ << "[ ";
+    if (v.size() > 0) {
+      auto last = v.end();
+      --last;
+      for (auto it = v.begin(); it != last; ++it) {
+        *output_ << *it << ", ";
+      }
+    }
+    *output_ << v.back() << " ]";
```
SteveBronder, line 410: Same code is replicated across multiple functions. Is there a reason you got rid of `write_vector`?

src/stan/callbacks/json_writer.hpp
```diff
@@ -410,13 +347,91 @@ class json_writer {
...
+  }
+
+  /**
+   * Write a key-value pair where the value is a vector to be made a list.
+   * @param key Name of the value pair
+   * @param values vector to write.
+   */
+  void write(const std::string& key,
+             const std::vector<std::complex<double>>& v) {
```
SteveBronder, line 419: Need to make these virtual in `structured_writer`?

SteveBronder (changes requested):
Few small changes

src/stan/callbacks/json_writer.hpp
WardBrian: The value type is printed differently in each case - this one uses `<<`, the vector<string> uses `process_string()`, etc. 

Write_vector was also just four functions which were the same as these but called from a template. To use this as a derived class we needed un-templated versions

src/stan/callbacks/json_writer.hpp
WardBrian: Done for all the functions

src/stan/callbacks/json_writer.hpp
SteveBronder: Oh for NA etc.

I'd still call it `write_` and have the signature like 

```cpp
template <typename T, stan::require_integral_t<T>* = nullptr>
  void write_int_like(const std::string& key, T value) 
```

SteveBronder (approved):
Looks good besides the int thing which I'll leave as optional

PR #3230, opened 2023-09-13
Feature/3181 json hmc tuning params

mitzimorris:
#### Submission Checklist

#### Summary

Add hooks to adaptive sampler methods to write adapted metric to JSON file.

#### Intended Effect

Make it easier to get adapted metric.

#### How to Verify

Unit tests
#### Side Effects

N/A

#### Documentation

N/A

#### Copyright and Licensing


mitzimorris:
@SteveBronder and @WardBrian - current implementation.  needs unit tests for unit_e_metric.  would appreciate feedback on best way to do method signature.

src/stan/services/sample/hmc_nuts_dense_e_adapt.hpp
```diff
@@ -219,14 +349,15 @@ int hmc_nuts_dense_e_adapt(
     callbacks::interrupt& interrupt, callbacks::logger& logger,
     std::vector<InitWriter>& init_writer,
     std::vector<SampleWriter>& sample_writer,
-    std::vector<DiagnosticWriter>& diagnostic_writer) {
+    std::vector<DiagnosticWriter>& diagnostic_writer,
+    std::vector<callbacks::json_writer<Stream, Deleter>>& metric_writer) {
```
WardBrian, line 353: Here we should just have the metric_writer itself be a template, similar to SampleWriter etc

mitzimorris:
in services layer using templated arg "MetricWriter",  and using `stan::callbacks::structured_writer` in services::util::mcmc_writer.hpp` and `mcmc::hmc::base_hmc.hpp`, and then in unit test, calling with a `json_writer` results in compiler error:

```
src/stan/services/sample/hmc_nuts_diag_e_adapt.hpp:545:5: note: candidate function template not viable: no known conversion from 'std::vector<stan::callbacks::json_writer<std::stringstream, deleter_noop>>' (aka 'vector<json_writer<basic_stringstream<char>, deleter_noop>>') to 'std::vector<callbacks::structured_writer> &' for 27th argument
int hmc_nuts_diag_e_adapt(
    ^
```

not enough of a types/template ninja for this.

WardBrian:
> unit tests must declare vectors as type `structured_writer`

This doesn’t sound totally right to me. Would you mind pushing the code as you have it?

mitzimorris:
pushed code w/ compiler error - trying

```
> python3 runTests.py src/test/unit/services/sample/hmc_nuts_diag_e_adapt_parallel_match_test.cpp
```

generates this error:

```
/unit/services/sample/hmc_nuts_diag_e_adapt_parallel_match_test.cpp -o test/unit/services/sample/hmc_nuts_diag_e_adapt_parallel_match_test.o
src/test/unit/services/sample/hmc_nuts_diag_e_adapt_parallel_match_test.cpp:96:21: error: no matching function for call to 'hmc_nuts_diag_e_adapt'
  int return_code = stan::services::sample::hmc_nuts_diag_e_adapt(
                    ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
src/stan/services/sample/hmc_nuts_diag_e_adapt.hpp:545:5: note: candidate function template not viable: no known conversion from 'std::vector<stan::callbacks::json_writer<std::stringstream, deleter_noop>>' (aka 'vector<json_writer<basic_stringstream<char>, deleter_noop>>') to 'std::vector<callbacks::structured_writer> &' for 27th argument
```

src/stan/services/sample/hmc_nuts_diag_e_adapt.hpp
```diff
@@ -565,7 +553,7 @@ int hmc_nuts_diag_e_adapt(
     std::vector<InitWriter>& init_writer,
     std::vector<SampleWriter>& sample_writer,
     std::vector<DiagnosticWriter>& diagnostic_writer,
-    std::vector<callbacks::json_writer<Stream, Deleter>>& metric_writer) {
+    std::vector<callbacks::structured_writer>& metric_writer) {
```
WardBrian, line 555: That error is because you didn't use a template inside this std::vector - you did the right thing for `hmc_nuts_dense_e_adapt` but it looks wrong in these other places

mitzimorris: thanks!

mitzimorris:
ready for re-review.

src/stan/mcmc/hmc/hamiltonians/unit_e_point.hpp
WardBrian, line 1: Is there a reason this file needed changing? It doesn't seem like the behavior with respect to the API changed

src/stan/services/util/mcmc_writer.hpp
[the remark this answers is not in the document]
WardBrian: You could also add an overloaded constructor which just sets the `metric_writer` to the no-op base class. This would remove the need to change the code in fixed_param, for example.

src/stan/services/util/run_sampler.hpp
```diff
@@ -45,7 +47,9 @@ void run_sampler(stan::mcmc::base_mcmc& sampler, Model& model,
                  size_t num_chains = 1) {
   Eigen::Map<Eigen::VectorXd> cont_params(cont_vector.data(),
                                           cont_vector.size());
-  services::util::mcmc_writer writer(sample_writer, diagnostic_writer, logger);
+  callbacks::structured_writer dummy_metric_writer;
+  services::util::mcmc_writer writer(sample_writer, diagnostic_writer,
+                                     dummy_metric_writer, logger);
```
WardBrian, line 52: These kinds of changes are the thing we could avoid by adding another constructor to mcmc_writer

src/test/unit/services/sample/hmc_nuts_dense_e_adapt_parallel_match_test.cpp
```diff
@@ -97,26 +114,38 @@ TEST_F(ServicesSampleHmcNutsDenseEAdaptParMatch, single_multi_match) {
...
     auto par_str = par_parameters[i].get_stream().str();
     auto sub_par_str = par_str.substr(par_str.find("Elements") - 1);
     std::istringstream sub_par_stream(sub_par_str);
     Eigen::MatrixXd par_mat
         = stan::test::read_stan_sample_csv(sub_par_stream, 80, 9);
     par_res.push_back(par_mat);
+
+    par_metrics.push_back(ss_metric[i].str());
+    rapidjson::Document document;
+    ASSERT_FALSE(document.Parse<0>(par_metrics[i].c_str()).HasParseError());
```
WardBrian, line 128: [Optional] it might be nice to pull this "assert that the JSON in this string parses" logic into a test utility function/macro so we can re-use it rather than needing to remember the RapidJSON incantation.

src/stan/services/util/mcmc_writer.hpp
WardBrian, line 1: On further inspection, is it even necessary that this be inside mcmc_writer? It looks like `run_adaptive_sampler` uses it _outside_ of the mcmc_writer only

src/stan/mcmc/hmc/hamiltonians/unit_e_point.hpp
mitzimorris: yes - the member variable `inv_e_metric` is accessed here:
https://github.com/stan-dev/stan/blob/9ec019bb96ed12e1de2aee35d6e84fa69fbe785f/src/stan/mcmc/hmc/base_hmc.hpp#L70-L75

there's no reason why the unit_e_point couldn't have been written this way - the unit metric is set, just never updated.

WardBrian (changes requested):
A few more questions/comments - mostly on the internals, the API/overloads of the services functions all look good at this point!

src/stan/services/util/mcmc_writer.hpp
mitzimorris: by this, you mean diagnostic writer?

src/test/unit/services/sample/hmc_nuts_dense_e_adapt_parallel_match_test.cpp
```diff
@@ -75,18 +92,18 @@ TEST_F(ServicesSampleHmcNutsDenseEAdaptParMatch, single_multi_match) {
...
-      interrupt, logger, init, par_parameters, diagnostic);
-
+      interrupt, logger, init, par_parameters, diagnostics, metrics);
   EXPECT_EQ(0, return_code);
 
   int num_output_lines = (num_warmup + num_samples) / num_thin;
   EXPECT_EQ((num_warmup + num_samples) * num_chains, interrupt.call_count());
   for (int i = 0; i < num_chains; ++i) {
-    stan::test::unit::instrumented_writer seq_init;
+    stan::test::unit::instrumented_writer seq_init;  // MM: what does this do?
```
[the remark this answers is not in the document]
mitzimorris: this hook is something that would be useful to expose, somehow - people want to know what the inits were.

src/stan/services/util/mcmc_writer.hpp
WardBrian: Yes

mitzimorris: will try to simplify!

mitzimorris: you're right - don't need metric_writer in mcmc_writer.  gone.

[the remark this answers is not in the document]
mitzimorris: reverted `mcmc_writer` to how it was before - set of calls in `mcmc_writer` class doesn't include `write_metric`.

src/stan/services/sample/hmc_nuts_unit_e_adapt.hpp
```diff
@@ -178,20 +232,91 @@ int hmc_nuts_unit_e_adapt(
...
+    double delta, double gamma, double kappa, double t0,
+    callbacks::interrupt& interrupt, callbacks::logger& logger,
+    std::vector<InitWriter>& init_writer,
+    std::vector<SampleWriter>& sample_writer,
+    std::vector<DiagnosticWriter>& diagnostic_writer) {
+  std::vector<stan::callbacks::structured_writer> dummy_metric_writer;
+  dummy_metric_writer.reserve(num_chains);
+  for (size_t i = 0; i < num_chains; ++i) {
+    dummy_metric_writer.emplace_back(stan::callbacks::structured_writer());
+  }
```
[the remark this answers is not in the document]
mitzimorris: changed everywhere.

src/stan/services/util/run_sampler.hpp
mitzimorris: not needed after all.

src/stan/services/sample/hmc_nuts_diag_e_adapt.hpp
```diff
@@ -101,18 +99,130 @@ int hmc_nuts_diag_e_adapt(
...
+    double stepsize_jitter, int max_depth, double delta, double gamma,
+    double kappa, double t0, unsigned int init_buffer, unsigned int term_buffer,
+    unsigned int window, callbacks::interrupt& interrupt,
+    callbacks::logger& logger, callbacks::writer& init_writer,
+    callbacks::writer& sample_writer, callbacks::writer& diagnostic_writer,
+    callbacks::structured_writer& metric_writer) {
+  stan::io::dump dmp
+      = util::create_unit_e_diag_inv_metric(model.num_params_r());
+  stan::io::var_context& unit_e_metric = dmp;
+  return hmc_nuts_diag_e_adapt(
```
WardBrian, line 212: Question: is it necessary to provide this combination (where you don't initialize the metric but do want it written out in the end)?

Just trying to minimize the boilerplate if possible but if the existing calling structure requires it to be so I understand

WardBrian (approved):
Two small questions but I think this is good!

src/stan/services/sample/hmc_nuts_diag_e_adapt.hpp
mitzimorris: yes, absolutely we want to do this - the default is to start from the unit metric and then let it evolve during adaptation.  this would be the starting point in trying to figure out what a good metric is.

WardBrian: Oh, surely, the end user probably wants this option more than many others. What I meant more so was do we actually need this overload provided for everything to compile on our end (e.g., in cmdstan we never call this, just the one with `num_chains`, right?)

mitzimorris:
re your question about not breaking RStan and PyStan - I think I should add back the old signature to run_adaptive_sampler.

src/stan/services/sample/hmc_nuts_diag_e_adapt.hpp
mitzimorris: perhaps only the unit tests will break?  I can test against CmdStan.

mitzimorris:
for multi-chain adaptive NUTS-HMC, there are now only 2 signatures:  with and without pre-specified metric, both of which now also have add'l arg for metric_writer.

updated the unit tests accordingly.

if CmdStan is the only interface that calls these methods, this shouldn't mess anything up.

mitzimorris:
performance tests now failing because current CmdStan calls to adaptive sampler don't match new services API.

mitzimorris:
very confused.  what is the 3rd?

WardBrian:
In develop, there are 4 overloads of `hmc_nuts_diag_e_adapt`. 

Back in [e4f62f0](https://github.com/stan-dev/stan/pull/3230/commits/e4f62f05ef3ab8f7f771af109993e5d5b38547ca), there were 8 overloads of `hmc_nuts_diag_e_adapt`, including 4 which accepted the new argument `metric_writer` (I miscounted when I said three).

However, of those 4 original overloads, I don't believe all of them are really in use besides for backwards compatibility, since several just call each other with extra arguments. If that is the case, we don't need to (and I would argue, should not) add functionality to the overloads which are already exposing some subset of the full feature set.

I haven't tried it myself, which is why I was asking if they were all necessary

mitzimorris:
my sad conclusion is that all of these are necessary.  reverted changes.

WardBrian (approved):

PR #3236, opened 2023-10-27
jenkins: cleanup old build logs

dylex:
Add jenkins configuration to remove old build logs and artifacts after some time to avoid running out of disk space. Feel free to tweak the specific numbers, as long as we don't keep them forever.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -95,6 +95,7 @@ pipeline {
         skipDefaultCheckout()
         preserveStashes(buildCount: 7)
         parallelsAlwaysFailFast()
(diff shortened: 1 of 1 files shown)


bob-carpenter:
Thanks, Dylan.  I don't think we even need these for 30 days if you need to recover more space.  I'll approve and this can be merged when tests pass.

dylex:
Thanks. It was mainly Math that was a problem (had almost 500G of logs), for which I've added this already. Others are much smaller, but it would be good to propagate it into all the other packages eventually.

bob-carpenter (approved):
Thanks!

Issue #3242, opened 2023-11-16
Add resampling options for multi-Pathfinder


avehtari:
1. Currently, multi-Pathfinder always resamples. When the approximation is far from the true target and especially in case of high dimensional posteriors, one weight can dominate and resampling returns only copies of a single draw. Furthermore, in edge cases, Pareto-smoothing or computation of weights can fail and produce completely unexpected results. To allow 1) further diagnostics in such problematic cases and 2) have more unique draws to initialize MCMC to improve MCMC convergence diagnostics
 - [ ] add boolean option `resample` with default value `TRUE`, and value `FALSE` disabling the resampling

2. For cases where we don't care about doing further diagnostics, but want a minimum number of unique draws, allow resampling without replacement (and then set `num_draws` to the number of unique draws needed, e.g. for initializing MCMC)
 - [ ] add boolean option `with_replacement` with default value `TRUE`, and value `FALSE` using without replacement resampling

WardBrian:
What is returned when resample is false?

avehtari:
When resample is false, return the draws from the normal approximations (skip lines 173-186 and index with consecutive numbers on line 188 of multi.hpp)

WardBrian:
I still don't think I understand how this is different from setting `save_single_paths` to true and using those draws

avehtari:
- I wasn't aware of `save_single_paths` option (after looking the doc, I think it's not well documented (one part gives an impression of that it saves the paths and not the draws), and also it's not currently working in CmdStanR (https://github.com/stan-dev/cmdstanr/issues/878#issuecomment-1870108851)
- `save_single_paths` saves also the paths, which I'm not interested in, and to load the draws from the individual csv's would require additional code from the user or method in CmdStanR, so it would not be as clean solution

WardBrian:
#3249

PR #3249, opened 2024-01-02
Adds option to turn off psis resampling and lp calculation for pathfinder

SteveBronder:
#### Submission Checklist

#### Summary

Adds the request from @avehtari in #3242 to add an option for turning off resampling and calculating the lp.

 If `psis_resample` is false then psis resampling is not performed and the full set of samples is returned from all the individual pathfinders. 

If `return_lp` is false then the lp values are not calculated for each sample. I think a lot of upstream packages assume that a `lp__` column at least exists so here I just set the column of lp values to `NaN`. This also means psis resampling is turned off since we cannot calculate the lp ratio and the full set of samples is returned from all the individual pathfinders. 

Something questionable here is that since we have to calculate the lp values for the set of samples we compute the elbo metric on we just get those lp calculations for free. So here I still include those lp values even if `return_lp` is false. But that could be weird for users. I documented it, but another approach here would be to just explicitly set all lp values in that column to `NaN`.

#### How to Verify

Tests were added for the new options and can be run with 

```bash
python3 ./runTests.py ./src/test/unit/services/pathfinder/normal_glm_test.cpp
```

#### Documentation

Once we open the cmdstan PR we should add documentation for these new options

#### Copyright and Licensing

### DIFF
--- a/src/stan/callbacks/unique_stream_writer.hpp
+++ b/src/stan/callbacks/unique_stream_writer.hpp
@@ -136,7 +136,7 @@ class unique_stream_writer final : public writer {
    * @param[in] v Values in a block representing an Eigen column vector
    */
   virtual void operator()(
-      const Eigen::Block<Eigen::Matrix<double, -1, -1>, -1, 1, true>) {
+      const Eigen::Block<Eigen::Matrix<double, -1, -1>, -1, 1, true>& values) {
     if (output_ == nullptr)
       return;
     Eigen::IOFormat CommaInitFmt(Eigen::StreamPrecision, Eigen::DontAlignCols,
@@ -153,7 +153,7 @@ class unique_stream_writer final : public writer {
    * @param[in] v Values in a block representing an Eigen row vector
    */
   virtual void operator()(
-      const Eigen::Block<Eigen::Matrix<double, -1, -1>, 1, -1, true>) {
+      const Eigen::Block<Eigen::Matrix<double, -1, -1>, 1, -1, true>& values) {
     if (output_ == nullptr)
       return;
     Eigen::IOFormat CommaInitFmt(Eigen::StreamPrecision, Eigen::DontAlignCols,
--- a/src/stan/callbacks/unique_stream_writer.hpp
+++ b/src/stan/callbacks/unique_stream_writer.hpp
@@ -118,8 +118,8 @@ class unique_stream_writer final : public writer {
   /**
    * Comma formatter for writing Eigen matrices
    */
-  Eigen::IOFormat CommaInitFmt(Eigen::StreamPrecision, Eigen::DontAlignCols,
-                               ", ", "", "", "\n", "", "");
+  Eigen::IOFormat CommaInitFmt{Eigen::StreamPrecision, Eigen::DontAlignCols,
+                               ", ", "", "", "\n", "", ""};
 
   /**
    * Output stream
--- a/src/stan/callbacks/unique_stream_writer.hpp
+++ b/src/stan/callbacks/unique_stream_writer.hpp
@@ -106,7 +106,7 @@ class unique_stream_writer final : public writer {
    * parameters in the rows and samples in the columns. The array is then
    * transposed for the output.
    */
-  void operator()(const Eigen::Array<double, -1, -1>& vals) {
+  void operator()(const Eigen::Array<double, -1, -1>& values) {
     if (output_ == nullptr)
       return;
     Eigen::IOFormat CommaInitFmt(Eigen::StreamPrecision, Eigen::DontAlignCols,
--- a/src/stan/callbacks/unique_stream_writer.hpp
+++ b/src/stan/callbacks/unique_stream_writer.hpp
@@ -91,16 +91,6 @@ class unique_stream_writer final : public writer {
   void operator()(const Eigen::Ref<Eigen::Matrix<double, -1, -1>>& values) {
     if (output_ == nullptr)
       return;
-    const bool is_row_vector = values.rows() == 1 && values.cols() > 1;
-    Eigen::IOFormat CommaInitFmt;
-    if (is_row_vector) {
-      CommaInitFmt = Eigen::IOFormat(Eigen::StreamPrecision, Eigen::DontAlignCols,
-                                "\n", ", ", "", "", "", "\n");
-    } else {
-      CommaInitFmt = Eigen::IOFormat(Eigen::StreamPrecision, Eigen::DontAlignCols,
-                                ", ", "", "", "\n", "", "");
-    }
-    
     *output_ << values.transpose().format(CommaInitFmt);
   }
 
@@ -126,6 +116,12 @@ class unique_stream_writer final : public writer {
   }
 
  private:
+  /**
+   * Comma formatter for writing Eigen matrices 
+   */
+  Eigen::IOFormat CommaInitFmt(Eigen::StreamPrecision, Eigen::DontAlignCols,
+                            ", ", "", "", "\n", "", "");    
+
   /**
    * Output stream
    */
--- a/src/stan/callbacks/unique_stream_writer.hpp
+++ b/src/stan/callbacks/unique_stream_writer.hpp
@@ -97,6 +97,24 @@ class unique_stream_writer final : public writer {
   }
 
   /**
+   * Writes multiple rows and columns of values in csv format.
+   *
+   * Note: the precision of the output is determined by the settings
+   *  of the stream on construction.
+   *
+   * @param[in] values An array of values. The input is expected to have
+   * parameters in the rows and samples in the columns. The array is then
+   * transposed for the output.
+   */
+  void operator()(const Eigen::Array<double, -1, -1>& vals) {
+    if (output_ == nullptr)
+      return;
+    Eigen::IOFormat CommaInitFmt(Eigen::StreamPrecision, Eigen::DontAlignCols,
+                                 ", ", "", "", "\n", "", "");
+    *output_ << values.transpose().format(CommaInitFmt);
+  }
+
+  /**
    * Writes the comment_prefix to the stream followed by a newline.
    */
   void operator()() {
--- a/src/stan/callbacks/unique_stream_writer.hpp
+++ b/src/stan/callbacks/unique_stream_writer.hpp
@@ -77,6 +77,7 @@ class unique_stream_writer final : public writer {
       return;
     write_vector(values);
   }
+  
   /**
    * Writes a set of values in csv format followed by a newline.
    *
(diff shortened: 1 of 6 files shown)


src/stan/services/pathfinder/multi.hpp
```diff
@@ -158,53 +167,68 @@ inline int pathfinder_lbfgs_multi(
...
+    std::string psis_time_str = std::string(time_header.size(), ' ')
+                                + std::to_string(psis_delta_time)
+                                + " seconds (PSIS)";
+    parameter_writer(psis_time_str);
+    std::string total_time_str
+        = std::string(time_header.size(), ' ')
+          + std::to_string(pathfinders_delta_time + psis_delta_time)
+          + " seconds (Total)";
+    parameter_writer(total_time_str);
+  }
```
WardBrian, line 231: It seems like we would want to print the total time even when we did not perform PSIS?

src/test/unit/services/pathfinder/util.hpp
WardBrian, line 1: Why does this file change?

src/stan/services/pathfinder/multi.hpp
```diff
@@ -77,6 +77,14 @@ namespace pathfinder {
  * @param[in,out] parameter_writer output for parameter values
  * @param[in,out] diagnostic_writer output for diagnostics values,
  * `error_codes::SOFTWARE` for failures
+ * @param[in] return_lp Whether single pathfinder should return lp calculations.
```
WardBrian, line 80: I don't know about the name `return_lp`, since it (and this sentence, IMO) suggests the calculations are being done, and this flag just determines whether we share them, rather than the reality that it controls whether or not the log probability for each sample is even calculated.

src/stan/services/pathfinder/multi.hpp
```diff
@@ -77,6 +77,14 @@ namespace pathfinder {
...
+ *  If `true`, calculates the joint log probability for each sample.
+ *  If `false`, (`num_draws` - `num_elbo_draws`) of the joint log probability
+ *  calculations will be `NA` and psis resampling will not be performed.
+ * @param[in] psis_resampling If `true`, psis resampling is performed over the
+ *  samples returned by all of the individual pathfinders and `num_multi_draws`
+ *  samples are returned. If `false`, no psis resampling is performed
+ *  and (`num_paths` * `num_draws`) samples are returned.
```
WardBrian, line 87: They're not really returned, but written to the output writer. 

Might be worth noting that this is equivalent to concatenating the end values of of the `single_path_parameter_writer`s if they were specified, with less computation

WardBrian (changes requested):
A few comments. Boolean flags generally make me nervous, especially when they are interdependent or change the sizes of the return values (and these do both!) but I don't see any other reasonable way of fulfilling this request

src/stan/services/pathfinder/multi.hpp
SteveBronder: Yeah I think I'm going to change this to `calculate_lp`

src/stan/services/pathfinder/multi.hpp
SteveBronder: Agree going to update the docs. return is def not the right word

src/stan/services/pathfinder/multi.hpp
SteveBronder: I thought to leave it out since the only reported time is `Pathfinder`. I could just return `Pathfinder` as total time when psis is turned off

src/test/unit/services/pathfinder/util.hpp
SteveBronder: I need to read in an array to the `values_` for multi pathfinder. I actually need to add this to  `writer` and `unique_stream_writer`

WardBrian (approved):

src/stan/services/pathfinder/multi.hpp
```diff
@@ -158,53 +169,69 @@ inline int pathfinder_lbfgs_multi(
...
-      boost::random::discrete_distribution<Eigen::Index, double>>
-      rand_psis_idx(rng,
-                    boost::random::discrete_distribution<Eigen::Index, double>(
-                        boost::iterator_range<double*>(
-                            weight_vals.data(),
-                            weight_vals.data() + weight_vals.size())));
-  for (size_t i = 0; i <= num_multi_draws - 1; ++i) {
-    parameter_writer(samples.col(rand_psis_idx()));
+  double psis_delta_time = 0;
+  if (psis_resample && calculate_lp) {
```
andrjohns, line 183: Possibly a bit cheeky, but while you're in here would it be much of a pain to move this resampling process to a separate function (possibly living in `psis.hpp`?) which just gets called here? It would be great if I could perform the resampling with multiple existing single-pathfinder results.

I can also open a PR to do it once this one is merged, if you want to keep this one focused on the resampling option, just let me know!

For context, I can't use the TBB for parallelism in [StanEstimators](https://github.com/andrjohns/StanEstimators), so to parallelise multi-pathfinder I'll have to run multiple single-paths and then perform the resampling on the results.

SteveBronder: Yeah that's reasonable!

SteveBronder: Actually could we do this in a separate PR? Looking at it rn I'm not sure if you want back just the resampling weights (which would be just calling `psis_weights` or if you want a function that does the resampling and writes to a `parameter_writer`

src/stan/callbacks/unique_stream_writer.hpp
```diff
@@ -96,6 +96,24 @@ class unique_stream_writer final : public writer {
...
+   * Writes multiple rows and columns of values in csv format.
+   *
+   * Note: the precision of the output is determined by the settings
+   *  of the stream on construction.
+   *
+   * @param[in] values An array of values. The input is expected to have
+   * parameters in the rows and samples in the columns. The array is then
+   * transposed for the output.
+   */
+  void operator()(const Eigen::Array<double, -1, -1>& values) {
```
WardBrian, line 109: This overload is now causing failures because it is ambiguous (I guess because of the implicit conversion constructor between the Array and Matrix classes?)

This may be a silly question, but because of where it is used, we never actually need the memory as an Array again, so is there some way we could move-construct a matrix and then call the existing overload?

src/stan/services/pathfinder/multi.hpp
andrjohns: Of course! I'll open a PR once this is in

src/stan/callbacks/unique_stream_writer.hpp
```diff
@@ -77,6 +77,69 @@ class unique_stream_writer final : public writer {
...
+   * Note: the precision of the output is determined by the settings
+   *  of the stream on construction.
+   *
+   * @param[in] v Values in an Eigen vector
+   */
+  void operator()(const Eigen::Matrix<double, -1, 1>& values) {
+    if (output_ == nullptr)
+      return;
+    Eigen::IOFormat CommaInitFmt(Eigen::StreamPrecision, Eigen::DontAlignCols,
+                                 ", ", "", "", "\n", "", "");
```
WardBrian, line 92: I believe you can create a different formatter such that the transpose is not necessary - not sure if this actually affects the generated code:

```c++
Eigen::IOFormat CommaInitFmt(Eigen::StreamPrecision, Eigen::DontAlignCols,
                                 "\n", ", ", "", "", "", "\n");
```

src/stan/callbacks/writer.hpp
WardBrian, line 1: Even without the ambiguity issue we're currently facing in CI, I still think I prefer a design which doesn't handle `Eigen::Arrays` but forces the calling code to coerce that into a Matrix.

Alternatively, could these take `Eigen::DenseBase`?

WardBrian (changes requested):

src/stan/callbacks/writer.hpp
WardBrian: DenseBase still requires a template parameter,  so it's a no go. 

I've confirmed it all works if you make `samples` in the current code a `Map<Array<...>>` with the underlying data being a Matrix, and that avoids the copy we are worried about

SteveBronder: I think I'm going to just make everything matrices. It would be nice to support arrays but yeah I thiink it gets too confusing

WardBrian: Even with the latest changes, I still get a failure building cmdstan: 

```
/home/brian/Dev/cpp/stan/src/stan/services/pathfinder/multi.hpp:208:23: error: call of ‘(stan::callbacks::unique_stream_writer<std::basic_ofstream<char> >) (Eigen::DenseBase<Eigen::Matrix<double, -1, -1> >::ColXpr)’ is ambiguous
```

I think because this is convertable to either the column vector overload or the array.

SteveBronder: Yes just fixed this up

WardBrian:
I still think adding more overloads to writer (e.g. for Block) is a very heavy handed move. 

As far as I can tell, things would work fine if there was a single version that took `const Eigen::Ref<const Eigen::Matrix<double, -1, -1>>& ` replacing the previous MatrixXd overload. Am I missing something?

SteveBronder:
Let me try that. I think we will still need the vector and row vector versions but that might work nicely for the block overloads

WardBrian:
Passing a RowVector to a MatrixXd is free, right?

SteveBronder:
Printing row vs. column vectors is the issue. I think we just need to change the CommaFormat

WardBrian:
The writer class is pretty explicit that `The input is expected to have parameters in the rows and samples in the columns`. So imo, if there is anything which was passing a wrongly-shaped vector, we should fix that call site

SteveBronder:
That's fine I removed that logic. Should be good to go now

src/stan/callbacks/unique_stream_writer.hpp
```diff
@@ -117,6 +115,12 @@ class unique_stream_writer final : public writer {
   }
 
  private:
+  /**
+   * Comma formatter for writing Eigen matrices
+   */
+  Eigen::IOFormat CommaInitFmt(Eigen::StreamPrecision, Eigen::DontAlignCols,
+                               ", ", "", "", "\n", "", "");
```
WardBrian, line 122:
```suggestion
  Eigen::IOFormat CommaInitFmt = Eigen::IOFormat(Eigen::StreamPrecision, Eigen::DontAlignCols,
                               ", ", "", "", "\n", "", "");
```

WardBrian (approved):
Latest push has a syntax error and I think some test code can be cleaned up, but otherwise these changes look good

PR #3250, opened 2024-01-08
uses Eigen's internal multiindexing for multi indexing on vectors and matrices

SteveBronder:
#### Submission Checklist

#### Summary

Uses Eigen 3.4's new non-contiguous indexing expressions to have the `rvalue()` functions that use `multi_index` return an expression of multi indexing instead of eagerly creating a new matrix from the indices.

In the Stan language doing a loop like the code in (a) is faster than doing a multi index like in (b) because we have to make a giant matrix in (b) for the multi index. 

(a)
```stan
for (i in 1:N)
  mu[i] = beta * X[idx[i]];
```

(b)
```stan
mu = beta * x[idx]
```

Eigen 3.4 introduced an expression for [non-contiguous indexing](https://eigen.tuxfamily.org/dox/group__TutorialSlicingIndexing.html) which we can utilize to not have a big matrix creation in (b).

#### Intended Effect

Make vectorized multi indexing as fast as the loop

#### How to Verify

Tests pass for current rvalue tests with no changes

#### Side Effects

Nope

#### Copyright and Licensing


src/stan/model/indexing/rvalue.hpp
```diff
@@ -144,25 +144,31 @@ inline auto rvalue(Vec&& v, const char* name, index_uni idx) {
...
  * @param[in] idx Sequence of integers.
  * @throw std::out_of_range If any of the indices are out of bounds.
  * @throw std::invalid_argument If the value size isn't the same as
  * the indexed size.
  */
-template <typename EigVec, require_eigen_vector_t<EigVec>* = nullptr>
-inline auto rvalue(EigVec&& v, const char* name, const index_multi& idx) {
+template <typename EigVec, typename MultiIndex,
+          require_eigen_vector_t<EigVec>* = nullptr,
+          require_same_t<MultiIndex, index_multi>* = nullptr>
```
WardBrian, line 157: What is the purpose of the template if you're constraining it to one concrete type?

src/stan/model/indexing/rvalue.hpp
```diff
@@ -144,25 +144,31 @@ inline auto rvalue(Vec&& v, const char* name, index_uni idx) {
...
+inline auto rvalue(EigVec&& v, const char* name, MultiIndex&& idx) {
```
WardBrian, line 158: idx can no longer be `const`?

src/test/unit/model/indexing/rvalue_test.cpp
```diff
@@ -719,7 +719,7 @@ void vector_multi_test() {
   test_out_of_range(v, index_multi(ns));
 }
 
-TEST(ModelIndexing, rvalueVectorMulti) { vector_multi_test<Eigen::VectorXd>(); }
+TEST(ModelIndexing, x) { vector_multi_test<Eigen::VectorXd>(); }
```
WardBrian, line 722: Necessary?

WardBrian (commented):
A few questions - might be easier to explain next week over a board.

Have you fed this into your benchmarking framework to make sure it has the desired effect?

src/stan/model/indexing/rvalue.hpp
SteveBronder: Perfect forwarding

src/stan/model/indexing/rvalue.hpp
```diff
@@ -144,25 +144,31 @@ inline auto rvalue(Vec&& v, const char* name, index_uni idx) {
...
-              return v_ref.coeff(idx.ns_[i] - 1);
-            });
+      [](auto&& v_ref, auto&& idx_inner) {
+        Eigen::Map<const Eigen::Array<int, -1, 1>> idx2(idx_inner.ns_.data(),
+                                                        idx_inner.ns_.size());
+        return std::forward<decltype(v_ref)>(v_ref)(idx2 - 1);
       },
-      stan::math::to_ref(v));
+      std::forward<fwd_t>(stan::math::to_ref(std::forward<EigVec>(v))),
+      std::forward<MultiIndex>(idx));
```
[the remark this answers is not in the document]
SteveBronder: Yep!

src/stan/model/indexing/rvalue.hpp
SteveBronder: The && is a perfect forwarding template parameter so it could be possibly const if the input is const

src/test/unit/model/indexing/rvalue_test.cpp
SteveBronder: Oops, I think this was a typo!

WardBrian:
@SteveBronder did you get a chance to try this in your stan-benchmark playground?

SteveBronder:
I'm waiting till I have the math pr for this up as well so that I can benchmark this and the new matrix type all at once

WardBrian (approved):
@SteveBronder and I discussed some benchmarking he did (which hopefully he can post a nice graph of) which showed that this change did not immediately lead to the improvements we thought it might - that will require some specializations in the arena matrix representation.

But, it also didn't make things any worse, so I'm approving this on its own now.

SteveBronder:
Running some benchmarks idt we should merge this PR yet. The benchmark can be found [here](https://github.com/SteveBronder/stan-perf/tree/main/benchmarks/eigen_multi_index). I took the code that brms generates and compared the new and current multi indexing to a loop as well as to a raw call from Eigen that does the same thing.

The plot has facets by the number of group (M) and plots vectors of size N against the average time for each combination of M and N. The indices are set up so that each inner index does the worst case scenario of random access to elements in the vectors. 

Each line represents

1. Code generated from a a for loop in Stan
2. Code generated from a multi index in Stan with the current multi index code
3. Code generated from a multi index in Stan with the new multi index code
4. Code generated from a multi index in Stan with the new multi index code, removing any checks
5. Using plain eigen to do the multi index

The results were very suprising to me

image

Looping is very consistent with linear time, but it looks like our multi indexing has a kink in it once we cross 50K observations or so irrespective of the number of groups. The main reason the new and current multi indexing is slower is because of the checks, which you can see from the `nocheck` time

The loop does not have to do any checks. For multi index we have to check that the index is valid within the range of the size of the vector. For the current multi index we do that check as we are creating the index

```c++
        return plain_type_t<EigVec>::NullaryExpr(
            idx.ns_.size(), [name, &idx, &v_ref](Eigen::Index i) {
              math::check_range("vector[multi] indexing", name, v_ref.size(),
                                idx.ns_[i]);
              return v_ref.coeff(idx.ns_[i] - 1);
            });
```

but in the new code we do that check outside of the loop

```c++
  for (auto idx_i : idx.ns_) {
    math::check_range("vector[multi] indexing", name, v_size, idx_i);
  }
```

I think iterating over the index for the check is causing the main slowdown. The kink in the graph after 50K most likely happens because the index and vectors no longer fit in the lower levels of cache. Having that check inside or outside of the index creation most likely explains why this PRs multi indexing is slower than the current multi indexing

The nocheck benchmark does a few things that I think we could do in Stan, from most to least performance gain:

1. Removes the check. This could be done with a flag defined at compile time. We could also do something smart in the compiler if we wanted to
    - Since we know the size of the containers we will be indexing into we could actually have a check in the constructor that validates the size once. Then we know that the index will be correct for all the future operations.
2. Removes `multi_index`. The `multi_index` class is really just a holder for a vector with nothing else, but constructing that class requires a copy of the indexing vector. We would need to write `rvalue()` functions that accept vectors as the index and then the compiler would need to generate the right code.
3. Removes the `deep_copy(mu)` normallly generated by the code. We don't manipulate `mu` here so it is safe to remove. When and when not to deep copy is kind of tricky. We want the deep copy when
    - The left hand side and right hand side use the same object (mu here for the `+=`)
    - The right hand side accesses the data used in the lhs in a non-linearly increasing and/or non-sequential order 
        - (ex `mu[1:3] += mu[{1, 1, 1}] * b` where the first indexed value changes after the first assignment)

I think 1:3 are very feasible and the speed gain seems very worth it

SteveBronder:
But my end recommendation is we close this PR without merge. I can write up a seperate PR that does 1 and 2 from the list above. We can add those signatures and then add the necessary components from the compiler.

The good news is we have a reasonable path forward that should give a very nice speedup to any brms model!

WardBrian:
I am still not sure if the benchmark you posted represents @bob-carpenter's original case he was concerned with, since it doesn't ever end up calling a lpdf. The concern as I understand it was when these matrices are constructed for use in a distribution, one way of writing it ends up leading to an un-vectorized lpdf, which goes against the other advice we give

SteveBronder:
Doing the multi index inside an lpdf would give the same effects

SteveBronder:
Actually @WardBrian I want to reopen this. Running the benchmarks on my mac, just turning on `STAN_NO_RANGE_CHECKS` with the new code gives us a very good speedup.

image

bob-carpenter:
The plots make it look like the Eigen approach is the best.  Am I missing something?  

Also, what specifically are you evaluating?  I see the label on one of the plots involves multi-indexing with an array and then elementwise multiplication?  Usually we just do the indexing for random effects, so I'm not sure where the elementwise multiply is coming from.  Normally it'd just be `foo[idxs] + bar[idxs2] + ...` for varying effects.

I had thought part of the problem with multi-indexing was creating an intermediate, but it's now returning a template expression created by binding.

```cpp
template <typename EigVec, require_eigen_vector_t<EigVec>* = nullptr>
inline auto rvalue(EigVec&& v, const char* name, const index_multi& idx) {
  return stan::math::make_holder(
      [name, &idx](auto& v_ref) {
        return plain_type_t<EigVec>::NullaryExpr(
            idx.ns_.size(), [name, &idx, &v_ref](Eigen::Index i) {
              math::check_range("vector[multi] indexing", name, v_ref.size(),
                                idx.ns_[i]);
              return v_ref.coeff(idx.ns_[i] - 1);
            });
      },
      stan::math::to_ref(v));
}
```

Instead, it looks like the issue is creating the nested `NullaryExpr` for every single argument.  Won't that be at least as expensive as the range check?  

With the Eigen approach, how do you avoid the `- 1` for indexing?

SteveBronder:
> The plots make it look like the Eigen approach is the best. Am I missing something?

The code for all of the benchmarks are [here](https://github.com/SteveBronder/stan-perf/tree/main/benchmarks/eigen_multi_index). The Eigen benchmark is just doing 

```
Eigen::Matrix<double, -1, 1> mu = r_1_1(J_1_map).array() * Z_1_1.array() +
     r_1_2(J_1_map).array() * Z_1_2.array();
```

Which is a baseline for "If we remove all of the stan overhead like checks, indexing from 1, expression safety, etc. what is the best possible performance we could get?"

> Also, what specifically are you evaluating?

I have all the benchmark code in the link above and the comment above (comment [here](https://github.com/stan-dev/stan/pull/3250#issuecomment-2075380874)) gives all the info for each benchmark. But tldr each line is for 

```
Each line represents

1. (loop) Code generated from a for loop in Stan
2. (orig) Code generated from a multi index in Stan with the current multi index code
3. (new) Code generated from a multi index in Stan with the new multi index code
4. (newnocheck) Code generated from a multi index in Stan with the new multi index code, removing any checks
5. (eigen) Using plain eigen to do the multi index

>  I see the label on one of the plots involves multi-indexing with an array and then elementwise multiplication? Usually we just do the indexing for random effects, so I'm not sure where the elementwise multiply is coming from. Normally it'd just be foo[idxs] + bar[idxs2] + ... for varying effects.

The elemenwise multiplication and multi index is what brms generates for models that do random effects like `y ~ (effect | group)`

> Instead, it looks like the issue is creating the nested NullaryExpr for every single argument. Won't that be at least as expensive as the range check?

The Nullary expression is just done once and is then returned so it should not be expensive. After all these benchmarks I'm really convinced the slowdown is from the range check. In the graph above removing the range check gives about a magnitude of performance speedup. You can see in the graph in the comment above all of the stan versions that do the range check sit near one another, but removing that range check with `STAN_NO_RANGE_CHECK` we get much closer to the baseline of a plain Eigen call

> With the Eigen approach, how do you avoid the - 1 for indexing?

It just uses a zero based index instead of Stan's 1 based index since it's a baseline of how fast this operation is minus any Stan changes.

bob-carpenter:
Thanks for all the clarification.  That makes a lot more sense now.

> The elemenwise multiplication and multi index is what brms generates for models that do random effects like y ~ (effect | group)

Aha!  A varying slope model.  I just couldn't think of where we'd want to do a multi-index and then elementwise multiply.  In any case, it sounds like the Eigen baseline isn't something that'd be workable in Stan due to the indexing-from-1 problem.

Sorry, but no insight about how to fix it easily.  It does seem like the current Stan approach is allocating a new class for each index.  Isn't that causing a lot of overhead or is the overhead all compiled away somehow?  I would've thought you could get away with just one level of indirection here.

SteveBronder:
Np!

> In any case, it sounds like the Eigen baseline isn't something that'd be workable in Stan due to the indexing-from-1 problem.

Totally agree. I just have it there as a sort of "if we could do the most performance thing what would that look". It's not a feasible answer but moreso for comparison

> It does seem like the current Stan approach is allocating a new class for each index. Isn't that causing a lot of overhead or is the overhead all compiled away somehow? 

It does make a new class but that class is an expression so there won't be any actual allocation from value. There's a small amount of overhead but most of it is compiled away

I think we should merge this PR and if people want speed they can define the macro to remove the range checks. Idt there's any other way for us to make multi indexing fast

PR #3251, opened 2024-01-15
[BUGFIX] Replace STL regex with Boost for MinGW bug

andrjohns:
#### Submission Checklist

#### Summary

Depending on the user's locale, using `std::regex` with the MinGW GCC in RTools will result in a memory leak and the program hangs.

This has previously been reported as bug to MinGW [here](https://sourceforge.net/p/mingw-w64/mailman/mingw-w64-public/thread/CABFfbXvurKZjO8tjrFrirwy_Spego-nRvOksjkmnDtYsESu%3DYQ%40mail.gmail.com/#msg37203838), but it is still an issue.

For Stan, this can cause a hang when parsing JSON data (seen recently as the cause of this [downstream issue](https://github.com/andrjohns/StanEstimators/issues/1)). We can instead use the Boost Regex headers as a drop-in replacement, which I've verified resolve the issue.

#### Intended Effect

Avoid hanging under certain locales

#### Side Effects

N/A

#### Documentation

N/A

#### Copyright and Licensing

### DIFF
--- a/src/stan/io/json/json_data_handler.hpp
+++ b/src/stan/io/json/json_data_handler.hpp
@@ -16,7 +16,7 @@
 #include <utility>
 #include <vector>
 #include <boost/algorithm/string.hpp>
-#include <regex>
+#include <boost/regex>
 
 namespace stan {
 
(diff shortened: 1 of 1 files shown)


andrjohns:
GCC bug listing here as well: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=98723

WardBrian (approved):
I think we would be fine to include this in a 2.34.1 if we end up needing one, so it should be good to merge now. 

Thanks!

andrjohns:
Awkward, but it looks like I don't have permissions for `stan`:

image


Can I get a grown-up to merge this for me?

syclik:
@andrjohns, merged! That looks like an oversight. 

@WardBrian, can we give @andrjohns permissions?

WardBrian:
So far as I can tell he has them:
image

Maybe it was a github hiccup?

syclik:
Thanks for checking! I hope it was. @andrjohns please let people know if it happens again!

PR #3259, opened 2024-01-25
Only swallow domain_errors in various algorithms 

### LINKED ISSUE #3258: Remove places where exceptions are unconditionally swallowed.
#### Summary:

There are a few places where we catch `std::exception` and do not end up re-throwing it. This prevents features like https://github.com/stan-dev/stanc3/issues/1393 (an `exit()` function in the language), and leads to some unexpected behavior.

For example, if an indexing exception occurs during initialization, the program halts. If it occurs during _sampling_, it's all good. Run this program a few times to get a sense of this:

```stan
parameters {
  real mu;
}
model {
  mu ~ normal(0, 1);
  
  if (mu < 0) {
    array[1] int sigma;
    print(sigma[2]);
  }
}
```

#### Description:

I believe I have found the relevant places that need to change and highlighted them here: https://gist.github.com/WardBrian/a21b6e650606e562ca1466eedaa3eb47

When a similar change was made in https://github.com/stan-dev/stan/pull/2307, the choice was to not catch anything that isn't a `std::domain_error`. This is what `reject()` throws in the language, and a lot of our math functions as well, and generally seems like a good choice for "recoverable".

#### Current Version:
v2.34

WardBrian:
#### Submission Checklist

#### Summary

Closes #3258. This changes any place which caught `std::exception` from the model to only catch `std::domain_error`, matching what is done during initialization. 

This prevents odd behavior like an indexing exception that only occurs randomly being swallowed if initialization happens to pass, and allows things like an `exit()` function in the language.

#### Intended Effect

Cause certain exceptions to terminate the inference algorithms.

#### How to Verify

This model will usually initialize but quickly fail now: 

```stan
parameters {
  real mu;
}
model {
  mu ~ normal(0, 1);

  if (mu > 1) {
    array[1] int sigma;
    print(sigma[2]);
  }
}
```

#### Side Effects

I've installed handlers in the service functions, but code that calls the internals of these algorithms (e.g `run_adaptive_sampler`) _not_ through the services will need to be ready to catch exceptions

Additionally, if any Math functions raise an exception which is intended to be recoverable but is not a domain_error, they will need updating. Because this is the current behavior in initialization, I suspect we have probably already smoked any of these out.

#### Copyright and Licensing

Simons Foundation

### DIFF
--- a/src/stan/services/optimize/bfgs.hpp
+++ b/src/stan/services/optimize/bfgs.hpp
@@ -96,7 +96,16 @@ int bfgs(Model& model, const stan::io::var_context& init,
   if (save_iterations) {
     std::vector<double> values;
     std::stringstream msg;
-    model.write_array(rng, cont_vector, disc_vector, values, true, true, &msg);
+    try {
+      model.write_array(rng, cont_vector, disc_vector, values, true, true,
+                        &msg);
+    } catch (const std::exception& e) {
+      if (msg.str().length() > 0) {
+        logger.info(msg);
+      }
+      logger.error(e.what());
+      return error_codes::SOFTWARE;
+    }
     if (msg.str().length() > 0)
       logger.info(msg);
 
@@ -119,7 +128,13 @@ int bfgs(Model& model, const stan::io::var_context& init,
           "  # evals"
           "  Notes ");
 
-    ret = bfgs.step();
+    try {
+      ret = bfgs.step();
+    } catch (const std::exception& e) {
+      logger.error(e.what());
+      return error_codes::SOFTWARE;
+    }
+
     lp = bfgs.logp();
     bfgs.params_r(cont_vector);
 
@@ -150,8 +165,16 @@ int bfgs(Model& model, const stan::io::var_context& init,
     if (save_iterations) {
       std::vector<double> values;
       std::stringstream msg;
-      model.write_array(rng, cont_vector, disc_vector, values, true, true,
-                        &msg);
+      try {
+        model.write_array(rng, cont_vector, disc_vector, values, true, true,
+                          &msg);
+      } catch (const std::exception& e) {
+        if (msg.str().length() > 0) {
+          logger.info(msg);
+        }
+        logger.error(e.what());
+        return error_codes::SOFTWARE;
+      }
       // This if is here to match the pre-refactor behavior
       if (msg.str().length() > 0)
         logger.info(msg);
@@ -164,7 +187,16 @@ int bfgs(Model& model, const stan::io::var_context& init,
   if (!save_iterations) {
     std::vector<double> values;
     std::stringstream msg;
-    model.write_array(rng, cont_vector, disc_vector, values, true, true, &msg);
+    try {
+      model.write_array(rng, cont_vector, disc_vector, values, true, true,
+                        &msg);
+    } catch (const std::exception& e) {
+      if (msg.str().length() > 0) {
+        logger.info(msg);
+      }
+      logger.error(e.what());
+      return error_codes::SOFTWARE;
+    }
     if (msg.str().length() > 0)
       logger.info(msg);
     values.insert(values.begin(), lp);
--- a/src/stan/services/optimize/bfgs.hpp
+++ b/src/stan/services/optimize/bfgs.hpp
@@ -172,7 +172,6 @@ int bfgs(Model& model, const stan::io::var_context& init,
         parameter_writer(values);
       }
     }
-
   } catch (const std::exception& e) {
     logger.error(e.what());
     return error_codes::SOFTWARE;
--- a/src/stan/services/optimize/bfgs.hpp
+++ b/src/stan/services/optimize/bfgs.hpp
@@ -114,74 +114,68 @@ int bfgs(Model& model, const stan::io::var_context& init,
   }
   int ret = 0;
 
-  while (ret == 0) {
-    interrupt();
-    if (refresh > 0
-        && (bfgs.iter_num() == 0 || ((bfgs.iter_num() + 1) % refresh == 0)))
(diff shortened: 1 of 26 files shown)


WardBrian:
This is failing downstream due to https://github.com/stan-dev/math/issues/1039: Some CVODES errors are still thrown as generic `runtime_error`s in https://github.com/stan-dev/math/blob/develop/stan/math/prim/err/check_flag_sundials.hpp

@syclik - the discussion from back then suggested these becoming `domain_error`s, but they didn't. Was there some reason for this, or should I go ahead and open a PR to math which does that?

src/stan/services/optimize/bfgs.hpp
```diff
@@ -119,7 +128,13 @@ int bfgs(Model& model, const stan::io::var_context& init,
           "  # evals"
           "  Notes ");
 
-    ret = bfgs.step();
+    try {
+      ret = bfgs.step();
+    } catch (const std::exception& e) {
+      logger.error(e.what());
+      return error_codes::SOFTWARE;
+    }
```
SteveBronder, line 136: Can we move this to the outer part of the while loop?

src/stan/services/optimize/lbfgs.hpp
```diff
@@ -123,7 +123,12 @@ int lbfgs(Model& model, const stan::io::var_context& init,
           "  # evals"
           "  Notes ");
 
-    ret = lbfgs.step();
+    try {
+      ret = lbfgs.step();
+    } catch (const std::exception& e) {
+      logger.error(e.what());
+      return error_codes::SOFTWARE;
+    }
```
SteveBronder, line 131: Same here. Generally for any while loops I'd like to just have the try on the outside

src/stan/services/pathfinder/multi.hpp
```diff
@@ -117,28 +117,33 @@ inline int pathfinder_lbfgs_multi(
...
+                    model, *(init[iter]), random_seed, stride_id + iter,
+                    init_radius, history_size, init_alpha, tol_obj, tol_rel_obj,
+                    tol_grad, tol_rel_grad, tol_param, num_iterations,
+                    num_elbo_draws, num_draws, save_iterations, refresh,
+                    interrupt, logger, init_writers[iter],
+                    single_path_parameter_writer[iter],
+                    single_path_diagnostic_writer[iter], calculate_lp);
+            if (unlikely(std::get<0>(pathfinder_ret) != error_codes::OK)) {
+              logger.error(std::string("Pathfinder iteration: ")
+                           + std::to_string(iter) + " failed.");
```
SteveBronder, line 135: If we are already throwing in the individual pathfinder do we need this message?

src/stan/services/pathfinder/single.hpp
```diff
@@ -820,7 +820,16 @@ inline auto pathfinder_lbfgs_single(
         logger.info(lbfgs_ss);
         lbfgs_ss.str("");
       }
-      throw e;
+      if (ReturnLpSamples) {
+        // we want to terminate multi-path pathfinder during these unrecoverable
+        // exceptions
+        throw;
```
SteveBronder, line 826: Why? One pathfinder can fail while the others succeed?

src/stan/services/pathfinder/single.hpp
```diff
@@ -820,7 +820,16 @@ inline auto pathfinder_lbfgs_single(
...
+      } else {
+        logger.error(e.what());
```
SteveBronder, line 828: `path_num` should be appended to this message?

SteveBronder (approved):
I think this is fine, one thing I would like clarified is what happens to things like `std::out_of_range` errors.

src/stan/services/optimize/lbfgs.hpp
WardBrian: Moved both

src/stan/services/pathfinder/single.hpp
WardBrian: If `ReturnLPSamples` is false, we're always in single path pathfinder, right? If so, I don't think the path number is important

src/stan/services/pathfinder/multi.hpp
WardBrian: This can still happen if e.g. a path fails to initialize. That isn't fatal to the entire algorithm, so it will trigger this message, which I think is still valuable.

WardBrian:
`std::out_of_range` would be considered fatal by this and lead to the algorithms terminating. Is there something else you're looking for clarification on about them?

src/stan/services/pathfinder/single.hpp
SteveBronder: Whoops! Wrong order that is correct

src/stan/services/pathfinder/single.hpp
SteveBronder: > If one path does something unrecoverable like indexes into a vector out of bounds, I think the right thing to do is stop and say "something is clearly wrong with your model".

Are only unrecoverable errors able to be thrown here? Like if the call to log_prob throws it could be from something recoverable like one of the ode solvers failing for a given set of parameters

WardBrian: `pathfinder_impl` catches `std::domain_error`s, which is what the ODE solvers throw (and is generally what the math library considers 'recoverable')

WardBrian:
Those have been considered unrecoverable (at least during initialization) for a while. `invalid_argument` is thrown for things like `check_square`, which won't change if you just 'try again'.

I think if there are usages in Math we want to be recoverable, we should change those to match the others

SteveBronder:
Okay looking at this again I think this is fine to merge 👍

PR #3261, opened 2024-01-25
Add option to output Hessian, disable LP calculations in Laplace

### LINKED ISSUE #3215: Allow drawing from the Laplace approximation and Pathfinder without evaluating log_p
#### Summary:

Currently, when drawing from the Laplace approximation, log_p and log_q are always evaluated for each draw. log_p and log_q are useful for diagnosing the accuracy of the Laplace approximation, but there are cases where we trust that the normal approximation at the mode is sufficient and then we would not need log_p and log_q, and evaluation of log_p for all draws usually takes much more time than finding the mode, computing hessian, and drawing from the normal take. So it would be useful to make it optional and output just 0 or NA to the csv when the user so requests.

The relevant part in Stan services is 
https://github.com/stan-dev/stan/blob/3c2b7bad93e37e787fd456e712996b19ef84de1f/src/stan/services/optimize/laplace_sample.hpp#L110

Naturally the change would need to make also in stan/arguments, cmdstan, and other interfaces, too

#### Current Version:
v2.32.2

### LINKED ISSUE #3260: Add an argument to write Hessian from Laplace algorithm to a file
Laplace algorithm computes Hessian at the mode, but it is not available for the user. The Hessian would be useful for 
- computing the normalization term of the normal approximation, which would help to estimate marginal likelihood (we have a use case)
- computing normal approximation based sd and SE without drawing from the normal distribution, which would be useful for very big data and close to normal posterior
- for algorithm development for algorithms that start with mode and Hessian, like CCD

WardBrian:
#### Submission Checklist

#### Summary
Closes #3215 
Closes #3260 
#### Intended Effect
This adds two new options to the laplace approximation service function. The first is a boolean `calculate_lp`, similar to #3249 

The second is a structured writer which will save the result of the evaluation at the mode, namely it will produce a JSON file like
```
{
  "lp_mode" : 0,
  "gradient" : [ 0, 0 ],
  "Hessian" : [ [ -2.77778, 2.22222 ], [ 2.22222, -2.77778 ] ]
}
```

#### How to Verify

Tests added

#### Side Effects

I left the existing function for backwards compatibility. The above arguments default to `true` and a no-op writer, respectively.

#### Documentation

Added inline documentation, online doc will come with the cmdstan PR.

#### Copyright and Licensing

Simons Foundation

### DIFF
--- a/src/stan/services/optimize/laplace_sample.hpp
+++ b/src/stan/services/optimize/laplace_sample.hpp
@@ -4,6 +4,7 @@
 #include <stan/callbacks/interrupt.hpp>
 #include <stan/callbacks/logger.hpp>
 #include <stan/callbacks/writer.hpp>
+#include <stan/callbacks/structured_writer.hpp>
 #include <stan/math/rev.hpp>
 #include <stan/services/error_codes.hpp>
 #include <stan/services/util/create_rng.hpp>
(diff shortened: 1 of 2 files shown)


SteveBronder (approved):
Good!

avehtari:
Great!

PR #3262, opened 2024-01-26
Properly allow multiple downstream jenkins runs

WardBrian:
We had an || instead of an && so this was always `true`, which led to problems when multiple Math PRs were running at once. 

Cmdstan has the same issue currently, I'll open a separate PR


serban-nicusor-toptal (approved):
Nice catch, thanks!

PR #3266, opened 2024-02-06
Improved rhat diagnostic

aleksgorica:
#### Submission Checklist

#### Summary
Changes in compute_potential_scale_reduction, based on [Vehtari](https://arxiv.org/abs/1903.08008)
 Rhat computattion moved in rhat function
Added function rank_transform
Changed tests values to match new output

#### Intended Effect
**rank_transform:** Computes normalized average ranks for draws. Transforming them to normal scores using inverse normal transformation and a fractional offset.

**rhat:** computes rhat like before, the computatoinal part is just moved in a new function.

**compute_potential_scale_reduction:** copies draws in matrix object, then computes bulk rhat and tail rhat and returns the maximum

#### How to Verify

Compare the results with Arviz rhat function
[arviz](https://python.arviz.org/en/stable/api/generated/arviz.rhat.html)

#### Copyright and Licensing

Aleks Stepančič


bob-carpenter:
Hi, @aleksgorica and welcome to the Stan project.  

In terms of process, there should be an issue specifying the feature which this PR addresses.  Please add an issue.

I would also suggest not removing the old functionality but instead just adding new functionality for the ranked version.  That way, it won't break backward compatibility when it's added.

When it's ready to review for inclusion, please ping me and I can do it.

src/stan/analyze/mcmc/compute_potential_scale_reduction.hpp
```diff
@@ -16,6 +17,99 @@
...
+    // Handle ties by averaging ranks
+    int j = i;
+    double sumRanks = 0;
+    int count = 0;
+
+    while (j < size && valueWithIndex[j].first == valueWithIndex[i].first) {
+      sumRanks += j + 1;  // Rank starts from 1
+      ++j;
+      ++count;
+    }
```
SteveBronder, line 58: Right now this while loop while always go off at least once, can you start `j` at `j = i + 1` etc to avoid this? Then that while loop only happens for cases of duplicates

src/stan/analyze/mcmc/compute_potential_scale_reduction.hpp
```diff
@@ -16,6 +17,99 @@
...
+      rankMatrix(row, col) = boost::math::quantile(dist, p);
+    }
+    i = j - 1;  // Skip over tied elements
+  }
+  return rankMatrix;
+}
+
+/**
+ * Computes square root of marginal posterior variance of the estimand by
+ * weigted average of within-chain variance W and between-chain variance B.
```
SteveBronder, line 77:
```suggestion
 * Computes square root of marginal posterior variance of the estimand by the
 * weighted average of within-chain variance W and between-chain variance B.
```

src/stan/analyze/mcmc/compute_potential_scale_reduction.hpp
```diff
@@ -16,6 +17,99 @@
...
+                                           boost::accumulators::tag::variance>>
+        acc_draw;
+    for (int n = 0; n < num_draws; ++n) {
+      acc_draw(draws(n, chain));
+    }
+    chain_mean(chain) = boost::accumulators::mean(acc_draw);
+    acc_chain_mean(chain_mean(chain));
+    chain_var(chain)
+        = boost::accumulators::variance(acc_draw) * unbiased_var_scale;
+  }
```
SteveBronder, line 104: It doesn't look like you need an online way to update and view the current mean and variance. You can calculate the mean for each chain via `draws.colwise().mean()` and `draws.array().mean()` for the overall mean. Then a loop for the chain variance calculation.

Now that I'm reading this again should that be named `acc_chain_variance because of line 105? I've never used boost accumulators before so not sure if I'm following all the way here

SteveBronder (commented):
Just leaving comments for now. Not sure if we want to modify this code directly or have a new function

src/stan/analyze/mcmc/compute_potential_scale_reduction.hpp
```diff
@@ -16,6 +17,150 @@
...
+/**
+ * Computes normalized average ranks for draws. Transforming them to normal
+ * scores using inverse normal transformation and a fractional offset. Based on
+ * paper https://arxiv.org/abs/1903.08008
+ * @param draws stores chains in columns
+ * @return normal scores for average ranks of draws
+ *
+ */
+
+Eigen::MatrixXd rank_transform(const Eigen::MatrixXd& draws) {
```
SteveBronder, line 29: For the error on jenkins. This makes it so that there are not multiple definitions for different translation units

```suggestion
inline Eigen::MatrixXd rank_transform(const Eigen::MatrixXd& draws) {
```

src/stan/analyze/mcmc/compute_potential_scale_reduction.hpp
```diff
@@ -16,6 +17,150 @@
...
+/**
+ * Computes square root of marginal posterior variance of the estimand by the
+ * weigted average of within-chain variance W and between-chain variance B.
+ *
+ * @param draws stores chains in columns
+ * @return square root of ((N-1)/N)W + B/N
+ *
+ */
+
+inline double rhat(const Eigen::MatrixXd& draws) {
```
SteveBronder, line 77: I think `chains` is the better name here since it's chains of draws

```suggestion
inline double rhat(const Eigen::MatrixXd& chains) {
```

src/stan/analyze/mcmc/compute_potential_scale_reduction.hpp
```diff
@@ -16,6 +17,150 @@
...
+  double total_mean = chain_mean.mean();
+  double var_between = num_draws
+                       * (chain_mean.array() - total_mean).square().sum()
+                       / (num_chains - 1);
+  double var_sum = 0;
+  for (Eigen::Index col = 0; col < num_chains; ++col) {
+    var_sum += (draws.col(col).array() - chain_mean(col)).square().sum()
+               / (num_draws - 1);
+  }
+  double var_within = var_sum / num_chains;
```
SteveBronder, line 92: A little more advanced but we can also do this like the below. I added comments to explain the Eigen parts but delete them if you change this.

Also in the below I assume you change `chain_mean` to be Eigen::RowVectorXd`

```suggestion
  double within_variance =
     // Divide each row by chains and get sum of squares for each chain (getting a vector back)
      ((chains.array().rowwise() - chain_means.array())
           .square()
           .colwise()
           .sum() /
       // divide each sum of square by num_draws, sum the sum of squares, and divide by num chains
       (num_draws - 1.0))
          .sum() / num_chains;

```

src/stan/analyze/mcmc/compute_potential_scale_reduction.hpp
```diff
@@ -16,6 +17,150 @@
...
+    for (int row = 0; row < num_draws; ++row) {
+      matrix(row, col) = draws[col][row];
+    }
+  }
+
+  double rhat_bulk = rhat(rank_transform(matrix));
+  double rhat_tail = rhat(rank_transform(
+      (matrix.array() - math::quantile(matrix.reshaped(), 0.5)).abs()));
+
+  return std::max(rhat_bulk, rhat_tail);
```
SteveBronder, line 161: Question for @avehtari 

Do we want to just return the max or should we return a pair so the user can see the bulk and tail rhats?

src/stan/analyze/mcmc/compute_potential_scale_reduction.hpp
```diff
@@ -31,6 +176,7 @@ namespace analyze {
  * @param sizes stores sizes of chains
  * @return potential scale reduction for the specified parameter
  */
+
 inline double compute_potential_scale_reduction(
```
SteveBronder, line 180: @bob-carpenter Should we be changing the old code at all? Do we want to always use the new rhat?

src/stan/analyze/mcmc/compute_potential_scale_reduction.hpp
```diff
@@ -124,6 +275,40 @@ inline double compute_potential_scale_reduction(
...
+ * Current implementation assumes draws are stored in contiguous
+ * blocks of memory.  Chains are trimmed from the back to match the
+ * length of the shortest chain.
+ *
+ * @param draws stores pointers to arrays of chains
+ * @param sizes stores sizes of chains
+ * @return potential scale reduction for the specified parameter
+ */
+inline double compute_split_potential_scale_reduction_rank(
+    std::vector<const double*> draws, std::vector<size_t> sizes) {
```
SteveBronder, line 297: We want these arguments to come in as constant references. As written this will make a hard copy of the input vectors when you call this function. Making the arguments references (`&`) means the function will just use the already existing object without making a copy and `const` means the arguments will be constant in the function (i.e. we will not modify them)

```suggestion
inline double compute_split_potential_scale_reduction_rank(
    const std::vector<const double*>& draws, const std::vector<size_t>& sizes) {
```

We want containers (like `std::vector` or `Eigen::MatrixXd` or `std::map`) to be passed by constant reference. Small types such as `double`, `int`, `std::size_t` etc. can be passed by value as making copies of them is trivial. Happy to explain this more if you like but don't want to overload you with info. A nice place to read about things like this is Scott Meyers "[Effective Modern C++](https://www.amazon.com/Effective-Modern-Specific-Ways-Improve/dp/1491903996)" which if you google should be easy to find a free copy of online. 

This comment applies to all the function signatures you added here

src/stan/analyze/mcmc/compute_potential_scale_reduction.hpp
```diff
@@ -124,6 +275,40 @@ inline double compute_potential_scale_reduction(
...
+ * https://arxiv.org/abs/1903.08008
+ *
+ * When the number of total draws N is odd, the (N+1)/2th draw is ignored.
+ *
+ * See more details in Stan reference manual section "Potential
+ * Scale Reduction". http://mc-stan.org/users/documentation
+ *
+ * Current implementation assumes draws are stored in contiguous
+ * blocks of memory.  Chains are trimmed from the back to match the
+ * length of the shortest chain.
```
SteveBronder, line 290: You use [`split_chains`](https://github.com/stan-dev/stan/blob/51ad447d6965a59da24c6365c078450bdaebea9b/src/stan/analyze/mcmc/split_chains.hpp#L25) which also assumes each chain is the same length

src/stan/analyze/mcmc/compute_potential_scale_reduction.hpp
```diff
@@ -124,6 +275,40 @@ inline double compute_potential_scale_reduction(
...
+  int num_chains = sizes.size();
+  size_t num_draws = sizes[0];
+  for (int chain = 1; chain < num_chains; ++chain) {
+    num_draws = std::min(num_draws, sizes[chain]);
+  }
+
+  std::vector<const double*> split_draws = split_chains(draws, sizes);
+
+  double half = num_draws / 2.0;
+  std::vector<size_t> half_sizes(2 * num_chains, std::floor(half));
```
SteveBronder, line 307: Just to make it more clear you are using floating point division and then taking the floor to get the index
```suggestion
  std::size_thalf = std::floor(num_draws / 2.0);
  std::vector<size_t> half_sizes(2 * num_chains, half);
```

SteveBronder (changes requested):
I think this is good! The suggested changes mostly cover some C++ things, some Qs about the codes math, and handling some edge cases.

As a first PR this is very good so far!

src/stan/analyze/mcmc/compute_potential_scale_reduction.hpp
avehtari: It is useful for the user to know both. A diagnostic message could be simplified by reporting if the max of these is too low, but otherwise I would prefer that both would be available for the user. Making them both available does change the io via csv and changing csv structures need to be considered carefully

SteveBronder: Okay then @aleksgorica can you have this return back an `std::pair`?

src/stan/analyze/mcmc/compute_potential_scale_reduction.hpp
```diff
@@ -16,6 +17,150 @@
...
+    }
+  }
+
+  if (are_all_const) {
+    // If all chains are constant then return NaN
+    // if they all equal the same constant value
+    if (init_draw.isApproxToConstant(init_draw(0))) {
+      return std::numeric_limits<double>::quiet_NaN();
+    }
+  }
```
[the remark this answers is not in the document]
aleksgorica: Yes, you are correct. The current implementation fails if different chains are constant. For example, `C1: [1, 1, 1]; C2: [2, 2, 2]` would have a within-variance of 0, and the `rhat` function would return `inf` due to division by zero. I think the best way to correct this is to check if there exists a non-constant chain.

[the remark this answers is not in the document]
bob-carpenter: Doing something intelligent around constant chains would be a big improvement on our current NaN behavior.  But I'm not sure what that is as there's not a number that makes sense as the ESS.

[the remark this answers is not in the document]
avehtari: If all chains have the same constant value, we can't make the difference between all chains being stuck or variable actually being constant (e.g. diagonal of correlation matrix) as Stan doesn't tag the variables. In that case diagnostics in R return NA. If the chains have different constant values, then the variable can't be a true constant, and Rhat Inf is fine.

src/stan/analyze/mcmc/compute_potential_scale_reduction.hpp
```diff
@@ -71,34 +159,104 @@ inline double compute_potential_scale_reduction(
...
+      if (!std::isfinite(draws(n))) {
+        return std::numeric_limits<double>::quiet_NaN();
+      }
+      draws_matrix(n, chain) = draws(n);
+    }
+
+    init_draw(chain) = draws(0);
+    if (!draws.isApproxToConstant(draws(0))) {
+      are_all_const |= true;
+    }
```
SteveBronder, line 225:
```suggestion
are_all_const |= !draws.isApproxToConstant(draws(0))
```

src/test/unit/analyze/mcmc/compute_potential_scale_reduction_test.cpp
```diff
@@ -54,11 +53,75 @@ TEST_F(ComputeRhat, compute_potential_scale_reduction) {
...
+        << ", parameter: " << chains.param_name(index);
+    ASSERT_NEAR(expected_tail_rhat, computed_tail_rhat, 1e-4)
+        << "Tail Rhat mismatch for index: " << index
+        << ", parameter: " << chains.param_name(index);
+    // ASSERT_NEAR(
+    //     rhat(index - 4),
+    //     stan::analyze::compute_potential_scale_reduction_rank(draws, sizes),
+    //     1e-4)
+    //     << "rhat for index: " << index
+    //     << ", parameter: " << chains.param_name(index);
```
SteveBronder, line 122: Why is this commented out?

SteveBronder (changes requested):
Looking good! Very small modifications then I think this is good to go!

Also remove or re-add any commented out code. There's a lot in the tests

src/test/unit/analyze/mcmc/compute_potential_scale_reduction_test.cpp
aleksgorica: I've replaced the `rhat` check, which used `max(rhat_bulk, rhat_tail)`, with a check that both `rhat_bulk` and `rhat_tail` are correct, which is why I've removed or commented out the `rhat` tests in the rank version.

SteveBronder:
@aleksgorica this code all looks good! The only thing to fix now is the API. We want to keep all the old code and functions the same for backwards compatibility and have all of this in new functions. For instance `compute_potential_scale_reduction()` should have the same results as previously. 

So the new code you have in `compute_potential_scale_reduction_rank` etc. is good and you can revert your changes to `compute_potential_scale_reduction`. New code in the future can use your `compute_potential_scale_reduction_rank` for the new estimate. 

Once that is changed then I think you are good to merge!

aleksgorica:
> @aleksgorica this code all looks good! The only thing to fix now is the API. We want to keep all the old code and functions the same for backwards compatibility and have all of this in new functions. For instance `compute_potential_scale_reduction()` should have the same results as previously.
> 
> So the new code you have in `compute_potential_scale_reduction_rank` etc. is good and you can revert your changes to `compute_potential_scale_reduction`. New code in the future can use your `compute_potential_scale_reduction_rank` for the new estimate.
> 
> Once that is changed then I think you are good to merge!

Okay, I hope I have understood correctly, I have just reverted the changes in the original non-rank functions to the previous code. However, I would like to know why that is necessary since I believe the new code yields the same results as the previous code according to the tests, and it is also a bit better written.

SteveBronder:
@bob-carpenter I rewrote this to dispatch to the old cold from the previous API

@aleksgorica there's one signature missing for `split_potential_scale_reduction_rank` so it matches the API of `split_potential_scale_reduction`. See below. Can you add that signature and have the current `split_potential_scale_reduction` signature call your new `split_potential_scale_reduction_rank` returning bulk rhat?

https://github.com/stan-dev/stan/blob/develop/src/stan/mcmc/chains.hpp#L219

SteveBronder:
Actually I reverted my old PR. Changing all the old functions to use the new one causes cmdstan tests to fail. 

Let's have the `_rank` versions like you made here. We just need one last signature for

```c++
double split_potential_scale_reduction(
      const Eigen::Matrix<Eigen::VectorXd, Dynamic, 1>& samples)
```

Then we are good! After this is in we can then change cmdstan etc. to use the new ranked version. I think this is better because we need to change the output anyway to report back both the bulk and tail rhat

aleksgorica:
> Actually I reverted my old PR. Changing all the old functions to use the new one causes cmdstan tests to fail.
> 
> Let's have the `_rank` versions like you made here. We just need one last signature for
> 
> ```c++
> double split_potential_scale_reduction(
>       const Eigen::Matrix<Eigen::VectorXd, Dynamic, 1>& samples)
> ```
> 
> Then we are good! After this is in we can then change cmdstan etc. to use the new ranked version. I think this is better because we need to change the output anyway to report back both the bulk and tail rhat

Ok, but why should we even keep ```double split_potential_scale_reduction(
      const Eigen::Matrix<Eigen::VectorXd, Dynamic, 1>& samples)``` or add ```double split_potential_scale_reduction_rank(
      const Eigen::Matrix<Eigen::VectorXd, Dynamic, 1>& samples)```, when it is declared private and there are no written tests for the function and no known references? Would we break backward compatibility by removing the function?

SteveBronder:
Oh, your right I didn't see that it was a private member that is never called. Let's leave it for now but idt you need to write a rank version for that one.

SteveBronder:
@aleksgorica you should have received an email that adds you to the stan project :) clicking on the link in that email from github should then allow you to press the "merge pull request" button

aleksgorica:
Thank you all for accepting me to Stan project :). I had really great time working on the pull request. Also thank you @SteveBronder and @bob-carpenter for mentoring.

But github doesn't allow me to merge pull request.

bob-carpenter:
I just verified that Steve gave you permission to merge.  Can you verify that works on your end by merging this?

Thanks, and welcome to the Stan team!

SteveBronder:
Did you click on the link in the email that was sent to you from github? After you click that and refresh the page you should have permission to merge

aleksgorica:
Yes, I clicked multiple times, now it always redirects me to [https://github.com/stan-dev/stan](url). But it still does not allow me to merge. If I go to [https://github.com/settings/organizations](url), there is written that I am outside collaborator on 1 repository for Stan.

SteveBronder:
Sorry you should have just gotten a new email that should give you the right permissions

PR #3286, opened 2024-05-16
Add 1 to random seed 

### LINKED ISSUE #3285: Possible issue with new RNG?
#### Description:

Not sure if this is a bug or just me doing things wrong, but it looks like the new RNG doesn't interact with the Math library in the same way - the return is always constant:

```cpp
#include <iostream>
#include <stan/services/util/create_rng.hpp>
#include <stan/math.hpp>

int main() {
  using stan::math::normal_rng;
  stan::rng_t rng(0);
  std::cout << "stan::rng_t: "
            << normal_rng(5, 10, rng) << ", "
            << normal_rng(5, 10, rng) << ", "
            << normal_rng(5, 10, rng) << ", "
            << normal_rng(5, 10, rng)
            << std::endl;

  boost::ecuyer1988 rng2(0);
  std::cout << "boost::ecuyer1988: "
            << normal_rng(5, 10, rng2) << ", "
            << normal_rng(5, 10, rng2) << ", "
            << normal_rng(5, 10, rng2) << ", "
            << normal_rng(5, 10, rng2)
            << std::endl;
  return 0;
}
```

Gives:
```
stan::rng_t: 5, 5, 5, 5
boost::ecuyer1988: -4.52988, 8.12959, 2.12997, 5.78816
```

#### Current Version:
v2.35.0

SteveBronder:
#### Submission Checklist

#### Summary

Fixes #3285 by just adding 1 to the given seed

#### How to Verify

```bash
python3 ./runTests.py ./src/test/unit/services/util/create_rng_test.cpp
```

#### Copyright and Licensing

Please list the copyright holder for the work you are submitting (this will be you or your assignee, such as a university or company): Simon's Foundation

### DIFF
--- a/src/test/unit/services/util/create_rng_test.cpp
+++ b/src/test/unit/services/util/create_rng_test.cpp
@@ -28,3 +28,12 @@ TEST(rng, initialize_with_zero) {
   rng2();
   EXPECT_NE(rng1, rng2);
 }
+
+TEST(rng, run_twice_with_minus_one_seed) {
+  stan::rng_t rng1 = stan::services::util::create_rng(-1, 0);
+  EXPECT_EQ(rng1(), rng1());
+}
+TEST(rng, run_twice_with_zero_seed) {
+  stan::rng_t rng1 = stan::services::util::create_rng(0, 0);
+  EXPECT_NE(rng1(), rng1());
+}
--- a/src/stan/services/util/create_rng.hpp
+++ b/src/stan/services/util/create_rng.hpp
@@ -26,7 +26,7 @@ namespace util {
  * @return an stan::rng_t instance
  */
 inline rng_t create_rng(unsigned int seed, unsigned int chain) {
-  rng_t rng(seed + chain);
+  rng_t rng(seed + chain + 1);
   return rng;
 }
 


src/test/unit/services/util/create_rng_test.cpp
```diff
@@ -28,3 +28,12 @@ TEST(rng, initialize_with_zero) {
   rng2();
   EXPECT_NE(rng1, rng2);
 }
+
+TEST(rng, run_twice_with_minus_one_seed) {
+  stan::rng_t rng1 = stan::services::util::create_rng(-1, 0);
```
WardBrian, line 33: We shouldn't pass -1 to something expecting unsigned ints IMO (even in a test)

SteveBronder: Should we have a check for this to fail then? I just want to have a test to show why we added a +1 to the rng

WardBrian: If you want to rely on the wrapping behavior, I think explicitly passing UINT_MAX is better

However, this is part of why I don't know if I like this solution. If a user passes UINT_MAX, they are effectively passing 0. If we don't want that to be possible, the right thing to do seems to be actually prevent it somewhere (e.g. cmdstan's argument parser could do that)

WardBrian:
I've read the boost docs and couple papers on MIXMAX. The actual RNG state is 128 bits, and it does require at least 1 nonzero bit in the seed to function. 

Luckily for us, the boost seed taken in is a uint64_t, and our seeds are only uint32s, so we can create the seed like
```c++
uint64_t seed_boost = static_cast<uint64_t>(seed) + static_cast<uint64_t>(chain) + 1;
```
and guarantee that there is always a nonzero bit, without the issue of wrapping around leading to another zero appearing at the top of the range.

This will 1) require updating all the tests which use fixed seeds and 2) be rather difficult if not impossible to test on its own, but I think both of those are fine

nhuurre:
> Luckily for us, the boost seed taken in is a uint64_t, and our seeds are only uint32s, so we can create the seed like

`boost::random::mixmax` also has a constructor that takes four uint32s. Why not use that?
```cpp
inline rng_t create_rng(unsigned int seed, unsigned int chain) {
  rng_t rng(1, 1, seed, chain);
  return rng;
}
```
And if you're going to create a 64-bit seed, I think concatenating the 32-bit parts is better than summing
```cpp
uint64_t seed_boost = (static_cast<uint64_t>(seed) << 32) + static_cast<uint64_t>(chain) + 1;
```

WardBrian:
@nhuurre I agree after reading more. 

A feature of mixmax which was described by the boost authors is that `seed` and `seed+1` are independent in the case of calling the 4-argument version with `(0,0, seed<<32, seed)`. 

If that remains true for values other than 0 for the first two arguments (e.g.  `(0,1, seed<<32, seed)`), then I think we should just call the 4 argument constructor with `(0, 1, user_seed, chain_id)`. 
I'm inquiring now. If there is some magic in the first two arguments being zero, we can do the shift-by-31 and add 1 you describe.

SteveBronder:
Going to close this as it sounds like there are better answers

PR #3295, opened 2024-06-27
Remove support for ~/.config makefile inclusion

WardBrian:
#### Submission Checklist

#### Summary

Counterpart to https://github.com/stan-dev/math/pull/3090

#### Copyright and Licensing

### DIFF
--- a/makefile
+++ b/makefile
@@ -2,16 +2,12 @@
 # Stan
 # -----------------
 #
-# To customize your build, set make variables in either:
-#    ~/.config/stan/make.local
-#    make/local
-# Variables in make/local is loaded after ~/.config/stan/make.local
+# To customize your build, set make variables in make/local
 
 
 ## 'help' is the default make target.
 help:
 
--include $(HOME)/.config/stan/make.local  # user-defined variables
 -include make/local                       # user-defined variables
 
 MATH ?= lib/stan_math/


wds15:
How come you want to remove this functionality? I have used this a few times to set things from my home dir as I am not allowed on my cluster to edit the cmdstan install dir. So how can users set stan makefile matters in the future from a user writable directory (meaning their home)? I do have to use the cmdstan installed into the system for compliance reasons.

WardBrian:
Hi @wds15 -

Some explanation was given in https://github.com/stan-dev/math/issues/3088, but in short there are 3 main reasons:

- This 'feature' is almost completely undocumented. There were ~two mentions of it in the math developer documentation, none in the users guides.
- It [causes issues on Windows machines](https://discourse.mc-stan.org/t/cmdstan-path-has-not-been-set-yet-windows-cmdstanr-0-8/35398/7?u=wardbrian)
- Global config is in general hard to debug and having only one place that users can include special variables makes it easier to find config issues when they arise


If you are in a situation, as you describe, where the HOME folder behavior is desirable, you should ask the system administrators to add a `make/local` which defers to user's folders. For example, just adding the line this PR removes:

```make
-include $(HOME)/.config/stan/make.local 
```

to the global installation's `make/local` would opt back in to this behavior while not requiring the project as a whole to support it.

wds15:
I see. Thanks for the explanation and the workaround to get the old behaviour.

Could the workaround be added to the example make/local which is distributed possibly? If not enabled, then at least in a commented out variant?

I consider it standard behaviour that programs respect config files residing it in users home directories...but you are right in that this specific option was not documented and I took it from the source.

mitzimorris (approved):
lgtm!

PR #3302, opened 2024-07-23
Bugfix/3301 stan csv reader skip warmup

mitzimorris:
#### Submission Checklist

- [x] Run unit tests: `./runTests.py src/test/unit`
- [x] Run cpplint: `make cpplint`
- [x] Declare copyright holder and open-source license: see below

#### Summary

Added logic to skip warmup draws, if any present, more checks on consistency between config and output.

#### Intended Effect

see #3301 

#### How to Verify

Unit tests

#### Side Effects

N/A

#### Documentation

N/A

#### Copyright and Licensing

Please list the copyright holder for the work you are submitting (this will be you or your assignee, such as a university or company):
Columbia University

By submitting this pull request, the copyright holder is agreeing to license the submitted work under the following licenses:
- Code: BSD 3-clause (https://opensource.org/licenses/BSD-3-Clause)
- Documentation: CC-BY 4.0 (https://creativecommons.org/licenses/by/4.0/)

PR #3305, opened 2024-07-31
Feature/3299 improve ess mcse

mitzimorris:
#### Submission Checklist

#### Summary

Add split-rank-folded ESS .

#### Intended Effect

Expose new split-Rhat and split-ESS for CmdStan

#### How to Verify

unit tests

#### Side Effects

N/A

#### Copyright and Licensing


mitzimorris:
@SteveBronder - I would really appreciate your eyeballs on the ESS calculations - a preliminary review?

@WardBrian - does this API look OK?

also @avehtari - I implemented tail ESS - current logic is that if tail ESS returns NaN, set if to bulk ESS, as this seems to be what the posterior package does.  is this correct?

the current set of unit tests test split R-hat and split-ESS against what's in posterior for a run of 2 chains on the eight_schools model.   suggestions for further tests welcome.

WardBrian:
I think the new APIs look nice. I still have some concerns about the implementations, especially things like sample means vs population means, and how difficult it is to tell them apart just reading the code. For example, split_rank_normalized_rhat.hpp has 

`chain_mean(i) = chain_col.mean();`

and 

`double mean_var = math::mean(chain_var);`

just a few lines apart -- are these the same mean as each other? I think in this case they are, but it's worth making obvious. Same for variances. A few small named functions for these would go a long way, I think

avehtari:
> also @avehtari - I implemented tail ESS - current logic is that if tail ESS returns NaN, set if to bulk ESS, as this seems to be what the posterior package does. is this correct?

In posterior package if tail ESS is NaN, then it's NaN. Can you provide a link to the posterior package code which made you think otherwise?

mitzimorris:
> In posterior package if tail ESS is NaN, then it's NaN. Can you provide a link to the posterior package code which made you think otherwise?

you're correct - this is what posterior will do.  I'm puzzled as to why the implementation in this PR fails on a test of the eight schools model and data, run with 2 chains, 1000 iters each but thinned by a factor of 2.   the parameter `tau` has a low effective bulk ESS, and CmdStanR reports bulk and tail having the same value.   for the implementation in this PR,  the tail 05 ESS is NaN, and the tail 95 ESS is 486.   In CmdStanR,  the tail ESS is 71 - same as bulk ESS.  more testing needed.

update:  caused by differences in computing quantiles between Eigen and R.

mitzimorris:
ready for first round of review.

mitzimorris:
@SteveBronder and @WardBrian - ready for first round of review

mitzimorris:
todo note: document the functions in `chainset.hpp`.

mitzimorris:
this upstream CmdStan test failed:

> test/interface/fixed_param_sampler_test --gtest_output="xml:test/interface/fixed_param_sampler_test.xml" failed

mitzimorris:
@WardBrian or @dylex - upstream tests are failing - is there a Jenkins problem?  workspace?

WardBrian:
I’m not seeing anything abnormal - latest run failed due to excess compiler warnings

mitzimorris:
> I’m not seeing anything abnormal - latest run failed due to excess compiler warnings

this is what I'm seeing:

```
continuous-integration/jenkins/pr-merge failed:  https://jenkins.flatironinstitute.org/blue/organizations/jenkins/Stan%2FStan/detail/PR-3305/27/pipeline
```

WardBrian:
Yes, because this PR is causing 4 new compiler warnings which is crossing the threshold of what we allow in CI:

https://jenkins.flatironinstitute.org/job/Stan/job/Stan/job/PR-3305/27/pipeline_clang/folder.1689781376/

mitzimorris:
thanks - fixed!  (this time for sure!)

lib/stan_math
WardBrian, line 1: Shouldn't be touching the math submodule, needs to be resolved

src/stan/analyze/mcmc/check_chains.hpp
```diff
@@ -0,0 +1,43 @@
...
+#include <utility>
+#include <vector>
+
+namespace stan {
+namespace analyze {
+
+/**
+ * Checks that values across all matrix columns finite and non-identical.
+ *
+ * @param chains matrix of draws, one column per chain
```
WardBrian, line 17: Just to clarify, this holds the draws for one single variable from across all the chains?

src/stan/analyze/mcmc/compute_potential_scale_reduction.hpp
```diff
@@ -17,156 +17,6 @@
...
-}
-
-/**
- * Computes square root of marginal posterior variance of the estimand by the
- * weigted average of within-chain variance W and between-chain variance B.
- *
- * @param chains stores chains in columns
- * @return square root of ((N-1)/N)W + B/N
- */
-inline double rhat(const Eigen::MatrixXd& chains) {
```
WardBrian, line 73: I am a little surprised to see all these deletions -- is the idea not to leave existing code mostly untouched? (To be clear, I'm fine if that _isn't_ the idea, but then there are probably a lot more deletions we should be considering in this refactor)

src/stan/analyze/mcmc/split_rank_normalized_ess.hpp
```diff
@@ -0,0 +1,148 @@
...
+ *
+ * See more details in Stan reference manual section "Potential
+ * Scale Reduction". http://mc-stan.org/users/documentation
+
+ * @param chains matrix of per-chain draws, num_iters X chain
+ * @return potential scale reduction
+ */
+inline std::pair<double, double> split_rank_normalized_ess(
+    const Eigen::MatrixXd& chains) {
+  Eigen::MatrixXd split_draws_matrix = split_chains(chains);
```
WardBrian, line 118: This seems to differ from the posterior function, which calculates the quantiles and _then_ splits the chains -- just to check, these operations are interchangeable, right?

src/test/unit/analyze/mcmc/compute_potential_scale_reduction_test.cpp
WardBrian, line 1: Did the deleted tests from here end up anywhere?

src/stan/mcmc/chainset.hpp
```diff
@@ -0,0 +1,557 @@
...
+namespace mcmc {
+using Eigen::Dynamic;
+
+/**
+ * An <code>mcmc::chainset</code> object manages the post-warmup draws
+ * across a set of MCMC chains, which all have the same number or samples.
+ *
+ * <p><b>Storage Order</b>: Storage is column/last-index major.
+ */
+template <typename Unused = void*>
```
WardBrian: Why?

src/stan/mcmc/chainset.hpp
```diff
@@ -0,0 +1,557 @@
...
+    if (stan_csv.samples.rows() != thinned_samples(stan_csv))
+      return false;
+    return true;
+  }
+
+  /**
+   * Process first chain: record header, thinned samples,
+   * add samples to vector chains.
+   */
+  void initFromStanCsv(const stan::io::stan_csv& stan_csv) {
```
WardBrian: Is there a reason to camelCase this function while the rest are snake_case?

src/stan/io/stan_csv_reader.hpp
WardBrian, line 1: These are the changes from #3302, right? Is there a reason why that was causing CI to fail in cmdstan and this PR is not?

WardBrian (commented):
Thanks for tackling this

src/stan/analyze/mcmc/check_chains.hpp
mitzimorris: yes.

src/stan/analyze/mcmc/compute_potential_scale_reduction.hpp
mitzimorris: these haven't been deleted, they've just been moved to file `split_rank_normalized_rhat`.
this was done for 2 reasons: 
1. so that `rank_transform` can be used by split_rank_normalized_ess as well
2. the Stan modules sometimes have 1 function per file and sometimes are a whole set of functions in a file. when the functions are as complex as computing Rhat and ESS, the fewer functions per file, the easier it is for me to deal with them.  I realize this is solely personal taste.

src/stan/analyze/mcmc/split_rank_normalized_ess.hpp
mitzimorris: this follows what the split_rank_normalized_rhat does.
but yes, interchangeable because the call on line 130 is operating on the data vector; arrangement of rows, columns doesn't matter.

src/stan/io/stan_csv_reader.hpp
mitzimorris: This was failing in cmdstan because this PR added a new CSV test file, which uncovered a problem with the metadata handling - the recoding of boolean config flags for `true` and `false` triggered the error.

src/test/unit/analyze/mcmc/compute_potential_scale_reduction_test.cpp
mitzimorris: yes, they're all in `chainset_test.hpp`; using the eight_schools and bernoulli models.

src/stan/analyze/mcmc/compute_potential_scale_reduction.hpp
WardBrian: That sounds fine, but I think it would make sense to break up even further if so. For backwards compatibility, you can just `#include` the new files into this and existing includes will continue working

src/stan/mcmc/chainset.hpp
mitzimorris: you caught me using ChatGPT to refactor the code.  it likes camel case.  will need to update instructions.  ;-)
and will change this to snake_case

src/stan/mcmc/chainset.hpp
mitzimorris: that's copied from `chains.hpp` - will remove.

lib/stan_math
mitzimorris: this was totally a mistake.  what is the recommended way to undo this?

WardBrian: I think you can reset it with git checkout as if it's a normal file (https://stackoverflow.com/questions/215718/how-can-i-reset-or-revert-a-file-to-a-specific-revision) but I'm honestly unsure

src/stan/analyze/mcmc/compute_potential_scale_reduction.hpp
mitzimorris: new files included.

lib/stan_math
WardBrian: You can possibly also just merge develop into this branch and discard the conflict that arises in the submodule in favor of the one in develop?

mitzimorris: yes, that worked.

src/stan/mcmc/chainset.hpp
```diff
@@ -0,0 +1,556 @@
...
+#include <vector>
+#include <cstdlib>
+
+namespace stan {
+namespace mcmc {
+using Eigen::Dynamic;
+
+/**
+ * An <code>mcmc::chainset</code> object manages the post-warmup draws
+ * across a set of MCMC chains, which all have the same number or samples.
```
SteveBronder:
```suggestion
 * across a set of MCMC chains, which all have the same number of samples.
```

src/stan/mcmc/chainset.hpp
```diff
@@ -0,0 +1,556 @@
...
+ *
+ * <p><b>Storage Order</b>: Storage is column/last-index major.
```
SteveBronder: Do you mean something like the below?

```suggestion
 @note Samples are stored in column major order where each column is a sample of a returned value from the chain
```

src/stan/mcmc/chainset.hpp
```diff
@@ -0,0 +1,556 @@
...
+
+  static bool is_valid(const stan::io::stan_csv& stan_csv) {
+    if (stan_csv.header.empty())
+      return false;
+    if (stan_csv.samples.size() == 0)
+      return false;
+    if (stan_csv.samples.rows() != thinned_samples(stan_csv))
+      return false;
+    return true;
+  }
```
SteveBronder: If these are static can they exist outside of the class? I think these are fine/good to expose as utility functions

src/stan/mcmc/chainset.hpp
```diff
@@ -0,0 +1,556 @@
...
+        throw std::invalid_argument(ss.str());
+      }
+    }
+    chains_.push_back(stan_csv.samples);
+  }
+
+ public:
+  explicit chainset(const stan::io::stan_csv& stan_csv) {
+    init_from_stan_csv(stan_csv);
+  }
```
SteveBronder, line 114: I would just use an Member initializer list. I think it's reasonable to assume the user will bring in a valid `stan_csv` so idt this function should have to check that

```suggestion
  explicit chainset(const stan::io::stan_csv& stan_csv) : 
    param_names_(stan_csv.header),
    num_samples_(thinned_samples(stan_csv)),
    chains_(stan_csv.samples) {}
```

src/stan/mcmc/chainset.hpp
```diff
@@ -0,0 +1,556 @@
...
+
+  explicit chainset(const std::vector<stan::io::stan_csv>& stan_csv) {
+    if (stan_csv.empty())
+      return;
+    init_from_stan_csv(stan_csv[0]);
+    for (size_t i = 1; i < stan_csv.size(); ++i) {
+      add(stan_csv[i]);
+    }
+  }
```
SteveBronder, line 123: Can we write a `thinned_samples` function with a signature `thinned_samples(const std::vector<stan::io::stan_csv>&)`? Also add a function like the below so that we can use a member initialization list

```c++
std::vector<Eigen::MatrixXd> extract_samples(const std::vector<stan::io::stan_csv>&)`
```

```suggestion
  explicit chainset(const std::vector<stan::io::stan_csv>& stan_csvs) :
    param_names_(stan_csvs[0].header),
    num_samples_(thinned_samples(stan_csvs)),
    chains_(extract_samples(stan_csvs)) {}
```

src/stan/mcmc/chainset.hpp
```diff
@@ -0,0 +1,556 @@
...
+   * Report number of parameters per chain.
+   * @return size of parameter names vector.
+   */
+  inline int num_params() const { return param_names_.size(); }
+
+  /**
+   * Report number of samples (draws) per chain.
+   * @return rows per chain
+   */
+  inline int num_samples() const { return num_samples_; }
```
SteveBronder, line 141: `std::vector` returns `size_t` which is a `unsigned int`. I like using `Eigen::Index` instead of `int` since it's a `long int` and it's a reasonably safe int to convert to

```suggestion
  inline int num_chains() const { return chains_.size(); }

  /**
   * Report number of parameters per chain.
   * @return size of parameter names vector.
   */
  inline int num_params() const { return param_names_.size(); }

  /**
   * Report number of samples (draws) per chain.
   * @return rows per chain
   */
  inline int num_samples() const { return num_samples_; }
```

src/stan/mcmc/chainset.hpp
```diff
@@ -0,0 +1,556 @@
...
+ * An <code>mcmc::chainset</code> object manages the post-warmup draws
+ * across a set of MCMC chains, which all have the same number or samples.
+ *
+ * <p><b>Storage Order</b>: Storage is column/last-index major.
+ */
+class chainset {
+ private:
+  size_t num_samples_;
+  std::vector<std::string> param_names_;
+  std::vector<Eigen::MatrixXd> chains_;
```
SteveBronder, line 64: Instead of an std::vector of chains with inner matrices of size `(Samples, Parameters)` can we just have one large matrix of size `(Samples * Chains, Parameters)`. I think this will be nicer for extracting information for each parameter

src/stan/mcmc/chainset.hpp
```diff
@@ -0,0 +1,556 @@
...
+  /**
+   * Compute sample variance for specified parameter across all chains.
+   * 1 / (N - 1) * sum((theta_n - mean(theta))^2)
+   *
+   * @param name parameter name
+   * @return sample variance
+   */
+  double variance(const std::string& name) const {
+    return variance(index(name));
+  }
```
SteveBronder, line 254: Where will these functions this used? I know the other class had these, but where they actually used? (I genuinely don't know)

SteveBronder (changes requested):
Few comments about `chainset`

src/stan/mcmc/chainset.hpp
mitzimorris: these are used by downstream analysis packages - afaik, only the CmdStan `stansummary` utility.

at the point where CmdStan became its own module, these stayed as part of Stan.  you'd have to ask the devs who were there - I don't know.

but I will edit the PR summary and description to better document all the changes in this PR.

src/stan/mcmc/chainset.hpp
mitzimorris: this data structure is convenient for computing Rhat, because you need both w/in chain and cross-chain variances.

src/stan/mcmc/chainset.hpp
mitzimorris: done.

src/stan/mcmc/chainset.hpp
mitzimorris: I think it's safer to check - class `stan::io::stan_csv_reader` doesn't do these checks and they're not particularly expensive.

src/stan/mcmc/chainset.hpp
mitzimorris: fixed in PR #3310

src/stan/mcmc/chainset.hpp
mitzimorris: fixed in PR https://github.com/stan-dev/stan/pull/3310

src/stan/mcmc/chainset.hpp
mitzimorris: fixed in PR https://github.com/stan-dev/stan/pull/3310

src/stan/mcmc/chainset.hpp
mitzimorris: this isn't a very expensive check - but maybe it should throw an error instead of returning true/false.

mitzimorris:
closing this PR - see comment https://github.com/stan-dev/stan/pull/3310#issuecomment-2389091218

Issue #3301, opened 2024-07-22
stan_csv_reader should skip warmup draws


mitzimorris:
#### Summary:

As currently written,  when `stan::io::stan_csv_reader` encounters a file with saved warmup draws,
the parser will return both warmup and sample draws, but no adaptation information.

Instead, the reader should check the parsed metatdata to determine whether or not there are saved warmup draws.
If there are, these can be discarded, as the consumer of the resulting `stan_csv` object - `stan::mcmc::chains` doesn't need them.

#### Description:
The parser expects to find the following element in the following order:

- metadata
- header
- adaptation
- samples, followed by timing.

Function `read_adaptation` is called after the header line has been consumed.  If the following line is not a comment block, it returns.  An error is written to the supplied stream `out` and the parser then calls `read_samples`.   Read samples then reads both the warmup and post-warmup draws.

The returned object contains warmup and post-warmup draws, but no information about step-size or metric.

#### Reproducible Steps:

1. Get outputs from 2 runs of a Stan model, w/ and w/out `save_wamup=true`.
2. Parse both with `stan_csv_reader`, examine the returned `stan_csv` object, e.g.:

```
TEST_F(McmcChains, stan_csv) {
  std::cout << "bernoulli_default " << bernoulli_default.metadata.num_samples;
  std::cout << " sample size " << bernoulli_default.samples.rows();
  std::cout << " step size " << bernoulli_default.adaptation.step_size << std::endl;
  std::cout << " bernoulli_warmup " << bernoulli_warmup.metadata.num_samples;
  std::cout << " sample size " << bernoulli_warmup.samples.rows();
  std::cout << " step size " << bernoulli_warmup.adaptation.step_size << std::endl;
}
```

#### Current Output:
bernoulli_default 1000 sample size 1000 step size 0.932037
bernoulli_warmup 1000 sample size 2000 step size 0

#### Expected Output:
Both should have sample size 1000, and step size close to 0.93.

#### Additional Information:
Provide any additional information here.

#### Current Version:
v2.35.0

mitzimorris:
additional problems found with stan_csv_reader.hpp - extend to properly parse ADVI samples - per CmdStan https://github.com/stan-dev/cmdstan/pull/972

PR #3310, opened 2024-09-25
Feature/3299 diagnostics chainset

mitzimorris:
#### Submission Checklist

#### Summary

This PR uses the improved R-hat diagnostics (https://github.com/stan-dev/stan/pull/3266) and extends rank-normalization to the computation of ESS-bulk and ESS-tail.   These diagnostics are computed on samples from one or more chains, which entails parsing the sample out of a stan-csv file and assembling it into an Eigen MatrixXd object.

The work done for this PR is:

- refactoring of the functionality in `stan/analyze/mcmc` so that both ESS and Rhat can use rank-normalization
- addition of a new class `stan/mcmc/chainset.hpp` which stores a set of samples as a std::vector of Eigen MatrixXds.
- changes to the `stan_csv_reader` to fix bug https://github.com/stan-dev/stan/issues/3301, and to distinguish between ADVI and NUT-HMC samples (because CmdStan PR https://github.com/stan-dev/cmdstan/pull/972 was done in order to get summaries from ADVI outputs).

#### Intended Effect

The consumer of this functionality is CmdStan's `stansummary` utility, which, given a list of Stan CSV filenames, assembles them into a chainset object and then calls functions on the chainset object to get summary stats and diagnostics.

#### How to Verify

unit tests in this repo; unit tests in CmdStan repo test end-to-end parsing and reporting

#### Side Effects

N/A

#### Documentation

Documentation for CmdStan bin/stansummary to be done in docs repo, and CmdStanPy docs (CmdStanPy's `summary` function wraps CmdStan's `stansummary`).

#### Copyright and Licensing


mitzimorris:
Created a new PR because accidentally included some version of Stan math_lib in https://github.com/stan-dev/stan/pull/3305 and don't have the GitHub chops to undo its fubarness.

mitzimorris:
CmdStan tests failed - correctly!
offending test is here:  https://github.com/stan-dev/cmdstan/blob/33e27a4f7b72c7df558d982a33bfe33ce0b14211/src/test/interface/variational_output_test.cpp#L53

ADVI output format is 1st line is the ADVI estimate, followed by 1000 samples.
summary of *samples* should only look at the sample, not the ADVI mean.

WardBrian:
> Created a new PR because accidentally included some version of Stan math_lib in #3305 and don't have the GitHub chops to undo its fubarness.

Seems like a good move. Only downside is we lost the review history, which makes it difficult to track what has changed

mitzimorris:
agreed - will continue to add in changes suggested/requested in #3305

WardBrian:
Sounds good! Could you ping me when you think all of the 'old' review is resolved and I can take a look at the branch then?

src/stan/io/stan_csv_reader.hpp
```diff
@@ -348,36 +351,55 @@ class stan_csv_reader {
...
         *out << "Warning: non-fatal error reading samples" << std::endl;
     }
-
+    if (data.metadata.method == "sample") {
+      int expected_samples = data.metadata.num_samples / data.metadata.thin;
+      if (expected_samples != data.samples.rows()) {
+        std::stringstream msg;
+        msg << "Error reading samples, expecting " << expected_samples
+            << " samples, found " << data.samples.rows();
+        throw std::invalid_argument(msg.str());
```
WardBrian: I'm not sure of the value add of checking this -- in particularly, it would prevent us from summarizing a csv file that didn't come from Stan, which is otherwise fine (and useful, I would argue).

Because we're now setting the `badbit` exception flag (https://github.com/stan-dev/cmdstan/pull/1249) we don't need to worry as much about malformed CSVs like this going undetected

src/stan/mcmc/chainset.hpp
```diff
@@ -0,0 +1,567 @@
...
+ *
+ * @param stan_csv parsed csv file object
+ * @return expected number of draws
+ */
+size_t thinned_samples(const stan::io::stan_csv& stan_csv) {
+  size_t thinned_samples = stan_csv.metadata.num_samples;
+  if (stan_csv.metadata.thin > 1) {
+    thinned_samples = thinned_samples / stan_csv.metadata.thin;
+  }
+  return thinned_samples;
```
WardBrian: Another place to consider if we want to enforce metadata being present -- I would prefer to default to the number of rows being the number of samples if metadata is absent

src/stan/mcmc/chainset.hpp
```diff
@@ -0,0 +1,567 @@
...
+  /**
+   * Compute median value of specified parameter across all chains.
+   *
+   * @param index parameter index
+   * @return median
+   */
+  double median(const int index) const {
+    Eigen::MatrixXd draws = samples(index);
+    std::vector<double> sorted(draws.data(), draws.data() + draws.size());
+    std::sort(sorted.begin(), sorted.end());
```
WardBrian: Optional: I think there is a faster way than sorting. `std::nth_element` looks close, but not perfect

src/stan/mcmc/chainset.hpp
```diff
@@ -0,0 +1,567 @@
...
+   */
+  double quantile(const int index, const double prob) const {
+    // Ensure the probability is within [0, 1]
+    if (prob < 0.0 || prob > 1.0) {
+      throw std::out_of_range("Probability must be between 0 and 1.");
+    }
+    Eigen::MatrixXd draws = samples(index);
+    std::vector<double> sorted(draws.data(), draws.data() + draws.size());
+    std::sort(sorted.begin(), sorted.end());
+    size_t idx = static_cast<size_t>(prob * (sorted.size() - 1));
```
WardBrian: Combined with above: worth writing a helper that is outside this class and can be used for this and the couple median calculations

src/stan/mcmc/chainset.hpp
```diff
@@ -0,0 +1,567 @@
...
+   * @return pair (bulk_ess, tail_ess)
+   */
+  double mcse_sd(const int index) const {
+    Eigen::MatrixXd s = samples(index);
+    Eigen::MatrixXd s2 = s.array().square();
+    double ess_s = analyze::split_rank_normalized_ess(s).first;
+    double ess_s2 = analyze::split_rank_normalized_ess(s2).first;
+    double ess_sd = std::min(ess_s, ess_s2);
+    return sd(index)
+           * std::sqrt(std::exp(1) * std::pow(1 - 1 / ess_sd, ess_sd - 1) - 1);
```
WardBrian: We can use `e()` from Stan Math instead of exp(1)
https://github.com/stan-dev/math/blob/9b1f08be7e76c6468fc2fcf2f75d641a619ae349/stan/math/prim/fun/constants.hpp#L20C24-L20C29

src/stan/mcmc/chainset.hpp
```diff
@@ -0,0 +1,567 @@
...
+      ss << "Bad index " << index << ", should be between 0 and "
+         << (num_chains() - 1);
+      throw std::invalid_argument(ss.str());
+    }
+    if (index < 0 || index >= param_names().size()) {
+      std::stringstream ss;
+      ss << "Bad index " << index << ", should be between 0 and "
+         << (num_params() - 1);
+      throw std::invalid_argument(ss.str());
+    }
```
WardBrian: Rest of this file is not doing bounds checking -- either behavior seems fine but should be consistent

src/test/unit/io/bernoulli_corrupt.csv
WardBrian, line 1: Re previous comment, I'm not sure this kind of corruption represents something cmdstan can actually generate now

src/test/unit/io/test_csv_files/missing_draws.csv
WardBrian, line 1: Save as above "corrupt" csv?

src/stan/io/stan_csv_reader.hpp
mitzimorris: removed.

src/stan/mcmc/chainset.hpp
mitzimorris: implemented.  note, other than for median, we'll likely be calling `quantiles`, and use of std::sort is unavoidable.

WardBrian: we avoid it in the existing `chains.hpp` which uses `boost::accumulators::quantile`. is there a reason to change?

src/stan/mcmc/chainset.hpp
mitzimorris: gone.

src/stan/mcmc/chainset.hpp
mitzimorris: for median and `quantile`, std::nth_element is fast.
for `quantiles`, is sorting the entire array once better than iterating on a bunch of partial sorts?

the code in `chains.hpp` `quantiles` function also is just a big loop over the array of probs passed in.

WardBrian: Assuming the boost quantile code is one of the O(n) implementations, for p quantiles that will be O(np)

Sorting is O(n log(n)), so if p > log(n), I would expect the boost version with a loop to be faster, but this is where constant factors would start to mater for small n, p

mitzimorris: my other concern, w/r/t quantiles, is to avoid the copy that's required in setting up the nth_element finding array.
also, N is usually 4000, so logN is 12, and the default quantiles call has 3 args.  so a difference of 9, plus whatever constant, plus 2 extra copy operations.   I would prefer to leave it as is.

WardBrian: This sort also requires a copy right?

By "leave it as is" does that mean the boost implementation?

mitzimorris: for `quantiles`, using std::sort requires a copy, but then you can use the sorted array to get all requested quantiles.
added `quantile`, which uses `nth_element`, so one copy but less sorting.  `median` calls `quantile`.

WardBrian: Does this actually compute the same quantiles as the current code? There are a fair number of subtleties, cf https://github.com/stan-dev/cmdstanpy/issues/762#issuecomment-2150061089

mitzimorris: this computes the same quantiles as the implementation in R's `posterior` package, which is what CmdStanR uses.  

latest commit is using `nth_element` everywhere.   I don't think we need the boost accumulators includes anymore.

mitzimorris: https://github.com/stan-dev/stan/blob/021ac040c0d18dd97c5311b033b5e58c21502731/src/test/unit/mcmc/chainset_test.cpp#L179

WardBrian: Okay. Probably still worth noting to users that this will be different than the numbers they've been seeing before

src/stan/mcmc/chainset.hpp
mitzimorris: the new summary will include MAD, and ESS_bulk and ESS_tail, and won't include ESS/second; we will need to document all of this, as well as the fact that quantiles are calculated using different libraries than before.

mitzimorris:
@WardBrian @SteveBronder I've made changes and have a branch of CmdStan that works with updated code.
can this go in today?

WardBrian:
It's a bit tricky to approve because of the size. Overall:

- changes to `src/stan/io/stan_csv_reader.hpp` look great, would be happy to merge if they were independent
- Moving the existing computations to their own files seems like a good idea, would be happy to merge
- new tests all look good
- I'm not sure why any changes were made to `chains.hpp` -- just deleting dead code?
- The idea behind the new chainset class is well motivated (I like having fewer edge cases/more up-front validation) but I'm pretty nervous about changing both the code organization and the behavior at the same time (see above on quantiles, etc). It also means we can't test the new one against the old one

It would be nice if _every_ function in chainset looked like the `split_rank_normalized_ess` functions, where the actual computation lived somewhere else. That way it is easy to independently test, discuss, change, etc. 

It also seems odd to me to not expose the older forms of ess etc in the chainset class, if they're similarly small functions. I understand the primary consumer may not care as it stands, but if we want to say "here's a better API for chains.hpp", including the stuff chains.hpp can already do seems nice

mitzimorris:
>  I'm not sure why any changes were made to chains.hpp -- just deleting dead code?

https://github.com/stan-dev/stan/pull/3266 added call for split r-hat to chains.hpp

when I first started working on this PR, I thought that we could add rank-normalized ESS and MCSE to chains.hpp as well.
I can revert chains.hpp to the state that it was in before #https://github.com/stan-dev/stan/pull/3266.

mitzimorris:
> I'm pretty nervous about changing both the code organization and the behavior at the same time (see above on quantiles, etc). It also means we can't test the new one against the old one

we know that the behavior will be different because the old version was implemented before the 2019 improved r-hat paper came along.   for this PR, I ran the example output CSV files through CmdStanR, which uses posterior, which provides a correct implementation of the rank-normalized rhat.  (Aki / Paul Beurkner / Johah ).

WardBrian:
I'm not referring to the new ESS/R-Hat in that comment -- those are entirely new things being computed and reported to the user. But it does seem weird (and potentially impactful) if the 95% interval changes based on just updating stansummary and running on the same outputs as before

mitzimorris:
this is what the quantile code from `chains.hpp` computes:

```
    using boost::accumulators::accumulator_set;
    using boost::accumulators::left;
    using boost::accumulators::quantile;
    using boost::accumulators::quantile_probability;
    using boost::accumulators::right;
    using boost::accumulators::stats;
    using boost::accumulators::tag::tail;
    using boost::accumulators::tag::tail_quantile;
    double M = x.rows();
    size_t cache_size = M;

    if (prob < 0.5) {
      accumulator_set<double, stats<tail_quantile<left> > > acc(
          tail<left>::cache_size = cache_size);
      for (int i = 0; i < M; i++)
        acc(x(i));
      return quantile(acc, quantile_probability = prob);
    }
    accumulator_set<double, stats<tail_quantile<right> > > acc(
        tail<right>::cache_size = cache_size);
    for (int i = 0; i < M; i++)
      acc(x(i));
    return quantile(acc, quantile_probability = prob);
  }
```

the new code is more efficient - but possibly different behavoir.

your call on what to do.

WardBrian:
> the new code is more efficient 

Do you have timing on that? The latest commit seems to just call quantile in a loop, so I'd expect them to end up pretty much even

> but possibly different behavoir.

Definitely different -- here's a super minimal CSV file:

```csv
foo
100
101
102
103
104
105
106
107
108
109
110
```

current stansummary:
```
      Mean  MCSE  StdDev  5%  50%  95%  N_Eff  N_Eff/s  R_hat

foo    105   1.8     3.3 100  105  110    3.4      inf    2.8
```

the cmdstan PR based on this branch:

```
      Mean  MCSE  StdDev  MAD   5%  50%  95%  N_Eff_bulk  N_Eff_tail  R_hat

foo    105   1.5     3.3  4.4  100  105  109         5.0         5.0    2.1
```

mitzimorris:
this is not surprising - the super-minimal CSV file has 10 rows.  we're advertising different computation.  we do this all the time.

mitzimorris:
also, trying to run minimal CSV file  "foo.csv" through bin/stansummary from 2.35.0 release bombs.  how did you run this test?

mitzimorris:
here's another test:  current release vs new branch - quantiles are the same.

```
 ~/.cmdstan/cmdstan-2.35.0> bin/stansummary src/test/interface/example_output/bernoulli_chain_1.csv 
Inference for Stan model: bernoulli_model
1 chains: each with iter=(1000); warmup=(0); thin=(1); 1000 iterations saved.

Warmup took 0.0090 seconds
Sampling took 0.023 seconds

                Mean     MCSE   StdDev     5%   50%   95%  N_Eff  N_Eff/s  R_hat

lp__            -7.3  3.7e-02     0.77   -9.1  -7.0  -6.8    443    19275    1.0
accept_stat__   0.90  4.6e-03  1.5e-01   0.57  0.96   1.0   1026    44597   1.00
stepsize__       1.0      nan  4.9e-15    1.0   1.0   1.0    nan      nan    nan
treedepth__      1.4  1.6e-02  4.9e-01    1.0   1.0   2.0    895    38919   1.00
n_leapfrog__     2.3  3.4e-02  9.6e-01    1.0   3.0   3.0    799    34758   1.00
divergent__     0.00      nan  0.0e+00   0.00  0.00  0.00    nan      nan    nan
energy__         7.8  5.1e-02  1.0e+00    6.8   7.5   9.9    411    17865    1.0

theta           0.26  6.1e-03     0.12  0.079  0.25  0.47    384    16683   1.00

```

new version:
```
(base) ~/github/stan-dev/cmdstan (feature/1263-new-rhat-summary)> bin/stansummary src/test/interface/example_output/bernoulli_chain_1.csv 
Inference for Stan model: bernoulli_model
1 chains: each with iter=1000; warmup=1000; thin=1; 1000 iterations saved.

Warmup took 0.0090 seconds
Sampling took 0.023 seconds

                Mean    MCSE   StdDev    MAD     5%   50%   95%  N_Eff_bulk  N_Eff_tail  R_hat

lp__            -7.3   0.034  7.7e-01   0.30   -9.1  -7.0  -6.8         519         503    1.0
accept_stat__   0.90  0.0041  1.5e-01  0.064   0.57  0.96   1.0        1284         941   1.00
stepsize__       1.0     nan  2.4e-15   0.00    1.0   1.0   1.0         nan         nan    nan
treedepth__      1.4     nan  4.9e-01   0.00    1.0   1.0   2.0         nan         nan    nan
n_leapfrog__     2.3   0.034  9.6e-01   0.00    1.0   3.0   3.0         790         786   1.00
divergent__     0.00     nan  0.0e+00   0.00   0.00  0.00  0.00         nan         nan    nan
energy__         7.8   0.047  1.0e+00   0.74    6.8   7.5   9.9         489         486    1.0

theta           0.26  0.0063  1.2e-01   0.12  0.079  0.25  0.47         361         395    1.0

Samples were drawn using hmc with nuts.
For each parameter, N_Eff_bulk and N_Eff_tail measure the effective sample size 
for the entire sample and for the the .05 and .95 tails, respectively, 
and R_hat_bulk and R_hat_tail measure the potential scale reduction 
on split chains, (at convergence will be very close to 1).
```

WardBrian:
I used the current develop branch

I’m not saying we’re obligated to report the same numbers from now until the end of days, but I just think if we do make a change we should have a reason why the new number is better/more correct, not just that we like the code that calculates it more

mitzimorris:
I have checked the existing unit tests and the behavior on real output files.   I think this is OK.

bob-carpenter:
> I’m not saying we’re obligated to report the same numbers from now until the end of days, but I just think if we do make a change we should have a reason why the new number is better/more correct, not just that we like the code that calculates it more

The quantiles shouldn't change.  [edit: this is wrong---see below] But the R-hat and ESS will.  That's one of the motivations for the PR---to switch over to our improved versions of these as described in this paper:

Vehtari, Gelman, Simpson, Carpenter, Bürkner. 2019.
Rank-normalization, folding, and localization: An improved Rˆ for assessing convergence of MCMC
https://arxiv.org/abs/1903.08008

I believe @avehtari has already switched over RStan and PyStan and we're just waiting for this PR to switch over CmdStan and hence CmdStanPy and CmdStanR.  I believe arViz also does these new calculations.

We could keep our old definitions and report two versions of R-hat (old, old per second, new) and three versions of ESS (old, new bulk, new tail), but I think it'd be confusing for users.

bob-carpenter:
As another comment, I don't think we should be reporting ESS for sampler parameters other than `lp__`.

avehtari:
> I believe @avehtari has already switched over RStan and PyStan and we're just waiting for this PR to switch over CmdStan and hence CmdStanPy and CmdStanR. I believe arViz also does these new calculations.

 CmdStanR can call CmdStan diagnostics, but CmdStanR is also using `posterior` package, which has new Rhat and bulk- and tail-ESS, and it is more handy to use `posterio` package diagnostics than `CmdStan` diagnostics. `ArviZ`, which is used by Python and Julia users, has these new functions, too. RStan is not yet using `posterior` package.

> We could keep our old definitions and report two versions of R-hat (old, old per second, new) and three versions of ESS (old, new bulk, new tail), but I think it'd be confusing for users.

The old ESS which (called `ess_basic()` in `posterior` package) is still needed for computing MCSE for mean estimate. I think the issue is that reporting everything by default would be too much, but CmdStan doesn't have a nice way for the users to define which diagnostic quantities they want to see, and thus we need to make some choices. 

> As another comment, I don't think we should be reporting ESS for sampler parameters other than lp__.

I agree. Same for Rhat.

WardBrian:
> The quantiles shouldn't change.

My point is, they do. I believe the test cases we are using are not specific enough to see that, but doing a bit of testing on the side:

- Running the bernoulli model
- Running the current stansummary and this branch's version, with some extra decimals to be able to actually tell the difference
- On basically every run I've observed so far, they disagree on the `50%` quantile by the 4th decimal place

Here's a CSV file: https://gist.github.com/WardBrian/41bbf1b57829efa32e6aa58c5c493f57

Current stansummary (with --sig_figs=4)

```
                     Mean       MCSE     StdDev         5%        50%        95%  N_Eff    N_Eff/s      R_hat
theta           2.593e-01  7.266e-03  1.251e-01  8.379e-02  2.429e-01  4.830e-01  296.7      59330      1.003
```

new stansummary (with --sig_figs=4)

```
                  Mean       MCSE  StdDev      MAD       5%     50%     95%  N_Eff_bulk  N_Eff_tail   R_hat
theta           0.2593  7.263e-03  0.1251   0.1343  0.08379  0.2425  0.4830       296.9       446.6   1.008
```


0.2429 != 0.2425

CmdStanPy uses `--sig_figs=6` by default, so end users _will_ see different numbers.

Again, we can decide we _want_ to make this change, but it just happening incidentally doesn't sit right with me, hence me not merging as-is.

avehtari:
> N_Eff_bulk  N_Eff_tail

It would be better to use terms ESS_bulk and ESS_tail, as 1) N has different meaning in the paper describing these quantities, and 2) N_eff makes people to misread it as "Number of EFFective samples" which is wrong, and the correct term is "Effective Sample Size", 3) CmdStanR using posterior is showing ESS_bulk and ESS_tail.

mitzimorris:
> The old ESS which (called ess_basic() in posterior package) is still needed for computing MCSE for mean estimate.

@avehtari  - could you check the implementations of mcse and mcse_sd?

https://github.com/stan-dev/stan/blob/dbf6a9f80ae2cd38e64fb627080ead01bc81dad7/src/stan/mcmc/chainset.hpp#L424-L467

WardBrian:
> It would be better to use terms ESS_bulk and ESS_tail

Yes, there is an existing cmdstan issue for this https://github.com/stan-dev/cmdstan/issues/916, which @mitzimorris will address as part of the cmdstan pr that accompanies this

mitzimorris:
@WardBrian regarding the implementation of quantiles, we need a better understanding how the boost accumulators library computes quantiles - it is returning the value of a the nth draw or interpolating?  and which is appropriate?

I spent yesterday looking at the boost code and don't understand it well enough to make that call.   changing to use C++ std::nth_element would make the code easier to understand and remove dependencies, but as you say, that's not a good reason to change behavoirs.

WardBrian:
I suspect the difference may just be as simple as the existing code prioritizes counting from the left if you're requesting the 50% or higher quantile, but I agree the boost reference is pretty opaque in this area

WardBrian:
There is also `stan::math::quantile` -- calling this would probably be the _ideal_ thing to do from a code standpoint, but it yields yet a third set of numbers...

src/test/unit/mcmc/chainset_test.cpp
```diff
@@ -0,0 +1,334 @@
...
+  // test against R implementation in pkg posterior (via cmdstanr)
+  Eigen::VectorXd s8_mean(10), s8_median(10), s8_sd(10), s8_mad(10), s8_q5(10),
+      s8_q95(10);
+  s8_mean << 7.95, 12.54, 7.82, 5.33, 7.09, 4.12, 5.72, 11.65, 8.80, 8.26;
+  s8_median << 8.00, 11.27, 7.39, 5.44, 6.64, 4.54, 5.93, 11.38, 8.28, 7.05;
+  s8_sd << 5.48, 9.57, 6.85, 8.39, 6.91, 6.57, 6.85, 7.76, 8.40, 5.53;
+  s8_mad << 5.49, 8.79, 6.39, 7.38, 5.98, 6.25, 6.59, 7.79, 7.59, 4.66;
+  s8_q5 << -0.46, -0.39, -3.04, -8.90, -3.31, -7.58, -5.84, 0.10, -4.15, 2.08;
+  s8_q95 << 17.01, 30.47, 19.25, 19.02, 18.72, 14.49, 16.04, 25.77, 22.71,
+      18.74;
```
WardBrian, line 247: I think this is simply too few decimal places to tell between the various implementations of quantile for this data. 

The one in stan math (added in https://github.com/stan-dev/math/pull/2398) claims to match R's implementation, but past a couple decimal places it gives different answers than this, which also claims to be tested against R's implementation. Math's is still only tested to 3 decimals, so it's hard to say.

If we were going to go with one, I'd say `stan::math::quantile`, but being sure they actually implement the same algorithm would be a nice justification

WardBrian: (`stan::math::quantile` is also well defined a 0 and 1., which stansummary currently prevents you from requesting, so we could relax that in this case)

src/stan/mcmc/chains.hpp
```diff
@@ -117,7 +117,6 @@ class chains {
     using boost::accumulators::tag::tail;
     using boost::accumulators::tag::tail_quantile;
     double M = x.rows();
-    // size_t cache_size = std::min(prob, 1-prob)*M + 2;
```
SteveBronder, line 120: Is there a reason to change anything in chains.hpp?

bob-carpenter:
> The quantiles shouldn't change. 

I spoke too soon.  The problem with quantiles is that reasonable definitions can differ.   When you don't have an exact item, a lot of algorithms are going to interpolate, and how you do that can differ.  For example if you only have two numbers [1, 3] and ask for a 50% quantile (i.e., a median), what is the answer?  You could return 1 or 3, which is what you get if you turn 50% and the size of 2 into an index.  Or you could interpolate between 1 and 3 and return 2.  The problem's the exact same thing at an edge, but it might not be 50%.  What do you do if I ask you for a 40% quantile of [1, 3]?  There you might be tempted to linearly interpolate and say 1.8.  That interpolation can often be made more stable if you use more elements around an element, so if we have four elements [2, 3, 5, 10], we might return a different answer than [3, 5] for the 50% quantile.  It's a mess!

The bottom line, though, is that in these situations, we really don't care.  The sampling algorithm didn't have the granularity to cleanly resolve the quantile so if we use one of the neightboring values or interpolate, it should be fine for our use.

> There is also stan::math::quantile -- calling this would probably be the ideal thing to do from a code standpoint, but it yields yet a third set of numbers.

See above.  That should also be fine.  Using a `quantile` function that works on a sorted list will be faster if we do a lot of quantiles.  Otherwise, using an accumulator boost style will be faster.  With an accumulator, if you are looking for the rank K element in a list, there's an O(N * log2(K)) algorithm (N for considering each element, log2(K) to insert it into position in the accumulator) that only consumes log2(K) memory.  It can get even more clever than that, see, e.g., the variants available here:

https://www.boost.org/doc/libs/1_86_0/doc/html/accumulators/user_s_guide.html#accumulators.user_s_guide.the_statistical_accumulators_library

WardBrian:
Calling `stan::math::quantile` seems like the best choice then. The code can also be made quite simple:

```c++
    Eigen::MatrixXd draws = samples(index);
    Eigen::Map<Eigen::VectorXd> map(draws.data(), draws.size());

    return stan::math::quantile(map, prob);
```

There is an overload for if `prob` is a single number or if it is a vector

src/stan/mcmc/chains.hpp
mitzimorris: I removed this commented out line after going through the blame - this code has always been there in its commented out form.   why is it here?   maybe to show one alternative - which would use less memory, run faster, and provide approximate as opposed to exact estimates.   I don't know.  so I got rid of this comment, and a few other comments that were clearly outdated.

mitzimorris: > Is there a reason to change anything in chains.hpp?

the one thing that needed to be changed was to remove the rank-normalized r-hat functions, which weren't being called by anything, since this PR changed the function arguments.

mitzimorris:
plugged most recent version of this branch into CmdStan and comparing CmdStanR's summary (via fit$print) against CmdStan's bin/stansummary on the test CSV files in `cmdstan/src/test/interface/example_outputs` and median and quantile values, as well as mean, sd, mad, ess, and rhat all match out to 9 digits.

src/stan/mcmc/chainset.hpp
```diff
@@ -0,0 +1,504 @@
+#ifndef STAN_MCMC_CHAINSET_HPP
+#define STAN_MCMC_CHAINSET_HPP
+
+#include <stan/io/stan_csv_reader.hpp>
+#include <stan/math/prim.hpp>
+#include <stan/math/prim/fun/constants.hpp>
+#include <stan/math/prim/fun/quantile.hpp>
```
WardBrian, line 7: Nit: These two includes shouldn't be necessary if we're also including math/prim above

src/stan/mcmc/chainset.hpp
```diff
@@ -0,0 +1,504 @@
...
+  /**
+   * Compute sample variance for specified parameter across all chains.
+   * 1 / (N - 1) * sum((theta_n - mean(theta))^2)
+   *
+   * @param index parameter index
+   * @return sample variance
+   */
+  double variance(const int index) const {
+    Eigen::MatrixXd draws = samples(index);
+    return (draws.array() - draws.mean()).square().sum() / (draws.size() - 1);
```
WardBrian, line 209: This is `stan::math::variance`

src/stan/mcmc/chainset.hpp
```diff
@@ -0,0 +1,504 @@
...
+   *
+   * Throws exception if specified probability is not between 0 and 1.
+   *
+   * @param index parameter index
+   * @param prob probability
+   * @return parameter value at quantile
+   */
+  double quantile(const int index, const double prob) const {
+    // Ensure the probability is within [0, 1]
+    if (prob <= 0.0 || prob >= 1.0) {
```
WardBrian, line 301: I think we can allow the endpoints using math's `quantile`. It also throws its own error if necessary, so we could just skip this entirely now.

src/stan/mcmc/chainset.hpp
```diff
@@ -0,0 +1,504 @@
...
+   * @param probs vector of probabilities
+   * @return vector of parameter values for quantiles
+   */
+  Eigen::VectorXd quantiles(const int index,
+                            const Eigen::VectorXd& probs) const {
+    if (probs.size() == 0)
+      return Eigen::VectorXd::Zero(0);
+    if (probs.minCoeff() <= 0.0 || probs.maxCoeff() >= 1.0) {
+      throw std::out_of_range("Probabilities must be between 0 and 1.");
+    }
```
WardBrian, line 339: Same note as above

src/stan/mcmc/chainset.hpp
```diff
@@ -0,0 +1,504 @@
...
+   * Computes the mean Monte Carlo error estimate for the central 90% interval.
+   * See https://arxiv.org/abs/1903.08008, section 4.4.
+   * Follows implementation in the R posterior package
+   *
+   * @param index parameter index
+   * @return pair (bulk_ess, tail_ess)
+   */
+  double mcse_mean(const int index) const {
+    double ess_bulk = analyze::split_rank_normalized_ess(samples(index)).first;
+    return sd(index) / std::sqrt(ess_bulk);
```
WardBrian, line 424: Question for @avehtari from above about which ESS this should use:

>  The old ESS which (called ess_basic() in posterior package) is still needed for computing MCSE for mean estimate.

?

src/stan/mcmc/chainset.hpp
```diff
@@ -0,0 +1,504 @@
...
+
+  /**
+   * Computes the split potential scale reduction (split Rhat) using rank based
+   * diagnostic for a set of per-chain draws, for bulk and tail Rhat.
+   * Based on paper https://arxiv.org/abs/1903.08008
+   *
+   * @param index parameter index
+   * @return pair (bulk_rhat, tail_rhat)
+   */
+  std::pair<double, double> split_rank_normalized_rhat(const int index) const {
```
WardBrian, line 370: (I still think it is worth adding functions for the "old" rhat and ess to this class, if they'd be this same size)

WardBrian (commented):
Ok, I had a few small things I noticed going through but otherwise I think this is good

mitzimorris:
comments look great - thanks!  per discussion with @SteveBronder, I think the way to proceed here is with 3 PRs:

1. fix issues with stan_csv_reader
2. refactor code added to stan::analyze::mcmc in https://github.com/stan-dev/stan/pull/3266 to add ESS_bulk, ESS_tail and MCSE calculations (to chains.hpp)
3. add `chainset.hpp` to code base; a replacement for `chains.hpp`.

WardBrian:
I agree that would have been a better (or at least probably faster) way to go originally, but this being as far along as it is it would also be fine by me to proceed with this one big PR. Whichever you feel like, at this point

mitzimorris:
I think 3 PRs would be better - a lot of work, but the grown-up thing to do here.  (especially since I've been advocating pragmatic programming lately).

src/stan/mcmc/chainset.hpp
avehtari: mcse_mean should use ess_mean, not ess_bulk
in the same way, mcse_sd should not use the rank normalized ess

avehtari:
> @avehtari - could you check the implementations of mcse and mcse_sd?

These should use ess_mean without rank normalization. The rank normalized versions are generic diagnostics that work also with infinite variance, but by definition of MCSE mcse_mean and mcse_sd should not use rank-normalization.

mitzimorris:
closing this per above comment - https://github.com/stan-dev/stan/pull/3310#issuecomment-2389091218

PR #3312, opened 2024-10-07
Feature/3299 improved ess rhat

mitzimorris:
#### Submission Checklist

#### Summary

Refactor of method for computing rank-normalized split bulk and tail rhat and addition of method to compute rank-normalized split bulk and tail ess  as described in https://arxiv.org/abs/1903.08008.

#### Intended Effect

Refactor the logic added to `stan/analyze/mcmc/compute_potential_scale_reduction.hpp` via PR https://github.com/stan-dev/stan/pull/3266. so that some of the logic can be reused to implement rank-normalized split ESS.

- added helper functions `rank_normalization` and `split_chains` and new functions `rank_normalized_split_rhat` and `rank_normalized_split_ess`, each in a separate file.

This PR will be followed by another refactor of the `stan::mcmc::chains` object.  In anticipation of that refactoring, both `rank_normalized_split_rhat` and `rank_normalized_split_ess` have a single argument which is an Eigen::MatrixXd of draws instead of vectors `draws`, `sizes`.

#### How to Verify

Unit tests based on running the same Stan CSV files through CmdStanR and using the current implementations in R's posterior library to get rhat and ess.

#### Side Effects

N/A

#### Documentation

N/A - when CmdStan's `stansummary` method is updated, will be documented in CmdStan docs.

#### Copyright and Licensing


mitzimorris:
this is ready for review.

src/stan/analyze/mcmc/rank_normalization.hpp
```diff
@@ -0,0 +1,61 @@
...
+      sum_ranks += j + 1;  // Rank starts from 1
+      ++j;
+      ++count;
+    }
+    double avg_rank = sum_ranks / count;
+    boost::math::normal_distribution<double> dist;
+    for (std::size_t k = i; k < j; ++k) {
+      double p = (avg_rank - 0.375) / (size + 0.25);
+      const Eigen::Index index = value_with_index[k].second;
+      rank_matrix(index) = boost::math::quantile(dist, p);
```
WardBrian, line 51: am I correct that this is `stan::math::inv_Phi`?

I know this is just moving code around, but if so that's probably worth calling instead

src/test/unit/mcmc/chains_test.cpp
```diff
@@ -925,14 +916,9 @@ TEST_F(McmcChains, blocker_split_potential_scale_reduction_rank) {
...
         << ", parameter: " << chains.param_name(index);
-
-    // ASSERT_NEAR(rhat(index - 4),
-    //             chains.split_potential_scale_reduction_rank(index), 1e-4)
-    //     << "rhat for index: " << index
-    //     << ", parameter: " << chains.param_name(index);
   }
 
-  for (int index = 0; index < chains.num_params(); index++) {
+  for (int index = 4; index < chains.num_params(); index++) {
```
WardBrian, line 921: Why is this file changing?

src/test/unit/analyze/mcmc/compute_potential_scale_reduction_test.cpp
WardBrian, line 1: Did these tests end up anywhere?

src/stan/analyze/mcmc/rank_normalization.hpp
mitzimorris: I'm not sure - this is a question for @aleksgorica or @avehtari

src/test/unit/mcmc/chains_test.cpp
mitzimorris: the test data are files `blocker1.csv` and `blocker2.csv` - these are CSV outputs from Stan version 1.3 with sampler params `lp__,`, `accept_stat__`, `stepsize__`, and `treedepth__`; except for `lp__`, the diagnostics will return `nan` and the test fails because `nan` != `nan`.

plus, removing commented out code

src/test/unit/analyze/mcmc/compute_potential_scale_reduction_test.cpp
mitzimorris: yes, in `split_normalized_rhat_test.cpp`

it seemed like the tests were highly redundant, so only test once.

WardBrian (approved):
Looks good -- if we get an answer on that inv_Phi question we can address separately

src/stan/analyze/mcmc/rank_normalization.hpp
avehtari: @WardBrian, can you clarify the question? I'm not certain to which part you refer to be `stan::math::inv_Phi()`

WardBrian: We're computing rank[index] by using the boost::math::quantile function where this `dist` object is just a standard normal -- I believe that is equivalent to calling our own inv_Phi

bob-carpenter:
> Looks good -- if we get an answer on that inv_Phi question we can address separately

Yes, `inv_Phi` is the quantile function for standard normal distributions, so this should be the same.  

Our internal `inv_Phi` is hard coded directly:

https://mc-stan.org/math/prim_2fun_2inv___phi_8hpp_source.html

and then in what looks like an identical implementation to me to first order, for OpenCL:

https://mc-stan.org/math/opencl_2kernels_2device__functions_2inv___phi_8hpp_source.html

src/stan/analyze/mcmc/compute_potential_scale_reduction.hpp
```diff
@@ -17,156 +17,6 @@
...
 namespace analyze {
 
-/**
- * Computes normalized average ranks for draws. Transforming them to normal
- * scores using inverse normal transformation and a fractional offset. Based on
- * paper https://arxiv.org/abs/1903.08008
- * @param chains stores chains in columns
- * @return normal scores for average ranks of draws
- */
-inline Eigen::MatrixXd rank_transform(const Eigen::MatrixXd& chains) {
```
mitzimorris, line 27: this function moved to file `rank_normalization.hpp`

mitzimorris:
@avehtari - in this PR, file `split_rank_normalized_ess.hpp`, function `ess` is a refactor of this:  https://github.com/stan-dev/stan/blob/2550bbd2bbde86352b9cf9e1bc520968e3a88788/src/stan/analyze/mcmc/compute_effective_sample_size.hpp#L73-L137

do you see any problems here?

avehtari:
1.  Why `ess` function is in `split_rank_normalize_ess.hpp` while it's needed also for non-rank-normalized ess?
2. The new code is dropping the following and I don't understand how it can be done
```
  for (int chain = 0; chain < num_chains; ++chain)
    acov_s(chain) = acov(chain)(1);
```
3. The initial monotone sequence part has been moved before "the improved estimate" lines
```
  // this is used in the improved estimate, which reduces variance
  // in antithetic case -- see tau_hat below
  if (rho_hat_even > 0)
    rho_hat_s(max_s + 1) = rho_hat_even;
```
and I'm not able to figure in my head if that is guaranteed to give the same answer. I can't remember, but I assume I had a reason to have it outside that earlier loop and in the place where I had it.

4. I (and Andrew) don't like `num_samples`. It should be either `num_draws`, `num_total_draws`, or `sample_size`. The best would be to have in the beginning
```
 const Eigen::Index num_chains = chains.cols();
 const Eigen::Index num_iterations = chains.rows();
```
and then later
```
double num_draws = num_chains * num_iterations;
```

mitzimorris:
re point 2 - yes, caught that.

the differences between this implementation and posterior are that missing step:

```
  for (int chain = 0; chain < num_chains; ++chain)
    acov_s(chain) = acov(chain)(1);
```

and the use of `stan::analyze::covariance`  instead of `stan::math::covariance`.

these have been fixed in the latest version of PR #3313, and unit tests added which verify that the old and new versions of ess compute the same answer.

I'm now going through the rank-normalized rhat and will add similar checks.

this raises the question of the why a different implementation of covariance?  is one ESS specific?

bob-carpenter:
> I (and Andrew) don't like num_samples

Andrew instilled that distinction in me, and I try to stick to it, but I've given up trying to get other people to recognize the distinction.  It's just too commonly used everywhere to fight.  Maybe I should write another crabby linguist blog post :-)

On the other hand, I think we should stick to this distinction in our own writing about Stan, at least in our doc.

I created an issue to deal with the dozens of uses in our docs that violate the @andrewgelman style guide.

> https://github.com/stan-dev/docs/issues/823

PR #3313, opened 2024-10-11
Feature/3299 chainset

mitzimorris:
#### Submission Checklist

#### Summary

This is a do-over of PR https://github.com/stan-dev/stan/pull/3310.   The overall goal is to make the improved R-hat diagnostics (https://github.com/stan-dev/stan/pull/3266), and rank-normalized ESS bulk and tail, (https://github.com/stan-dev/stan/pull/3312) available to CmdStan, following the definitions in https://arxiv.org/abs/1903.08008.

This PR adds a new class `stan::mcmc::chainset`.  A `chainset` object requires that all chains are the same size on construction, unlike the `chains<>` object.  This simplifies computation of split-Rhat and split-ESS diagnostics.   The `chainset` object provides rank-normalized split Rhat, rank-normalized split ESS which returns bulk and tail ESS, and functions `mcse` and `mcse_sd` which calculate the mean Monte Carlo standard error and its variance.

#### How to Verify

Unit tests for the chainset test performance against Stan CSV outputs.   The CmdStan branch https://github.com/stan-dev/cmdstan/tree/feature/1263-new-rhat-summary further checks that `chainset` provides the necessary functions to output the same set of summary statistics as the R interfaces.

#### Side Effects

N/A

#### Documentation

Documentation will be added to the Stan docset in a separate PR.

#### Copyright and Licensing


src/stan/mcmc/chainset.hpp
```diff
@@ -0,0 +1,510 @@
...
+#include <stdexcept>
+#include <string>
+#include <sstream>
+#include <utility>
+#include <vector>
+#include <cstdlib>
+
+namespace stan {
+namespace mcmc {
+using Eigen::Dynamic;
```
WardBrian:
```suggestion
```

src/stan/mcmc/chainset.hpp
```diff
@@ -0,0 +1,510 @@
...
+   *
+   * Throws exception if specified probability is not between 0 and 1.
+   *
+   * @param index parameter index
+   * @param prob probability
+   * @return parameter value at quantile
+   */
+  double quantile(const int index, const double prob) const {
+    // Ensure the probability is within [0, 1]
+    if (prob <= 0.0 || prob >= 1.0) {
```
WardBrian: stan::math::quantile already checks its bounds, so I don't know if this is necessary. At any rate, it actually allows 0 and 1, so we probably should too.

src/stan/mcmc/chainset.hpp
```diff
@@ -0,0 +1,510 @@
...
+   */
+  double mcse_sd(const int index) const {
+    if (num_samples() < 4
+        || !stan::analyze::is_finite_and_varies(samples(index)))
+      return std::numeric_limits<double>::quiet_NaN();
+    Eigen::MatrixXd s = samples(index);
+    Eigen::MatrixXd s2 = s.array().square();
+    double ess_s = analyze::ess(s);
+    double ess_s2 = analyze::ess(s2);
+    double ess_sd = std::min(ess_s, ess_s2);
```
WardBrian: I can't find a matching implementation in posterior. The one I see doesn't do this min calculation:

https://github.com/stan-dev/posterior/blob/90a07cce51fdde484e97596cdd186adf265d8f5e/R/convergence.R#L490-L506

I think it would be nice if MCSE for the mean and for the standard deviation both lived in their own files, like how the ess/rhat functions do, and this class just called them. That would allow for expanded documentation and testing

src/test/unit/mcmc/chainset_test.cpp
```diff
@@ -0,0 +1,334 @@
...
+  // test against R implementation in pkg posterior (via cmdstanr)
+  Eigen::VectorXd s8_mean(10), s8_median(10), s8_sd(10), s8_mad(10), s8_q5(10),
+      s8_q95(10);
+  s8_mean << 7.95, 12.54, 7.82, 5.33, 7.09, 4.12, 5.72, 11.65, 8.80, 8.26;
+  s8_median << 8.00, 11.27, 7.39, 5.44, 6.64, 4.54, 5.93, 11.38, 8.28, 7.05;
+  s8_sd << 5.48, 9.57, 6.85, 8.39, 6.91, 6.57, 6.85, 7.76, 8.40, 5.53;
+  s8_mad << 5.49, 8.79, 6.39, 7.38, 5.98, 6.25, 6.59, 7.79, 7.59, 4.66;
+  s8_q5 << -0.46, -0.39, -3.04, -8.90, -3.31, -7.58, -5.84, 0.10, -4.15, 2.08;
+  s8_q95 << 17.01, 30.47, 19.25, 19.02, 18.72, 14.49, 16.04, 25.77, 22.71,
+      18.74;
```
WardBrian: I think all of these tests could stand to have a digit or two more

src/test/unit/mcmc/chainset_test.cpp
```diff
@@ -0,0 +1,334 @@
...
+
+  Eigen::VectorXd rhat_8_schools_1_tail(10);
+  rhat_8_schools_1_tail << 1.005676523, 1.009670999, 1.00184184, 1.002222679,
+      1.004148161, 1.003218528, 1.009195353, 1.001426744, 1.003984381,
+      1.025817745;
+
+  for (size_t i = 0; i < 10; ++i) {
+    auto rhats = chain_1.split_rank_normalized_rhat(i + 7);
+    EXPECT_NEAR(rhats.first, rhat_8_schools_1_bulk(i), 0.05);
+    EXPECT_NEAR(rhats.second, rhat_8_schools_1_tail(i), 0.05);
```
WardBrian: Having an absolute tolerance of 0.05 completely invalidates this test. At that level of tolerance, both of these rhat vectors are indistinguishable from a vector of 1s

I think the number of digits in the expected vectors is reasonable, so it would be great if this test passed with the tolerance more like `1e-8`, or even `1e-10`, which it seems like most of the other uses of EXPECT_NEAR in the code base use

src/test/unit/mcmc/chainset_test.cpp
```diff
@@ -0,0 +1,334 @@
...
+  // test against R implementation in pkg posterior (via cmdstanr)
+  Eigen::VectorXd ess_8_schools_bulk(10);
+  ess_8_schools_bulk << 348, 370, 600, 638, 765, 608, 629, 274, 517, 112;
+  Eigen::VectorXd ess_8_schools_tail(10);
+  ess_8_schools_tail << 845, 858, 874, 726, 620, 753, 826, 628, 587, 108;
+
+  for (size_t i = 0; i < 10; ++i) {
+    auto ess = chain_2.split_rank_normalized_ess(i + 7);
+    EXPECT_NEAR(ess.first, ess_8_schools_bulk(i), 5);
+    EXPECT_NEAR(ess.second, ess_8_schools_tail(i), 5);
```
WardBrian: Similar question on tolerances -- why don't we expect some exact answer?

src/test/unit/mcmc/chainset_test.cpp
```diff
@@ -0,0 +1,334 @@
...
+  // test against R implementation in pkg posterior
+  Eigen::VectorXd s8_mcse_mean(10), s8_mcse_sd(10);
+  s8_mcse_mean << 0.288379, 0.4741815, 0.2741001, 0.3294614, 0.2473758,
+      0.2665048, 0.2701363, 0.4740092, 0.3621771, 0.3832464;
+  s8_mcse_sd << 0.1841825, 0.2854258, 0.192332, 0.2919369, 0.2478025, 0.2207478,
+      0.2308452, 0.2522107, 0.2946896, 0.3184745;
+
+  for (size_t i = 0; i < 10; ++i) {
+    auto mcse_mean = chain_2.mcse_mean(i + 7);
+    EXPECT_NEAR(mcse_mean, s8_mcse_mean(i), 0.5);
```
WardBrian: Again -- absolute tolerance is larger than most values in this array we're testing against

I think all the EXPECT_NEARs in this file need to be tightened, some replaced with EXPECT_FLOAT_EQ

src/test/unit/mcmc/chainset_test.cpp
```diff
@@ -0,0 +1,334 @@
...
+  stan::mcmc::chainset chain_1(eight_schools_1);
+  EXPECT_EQ(1, chain_1.num_chains());
+
+  Eigen::VectorXd mu_ac_posterior(10);
+  mu_ac_posterior << 1.00000000000, 0.19487668999, 0.05412049365, 0.07834048575,
+      0.04145609855, 0.04353962161, -0.00977255885, 0.00005175308,
+      0.01791577080, 0.01245035817;
+  auto mu_ac = chain_1.autocorrelation(0, "mu");
+  for (size_t i = 0; i < 10; ++i) {
+    EXPECT_NEAR(mu_ac_posterior(i), mu_ac(i), 0.05);
```
WardBrian: Tolerance of 0.05 when expected answer is 0.00005 -- no good!

WardBrian (changes requested):
I see this is still marked as draft, but most of my concerns are in the tests, so I hope this is still helpful at this stage

src/stan/mcmc/chainset.hpp
mitzimorris: I cannot figure out why this is written this way - perhaps I'm the person who copied it from somewhere else - mea culpa, but seriously, this looks quite wrong.  replacing with implementation in posterior, written by Aki.

src/test/unit/mcmc/chainset_test.cpp
mitzimorris: this isn't possible - the compound differences introduced between the  C++ implementation of rank-normalization, and computations of the between and within chain variances and the corresponding operations in R result in values that have at least 0.03 difference for these tests.

WardBrian: If that is the case, we shouldn't test against R at all, and instead compute some values to the desired precision and "bless" them as correct. It requires a lot more confidence in the existing C++ implementation, though, so I'd try to track down if we couldn't have C++ and R agree better first

src/test/unit/mcmc/chainset_test.cpp
mitzimorris: to compute ESS, the C++ implementation first does rank-normalization, then splits the chains and uses the Boost libraries to compute autocovariance at all lags.
because the underlying libraries may differ, the results differ, but not by much.   anything wildly off would be suspect, but given that the numbers we get for a dataset are relatively close, I think this is OK.

src/test/unit/mcmc/chainset_test.cpp
mitzimorris: not possible - mcse computations are based on ESS which is based on autocovariance.  different implementations produce slightly different values.   overall, results are consistent.

src/test/unit/mcmc/chainset_test.cpp
mitzimorris: made test more strict, but differences between R and C++ are to be expected.

WardBrian:
> made test more strict, but differences between R and C++ are to be expected.

I am not convinced that they are expected (I recently wrote a javascript version of ESS and RHat that matches stan's current C++ out to ~10 digits). I would want to see some explanation of exactly where and why they do to this degree, if we're claiming to implement the same things.

But, if these differences can be justified, then we shouldn't be testing against those values at all. The presence of those values in the tests visually implies we do match the R values (when we don't), and the resulting tests have very little power to catch issues. For example:

- the MCSE calculation needs the sample standard deviation. If you replace it with the population standard deviation (by deleting the -1 in the denominator), the tests all pass.
- if you replace it with something even more outlandish, like adding twenty to the denominator instead of subtracting one, the tests _also_ still pass

I pick this example not just to be a pain, but because replacing the in-line standard deviation calculation in mcse_mean function is something a future code editor may be likely to do (if we don't already change it during review for this!), and accidentally picking the wrong library function is not hard to imagine, so failing the tests if that mistake gets made would be very helpful. Similar arguments can be made for basically all the other tests having tighter tolerances

src/test/unit/mcmc/chainset_test.cpp
mitzimorris: added 2 more digits.

mitzimorris:
> But, if these differences can be justified, then we shouldn't be testing against those values at all.

I would be happy to find another way to test these.   what would you suggest?

WardBrian:
What previous implementations of diagnostics seem to do is just run the code once to get values and then "freeze" those in the test. I think #3312 deleted some tests in that style.

Doing this requires us to be pretty certain the implementation is correct, since we're making it the gold standard for all future versions of the code, which is why it would be important to understand what parts of the calculation are different than R

mitzimorris:
> which is why it would be important to understand what parts of the calculation are different than R

perhaps we should create an artificial dataset to convince ourselves of the goodness of ESS and Rhat.   what did you use for Javascript?

WardBrian:
For testing I ended up using the same blocker1.csv and blocker2.csv as the stan tests, but for initially getting them to agree I was just using an arbitrary CSV file, I think from the Bernoulli model?

mitzimorris:
@WardBrian this is ready for re-review.

the discrepencies between this code and the implementations in posterior package were due to:
- differences between `stan::math::covariance` and `stan::analyze::covariance`.   now using the latter.
- testing against outputs from centered eight_schools which contained only 500 draws - too few draws to resolve the variance.   instead now testing against 4000 draws from `bernoulli.stan` model everywhere.


changes per discussion here and in https://github.com/stan-dev/stan/pull/3312#issuecomment-2420180157

- further refactoring of code in `src/analyze/mcmc` into file `ess_basic` and `rhat_basic`, with corresponding unit tests
- unit tests of `ess_basic` and `rhat_basic` tests against both posterior pkg and old implementations in `src/analyze/mcmc/compute*.hpp`
- tests of new rhat, ess, mcse_mean and mcse_sd moved from `chainset_test.cpp` to unit tests for `src/analyze/mcmc`.
- tests for compatability with posterior pkg estimates use 4 chains of 1000 draws each.
- removed a bunch of test csv files no longer needed

at some point we should deprecate `chains.hpp` and corresponging `src/analyze/mcmc/compute*.hpp`.

src/stan/mcmc/chainset.hpp
```diff
@@ -420,11 +410,12 @@ class chainset {
...
-        || !stan::analyze::is_finite_and_varies(samples(index)))
-      return std::numeric_limits<double>::quiet_NaN();
-    double ess = analyze::ess(samples(index));
-    return sd(index) / std::sqrt(ess);
+    return analyze::mcse_mean(samples(index));
+    // if (num_samples() < 4
+    //     || !stan::analyze::is_finite_and_varies(samples(index)))
+    //   return std::numeric_limits<double>::quiet_NaN();
+    // double ess = analyze::ess(samples(index));
+    // return sd(index) / std::sqrt(ess);
```
WardBrian:
```suggestion

```

src/stan/mcmc/chainset.hpp
```diff
@@ -448,17 +439,18 @@ class chainset {
...
+    //     || !stan::analyze::is_finite_and_varies(samples(index)))
+    //   return std::numeric_limits<double>::quiet_NaN();
+    // Eigen::MatrixXd s = samples(index);
+    // Eigen::MatrixXd s2 = s.array().square();
+    // double ess_s = analyze::ess(s);
+    // double ess_s2 = analyze::ess(s2);
+    // double ess_sd = std::min(ess_s, ess_s2);
+    // return sd(index)
+    //        * std::sqrt(stan::math::e() * std::pow(1 - 1 / ess_sd, ess_sd - 1)
+    //                    - 1);
```
WardBrian:
```suggestion

```

src/stan/analyze/mcmc/ess.hpp
```diff
@@ -0,0 +1,105 @@
...
+
+namespace stan {
+namespace analyze {
+
+/**
+ * Computes the effective sample size (ESS) for the specified
+ * parameter across all chains.  The number of draws per chain must be > 3,
+ * and the values across all draws must be finite and not constant.
+ * The value returned is the minimum of ESS and (sample_sz * log10(sample_sz).
+ * Sample autocovariance is computed using Stan math library implmentation.
```
WardBrian: No longer true?

src/test/unit/mcmc/chainset_test.cpp
```diff
@@ -172,163 +142,70 @@ TEST_F(McmcChains, eight_schools_samples) {
...
+
+  double theta_sd_expect = 0.1215466867;
+  EXPECT_NEAR(theta_sd_expect, bern_chains.sd("theta"), 0.00001);
+
+  double theta_mad_expect = 0.1230906411;
+  EXPECT_NEAR(theta_mad_expect, bern_chains.max_abs_deviation("theta"),
+              0.00001);
+
+  double theta_mcse_mean_expect = 0.0032339916;
+  EXPECT_NEAR(theta_mcse_mean_expect, bern_chains.mcse_mean("theta"), 0.0001);
```
WardBrian: It is still incongruous to put 10 digits in the "expected" answer but only test four digits based on tolerances. 

Adding more precision shows that these two computed values aren't even the same out to 4 digits if you _round_ (expected rounds to .0032, computed rounds to .0033), only if you truncate. And, it still isn't precise enough to differentiate between using the sample variance and incorrectly using population variance

src/test/unit/analyze/mcmc/rhat_basic_test.cpp
```diff
@@ -0,0 +1,59 @@
...
+  Eigen::MatrixXd chains_lp;
+  Eigen::MatrixXd chains_theta;
+  std::vector<const double*> draws_theta;
+  std::vector<const double*> draws_lp;
+  std::vector<size_t> sizes;
+};
+
+TEST_F(RhatBasic, test_basic_rhat) {
+  double rhat_lp_basic_expect = 1.0001296;
+  double rhat_theta_basic_expect = 1.0029197;
```
WardBrian: Could you add a comment where these values came from?

src/test/unit/analyze/mcmc/mcse_test.cpp
```diff
@@ -0,0 +1,73 @@
...
+
+  std::stringstream out;
+  Eigen::MatrixXd chains_lp;
+  Eigen::MatrixXd chains_theta;
+  Eigen::MatrixXd chains_divergent;
+};
+
+TEST_F(MonteCarloStandardError, test_mcse) {
+  double mcse_mean_lp_expect = 0.020164778;
+  double mcse_mean_theta_expect = 0.0032339916;
```
WardBrian: Could you add a comment where these values came from?

src/test/unit/analyze/mcmc/mcse_test.cpp
```diff
@@ -0,0 +1,73 @@
...
+
+  double mcse_sd_lp_expect = 0.0355305;
+  double mcse_sd_theta_expect = 0.0021642137;
+  EXPECT_NEAR(mcse_mean_lp_expect, stan::analyze::mcse_mean(chains_lp), 0.0001);
+  EXPECT_NEAR(mcse_mean_theta_expect, stan::analyze::mcse_mean(chains_theta),
+              0.0001);
```
WardBrian: Insufficient precision

src/test/unit/analyze/mcmc/ess_basic_test.cpp
```diff
@@ -0,0 +1,61 @@
...
+  Eigen::MatrixXd chains_lp;
+  Eigen::MatrixXd chains_theta;
+  std::vector<const double*> draws_theta;
+  std::vector<const double*> draws_lp;
+  std::vector<size_t> sizes;
+};
+
+TEST_F(EssBasic, test_basic_ess) {
+  double ess_lp_expect = 1335.4137;
+  double ess_theta_expect = 1377.503;
```
WardBrian: Could you add a comment where these values came from?

src/stan/analyze/mcmc/ess.hpp
```diff
@@ -0,0 +1,105 @@
...
+#include <cmath>
+#include <vector>
+#include <limits>
+
+namespace stan {
+namespace analyze {
+
+/**
+ * Computes the effective sample size (ESS) for the specified
+ * parameter across all chains.  The number of draws per chain must be > 3,
```
WardBrian, line 16: I didn't realize we now had this _and_ `compute_effective_sample_size` -- could you comment (here and in the code, preferably) on what the differences are?
Should we mark `compute_effective_sample_size` as deprecated?

Ditto for rhat and `compute_potential_scale_reduction`

WardBrian (commented):
Thanks. Test structure is easier to follow now.

I still have some concerns about the level of precision being used in some of the tests, and some general requests for more comments

src/stan/mcmc/chainset.hpp
mitzimorris: good catch - gone

src/stan/mcmc/chainset.hpp
mitzimorris: ditto

src/stan/analyze/mcmc/ess.hpp
mitzimorris: changed.

src/test/unit/mcmc/chainset_test.cpp
mitzimorris: there are only 4000 draws in the sample.  that's a limited amount of resolution.  do you want a sample of 40000 draws?

WardBrian: I want a test to do at least one of two things:

1. Convince me we're calculating the correct quantity by re-creating (to high precision) the result of another package
2. Provide some reasonable certainty for future refactorings that they _continue_ to calculate the correct value

I would say right now it does neither. We aren't getting the same answer as R, nor are we testing tightly enough that we would catch simple bugs during a future code change.

4000 draws results in a limited amount of resolution that is useful for an end user, sure, but this is a deterministic calculation. For a given sample, if we were computing in infinite precision, there would be exactly one answer that properly implementing the algorithm would compute. We're not in infinite precision, of course, but I would expect R and C++ to get much closer than this. If they don't, then I think it is important to understand why so that we can either make them agree, or document why they do not

mitzimorris: > And, it still isn't precise enough to differentiate between using the sample variance and incorrectly using population variance

a review of the code will demonstrate that we are using sample variance, not population variance.

src/test/unit/analyze/mcmc/mcse_test.cpp
mitzimorris: this is as much precision as the data allows.

src/test/unit/analyze/mcmc/mcse_test.cpp
mitzimorris: added everywhere.

src/test/unit/mcmc/chainset_test.cpp
mitzimorris: I have changed the `rhat_basic_test` and `ess_basic_test` so that they check that the previous implementation of rhat and ess provides the same answer to 9 digits of accuracy.

the implementation of `mcse_mean` and `mcse_sd` match the implementations in R's posterior package.  they depend on the implementation of `ess` being correct, as demonstrated by the `ess_basic_test`.

mitzimorris: > I want a test to do at least one of two things:
> 
> 1. Convince me we're calculating the correct quantity by re-creating (to high precision) the result of another package
> 2. Provide some reasonable certainty for future refactorings that they _continue_ to calculate the correct value

see above - the calculations of rhat and ess are the same to 9 digits.

src/stan/analyze/mcmc/ess.hpp
WardBrian: It seems the CI failures are currently because of these warnings up in cmdstan -- probably worth not doing now, but instead circling back after cmdstan PR is merged. I think it's fine to leave in the docstrings

mitzimorris:
The unit tests now test basic ESS and Rhat against the values computed by CmdStan 2.35.0's bin/stansummary and the saved summary CSV file has been checked into the folder src/test/unit/analyze/mcmc/test_csv_files.

I have added unit tests for rank-normalization and splitting the chains - the latter did uncover a bug where splitting a chain with an odd number of draws would omit the last draw, not the (N+1)/2 draw; fixed - but this bug would not have been tickled by any of the existing tests since all CSV test sets had and even number of draws.

src/stan/analyze/mcmc/compute_effective_sample_size.hpp
```diff
@@ -182,8 +199,14 @@ inline double compute_effective_sample_size(std::vector<const double*> draws,
...
  * @return effective sample size for the specified parameter
  */
-inline double compute_split_effective_sample_size(
-    std::vector<const double*> draws, std::vector<size_t> sizes) {
+#if defined(__GNUC__) || defined(__clang__)
+__attribute__((deprecated))
+#elif defined(_MSC_VER)
+__declspec(deprecated)
+#endif
+inline double
```
WardBrian: Sorry, think I was too hasty to suggest warning here - it's making CI unhappy that cmdstan is now hitting these. We should probably do this eventually, but not in this PR.

I think _documenting_ them as deprecated is good to stay here, though.

src/stan/analyze/mcmc/compute_potential_scale_reduction.hpp
```diff
@@ -18,6 +18,8 @@ namespace stan {
 namespace analyze {
 
 /**
+ * DEPRECATED - re-implemented as function `rhat`.
```
WardBrian:
```suggestion
 * \deprecated use `rhat` instead
```
For consistency

src/stan/analyze/mcmc/mcse.hpp
```diff
@@ -0,0 +1,67 @@
...
+ * @return mcse
+ */
+inline double mcse_mean(const Eigen::MatrixXd& chains) {
+  const Eigen::Index num_draws = chains.rows();
+  if (chains.rows() < 4 || !is_finite_and_varies(chains))
+    return std::numeric_limits<double>::quiet_NaN();
+
+  double sd
+      = (chains.array() - chains.mean()).square().sum() / (chains.size() - 1);
+  return std::sqrt(sd / ess(chains));
```
WardBrian: This confused me when comparing to the posterior code, which is `sd / sqrt(ess)`, until I realized this variable is not actually the standard deviation, but the variance, in which case the code is correct. 

Would you mind renaming `sd`?

src/stan/mcmc/chainset.hpp
```diff
@@ -0,0 +1,486 @@
...
+   * @param chain chain index
+   * @param index column index
+   * @return vector of chain autocorrelation at all lags
+   */
+  Eigen::VectorXd autocorrelation(const int chain, const int index) const {
+    Eigen::MatrixXd s = samples(index);
+    Eigen::Map<const Eigen::VectorXd> chain_col(samples(chain).data(),
+                                                num_samples());
+    Eigen::VectorXd autocorr_col(num_samples());
+    stan::math::autocorrelation<double>(s.col(chain), autocorr_col);
```
WardBrian, line 464: Q: If ess and friends are using stan::analyze::autocorrelation, does it matter which one we use here?

I'm not sure if this is actually being called any more, so maybe not?

WardBrian (approved):
A few things I'd like to see before merge but then I think this can fly. Thank you for the diligence!

src/stan/analyze/mcmc/mcse.hpp
mitzimorris: changed to `sample_var`

src/stan/mcmc/chainset.hpp
mitzimorris: this matches what's being done in `chains.hpp`:  https://github.com/stan-dev/stan/blob/2550bbd2bbde86352b9cf9e1bc520968e3a88788/src/stan/mcmc/chains.hpp#L173-L188

PR #3318, opened 2024-11-27
chainset::quantiles function catches exceptions raised by stan::math::quantiles when sample contains NaN's

mitzimorris:
#### Submission Checklist

#### Summary

For samples where one or more draws is NaN, function `stan::math::quantiles` throws a `std::invalid_argument` exception.   Modify `quantile` and `quantiles` functions so that they catch this, as well as the `std::domain_error` exception (thrown if the requested probability is outside of interval [0, 1]) and return `NaN`.

#### Intended Effect

Prevent runtime exceptions being thrown by `stansummary` function.

#### How to Verify

Unit tests.

#### Side Effects

N/A

#### Documentation

N/A

#### Copyright and Licensing

Columbia University


WardBrian:
It seems like the stan csv reader is using istream’s operator>> to parse numbers, which as far as I can tell is not defined for nan or infinite values. I’m guessing the reason we were witnessing something on MacOS only is that the different platforms do different things in this case.

so:
- this change is good, but the test might need to be different
- csv_reader should probably use std::stod, rather than a stringstream and >>?

mitzimorris:
It looks like Ubuntu doesn't throw these exceptions.
Will revise PR accordingly.

WardBrian:
It's because ubuntu is reading the `nan` as 0, so it's not throwing an exception:
https://godbolt.org/z/b6G336qYP

WardBrian:
I suspect changing 
https://github.com/stan-dev/stan/blob/595ae416900093bdaa1309634c927a4814038b49/src/stan/io/stan_csv_reader.hpp#L337

to `samples(row, col) = std::stod(line);` will both make the test here work and fix a whole other set of bugs-lurking-in-the-shadows. Right now if some of the draws are nan, they're just being treated as zero (possibly in a platform-dependent way).

Not that this is performance critical, but I'd also think it would be faster (though there is a lot about the current code in that function that concerns me in _that_ regard)

mitzimorris:
just seeing this now.   will change the CSV reader.

mitzimorris:
> to `samples(row, col) = std::stod(line);` will both make the test here work and fix a whole other set of bugs-lurking-in-the-shadows. Right now if some of the draws are nan, they're just being treated as zero (possibly in a platform-dependent way).

one of the eight schools test files is triggering  a std::out_of_range error from std::stod.   indeed, the value is very close to zero - for some reason, there's an `accept_stat` value:  `6.77055e-309`.

WardBrian:
> from std::stod. indeed, the value is very close to zero

it seems using stold is able to parse that value, and then down casting to a double should work. The csv files were originally produced from double values, so I’m not worried about the down casting being lossy

WardBrian:
If the csv reader changes do work, I slightly preferred the rendition that caught logic_error to the current code that does a check that the quantiles function then repeats internally

src/stan/mcmc/chainset.hpp
```diff
@@ -5,6 +5,7 @@
 #include <stan/math/prim.hpp>
 #include <stan/math/prim/fun/constants.hpp>
 #include <stan/math/prim/fun/quantile.hpp>
+#include <stan/analyze/mcmc/check_chains.hpp>
```
WardBrian: Include not needed

src/stan/io/stan_csv_reader.hpp
```diff
@@ -334,7 +334,12 @@ class stan_csv_reader {
         for (int col = 0; col < cols; col++) {
           std::getline(ls, line, ',');
           boost::trim(line);
-          std::stringstream(line) >> samples(row, col);
+          try {
+            auto raw = std::stold(line);
```
WardBrian: thoughts on doing this in one line and not needing this temporary?

WardBrian (approved):
Two nits but otherwise looks good!
Thanks so much for tracking this down!

mitzimorris:
changes made - I leave it to you to merge.

WardBrian:
Many thanks!

PR #3325, opened 2024-12-13
Feature/reduce mem pathfinder

### LINKED ISSUE #3322: Use a local buffer in unique_writer for better write to disk performance
#### Summary:

Just a note on two optimizations we could do when writing parameters.

#### Description:

When sampling with nuts the [`mcmc_writer`](https://github.com/stan-dev/stan/blob/develop/src/stan/services/util/mcmc_writer.hpp#L98) generates the parameters it needs to write and then sends the parameters as a `std::vector` to a `writer`. For cmdstan the writer is the [`unique_stream_writer`](https://github.com/stan-dev/stan/blob/develop/src/stan/callbacks/unique_stream_writer.hpp#L75) that writes to disk.

We call this function decently often and I think we could do a few things to both speed it up while being a little nicer to the system so we can write faster.

1. `unique_stream_writer` for a `std::vector` calls its [`write_vector`](https://github.com/stan-dev/stan/blob/develop/src/stan/callbacks/unique_stream_writer.hpp#L143) member function. This function takes each value in the vector and one at a time write them to disk. Instead of many small calls for writing to disk we should have a string buffer inside of the `unique_stream_writer` that we write all of the values to first. Then we just write one large buffer to disk once instead of many times.

2. Inside of [`mcmc_writer::write_sample_params`](https://github.com/stan-dev/stan/blob/develop/src/stan/services/util/mcmc_writer.hpp#L100) we keep making the `std::vector<double> values` for each call. Instead we can put those vectors directly in the `mcmc_writer` class. That would make `mcmc_writer` stateful in terms of memory, but to my knowledge we never call an `mcmc_writer` instance from multiple threads and idt it would make sense to. If we did not want to add state to `mcmc_writer` we could also pass `mcmc_writer::write_sample_params` a local allocator to reuse memory. 

These optimizations would not really matter for small problems. But once we start getting to 100K+ parameter models reading and writing to disk can be an issue. 

#### Current Version:
v2.36.0

### LINKED ISSUE #3323: Ideas for saving memory in Pathfinder
#### Summary:
In multi pathfinder with a large number of samples we can end up making two large memory allocations when we only need one. Or perhaps even zero allocations.

We make a large matrix containing all of the samples from the individual pathfinders and an std vector of matrices to store the individual pathfinder samples. Both of these allocations will be `single_pathfinder_samples * num_pathfinders * num_params` in size. This can be huge if you have 100K+ parameters.

I think there are two ways to reduce the memory usage. In short one is to just return the things we need to make samples from the single pathfinders and the other is just to be clever with only making one large matrix.

1. When running single pathfinder from multi pathfinder, we will return everything we need to generate samples (including the rng state) along with the likelihoods. But not actual samples.  When we need to get the draws after doing psis etc. we can adjust the rng to the correct state and pass in the appropriate pathfinder parameters. This will require more computation, but should save a ton of memory. 

2. The second option would be to make one large matrix in multi pathfinder and make a parameter writer for each pathfinder that writes to chunks of the large matrix. So it would look something like the following

1. Make one large matrix `num_pathfinders * single_path_samples * parameters` at the beginning of pathfinder
2. Make a new writer that takes in a map to an Eigen matrix
3. Instead of single pathfinder returning the samples, have them write the samples to the new writer.

The main issue is what to do when a pathfinder fails. My current thought is to loop over the blocks of samples and if a pathfinder failed we move the last pathfinder that succeeded's samples  into that spot. So it would look something like the following

```c++
Eigen::Index total_samples = single_pathfinder_samples * num_pathfinders;
Eigen::Matrix<double, Eigen::Dynamic, Eigen::Dynamic> samples(num_params, total_samples);
std::vector<map_writer> map_writer;
for (int i = 0; i < num_pathfinders; i++) {
  map_writer.emplace_back(...); // setup pointers and map for each path here
}
// Call pathfinder
// ...
// Find bad pathfinders
std::vector<std::pair<bool, Eigen::Index>> pathfinders_start_idx;
for (int i = 0; i < num_pathfinders; i++) {
  // Should just be a pair but for splitting for clarity
  auto [pathfinder_success, individ_sample_start_idx] = check_samples_for_nan(samples, 
    i, single_pathfinder_samples, num_pathfinders);
  pathfinders_start_idx.emplace_back(pathfinder_success, individ_sample_start_idx);
}
// Move good pathfinders to earlier spots
for (auto& path_i :  pathfinders_start_idx) {
    if (!path_i.first) {
      // Find last good pathfinder
      auto last_good_path = std::find_if(pathfinders_start_idx.rbegin(), 
        pathfinders_start_idx.rend(), [](auto& path_j) { return path_i.first}
      move_good_to_bad(samples, path_i, *last_good_path);
      // Copy over last good path
      path_i = *last_good_path;
      // Remove the moved value
      pathfinders_start_idx.erase(last_good_path);
    }
}
// Make a map of only the successful samples
Eigen::Map<Eigen::Array<double, -1, -1>> sample_map(...);
// Do rest of PSIS etc.
```

SteveBronder:
#### Submission Checklist
closes #3323  
closes #3322 

#### Summary

This removes the large memory footprint for writing to csv for pathfinder. So large models should be nicer to the system memory. This may decrease performance a bit, but the memory savings for large models will be significant so I think it is worth it.

When running single pathfinder from pathfinder we now just return the pathfinder at the best elbo estimates instead of generating all of the constrained parameters within single pathfinder. The constrained parameters are now generated on the fly before we send them to the parameter writer. 

For pathfinder that does not do PSIS resampling we write to the parameter writer from within each single pathfinders thread. To avoid contention when writing these we use a `mutex` before doing a bulk write for each pathfinder instance. I thought about putting each pathfinders results to a seperate file and then combining them at the end. But at the end of the day I think this would still have the same problems as multiple writes to a single file. We only have one system and it can only handle so many writes at the same time. So idt the effect of splitting the writes to multiple files would be noticable for saving time. One possible idea would be to have a thread that is busy spinning while checking a multi producer single consumer queue of values to write to the parameter writer. 

#### Intended Effect

Have pathfinder use less memory when saving parameters to the writer

#### How to Verify

All previous tests pass

#### Side Effects

Reduce memory usage

#### Documentation

Added docs for function that generates parameters and writes them.

#### Copyright and Licensing


src/stan/callbacks/unique_stream_writer.hpp
```diff
@@ -77,6 +91,20 @@ class unique_stream_writer final : public writer {
...
+  /**
+   * Writes a set of values in csv format followed by a newline.
+   *
+   * Note: the precision of the output is determined by the settings
+   *  of the stream on construction.
+   *
+   * @param[in] values Values in a std::vector
+   * @param[in, out] ss ignored
+   */
+  void operator()(const std::vector<double>& values, std::stringstream& ss) {
```
WardBrian: I don't think I like these overloads --

1. They're not in the base class `writer`
2. Is there a reason we cant just have the buffer be a private member in the class or something similar?
3. The "Note" in the doc comment is wrong, since we make no attempt to make sure the settings of the stringstream are the same as the ostream. I think, among other things, this would break the way cmdstan sets the number of digits?

Separately, isn't `std::ostream` buffered internally anyway?

WardBrian (changes requested):

src/stan/callbacks/unique_stream_writer.hpp
SteveBronder: > Is there a reason we cant just have the buffer be a private member in the class or something similar?

I didn't want to introduce state directly in the class

> The "Note" in the doc comment is wrong, since we make no attempt to make sure the settings of the stringstream are the same as the ostream. I think, among other things, this would break the way cmdstan sets the number of digits?

I totally missed this. Yeah my through was that having a little buffer we reuse would save some memory, but idt it's worth the overhead of config etc. And we only use it for the case of a std vector really

src/stan/services/pathfinder/multi.hpp
```diff
@@ -22,6 +24,26 @@ namespace stan {
...
+  template <typename T>
+  void operator()(T&& t) {
+    std::lock_guard<std::mutex> lock(mut_.get());
+    writer.get()(t);
+  }
+  void operator()() {
+    std::lock_guard<std::mutex> lock(mut_.get());
+    writer.get()();
+  }
+};
```
SteveBronder: @WardBrian do you think this should be in it's own file or is it fine to just put this here?

The other alternative to this writer is to make a writer on a seperate thread that busy waits to watche an mpsc queue.

WardBrian: If it's not for general use it's probably fine to leave it here.

I haven't been following this PR _that_ closely, though. Do you have numbers on this versus a queue?

SteveBronder: I have a simple benchmark I'm running here

https://github.com/SteveBronder/stan-perf/tree/parallel-writer

Preliminary results seem like mpsc queue on a seperate thread is a good thing. The plot is faceted by the number of threads and the lines are for mpsc vs. mutex and the amount of "work" each main thread has to do between sending results to the writer thread. "0" means no additional work is done and the working thread just generates a random vector and then pushes to writer. The "16" means the work thread is doing a loop that multiplies that random vector with itself 16 times. One thing that is nice about the mpsc queue is that since we are pushing the write to disk off to another thread the work thread can do more work while instead of waiting for the write to disk.

image

SteveBronder:
Okay running it a bit more it looks like we definitely want the concurrent writer. The below is for the time to write to disk for mpsc vs mutex. The x axis is on the log scale but the y axis is not. The code is on the same branch as the last comment. Even with a small amount of loops there's a pretty huge difference in timing.

image

tbh I wonder if we should even do this concurrent writer for nuts with parallel chains. Even though we are writing to different files in that case, we only really have one disk still that we are writing to. So just pushing the operations for converting the vectors to char and then writing them to disk seems pretty useful.

SteveBronder:
I ran cmdstan with the [`normal_glm.stan`](https://github.com/stan-dev/stan/blob/develop/src/test/test-models/good/services/normal_glm.stan) model to get a ballpark of memory this saves versus the current implementation

```
## Develop Branch
	Command being timed: "./examples/normal_glm pathfinder num_psis_draws=20000 num_draws=20000 data file=./examples/normal_glm_test.json"
	User time (seconds): 0.96
	System time (seconds): 0.01
	Percent of CPU this job got: 99%
	Elapsed (wall clock) time (h:mm:ss or m:ss): 0:00.97
	Average shared text size (kbytes): 0
	Average unshared data size (kbytes): 0
	Average stack size (kbytes): 0
	Average total size (kbytes): 0
	Maximum resident set size (kbytes): 25168
	Average resident set size (kbytes): 0
	Major (requiring I/O) page faults: 0
	Minor (reclaiming a frame) page faults: 5378
	Voluntary context switches: 1
	Involuntary context switches: 2
	Swaps: 0
	File system inputs: 0
	File system outputs: 3808
	Socket messages sent: 0
	Socket messages received: 0
	Signals delivered: 0
	Page size (bytes): 4096
	Exit status: 0  
## This Branch
  Command being timed: "./examples/normal_glm pathfinder num_psis_draws=20000 num_draws=20000 data file=./examples/normal_glm_test.json"
	User time (seconds): 1.02
	System time (seconds): 0.00
	Percent of CPU this job got: 99%
	Elapsed (wall clock) time (h:mm:ss or m:ss): 0:01.04
	Average shared text size (kbytes): 0
	Average unshared data size (kbytes): 0
	Average stack size (kbytes): 0
	Average total size (kbytes): 0
	Maximum resident set size (kbytes): 18096
	Average resident set size (kbytes): 0
	Major (requiring I/O) page faults: 0
	Minor (reclaiming a frame) page faults: 4207
	Voluntary context switches: 1
	Involuntary context switches: 3
	Swaps: 0
	File system inputs: 0
	File system outputs: 3536
	Socket messages sent: 0
	Socket messages received: 0
	Signals delivered: 0
	Page size (bytes): 4096
	Exit status: 0
```

"Maximum resident set size" tells us the amount of memory used at peak during the program. So about 25mb currently and 18mb with this branch. Seems good!

lib/stan_math
WardBrian, line 1: Is this intentionally touching the submodule?

src/stan/callbacks/writer.hpp
```diff
@@ -46,6 +46,12 @@ class writer {
    */
   virtual void operator()(const std::string& message) {}
 
+  /**
+   * Checks if stream is valid.
+   */
+  virtual bool is_nonnull() const noexcept { return false; }
+  virtual const char* comment_prefix() const noexcept { return "# "; }
```
WardBrian: I don't like making this part of the writer API, especially if we ever hope to have non-csv-based IO that may not have a prefixed-comment style

What problem is it solving?

src/stan/callbacks/writer.hpp
```diff
@@ -46,6 +46,12 @@ class writer {
    */
   virtual void operator()(const std::string& message) {}
 
+  /**
+   * Checks if stream is valid.
+   */
+  virtual bool is_nonnull() const noexcept { return false; }
```
WardBrian: I would rather name this something like `is_valid` or `is_writable`

src/stan/callbacks/unique_stream_writer.hpp
```diff
@@ -114,6 +114,15 @@ class unique_stream_writer final : public writer {
     *output_ << comment_prefix_ << message << std::endl;
   }
 
+  /**
+   * Checks if stream is valid.
+   */
+  bool is_nonnull() const noexcept { return output_ != nullptr; }
```
WardBrian: [optional] Almost all the overloads in these files start with an if, that could be re-written to use this helper

src/stan/services/pathfinder/multi.hpp
```diff
@@ -105,136 +109,214 @@ inline int pathfinder_lbfgs_multi(
...
+                return;
+              }
+              lp_calls += std::get<2>(pathfinder_ret);
+              elbo_estimates.push_back(
+                  std::make_pair(iter, std::move(std::get<1>(pathfinder_ret))));
+            } else {
+              // For no psis, have single write to both single and multi writers
+              using multi_writer_t = stan::callbacks::multi_writer<
+                  SingleParamWriter,
+                  stan::callbacks::concurrent_writer<ParamWriter>>;
```
WardBrian: Q: wouldn't the concurrent_writer being wrapped around the multi_writer be what we really want here?

src/stan/services/pathfinder/multi.hpp
```diff
@@ -105,136 +109,214 @@ inline int pathfinder_lbfgs_multi(
...
-                    tol_grad, tol_rel_grad, tol_param, num_iterations,
-                    num_elbo_draws, num_draws, save_iterations, refresh,
-                    interrupt, logger, init_writers[iter],
-                    single_path_parameter_writer[iter],
-                    single_path_diagnostic_writer[iter], calculate_lp);
-            if (unlikely(std::get<0>(pathfinder_ret) != error_codes::OK)) {
-              logger.error(std::string("Pathfinder iteration: ")
-                           + std::to_string(iter) + " failed.");
-              return;
+            if (psis_resample && calculate_lp) {
```
WardBrian: Thoughts on hoisting this outside the parallel-for and having two separate ones? It probably doesn't matter for speed, but I think it could help readability

src/stan/services/pathfinder/multi.hpp
```diff
@@ -105,136 +109,214 @@ inline int pathfinder_lbfgs_multi(
...
+                return;
+              }
+              lp_calls += pathfinder_ret.second;
             }
-            individual_lp_ratios[iter] = std::move(std::get<1>(pathfinder_ret));
-            individual_samples[iter] = std::move(std::get<2>(pathfinder_ret));
-            lp_calls += std::get<3>(pathfinder_ret);
           }
         });
+    safe_write.wait();
```
WardBrian: Q: Rather than needing to call this, should the concurrent_writer type have a destructor that blocks?

src/stan/services/pathfinder/single.hpp
```diff
@@ -459,19 +460,18 @@ inline taylor_approx_t taylor_approximation(
...
-  return return_code;
+template <bool ReturnLpSamples, typename EigVec, typename ParamWriter>
+inline auto ret_pathfinder(int return_code, EigVec&& elbo_est,
+                           const std::atomic<size_t>& lp_calls,
+                           ParamWriter&& /* params */) noexcept {
+  if constexpr (ReturnLpSamples) {
+    return std::make_tuple(return_code, std::forward<EigVec>(elbo_est),
+                           lp_calls.load());
+  } else if constexpr (stan::callbacks::is_multi_writer_v<ParamWriter>) {
+    return std::pair(return_code, lp_calls.load());
```
WardBrian: The docstring of this function needs to be updated with the new template parameters and I think this needs to be explained -- is there a reason to not give this a name as another bool tparam?

src/stan/services/pathfinder/multi.hpp
```diff
@@ -105,136 +109,214 @@ inline int pathfinder_lbfgs_multi(
...
-
-  // if any pathfinders failed, we want to remove their empty results
-  individual_lp_ratios.erase(
-      std::remove_if(individual_lp_ratios.begin(), individual_lp_ratios.end(),
-                     [](const auto& v) { return v.size() == 0; }),
-      individual_lp_ratios.end());
-  individual_samples.erase(
-      std::remove_if(individual_samples.begin(), individual_samples.end(),
-                     [](const auto& v) { return v.size() == 0; }),
-      individual_samples.end());
```
WardBrian, line 156: Is this a behavior change or am I just missing where this idea lives now?

src/stan/callbacks/concurrent_writer.hpp
```diff
@@ -0,0 +1,121 @@
...
+      static_assert(1,
+                    "Unsupported type passed to concurrent_writer. "
+                    "This is an internal error. Please file an issue on the "
+                    "stan github repo.");
+    }
+  }
+  /**
+   * Writes a comment prefix to the writer.
+   */
+  void operator()() { str_messages_.push(writer.get().comment_prefix()); }
```
WardBrian: Oh I see, the `comment_prefix` exists so that this empty overload is implementable.

I think it would be better to keep that detail inside this class and have it contain something like an `std::atomic<int> null_writes_queued` which is incremented in this and decremented by the writer thread.


It also seems like this overload would primarily exist to control the formatting of the result, and anything calling this in a threaded context is likely gonna not get what it wants. Have you investigated why this overload gets called by pathfinder and what it is hoping to achieve?

WardBrian (changes requested):
This looks pretty good! 

A few minor comments and then some questions for my own understanding. 

The basic memory profile you posted above looks encouraging, have you done any more extensive timing benchmarks (including when STAN_THREADS is not set)? As long as this doesn't majorly slow down anything I think it's fine, but I'd hope to see some nice speed ups too

src/stan/callbacks/concurrent_writer.hpp
SteveBronder: Removed comment prefix function. Yeah it was just for this function and I like `null_writes_queued` much more

src/stan/services/pathfinder/multi.hpp
SteveBronder: Yeah I actually moved around a bunch of the if statements so things are a little more encapsulated. The loops are seperated and once we hit an if in multi there is a return early pattern which I think is easier to read

src/stan/services/pathfinder/multi.hpp
SteveBronder: ~I think it can be either/or but I like flipping them so more work is put off on the other thread~
No I did not notice at first. We need the concurrent writer made outside of the loop and the multi writer made inside of the loop.

src/stan/services/pathfinder/multi.hpp
SteveBronder: I like being explicit about it but I can also just make a destructor

src/stan/services/pathfinder/multi.hpp
SteveBronder: Yes so now we use a concurrent vector to hold these individual samples and do a `push_back()`. So there will never be a zero sized sample to erase.

src/stan/services/pathfinder/single.hpp
SteveBronder: I think I need to clean this up. ReturnLpSamples will only be true if we are not doing psis.

src/stan/services/pathfinder/multi.hpp
```diff
@@ -105,126 +108,132 @@ inline int pathfinder_lbfgs_multi(
...
   std::atomic<size_t> lp_calls{0};
+  // Save idx of pathfinder and it's elbo for resampling later
+  tbb::concurrent_vector<std::pair<Eigen::Index, internal::elbo_est_t>>
+      elbo_estimates;
+  elbo_estimates.reserve(num_paths + 10);
+  auto constrain_fun = [](auto&& constrained_draws, auto&& unconstrained_draws,
+                          auto&& model, auto&& rng) {
+    model.write_array(rng, unconstrained_draws, constrained_draws);
+    return constrained_draws;
+  };
```
WardBrian: I think these definitions can be moved below the special case

src/stan/services/pathfinder/multi.hpp
```diff
@@ -105,126 +108,132 @@ inline int pathfinder_lbfgs_multi(
...
+      start_pathfinders_time, std::chrono::steady_clock::now());
   const auto start_psis_time = std::chrono::steady_clock::now();
-  const size_t successful_pathfinders = individual_samples.size();
-  if (successful_pathfinders == 0) {
-    logger.info("No pathfinders ran successfully");
-    return error_codes::SOFTWARE;
-  }
-  if (refresh != 0) {
-    logger.info("Total log probability function evaluations:"
-                + std::to_string(lp_calls));
```
WardBrian, line 170: We're no longer printing this anywhere that I can see -- which might be fine, but in which case we can remove the `lp_calls` variable

src/stan/services/pathfinder/multi.hpp
```diff
@@ -105,126 +108,132 @@ inline int pathfinder_lbfgs_multi(
...
+      logger.error(e.what());
+      return error_codes::SOFTWARE;
+    }
+    // Writes are done in loop, so just return
+    return error_codes::OK;
+  }
   try {
     tbb::parallel_for(
         tbb::blocked_range<int>(0, num_paths), [&](tbb::blocked_range<int> r) {
+          auto toss_write = [](auto&&... /* x */) {};
```
WardBrian: I assume this has some performance advantage over just constructing a null writer, but I recommend using a name like `non_writer` and adding a comment that this gets written later in a separate loop

src/stan/services/pathfinder/single.hpp
```diff
@@ -857,111 +845,133 @@ inline auto pathfinder_lbfgs_single(
...
-      unconstrained_col = elbo_draws.col(i);
-      constrained_draws_mat.col(i).tail(num_unconstrained_params)
-          = constrain_fun(rng, unconstrained_col,
-                          approx_samples_constrained_col)
-                .matrix();
+    const auto end_pathfinder_time = std::chrono::steady_clock::now();
+    const double pathfinder_delta_time = stan::services::util::duration_diff(
+        start_pathfinder_time, end_pathfinder_time);
+    // For multi pathfinder, multi would write multiple end times
+    if constexpr (stan::callbacks::is_tee_writer_v<ParamWriter>) {
```
WardBrian: I'd prefer depending on a named tparam (even if it's the same as the ReturnLP one, which I think it is?)

src/stan/services/pathfinder/multi.hpp
SteveBronder: deleted

src/stan/services/pathfinder/single.hpp
SteveBronder: I changed this to depend on ReturnLP and added a static check for this section that we are working with a tee writer

SteveBronder:
Once https://github.com/stan-dev/cmdstan/pull/1311 is merged I'll kickoff the tests again and then this should be good to go

src/stan/callbacks/concurrent_writer.hpp
```diff
@@ -0,0 +1,133 @@
...
+      } else {
+        vec_str_messages_.push(t);
+      }
+    } else if constexpr (std::is_same_v<T, std::string>) {
+      str_messages_.push(std::forward<T>(t));
+    } else if constexpr (stan::is_eigen_vector<T>::value) {
+      eigen_messages_.push(std::forward<T>(t));
+    } else {
+      static_assert(
+          1,
```
WardBrian: Should this be 0?

src/stan/callbacks/stream_writer.hpp
```diff
@@ -68,7 +68,55 @@ class stream_writer : public writer {
...
+  }
+  /**
+   * Write a row of values in csv format.
+   *
+   * Note: the precision of the output is determined by the settings
+   *  of the stream on construction.
+   *
+   * @param[in] values A column vector of values.
+   */
+  virtual void operator()(const Eigen::Matrix<double, -1, 1>& values) {
```
WardBrian:
```suggestion
  void operator()(const Eigen::Matrix<double, -1, 1>& values) {
```

(I think? at least these three overloads should be consistent -- currently one doesn't have this and two do)

src/stan/callbacks/unique_stream_writer.hpp
```diff
@@ -88,12 +88,42 @@ class unique_stream_writer final : public writer {
...
+  }
+  /**
+   * Write a row of values in csv format.
+   *
+   * Note: the precision of the output is determined by the settings
+   *  of the stream on construction.
+   *
+   * @param[in] values A column vector of values.
+   */
+  virtual void operator()(const Eigen::Matrix<double, -1, 1>& values) {
```
WardBrian: Same q as above about `virtual`

src/stan/services/pathfinder/single.hpp
```diff
@@ -856,112 +875,140 @@ inline auto pathfinder_lbfgs_single(
...
-           ++i, ++j) {
-        constrained_draws_mat.col(i).head(2) = lp_draws.row(j).matrix();
-        unconstrained_col = new_draws.col(j);
-        constrained_draws_mat.col(i).tail(num_unconstrained_params)
-            = constrain_fun(rng, unconstrained_col,
-                            approx_samples_constrained_col)
-                  .matrix();
+  if constexpr (ReturnLpSamples) {
+    internal::elbo_est_t est_draws = internal::est_approx_draws<false>(
+        lp_fun, constrain_fun, rng, taylor_approx_best, num_draws,
```
WardBrian: I don't think the est_approx_draws function needs the constrain_fun argument any more? In fact, I don't think `pathfinder_impl` does either?

src/stan/services/pathfinder/single.hpp
```diff
@@ -856,112 +875,140 @@ inline auto pathfinder_lbfgs_single(
...
+          unconstrained_col = elbo_draws.col(i);
+          constrained_draws_vec.tail(num_unconstrained_params).transpose()
+              = constrain_fun(rng, unconstrained_col,
+                              approx_samples_constrained_col)
+                    .matrix();
+          parameter_writer(constrained_draws_vec);
+        }
+        for (Eigen::Index i = elbo_draws.cols(), j = 0; i < total_size;
+             ++i, ++j) {
+          constrained_draws_vec.head(2) = lp_draws.row(j);
```
WardBrian, line 932: Missing the `pathfinder__` column? (also from line 967 below)

src/stan/services/pathfinder/multi.hpp
```diff
@@ -104,138 +107,200 @@ inline int pathfinder_lbfgs_multi(
     std::vector<SingleDiagnosticWriter>& single_path_diagnostic_writer,
     ParamWriter& parameter_writer, DiagnosticWriter& diagnostic_writer,
     bool calculate_lp = true, bool psis_resample = true) {
+  using stan::services::pathfinder::internal::write_times;
   const auto start_pathfinders_time = std::chrono::steady_clock::now();
   std::vector<std::string> param_names;
   param_names.push_back("lp_approx__");
   param_names.push_back("lp__");
+  param_names.push_back("pathfinder__");
```
WardBrian: I think this should be `path__` or `path_num__`

Also, I think it is missing from some of the outputs depending on which branch you end up following here

src/stan/services/pathfinder/multi.hpp
```diff
@@ -104,138 +107,200 @@ inline int pathfinder_lbfgs_multi(
...
-  const double pathfinders_delta_time = stan::services::util::duration_diff(
-      start_pathfinders_time, end_pathfinders_time);
+  double pathfinders_delta_time = stan::services::util::duration_diff(
+      start_pathfinders_time, std::chrono::steady_clock::now());
   const auto start_psis_time = std::chrono::steady_clock::now();
-  const size_t successful_pathfinders = individual_samples.size();
-  if (successful_pathfinders == 0) {
-    logger.info("No pathfinders ran successfully");
-    return error_codes::SOFTWARE;
-  }
```
WardBrian, line 167: Are we no longer indicating an error if none of the individual paths succeed? That seems bad (question also applies to the special-case parallel for loop above)

WardBrian (commented):
Still trying to figure out some issues I'm seeing upstream, but these stood out to me:

src/stan/services/pathfinder/multi.hpp
SteveBronder: Adding that back. Also adding a warning if only N of M paths succeed

src/stan/services/pathfinder/multi.hpp
SteveBronder: I'm fine with that. For single I missed a branch

WardBrian:
Latest push seems to be passing all my tests in tinystan

SteveBronder:
Odd, cmdstan tests are passing locally for me. Going to run this one more time and then will investigate if that same failure pops up again

SteveBronder:
Also odd, this test is running locally for me even with threading on

```
[ RUN      ] CmdStan.pathfinder_40_draws
src/test/interface/pathfinder_test.cpp:88: Failure
Expected equality of these values:
  1
  count_matches(" seconds (PSIS)", output)
    Which is: 0
[  FAILED  ] CmdStan.pathfinder_40_draws (260 ms)
```

SteveBronder:
@serban-nicusor-toptal it looks like Jenkins is hanging when trying to spin up the windows unit job. Can you took a look?

Edit: nvm!

src/stan/callbacks/concurrent_writer.hpp
```diff
@@ -0,0 +1,212 @@
...
+  // The writing thread
+  std::thread thread_;
+  // Condition variable to signal the writing thread to continue
+  std::condition_variable cv;
+  // Maximum number of threads that can be in use
+  std::size_t max_threads{tbb::global_control::max_allowed_parallelism};
+  // Max capacity of queue
+  std::size_t max_capacity{1000 + max_threads};
+  // Threshold where the writing threads will wait for the queues to empty
+  std::size_t wait_threshold{max_capacity - (max_threads * 2)};
```
WardBrian: Won't this trigger UB if `max_threads` is over 1000? I realize this is unlikely

src/stan/callbacks/concurrent_writer.hpp
```diff
@@ -0,0 +1,212 @@
...
+template <typename Writer>
+struct concurrent_writer {
+  // A reference to the writer to write to
+  std::reference_wrapper<Writer> writer;
+  // Number of null writes queued
+  std::atomic<int> null_writes_queued{0};
+  // Queue for string messages
+  tbb::concurrent_bounded_queue<std::string> str_messages_{};
+  // Queue for vector of strings messages
+  tbb::concurrent_bounded_queue<std::vector<std::string>> vec_str_messages_{};
```
WardBrian: Thinking about this more -- any non-numeric content would be erroneous to write in a concurrent context, and should be written to the underlying writer to ensure ordering. IMO, we should remove this functionality for clarity

src/stan/callbacks/concurrent_writer.hpp
```diff
@@ -63,7 +57,8 @@ struct concurrent_writer {
   // Max capacity of queue
   std::size_t max_capacity{1000 + max_threads};
   // Threshold where the writing threads will wait for the queues to empty
-  std::size_t wait_threshold{max_capacity - (max_threads * 2)};
+  std::size_t wait_threshold{
+      (max_threads > 1000) ? 0 : (max_capacity - (max_threads * 2))};
```
WardBrian: I think we actually need this value to always be at least one, right? Otherwise we won't be able to send any messages at all

src/stan/callbacks/concurrent_writer.hpp
```diff
@@ -0,0 +1,168 @@
...
+#include <condition_variable>
+#include <functional>
+#include <string>
+#include <thread>
+#include <vector>
+
+namespace stan::callbacks {
+#ifdef STAN_THREADS
+/**
+ * Takes a writer and makes it thread safe via multiple queues.
```
WardBrian: optional:
```suggestion
 * Enables thread-safe writing of numeric values to a writer
```
I think this comment is when the class was trying to be more general

src/stan/services/pathfinder/multi.hpp
```diff
@@ -104,138 +107,258 @@ inline int pathfinder_lbfgs_multi(
...
+  }
+  // Save idx of pathfinder and it's elbo for resampling later
+  tbb::concurrent_vector<std::pair<Eigen::Index, internal::elbo_est_t>>
+      elbo_estimates;
+  elbo_estimates.reserve(num_paths + 10);
+  auto constrain_fun = [](auto&& constrained_draws, auto&& unconstrained_draws,
+                          auto&& model, auto&& rng) {
+    model.write_array(rng, unconstrained_draws, constrained_draws);
+    return constrained_draws;
+  };
```
WardBrian: optional: this definition could be moved way down (I think to ~line 268), which would help with clarity a little IMO

src/stan/services/pathfinder/multi.hpp
```diff
@@ -104,138 +107,258 @@ inline int pathfinder_lbfgs_multi(
...
+    double pathfinders_delta_time = stan::services::util::duration_diff(
+        start_pathfinders_time, std::chrono::steady_clock::now());
+    write_times<true, false>(parameter_writer, pathfinders_delta_time, 0);
+    // Writes are done in loop, so just return
+    return error_codes::OK;
+  }
+  // Save idx of pathfinder and it's elbo for resampling later
+  tbb::concurrent_vector<std::pair<Eigen::Index, internal::elbo_est_t>>
+      elbo_estimates;
+  elbo_estimates.reserve(num_paths + 10);
```
WardBrian: I don't think this `+ 10` is necessary

src/stan/services/pathfinder/multi.hpp
```diff
@@ -104,138 +107,258 @@ inline int pathfinder_lbfgs_multi(
...
           }
         });
+    if (unlikely(elbo_estimates.empty())) {
+      logger.error("No pathfinders ran successfully.");
+      return error_codes::SOFTWARE;
+    } else if (unlikely(elbo_estimates.size() < num_paths)) {
+      std::string msg = std::string("Only ")
+                        + std::to_string(elbo_estimates.size())
+                        + std::string(" of the ") + std::to_string(num_paths)
+                        + std::string(" pathfinders succeeded.");
```
WardBrian, line 206: Genuine question: I think a lot of the rest of the Stan codebase does this kind of formatting using a stringstream and `<<` -- does it matter?

src/stan/services/pathfinder/single.hpp
```diff
@@ -507,11 +499,11 @@ inline auto ret_pathfinder(int return_code, EigVec&& lp_ratio, EigMat&& samples,
...
  * approximation information.
  */
-template <typename RNG, typename LPFun, typename ConstrainFun,
-          typename AlphaVec, typename CurrentParams, typename CurrentGrads,
-          typename GradMat, typename ParamMat, typename Logger>
-auto pathfinder_impl(RNG&& rng, LPFun&& lp_fun, ConstrainFun&& constrain_fun,
-                     AlphaVec&& alpha, CurrentParams&& current_params,
+template <typename RNG, typename LPFun, typename AlphaVec,
+          typename CurrentParams, typename CurrentGrads, typename GradMat,
+          typename ParamMat, typename Logger>
```
WardBrian, line 490: The docstring of this function needs some cleanup of the `@tparam`s

src/stan/services/pathfinder/single.hpp
```diff
@@ -856,112 +876,142 @@ inline auto pathfinder_lbfgs_single(
...
+    names.push_back("lp__");
+    names.push_back("path__");
+    model.constrained_param_names(names, true, true);
+    if constexpr (InMultiPathfinder) {
+      static_assert(stan::callbacks::is_tee_writer_v<ParamWriter>,
+                    "ReturnLpSamples is false but the parameter_writer is not "
+                    "a tee_writer! "
+                    "Multi pathfinder assumes we use a tee writer here, if you "
+                    "intend to change this "
+                    "please make it clear why.");
```
WardBrian: Optional: Should this have an "if you are a user please report a bug" type ending?

src/test/unit/services/pathfinder/normal_glm_test.cpp
WardBrian, line 1: Noting, but optional: Might be nice to add a github actions job which runs some tests with `-DSTAN_THREADS` and a non-1 number of threads. We can make this and issue rather than do it here, IMO

cc @serban-nicusor-toptal

WardBrian (commented):
I am hopeful these will be my final comments before approving. It's looking real good at this point

src/stan/services/pathfinder/multi.hpp
SteveBronder: Not really and I just like this personally

src/stan/services/pathfinder/single.hpp
SteveBronder: I fixed it in the `@return`  but forgot to put it in the actual tparam. Fixed

src/stan/services/pathfinder/single.hpp
SteveBronder: No idt so. This assert can only happen at compile time so someone would have to change the C++

src/stan/services/pathfinder/single.hpp
WardBrian: @SteveBronder Line 473 just needs to be deleted

WardBrian (approved):

PR #3334, opened 2025-03-04
Add dispatcher object to callbacks to coordinate sampler outputs

mitzimorris:
#### Submission Checklist

#### Summary

Implement proposals in design-doc 32:  https://github.com/stan-dev/design-docs/blob/master/designs/0032-stan-output-formats.md, starting by adding a dispatcher class to the stan::callbacks namespace.

#### Intended Effect

* Simplify current stan::services APIs by replacing multiple output file arguments with a single dispatcher arg
* Allow for more different outputs.

#### How to Verify

unit tests

#### Side Effects

??

#### Copyright and Licensing

Columbia University


mitzimorris:
@SteveBronder - this is a WIP w/ help from Claude 3.7 sonnet, which doesn't seem to be very good.  if you have time, would like to discuss how it's got the output_controller class set up.

SteveBronder:
For this I think we would want to write these as C++ bindings to python. We want the R/Python/Julia/etc session to make a pointer that we than pass to a writer. I think tinystan has an impl of this

https://github.com/WardBrian/tinystan/blob/main/src/buffer.hpp#L30

mitzimorris:
clang auto-formatting is introducing errors for the linter.
cf this:

https://github.com/stan-dev/stan/pull/3334/commits/cef5f2be53de19eba9cc5b9761e9c38fbb30c1f6

mitzimorris:
here is a chat I've been having with Claude 3.7 about how the in-memory sample could be accessed by R and Python.
https://claude.ai/share/a4907ccd-7665-40bf-baa0-a240606b6bce

WardBrian:
Is there a design goal you have in mind for that which isn't answered by something like the existing code in tinystan?

In a bigger picture sense, I'm not sure that this would make sense to live in the Stan repo as opposed to a consumer (like tinystan or a future cmdstan-derived thing). I think focusing on what the API would look like (e.g. the router class, new sets of base classes, or improved services) is more important than on any specific instantiation of it (e.g. something that derives from one of the new base classes and stores its results in a buffer) at this point

--- 
Asides on Claude:

> Own its memory: The buffer should be allocated and managed by Stan

I can't see the files you provided as input to the model, but I'm guessing this is some AI-"yes-man" behavior where it doesn't seek to criticize. This is in fact a very difficult design to make work compared to the allocation happening externally. It is a general requirement that allocations and deallocations happen in the same library/module (especially if you want any hope of working on Windows, where we don't compile against the same libc as Python would), which also means that handing long-lived pointers between modules is fraught in general. Especially when one side is garbage collected, it is usually infinitely easier to have that side be in charge of memory that needs to be shared between them, otherwise you will end up with use-after-frees, double-frees, or lots of additional copy operations. 

One way to see this is to think about RAII and where the eigen matrix you're storing the draws in would get freed by C++. If you wrote wrapper calls to `services` like cmdstan does that create these in function scope, this will be dropped at return time. So you either need to have Python manage a pointer to the entire writer, or return a copy that needs to be separately freed later. Note that placing that copy into a numpy array usually requires _another copy_ to avoid numpy trying to free it on the wrong side of the module boundary...

This is why tinystan operates the way that it does. RStan lets R manage the memory (it gets this more or less 'for free' by using Rcpp) and Pystan writes back into python using a socket (Pystan2 was much more similar to RStan)

> Languages like R/Python often want column-oriented access

This is a more minor point, but this is false. Python is largely agnostic but actually defaults to row-major.

mitzimorris:
Closing this PR.
will reopen on renamed branch.

PR #3342, opened 2025-03-25
Deserializer for sum_to_zero matrix

WardBrian:
#### Submission Checklist

#### Summary

De/serialization to accompany https://github.com/stan-dev/math/pull/3169

#### Copyright and Licensing

Simons Foundation

### DIFF
--- a/src/stan/io/deserializer.hpp
+++ b/src/stan/io/deserializer.hpp
@@ -525,8 +525,8 @@ class deserializer {
 
   /**
    * Return the next simplex of the specified size (using one fewer
-   * unconstrained scalars). This particular transformation is linear, not
-   * requiring a Jacobian adjustment.
+   * unconstrained scalars), incrementing the specified reference with the
+   * log absolute Jacobian determinant.
    *
    * <p>See <code>stan::math::simplex_constrain(Eigen::Matrix,T&)</code>.
    *
@@ -550,8 +550,8 @@ class deserializer {
 
   /**
    * Return the next simplex of the specified size (using one fewer
-   * unconstrained scalars). This particular transformation is linear, not
-   * requiring a Jacobian adjustment.
+   * unconstrained scalars), incrementing the specified reference with the
+   * log absolute Jacobian determinant.
    *
    * <p>See <code>stan::math::simplex_constrain(Eigen::Matrix,T&)</code>.
    *
@@ -634,8 +634,8 @@ class deserializer {
 
   /**
    * Return the next zero sum vector of the specified size (using one fewer
-   * unconstrained scalars), incrementing the specified reference with the
-   * log absolute Jacobian determinant (no adjustment, in this case).
+   * unconstrained scalars). This particular transformation is linear, not
+   * requiring a Jacobian adjustment.
    *
    * <p>See <code>stan::math::sum_to_zero_constrain(Eigen::Matrix,T&)</code>.
    *
--- a/src/stan/io/deserializer.hpp
+++ b/src/stan/io/deserializer.hpp
@@ -592,11 +592,10 @@ class deserializer {
    * @tparam Jacobian Whether to increment the log of the absolute Jacobian
    * determinant of the transform.
    * @tparam LP Type of log probability.
-   * @tparam Sizes A parameter pack of integral types.
    * @param lp The reference to the variable holding the log
    * @param size Number of cells in zero sum vector to generate.
    * @return The next zero sum of the specified size.
-   * @throws std::invalid_argument if number of dimensions (`k`) is zero
+   * @throws std::invalid_argument if `size` is zero
    */
   template <typename Ret, bool Jacobian, typename LP,
             require_not_std_vector_t<Ret>* = nullptr>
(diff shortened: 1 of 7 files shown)


src/stan/io/deserializer.hpp
```diff
@@ -606,6 +605,32 @@ class deserializer {
         this->read<Ret>(size - 1), lp);
   }
 
+  /**
+   * Return the next zero sum matrix of the specified dimensions (requires N-1 *
+   * M-1 unconstrained scalars), incrementing the specified reference with the
```
bob-carpenter: N-1 * M-1 -> `(N - 1) * (M - 1)`

bob-carpenter: Can you just note that it doesn't require a Jacobian adjustment rather than saying it "increments the specified reference"?

src/stan/io/deserializer.hpp
```diff
@@ -606,6 +605,32 @@ class deserializer {
...
+   * log absolute Jacobian determinant (no adjustment, in this case).
+   *
+   * <p>See <code>stan::math::sum_to_zero_constrain(Eigen::Matrix,T&)</code>.
```
bob-carpenter: Space this out like other code Matrix,T -> Matrix, T

src/stan/io/deserializer.hpp
```diff
@@ -606,6 +605,32 @@ class deserializer {
...
+   *
+   * @tparam Ret The type to return.
+   * @tparam Jacobian Whether to increment the log of the absolute Jacobian
+   * determinant of the transform.
+   * @tparam LP Type of log probability.
+   * @param lp The reference to the variable holding the log
```
bob-carpenter: holding the log -> holding the log density

This is tricky terminologically.  A "log probability function" can be a pdf, in which case it returns a density (units of change in probability) as opposed to a probability, or a pdf, in which case it does return probabilities.  But what we don't have in Stan is `target` holding a log probability---it's always a density.  Our doc is a mess around all this.

src/stan/io/deserializer.hpp
```diff
@@ -606,6 +605,32 @@ class deserializer {
...
+   * @param N Number of rows in zero sum matrix to generate.
+   * @param M Number of columns in zero sum matrix to generate.
+   * @return The next zero sum of the specified size.
+   * @throws std::invalid_argument either `N` or `M` is zero
+   */
+  template <typename Ret, bool Jacobian, typename LP,
+            require_matrix_t<Ret>* = nullptr>
+  inline auto read_constrain_sum_to_zero(LP& lp, size_t N, size_t M) {
+    stan::math::check_positive("read_sum_to_zero", "N", N);
```
bob-carpenter, line 629: read_sum_to_zero -> read_constrain_sum_to_zero

src/stan/io/deserializer.hpp
```diff
@@ -606,6 +605,32 @@ class deserializer {
...
+   * @tparam LP Type of log probability.
+   * @param lp The reference to the variable holding the log
+   * @param N Number of rows in zero sum matrix to generate.
+   * @param M Number of columns in zero sum matrix to generate.
+   * @return The next zero sum of the specified size.
+   * @throws std::invalid_argument either `N` or `M` is zero
+   */
+  template <typename Ret, bool Jacobian, typename LP,
+            require_matrix_t<Ret>* = nullptr>
+  inline auto read_constrain_sum_to_zero(LP& lp, size_t N, size_t M) {
```
bob-carpenter, line 628: [question] Is the LP argument first?  I thought we'd included it last before.

src/stan/io/deserializer.hpp
```diff
@@ -1229,6 +1254,17 @@ class deserializer {
     return stan::math::sum_to_zero_free(this->read<Ret>(size));
   }
 
+  /**
+   * Read a serialized sum_to_zero matrix and unconstrain it
```
bob-carpenter: Sentences need to end with periods.  In other doc, you've ended parameter descriptions with period.  This we've been very inconsistent about.

src/stan/io/deserializer.hpp
```diff
@@ -1229,6 +1254,17 @@ class deserializer {
...
+   *
+   * @tparam Ret Type of output
+   * @return Unconstrained matrix
```
bob-carpenter: Missing doc for M and N.

src/test/unit/io/deserializer_test.cpp
```diff
@@ -714,6 +714,49 @@ TEST(deserializer_vector, sum_to_zero_jacobian) {
...
+  theta.push_back(-2.0);
+  theta.push_back(0.0);
+  stan::io::deserializer<double> deserializer(theta, theta_i);
+  double lp = 0.0;
+  Eigen::MatrixXd reference
+      = stan::math::sum_to_zero_constrain(stan::math::to_matrix(theta, 2, 2));
+  Eigen::MatrixXd phi(
+      deserializer.read_constrain_sum_to_zero<Eigen::MatrixXd, false>(lp, 3,
+                                                                      3));
+  for (int i = 0; i < phi.rows(); ++i) {
```
bob-carpenter: [optional] If `reference` and `phi` are the same size and same row-major vs. column major, it's easier to do this as

```cpp
for (int n = 0; n < phi.size(); ++n) {
  EXPECT_FLOAT_EQ(reference(n), phi(n));
}
```

src/test/unit/io/deserializer_stdvector_test.cpp
```diff
@@ -175,6 +175,50 @@ TEST(deserializer_array, sum_to_zero) {
   }
 }
 
+TEST(deserializer_array, sum_to_zero_matrix) {
```
bob-carpenter, line 178: I would like to see a boundary test for a 1 x 1 matrix, which has zero underlying unconstrained parameters and always just contains 0.

bob-carpenter (changes requested):
This basically looks good---just a few doc corrections and a request for a boundary test.

One thing I've found is that for things like this, it can help to write a very simple test framework that tests the round tripping and tests gradients and log densities.  Then you can use it to try out a bunch of examples.

src/stan/io/deserializer.hpp
WardBrian: For whatever reason, we've left off the "_constrain" in other error messages for these (e.g., "read_simplex" is also used instead of "read_constrain_simplex")

src/stan/io/deserializer.hpp
WardBrian: It is last in the math code, but first in the IO

WardBrian:
Thanks @bob-carpenter -- most of the doc comments you had also applied to the vector version (these were copied, primarily), so I've updated those as well.

We do have some test scaffolding for this in the `deserializer_free_test.hpp`, but it hasn't been extended to cover all the cases we test across these files. Would be worth doing at some point for sure.

src/stan/io/deserializer.hpp
bob-carpenter: Either way, it's probably not an error we want to show to a Stan User as it's not mentioning things in the Stan language.  Otherwise, I think we should just fix these.  Not a high priority as they're mainly for internal debugging.

bob-carpenter (approved):
Thanks for adding the boundary test.

bob-carpenter:
This looks good.  Feel free to merge when tests pass.

src/stan/io/deserializer.hpp
```diff
@@ -525,8 +525,8 @@ class deserializer {
 
   /**
    * Return the next simplex of the specified size (using one fewer
-   * unconstrained scalars), incrementing the specified reference with the
-   * log absolute Jacobian determinant.
+   * unconstrained scalars). This particular transformation is linear, not
```
spinkney: this is on the `simplex_constrain` and it is not linear, possibly a copy paste error

src/stan/io/deserializer.hpp
```diff
@@ -550,8 +550,8 @@ class deserializer {
...
```
spinkney: this is on the `simplex_constrain` and it is not linear, possibly a copy paste error

src/stan/io/deserializer.hpp
WardBrian: Good catch, definitely a copy/paste

PR #3345, opened 2025-04-27
Fix bug - stepsize from windowed adaptation is reset to 1 when `term_buffer` == 0

### LINKED ISSUE #3023: "term_buffer = 0" leads to "# Step size = 1" in output csv

mitzimorris:
#### Submission Checklist

#### Summary

The underlying bug occurs when `stepsize_adaptation.restart` is called immediately before `stepsize_adaptation.complete_adaptation`; the former resets value `x_bar_` to 0;  the latter returns `exp(x_bar_)` which is 1.   Added a check to `complete_adaptation` which doesn't update the stepsize when `x_bar_` is 0.

I found this bug after refactoring the behavior of the `windowed_adaptation` object in order to make sure that the adaptation schedule was being computed correctly.  The code uses names `phase1`, `phase2`, and `phase3` to track the adaptation schedule.  The `windowed_adaptation` object exposes a counter, `cur_iter_` which is incremented by the adapter objects, `adapt_diag_e_nuts` et al.

Unit tests on the `windowed_adaptation` check that the schedule is computed properly.

Unit tests on the `hmc_nuts_diag_e_adapt_sampler` check that the adapted stepsize matches the stepsize used during sampling.

#### Intended Effect

If sampler config sets 'term_buffer=0',  preserve the evolved stepsize from previous adaptation phases.

#### How to Verify

Unit tests

#### Side Effects

N/A

#### Documentation

N/A

#### Copyright and Licensing

Columbia University


WardBrian:
I think this is incorrect. The condition added here means that the stepsize adaptation is never restarted, even as the metric changes, which means the dual averaging is over the entire warmup phase, not just the most recent window.

I think the correct solution will need to modify `learn_variance` and `learn_covariance` such that they return `false` when the next window is size 0

WardBrian:
I _think_ the correct condition is to update the current `return true` in the `if (end_adaptation_window())` branch of the variance/covariance functions with `return adapt_term_buffer_ > 0 || adapt_window_counter_ < adapt_next_window_;`

At the end of metric adaptation, `adapt_window_counter_` = `adapt_next_window_ + 1`. If the term buffer is 0, we're done, otherwise we're doing the final stepsize adaptation. But this is subtle, and will require some delicate testing, I think

mitzimorris:
no, I think it's simpler than that - there's an off by one error in the logic of `end_adaptation_window`.
https://github.com/stan-dev/stan/blob/6d0d863b93c7c34dfec4b013863678b233dc9fc7/src/stan/mcmc/windowed_adaptation.hpp#L86-L100

mitzimorris:
I've added a bunch of tests to check that this logic doesn't affect the overall number of warmup iterations and that the stepsize is not reset to 1 at the final iteration.

WardBrian:
For the tests, I think it would be better to check that the third column isn't ==1 for the first sampling iteration, rather than relying on the string outputs.

I think there's also still an issue here, possibly a new one? I'm testing 3 configurations which should all have identical metric adaptation (at least, if my intuition here is correct), but the `term_buffer=0` case still doesn't. They do appear to have identical draws up until the 1000th iteration, as we would expect

| | `num_warmup=1050 adapt term_buffer=50`  | `num_warmup=1001 adapt term_buffer=1` | `num_warmup=1000 adapt term_buffer=0` |
|--------|--------|--------|--------|
| adapted metric | 0.463521 | 0.463521 | 0.411266 |

(this is for the bernoulli model with seed 1234).

I think the issue is that the `adaptation_window()` might have the same off-by-one error, meaning one extra iteration is included? Though I think that would mean there was an iteration of warmup where nothing is happening, so maybe the off-by-one isn't in the check but rather in setting the value to `num_warmup_ - adapt_term_buffer_ - 1` in the first place?

mitzimorris:
@WardBrian :   Happy Birthday!!!

mitzimorris:
> I think there's also still an issue here, possibly a new one? I'm testing 3 configurations which should all have identical metric adaptation (at least, if my intuition here is correct), but the term_buffer=0 case still doesn't. They do appear to have identical draws up until the 1000th iteration, as we would expect

agreed - there is definitely a problem here.  working on it.

mitzimorris:
@WardBrian - this is ready for review.

mitzimorris:
discretion is the better part of valor.  refactoring not worth it.  will submit fix for bug and unit tests in a new PR.

PR #3352, opened 2025-08-07
 Direct all std::vector assigns through size-aware version 

### LINKED ISSUE #3351: Assign not properly checking bounds on arrays of Eigen objects
The following model raises an eigen assert rather than a proper error:

```stan
parameters {
  array[1] vector[1] theta;
}
transformed parameters {
  array[1] vector[2] wrong_size;
  wrong_size = theta;
}
model {
  theta[1] ~ normal(0, 1);
}
```

If `theta` is moved to data, or if they aren't inside arrays, or if the _array_ sizes mismatch, the correct error gets raised.

I believe the call chain is 
https://github.com/stan-dev/stan/blob/develop/src/stan/model/indexing/assign.hpp#L63
https://github.com/stan-dev/stan/blob/develop/src/stan/model/indexing/access_helpers.hpp#L103

Where the second function does not check if it's values are the same size. I believe that the assign overload should probably disallow vectors the same way it disallows tuples, or at least disallow them if the vectors contain anything other than scalars?

WardBrian:
#### Submission Checklist

#### Summary

Closes #3351 

#### Intended Effect

More size mismatch errors are properly caught by Stan

#### How to Verify

New test added, which crashes on `develop` and passes on this branch.

#### Copyright and Licensing

Simons Foundation

### DIFF
--- a/src/stan/model/indexing/assign.hpp
+++ b/src/stan/model/indexing/assign.hpp
@@ -59,6 +59,7 @@ inline void assign(Tuple1&& x, Tuple2&& y, const char* name);
 template <
     typename T, typename U,
     require_t<std::is_assignable<std::decay_t<T>&, std::decay_t<U>>>* = nullptr,
+    require_all_not_std_vector_t<T, U>* = nullptr,
     math::require_all_not_tuple_t<T, U>* = nullptr>
 inline void assign(T&& x, U&& y, const char* name) {
   internal::assign_impl(x, std::forward<U>(y), name);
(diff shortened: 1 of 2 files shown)


src/stan/model/indexing/assign.hpp
```diff
@@ -776,15 +777,24 @@ inline void assign(Mat1&& x, Mat2&& y, const char* name, const Idx& row_idx,
...
+    if (is_stan_scalar_v<value_type_t<T>> || x.size() == 0) {
+      x = std::forward<U>(y);
+      return;
+    }
+  }
+
+  // If we've made it this far, we need x to have elements to assign to,
+  // and we know this is either a no-op or x has size 0
+  x.resize(y.size());
+  if constexpr (std::is_rvalue_reference_v<U&&>) {
```
SteveBronder: If we are trying to fix the case of the below does this do what we intend?

```
parameters {
  array[1] vector[1] theta;
}
transformed parameters {
  array[1] vector[2] wrong_size;
  wrong_size = theta;
}
```

idt we want to resize `wrong_size` here. Shouldn't this throw an error?

WardBrian: The first part of this function checks the sizes match if x is not size 0, and will throw an exception. 

So, when this line executes it is guaranteed that x.size == y.size OR x.size == 0. If the sizes match, this resize does nothing. 

But if x was size 0, and we can’t just directly assign to it, we need to resize it or else x[i] will blow up in the following loop

WardBrian: In particular, this is the test I added, and the test still asserts that this throws an error, it just no longer _crashes_.

SteveBronder (approved):

PR #3362, opened 2025-10-09
Allow min:max indexing to always succeed if it is an empty selection 

### LINKED ISSUE #3359: "Technically" out-of-range but actually-ok range indexing "unnecessarily" errors
Again a somewhat weird title - but as always it comes with an example. The following errors, but I think it should actually be fine.
However, whether or not this errors is probably a decision that can be made arbitrarily.

```stan
transformed data {
    vector[2] tmp = [1., 2.]';
    // This is fine and prints `[]`
    print(tmp[2:1]);
    // This is not fine and errors with `vector[min_max] min indexing: accessing element out of range. index 3 out of range; expecting index to be between 1 and 2`
    print(tmp[3:2]);
}
```

For example, the equivalent in python apparently **never** errors (which I find weird):

```python
[1, 2][1:0] # fine
[1, 2][2:1] # fine
[1, 2][1:2] # fine?!
[1, 2][2] # not fine
```
 
The equivalent in Julia (with the indices shifted due to Julia's 1-based indexing) does behave how I'd like Stan to behave (pretty please):
```julia
[1, 2][2:1] # fine, returns []
[1, 2][3:2] # fine, returns []
[1, 2][2:3] # not fine
[1, 2][3] # not fine
```

WardBrian:
#### Submission Checklist

#### Summary

Fixes #3359 

#### Intended Effect

A `x:y` selection where x is bigger than y results in an empty slice, even when `x` is greater than the size of the container.

#### Copyright and Licensing

Simons Foundation

### DIFF
--- a/src/stan/model/indexing/rvalue.hpp
+++ b/src/stan/model/indexing/rvalue.hpp
@@ -786,8 +786,8 @@ template <typename StdVec, typename Idx1, typename... Idxs,
           require_not_same_t<Idx1, index_uni>* = nullptr>
 inline auto rvalue(StdVec&& v, const char* name, const Idx1& idx1,
                    Idxs&&... idxs) {
-  using inner_type = plain_type_t<decltype(rvalue(v[rvalue_at(0, idx1) - 1],
-                                                  name, idxs...))>;
+  using inner_type = plain_type_t<decltype(
+      rvalue(v[rvalue_at(0, idx1) - 1], name, idxs...))>;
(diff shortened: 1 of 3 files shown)


WardBrian:
Note that this also allows doing these empty slices on size-0 containers. 

We could add e.g. `math::check_nonzero_size("vector[min_max] indexing", name, v);` to the else branches here if that is desirable

I think it's logically sound though, and UBSAN seems fine with it.

SteveBronder (approved):

Issue #3366, opened 2025-10-16
Exception: no more scalars to read with sum_to_zero_vector


mhollanders:
Good day,

I've written a Stan program that takes in an indicator `int<lower=0, upper=1> NB` to toggle between Poisson and negative binomial models. 

I model the overdispersion terms hierarchically and In the parameters block, I define:

```stan
parameters {
  vector<lower=0>[NB] phi_a, phi_t;  // NB overdispersion intercept and scale
  sum_to_zero_vector[NB* S] phi_z;  // overdispersion z-scores
}
```

When `NB = 0` I get the following error and the program fails to run: 

`Exception: no more scalars to read`

When I remove the indicator and replace it with 

```stan
parameters {
  sum_to_zero_vector[S] phi_z;  // overdispersion z-scores
}
```

It works fine. I tried to create a reprex to isolate the issue but didn't manage. 

Cheers,

Matt

WardBrian:
Hi @mhollanders thanks for reporting, this is something we should be catching.

Note that the fix will just mean your code will be rejected with a nicer error -- you still won't be able to declare a size-0 `sum_to_zero_vector`, since the number of free parameters for this type is one fewer than the declared size, like a simplex. 

To make one 'optional', you can either do `array[NB] sum_to_zero_vector[S]` or declare it as size-1 when it isn't needed, which will make it have no free parameters under the hood: `sum_to_zero_vector[NB ? S : 1]`

bob-carpenter:
Hi, @mhollanders.  Thanks for reporting.  As Brian says, Stan currently disallows `sum_to_zero_vector` of size zero.  Unlike the simplex case, a size zero, sum-to-zero vector exists mathematically if you assume the sum of a size-zero vector is 0, as is conventional.  Doing that in our code without branching all over the place is another matter.

Should we just close this as unplanned, @WardBrian?

WardBrian:
@bob-carpenter no, there is a legitimate bug here, this should have raised an error but what is in fact happening is it is doing 0 - 1 on an unsized type and trying to load MAX_INT items, which is why `no more scalars to read` shows up

The fix will just be an added check and some tidyness around sized-vs-unsized arithmetic, but we should leave it open until then

mhollanders:
Thank you!

PR #3367, opened 2025-10-17
update constraints to have checks for negative or zero sizes

SteveBronder:
#### Submission Checklist

#### Summary

Related to #3366 this adds checks to deserializer so that negative numbers or size 0 integers are not accepted by constraints when they would not be valid.

For a few of the constraints we do allow size zero matrices, but were previously not handling them nicely. For instance `cholesky_factor_cov` requests `(N * (N + 1)) / 2 + (M - N) * N` doubles from the deserializers vector. If `M=0` we could end up requesting negative amounts of memory.

#### Intended Effect

Check valid requests to deserializer early.

#### How to Verify

Tests were added to the deserializer's test suite and can be run with 
```bash
python3 ./runTests.py -j8 ./src/test/unit/io/deserializer_test.cpp
```

#### Documentation
No new docs

#### Copyright and Licensing

### DIFF
--- a/src/stan/io/deserializer.hpp
+++ b/src/stan/io/deserializer.hpp
@@ -180,6 +180,7 @@ class deserializer {
       return map_vector_t(nullptr, m);
     } else {
       check_r_capacity(m);
+      stan::math::check_nonnegative("read", "size", m);
       return map_vector_t(&scalar_ptr_increment(m), m);
     }
   }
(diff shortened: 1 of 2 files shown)


WardBrian (approved):
Looks good to me, thanks

SteveBronder:
@WardBrian now that we found the fix in `stanc3`, do you think this is necessary? This will only ever be used by the code generator, so unless the code generator has a bug in it then we should not need these.

WardBrian:
I don’t think they are strictly necessary, but it’s a bit odd that before this PR some of them have checks and some of them don’t

PR #3375, opened 2026-03-31
nthreads=4

drezap:
This is an experiment, I want to see how the previously implemented multithreading scales. I'm not too familiar with TBB, but I've done a little bit of multithreading, but I want to see how this scales. 

This is not intended to merge. I want to see the performance on the stan database models we have.

### DIFF
--- a/src/stan/mcmc/hmc/nuts/base_nuts.hpp
+++ b/src/stan/mcmc/hmc/nuts/base_nuts.hpp
@@ -245,7 +245,7 @@ namespace stan {
 
         // now wire up the fwd and bck build of the trees which
         // depends on single-core or multi-core run
-        const bool run_serial = stan::math::internal::get_num_threads() == 1;
+        const bool run_serial = stan::math::internal::get_num_threads() == 4;
 
         std::size_t fwd_idx = 0;
         std::size_t bck_idx = 0;


drezap:
And then how do I get it to run the test models to view performance?

SteveBronder:
I would recommend running this locally. The models and code to run them are all below 

https://github.com/stan-dev/performance-tests-cmdstan

drezap:
Thanks, that's weird none of this is cloning cleanly? None of the functions are showing up? Whatever, I'll figure it out

SteveBronder:
> Thanks, that's weird none of this is cloning cleanly? None of the functions are showing up? Whatever, I'll figure it out

This can happen with the submodule system we use. If we go into the cmdstan version in that repo and then down to the stan directory you can swap out the specific commit for the stan repo in cmdstan. I'm guessing the commit in that repo for cmdstan is also pretty old so you may need to do a pull to get to the most recent cmdstan version, then go into the stan directory and grab the commit you want for stan.

drezap:
Yeah, this is strange, for some reason, even if I download a zip, or clone, there's always directories missing. Even with stan/math. Not sure what's happening

drezap:
Alright we got it working, may be something with WSL

SteveBronder:
Have you done `git clone --recursive` to pull down the repo and it's submodules?

drezap:
yes but sometimes the submodules do not appear and i’m putting it together
manually, obnoxious. I think it’s a WSL problem, with MINGW64 git for
windows it works fine but I couldn’t get make to run. prior to SAS I was
all ubuntu or fedora. And there’s this issue where I was running make
build, and WSL wasn’t finding the build targets?

moreover, I did a recursive clone of the benchmarking library, and i’d type
./runBenchmarks (of whatever the command was, I’m not at a computer), and
autocomplete would work on WSL, but then, it would give me a “file not
found error,” for the executable even though autocomplete would work…
obnoxious. a lot of dev is done on unix/linux systems anyway, windows can
be unreliable. i was only ubuntu before.

PR #3379, opened 2026-04-16
Add convergence information to optimization outputs

WardBrian:
#### Submission Checklist

#### Summary

This adds a second column of diagnostic parameters to the optimize algorithms, `converged__`, which corresponds to the
per-iteration return of the algorithm (e.g., `0` is a sucessful step but not done, `-1` is a failure, positive numbers indicate different convergence criteria hit). These values are defined in `stan::optimize::TerminationCondition`

It also adds the message associated with the final value, which is currently printed to stdout, as a trailing comment in the CSV file

#### Intended Effect

Allow consumers of the output files to understand if optimization converged, c.f. https://github.com/stan-dev/cmdstanpy/pull/851#issuecomment-4256988417

#### Documentation

I will update the documentation for cmdstan after this is merged.

#### Copyright and Licensing

Simons Foundation

### DIFF
--- a/src/stan/services/optimize/newton.hpp
+++ b/src/stan/services/optimize/newton.hpp
@@ -5,6 +5,7 @@
 #include <stan/callbacks/logger.hpp>
 #include <stan/callbacks/writer.hpp>
 #include <stan/io/var_context.hpp>
+#include <stan/optimization/bfgs.hpp>
 #include <stan/optimization/newton.hpp>
 #include <stan/services/error_codes.hpp>
 #include <stan/services/util/initialize.hpp>
@@ -92,6 +93,8 @@ int newton(Model& model, const stan::io::var_context& init,
   parameter_writer(names);
 
   double lastlp = lp;
+  int ret = 0;
+
   for (int m = 0; m < num_iterations; m++) {
     if (save_iterations) {
       std::vector<double> values;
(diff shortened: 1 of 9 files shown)


src/stan/services/optimize/newton.hpp
```diff
@@ -121,7 +123,7 @@ int newton(Model& model, const stan::io::var_context& init,
     model.write_array(rng, cont_vector, disc_vector, values, true, true, &ss);
     if (ss.str().length() > 0)
       logger.info(ss);
-    values.insert(values.begin(), lp);
+    values.insert(values.begin(), {lp, 0});
```
SteveBronder: Newton converges whenever the difference in lp is less than `1e-8`. So why don't we just have an `int ret = -1` at the beginning of the loop like in lbfgs so that whenever `if (std::fabs(lp - lastlp) <= 1e-8)` goes off we set it to `0` so we know that newton converged

SteveBronder (changes requested):

src/stan/services/optimize/newton.hpp
WardBrian: Good suggestion, I've updated it to use the same numeric codes as lbfgs/bfgs for the two possible options (absolute tolerance reached and max iterations reached)

SteveBronder (approved):

PR #3400, opened 2026-08-07
Wrapped repeating code in a function

### LINKED ISSUE #2577: Clean up code: `.str().length()`

Salvatore1021:
#### Submission Checklist

#### Copyright and Licensing

### DIFF
--- a/src/stan/callbacks/logger.hpp
+++ b/src/stan/callbacks/logger.hpp
@@ -95,6 +95,19 @@ class logger {
    */
   virtual void fatal(const std::stringstream& message) {}
 };
+
+inline void log_if_nonempty(logger& log, const std::stringstream& message) {
+  if (message.str().length() > 0) {
+    log.info(message);
+  }
+}
+
+inline void log_if_nonempty(logger& log, const std::string& message) {
+  if (!message.empty()) {
+    log.info(message);
+  }
+}
+
 }  // namespace callbacks
 }  // namespace stan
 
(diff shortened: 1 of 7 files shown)


src/stan/callbacks/logger.hpp
```diff
@@ -95,6 +95,19 @@ class logger {
    */
   virtual void fatal(const std::stringstream& message) {}
 };
+
+inline void log_if_nonempty(logger& log, const std::stringstream& message) {
```
bob-carpenter, line 99: Given that this hard codes the "info" level, it'd be nice if that's in its name, as in `log_info_if_nonempty`.

bob-carpenter: If you could also add function doc, that'd be helpful.

src/stan/callbacks/logger.hpp
```diff
@@ -95,6 +95,19 @@ class logger {
...
+  if (message.str().length() > 0) {
```
bob-carpenter, line 100: If you reorder the two functions, this can be reduced to
`log_if_nonempty(logger, msg.str())`.  

I also think the variable should be called `logger` rather than `log`.

bob-carpenter (changes requested):
Thanks---this is useful and we should merge, but I just asked for a little doc and simplification in calling.

bob-carpenter:
There may also be a lot more places this comes up in the code, but I'm not sure.  I'm happy to take just this many fixes.

PR #3404, opened 2026-08-11
fix standalone generated quantities csv writer

### LINKED ISSUE #3403: Multi-chain `generate_quantities` writes the constrained parameters into the csv
The fix and issue text were assisted by Claude

#### Summary:

`gq_writer` has two `write_gq_values` overloads. The `std::vector` one
drops the first `num_constrained_params_` values before writing; the
Eigen one does not. Only the multi-chain path of `standalone_generate`
uses the Eigen overload, so `generate_quantities num_chains=N` writes
each row with the constrained parameters prepended, while the header
lists only the generated quantities.

`src/stan/services/util/gq_writer.hpp`, Eigen overload:

```cpp
model.write_array(rng, draw, values, false, true, &ss);
...
sample_writer_(values);                     // parameters + generated quantities
```

versus the `std::vector` overload:

```cpp
std::vector<double> gq_values(values.begin() + num_constrained_params_,
                              values.end());
sample_writer_(gq_values);                  // generated quantities only
```

The result is a CSV whose rows have `num_params` more fields than the
header has names, so any positional reader mis-assigns every column.

Only `generate_quantities num_chains=N` with `N > 1` in a single process
is affected, which in practice means the CmdStan command line. CmdStanR
and CmdStanPy both launch one process per chain, so each process runs
with `num_chains = 1` and is delegated to the single-chain overload;
RStan's `standalone_gqs()` calls that overload directly. All three
therefore take the `std::vector` path and are unaffected.

#### Reproducible Steps:

`gq_rng.stan`:

```stan
parameters {
  real theta;
}
model {
  theta ~ std_normal();
}
generated quantities {
  real u = uniform_rng(0, 1);
}
```

```
./gq_rng sample num_warmup=200 num_samples=4 output file=fit.csv random seed=1
cp fit.csv fit_1.csv; cp fit.csv fit_2.csv
./gq_rng generate_quantities fitted_params=fit.csv num_chains=2 output file=gqmc.csv random seed=999
```

#### Current Output:

`gqmc_1.csv` — one column name, two values per row; the leading value is
`theta`, a parameter:

```
u
-1.871546, 0.89641628
-0.60661605, 0.63008851
-0.36586517, 0.17265139
-0.96383407, 0.36561701
```

#### Expected Output:

```
u
0.89641628
0.63008851
0.17265139
0.36561701
```

#### Additional Information:

Fix is to slice in the Eigen overload as the `std::vector` one does:

```cpp
plain_type_t<EigVec> gq_values
    = values.tail(values.size() - num_constrained_params_);
sample_writer_(gq_values);
```

Note this changes the output of `generate_quantities num_chains=N` by
removing columns, so it deserves a changelog entry for anyone whose
parser adapted to the current behaviour.

Minor, separate: the two paths also disagree on the separator. The
Eigen writer emits `", "` where the `std::vector` writer emits `","`.
Not addressed here.

The regression test asserts that each row of a two-chain run has as
many comma-separated fields as the header has names.
It fails on `develop` and passes with the fix. PR #3402 changes the rng seed
behavior which changes the output, and after #3402 is merged, the test for this
needs to be updated.

#### Current Version:

v2.39.0 (also reproduced on `develop`: cmdstan `b6c5aa2`, stan
`d39d26ba7`).

avehtari:
Fixes #3403 

The fix and PR text were assisted by Claude

#### Submission Checklist

#### Summary

`gq_writer::write_gq_values` has two overloads; the
`std::vector` one drops the first `num_constrained_params_` values
before writing, the Eigen one wrote everything `write_array` returned.
The multi-chain path of `standalone_generate` uses the Eigen overload,
so `generate_quantities num_chains=N` produced rows with the constrained
parameters prepended and a header listing only the generated quantities.

#### Intended Effect

Multi-chain standalone `generate_quantities` writes one value per header
name, matching the single-chain path.

#### How to Verify

`./runTests.py src/test/unit/services/sample/standalone_gqs_parallel_test.cpp`

New test `genDraws_bernoulli_multi_chain_column_count`: for a two-chain
run, every row must have as many comma-separated fields as the header
has names. Fails on develop, passes with this PR.

Command line:

```
./model generate_quantities fitted_params=fit.csv num_chains=2 output file=gq.csv
```

`gq_1.csv` rows go from `theta_value, u_value` to `u_value`.

One further test is not included here. Once stan-dev/stan#3402 is
merged I will rebase and add
`genDraws_bernoulli_multi_chain_matches_single`, which asserts that
chain *i* of a multi-chain run reproduces a single-chain run with chain
id `1 + i`. It needs both stan-dev/stan#3402 and this PR for the column
alignment.

#### Side Effects

**Output change**: `generate_quantities num_chains=N` loses the spurious
leading parameter columns. Anyone whose reader adapted to the current
(header-inconsistent) layout will need to drop the workaround.

Not addressed here: the same two paths disagree on the CSV separator
(`", "` from the Eigen writer, `","` from the other). Cosmetic, and
fixing it means touching the writer classes, which are shared with other
methods.

#### Documentation

None.

#### Copyright and Licensing

Aki Vehtari

### DIFF
--- a/src/test/unit/services/sample/standalone_gqs_parallel_test.cpp
+++ b/src/test/unit/services/sample/standalone_gqs_parallel_test.cpp
@@ -92,7 +92,9 @@ Eigen::MatrixXd bernoulli_fit_draws() {
 }
 
 // drop the timing-dependent " Elapsed Time" line, which the test writers
-// emit without a comment prefix
+// emit without a comment prefix, and normalize the separator: the Eigen
+// writer path emits ", " where the std::vector path emits ",", which the
+// cross-path test below has to see through
 std::string data_lines(const std::string& csv) {
   std::stringstream in(csv), out;
   std::string line;
@@ -101,6 +103,7 @@ std::string data_lines(const std::string& csv) {
       continue;
     if (line.find("Elapsed Time") != std::string::npos)
       continue;
+    line.erase(std::remove(line.begin(), line.end(), ' '), line.end());
     out << line << "\n";
   }
   return out.str();
@@ -194,5 +197,32 @@ TEST_F(ServicesStandaloneGQ, genDraws_bernoulli_multi_chain_column_count) {
       rows++;
     }
     EXPECT_EQ(rows, 1000);
+  }
+}
+
+// Cross-path check: the multi-chain and single-chain code paths must agree.
+TEST_F(ServicesStandaloneGQ, genDraws_bernoulli_multi_chain_matches_single) {
+  Eigen::MatrixXd draws = bernoulli_fit_draws();
+  std::vector<std::stringstream> sample_ss(num_chains);
+  std::vector<test_writer> sample_writer;
+  sample_writer.reserve(num_chains);
+  std::vector<Eigen::MatrixXd> draws_vec;
+  for (int i = 0; i < num_chains; i++) {
+    sample_writer.emplace_back(
+        std::unique_ptr<std::stringstream, deleter_noop>(&sample_ss[i]), "");
+    draws_vec.push_back(draws);
+  }
+  EXPECT_EQ(
+      stan::services::standalone_generate(model, num_chains, draws_vec, 12345,
+                                          interrupt, logger, sample_writer),
+      stan::services::error_codes::OK);
+  for (int i = 0; i < num_chains; i++) {
+    std::stringstream single_ss;
+    test_writer single_writer(
+        std::unique_ptr<std::stringstream, deleter_noop>(&single_ss), "");
+    EXPECT_EQ(stan::services::standalone_generate(
+                  model, draws, 12345, interrupt, logger, single_writer, 1 + i),
+              stan::services::error_codes::OK);
+    EXPECT_EQ(data_lines(single_ss.str()), data_lines(sample_ss[i].str()));
   }
 }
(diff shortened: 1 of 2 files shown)


src/test/unit/services/sample/standalone_gqs_parallel_test.cpp
WardBrian, line 1: Besides fixing the merge conflict, mind also adding a test that you can get the exact same result from both overloads now?

avehtari: That test was already waiting and now included

avehtari:
Rebased onto develop now that #3402 has merged (force-pushed). Changes since this PR was opened:

- **Rebased** — the branch was based on pre-#3402 develop.
- **Dropped the duplicated test helpers.** The first version added its own `bernoulli_fit_draws()` and `test_writer`; #3402 now provides those (plus `data_lines`), so this PR adds only its test.
- **Added `genDraws_bernoulli_multi_chain_matches_single`** — chain *i* of a multi-chain run must reproduce a single-chain run at chain id `1 + i`. It needs #3402 for the per-chain streams and this PR for the column alignment, so it could not pass on either branch alone and was held back until #3402 landed.
- **Extended `data_lines()`** to normalize the CSV separator: the Eigen writer path emits `", "` where the `std::vector` path emits `","`, and the new cross-path test is the first one to compare the two. The separator inconsistency itself remains out of scope here.

The fix in `src/stan/services/util/gq_writer.hpp` is unchanged.

WardBrian (approved):

SteveBronder (approved):
Thanks!

PR #3405, opened 2026-08-12
Reduce Boost dependency footprint

andrjohns:
#### Submission Checklist

#### Summary

Related to [this Math PR](https://github.com/stan-dev/math/pull/3356), this PR removes/replaces non-math Boost includes with std-equivalents.

The following includes are removed:

  - `boost/random`: Replaced with `<random>` equivalents, expected test values also updated
  - `boost/algorithm`: Only basic string manip was needed, these have been implemented under `string_utils.hpp` instead
  - `boost/regex`: Only used to check whether variable name was valid, pattern was very simple so just used `std::isalpha` & `std::isalnum`
  - `boost/accumulators`: Replaced with Eigen functions

Now the only non-math Boost include is `boost/random/mixmax`. The `boost/accumulators` removal in particular is a big win, since that pulled in both `boost/mpl` and `boost/fusion` - these two libraries alone are ~20mb!

Let me know if it would be easier to review if I split the changes into separate PRs

#### Intended Effect

Reduce amount of Boost libraries/headers needed by Stan

#### Side Effects

Changing the rng usage from boost -> std will return different values from same seed.

#### Documentation

N/A

#### Copyright and Licensing

### DIFF
--- a/src/stan/analyze/mcmc/autocovariance.hpp
+++ b/src/stan/analyze/mcmc/autocovariance.hpp
@@ -2,9 +2,6 @@
 #define STAN_ANALYZE_MCMC_AUTOCOVARIANCE_HPP
 
 #include <stan/math/prim.hpp>
-#include <boost/accumulators/accumulators.hpp>
-#include <boost/accumulators/statistics/stats.hpp>
-#include <boost/accumulators/statistics/variance.hpp>
 #include <unsupported/Eigen/FFT>
 #include <complex>
 #include <vector>
@@ -87,16 +84,9 @@ void autocovariance(const Eigen::MatrixBase<DerivedA>& y,
   Eigen::FFT<T> fft;
   autocorrelation(y, acov, fft);
 
-  using boost::accumulators::accumulator_set;
-  using boost::accumulators::stats;
-  using boost::accumulators::tag::variance;
+  double variance = (y.array() - y.mean()).matrix().squaredNorm() / y.size();
 
-  accumulator_set<double, stats<variance>> acc;
-  for (int n = 0; n < y.size(); ++n) {
-    acc(y(n));
-  }
-
-  acov = acov.array() * boost::accumulators::variance(acc);
+  acov = acov.array() * variance;
 }
 
 /**
--- a/src/stan/io/string_utils.hpp
+++ b/src/stan/io/string_utils.hpp
@@ -109,6 +109,28 @@ inline void replace_first(std::string& s, std::string_view target,
   }
 }
 
+/**
+ * Returns true if `s` contains `needle` as a substring.
+ *
+ * @param s string to search
+ * @param needle substring to search for
+ * @return true if `needle` is found in `s`
+ */
+inline bool contains(std::string_view s, std::string_view needle) {
+  return s.find(needle) != std::string_view::npos;
+}
+
+/**
+ * Returns true if `s` starts with `prefix`.
+ *
+ * @param s string to check
+ * @param prefix prefix to look for
+ * @return true if `s` starts with `prefix`
+ */
+inline bool starts_with(std::string_view s, std::string_view prefix) {
+  return s.size() >= prefix.size() && s.compare(0, prefix.size(), prefix) == 0;
+}
+
 }  // namespace io
 }  // namespace stan
 #endif
--- a/src/stan/io/string_utils.hpp
+++ b/src/stan/io/string_utils.hpp
@@ -1,0 +1,103 @@
+#ifndef STAN_IO_STRING_UTILS_HPP
+#define STAN_IO_STRING_UTILS_HPP
+
+#include <algorithm>
+#include <cctype>
+#include <string>
+#include <vector>
+
+namespace stan {
+namespace io {
+
+/**
+ * Splits a string on any character found in `delims`.
+ *
+ * If `compress_delims` is false, every delimiter produces a split, so
+ * adjacent delimiters yield an empty token between them (e.g. splitting
+ * "a,,b" on "," gives {"a", "", "b"}).
+ *
+ * If `compress_delims` is true, runs of adjacent delimiters are treated
+ * as a single delimiter, except that a leading or trailing run still
+ * yields a single empty token at that end (matching the behavior of
+ * boost::split with token_compress_on).
+ *
+ * @param s string to split
+ * @param delims set of delimiter characters
+ * @param compress_delims whether to collapse adjacent delimiters
+ * @return the tokens found in `s`
+ */
+inline std::vector<std::string> split(const std::string& s,
+                                      const std::string& delims,
+                                      bool compress_delims = false) {
+  std::vector<std::string> tokens;
+  size_t start = 0;
+  while (true) {
+    size_t pos = s.find_first_of(delims, start);
+    if (pos == std::string::npos) {
+      tokens.push_back(s.substr(start));
+      break;
+    }
+    tokens.push_back(s.substr(start, pos - start));
+    start = pos + 1;
+  }
+  if (compress_delims) {
+    std::vector<std::string> compressed;
+    for (size_t i = 0; i < tokens.size(); ++i) {
+      // drop empty tokens produced by interior runs of delimiters,
+      // keeping a single empty token at either end
+      if (tokens[i].empty() && i > 0 && i + 1 < tokens.size())
+        continue;
+      compressed.push_back(std::move(tokens[i]));
+    }
+    tokens.swap(compressed);
+  }
+  return tokens;
+}
+
+/**
+ * Joins a vector of strings into a single string, separated by `sep`.
+ *
+ * @param parts strings to join
+ * @param sep separator inserted between parts
+ * @return the joined string
+ */
+inline std::string join(const std::vector<std::string>& parts,
+                        const std::string& sep) {
+  std::string result;
+  for (size_t i = 0; i < parts.size(); ++i) {
+    if (i > 0)
+      result += sep;
+    result += parts[i];
+  }
+  return result;
+}
+
+/**
+ * Trims leading and trailing whitespace from a string, in place.
+ *
+ * @param s string to trim
+ */
+inline void trim(std::string& s) {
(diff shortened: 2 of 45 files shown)


WardBrian:
> `boost/random/mixmax`

My coworker @DiamonDinoia has encouraged me to stop using this generator as well (can be a separate PR)

andrjohns:
> > `boost/random/mixmax`
> 
> My coworker @DiamonDinoia has encouraged me to stop using this generator as well (can be a separate PR)

Interesting! What's the recommended replacement? 

If we're looking at changing the RNG it would probably be best if we did the `boost/random` -> std random and the RNG change in the same release, since both are going to change any fixed expectations for downstream users. 

With the feature freeze coming up did you want to do the RNG change this release or hold off on the `boost/random` changes until the next release?

DiamonDinoia:
@vigna has a great page to guide the choice: https://prng.di.unimi.it/ look for how to choose a PRNG. 

If I were to implement the PRNG from scratch I would implement one of [xoshiro256++](https://prng.di.unimi.it/xoshiro256plusplus.c)/[xoshiro256**](https://prng.di.unimi.it/xoshiro256starstar.c) which are 10 liners. For example in c++: https://github.com/DiamonDinoia/simdrng/blob/main/include/simdrng/xoshiro_scalar.hpp 

If I can add a dependency I would use ChaCha20, AES in counter mode relies on having AES implemented in hw which not all platforms have. And I think Stan aims to support many no? 

I personally implemented those into my package here https://github.com/DiamonDinoia/simdrng I am still developing and cleaning up a few things. But I wanted to take the change to make a product placement. 


This package tests some generators and share some findings: https://github.com/vigna/modlin-rs

DiamonDinoia:
If you implement any of those generators, I am happy to help review.

src/stan/analyze/mcmc/autocovariance.hpp
```diff
@@ -87,16 +84,9 @@ void autocovariance(const Eigen::MatrixBase<DerivedA>& y,
...
-  using boost::accumulators::stats;
-  using boost::accumulators::tag::variance;
+  double variance = (y.array() - y.mean()).matrix().squaredNorm() / y.size();
 
-  accumulator_set<double, stats<variance>> acc;
-  for (int n = 0; n < y.size(); ++n) {
-    acc(y(n));
-  }
-
-  acov = acov.array() * boost::accumulators::variance(acc);
```
SteveBronder, line 99: We should use a little variance function like the below that uses welford's algorithm so we can be a bit more precise

https://godbolt.org/z/Porcq1z8W

src/stan/io/string_utils.hpp
```diff
@@ -0,0 +1,103 @@
...
+ * yields a single empty token at that end (matching the behavior of
+ * boost::split with token_compress_on).
+ *
+ * @param s string to split
+ * @param delims set of delimiter characters
+ * @param compress_delims whether to collapse adjacent delimiters
+ * @return the tokens found in `s`
+ */
+inline std::vector<std::string> split(const std::string& s,
+                                      const std::string& delims,
```
SteveBronder: Bit of code golf, but I think you can shorten this a good bit 

https://godbolt.org/z/hGTe59ber

src/stan/io/string_utils.hpp
```diff
@@ -0,0 +1,103 @@
...
+ * @param parts strings to join
+ * @param sep separator inserted between parts
+ * @return the joined string
+ */
+inline std::string join(const std::vector<std::string>& parts,
+                        const std::string& sep) {
+  std::string result;
+  for (size_t i = 0; i < parts.size(); ++i) {
+    if (i > 0)
+      result += sep;
```
SteveBronder:
```suggestion
    if (i > 0) {
      result += sep;
    }
```

src/stan/io/string_utils.hpp
```diff
@@ -0,0 +1,103 @@
...
+/**
+ * Joins a vector of strings into a single string, separated by `sep`.
+ *
+ * @param parts strings to join
+ * @param sep separator inserted between parts
+ * @return the joined string
+ */
+inline std::string join(const std::vector<std::string>& parts,
+                        const std::string& sep) {
+  std::string result;
```
SteveBronder, line 75: You will be allocating at least `parts.size()` sep + `parts[0]`


```suggestion
  std::string result;
  result.reserve(parts[0].size() + parts.size());
```

src/stan/io/string_utils.hpp
```diff
@@ -0,0 +1,103 @@
...
+/**
+ * Replaces the first occurrence of `target` in `s` with `replacement`,
+ * in place. Does nothing if `target` is not found.
+ *
+ * @param s string to modify
+ * @param target substring to replace
+ * @param replacement replacement text
+ */
+inline void replace_first(std::string& s, const std::string& target,
+                          const std::string& replacement) {
```
SteveBronder:
```suggestion
inline void replace_first(std::string_view& s, const std::string_view& target,
                          const std::string_view& replacement) {
```

src/stan/mcmc/chains.hpp
```diff
@@ -65,109 +57,35 @@ class chains {
...
 
     int M = std::min(x.size(), y.size());
-    for (int i = 0; i < M; i++)
-      acc(x(i), boost::accumulators::covariate1 = y(i));
-
-    return boost::accumulators::covariance(acc) * M / (M - 1);
+    double mx = x.head(M).mean();
+    double my = y.head(M).mean();
+    return ((x.head(M).array() - mx) * (y.head(M).array() - my)).sum()
+           / (M - 1.0);
```
SteveBronder, line 65: Like the variance I think it would be nice to add a welford style estimator for covariance here

https://godbolt.org/z/E7v9dbj59

src/stan/mcmc/chains.hpp
```diff
@@ -65,109 +57,35 @@ class chains {
...
-    for (int i = 0; i < probs.size(); i++) {
-      if (probs(i) < 0.5)
-        q(i) = quantile(acc_left, quantile_probability = probs(i));
-      else
-        q(i) = quantile(acc_right, quantile_probability = probs(i));
-    }
-    return q;
+    std::vector<double> probs_vec(probs.data(), probs.data() + probs.size());
+    std::vector<double> q = stan::math::quantile(x, probs_vec);
+    return Eigen::Map<const Eigen::VectorXd>(q.data(), q.size());
```
SteveBronder, line 88: This needs to be cast to a vector to use quantile? Could we update that function to also accept Eigen vectors?

SteveBronder (changes requested):
Few changes and I need to look over the ring buffer impl still

src/stan/analyze/mcmc/autocovariance.hpp
andrjohns: Interesting! I was looking into the speed/accuracy tradeoff from switching from vectorised Eigen to welford's, and it's showing the Eigen vectorised impl as being more accurate still: https://godbolt.org/z/G33637P6r.

I've also attached the version that did a timing comparison across a larger range as well.

I'm not overly-confident that Claude isn't missing something obvious though, are there any particular edge-cases or examples where you've seen the welford benefits that I can test with? I was always under the impression that the main motivator was for online/streaming cases rather than with a fully-materialised case.

[variance_accuracy.cpp](https://github.com/user-attachments/files/31095800/variance_accuracy.cpp)

andrjohns: We do have existing welford estimators for variance and covariance in the Math library that I could switch to here and for the covariance below:

  - [welford_var_estimator](https://github.com/stan-dev/math/blob/develop/stan/math/prim/fun/welford_var_estimator.hpp)
  - [welford_covar_estimator](https://github.com/stan-dev/math/blob/develop/stan/math/prim/fun/welford_covar_estimator.hpp)

src/stan/mcmc/chains.hpp
andrjohns: Added support in [this math PR](https://github.com/stan-dev/math/pull/3358)

src/stan/analyze/mcmc/autocovariance.hpp
roualdes: Relative to @andrjohns' file variance_accuracy.cpp.

* eigen_twopass is generally more accurate, but can fail in extreme cases. John Cook has a [post](https://www.johndcook.com/blog/2008/09/26/comparing-three-methods-of-computing-standard-deviation/) about this. It's not clear how common such cases are. 

* I've always written the variance calculations for welford_seq as
`m2 += -m2 * inv + delta * delta * inv * (1 - inv);` and then return `m2 * n / (n - 1);`, which proves a bit more accurate than welford_seq, and otherwise follows the pattern of the bullet point above.

andrjohns:
Quick comparison, with this PR plus Math PRs [3356](https://github.com/stan-dev/math/pull/3356) & [3357](https://github.com/stan-dev/math/pull/3357) the precompiled header on my system drops from 844MB down to  552MB!

WardBrian:
Presumably we can also remove some more pieces of boost and shrink the cmdstan tarball size

andrjohns:
> Presumably we can also remove some more pieces of boost and shrink the cmdstan tarball size

Yeah we can drop a ton! I've got one PR after this one to remove the usages in `cmdstan`. After that, we only need 18 Boost libs (dropping the boost folder from 81.5mb -> 28.1mb):

  - math
  - archive
  - mpl
  - config
  - preprocessor
  - core
  - type_traits
  - optional
  - mpi
  - integer
  - serialization
  - assert
  - throw_exception
  - smart_ptr
  - utility
  - static_assert
  - random
  - array


The `random` & `array` libs are only required by the `mixmax` RNG (removing them drops the folder to `27.2mb`)

The other 16 boost libs (~19mb) are all pulled in by MPI, so any non-mpi model would now only pull in the `math` lib from Boost

andrjohns:
(that's for the tarball, there are a few more libs used for tests in the `math` library that I'll get in a separate PR)

andrjohns:
I'm going to split this one into smaller PRs (per Boost dependency) so that review's a little easier

PR #3408, opened 2026-08-18
Migration to new jenkins system

dylex:
Migration to jenkins-new, significant cleanup of Jenkinsfile, mostly maintains the same behaviors and tests.

- Fake downstream_ branches replaced with downstream parameter and clearer branch targets.
- Docker imaged replaced by local Dockerfile, which is just an optimized version of the old one, with some things removed (and downgraded back to clang++-6.0, since the image was based on an older Dockerfile)
- `recordIssues` is removed in favor of more `-Werror` arguments, though this could be reconsidered if there are other important features.

Remaining possible work:

- Add M1 mac builds
- Review and cleanup email result targets, including stan-buildbot list

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -19,7 +19,6 @@ properties([
 def image = 'stanorg/ci:v1'
 def commit
 def runRemainingStages = false
-def runOpenCL = false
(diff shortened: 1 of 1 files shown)


WardBrian (approved):
Looks very clean, thanks!

WardBrian:
Actually, is the system smart enough to pull the dockerfile from the `math` submodule?

PR #3409, opened 2026-08-24
Replace Boost Regex with std::locale

andrjohns:
#### Submission Checklist

#### Summary

The validation rules for Stan variable names are simple enough that we don't need the `boost/regex` dependency and can use basic string checking utilities from the standard library instead

#### Intended Effect

Reduce dependency footprint

#### How to Verify

Current tests and models all still run/pass

#### Side Effects

N/A

#### Documentation

N/A

#### Copyright and Licensing

### DIFF
--- a/src/stan/io/json/json_data_handler.hpp
+++ b/src/stan/io/json/json_data_handler.hpp
@@ -171,7 +171,7 @@ class json_data_handler : public stan::json::json_handler {
    *  and contain only letters, numbers, or an underscore.
    */
   bool valid_varname(const std::string& name) {
-    static const std::locale locale;
+    static const std::locale locale = std::locale::classic();
     if (name.empty() || !std::isalpha(name[0], locale)) {
       return false;
     }
--- a/src/stan/io/json/json_data_handler.hpp
+++ b/src/stan/io/json/json_data_handler.hpp
@@ -5,8 +5,8 @@
 #include <stan/io/json/json_handler.hpp>
 #include <stan/io/json/rapidjson_parser.hpp>
 #include <stan/io/var_context.hpp>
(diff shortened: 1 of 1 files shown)


src/stan/io/json/json_data_handler.hpp
```diff
@@ -172,8 +171,13 @@ class json_data_handler : public stan::json::json_handler {
    *  and contain only letters, numbers, or an underscore.
    */
   bool valid_varname(const std::string& name) {
-    static const boost::regex re("[a-zA-Z][a-zA-Z0-9_]*");
-    return boost::regex_match(name, re);
+    static const std::locale locale;
+    if (name.empty() || !std::isalpha(name[0], locale)) {
```
WardBrian, line 175: Should we use `std::locale::classic()` instead? I think the "C" locale is correct here, at least until we'd allow unicode variable names

andrjohns: Ah true true

src/stan/io/json/json_data_handler.hpp
```diff
@@ -172,8 +171,13 @@ class json_data_handler : public stan::json::json_handler {
    *  and contain only letters, numbers, or an underscore.
    */
   bool valid_varname(const std::string& name) {
-    static const boost::regex re("[a-zA-Z][a-zA-Z0-9_]*");
-    return boost::regex_match(name, re);
+    static const std::locale locale = std::locale::classic();
```
WardBrian:
```suggestion
    static const std::locale& locale = std::locale::classic();
```

WardBrian (approved):

PR #3412, opened 2026-08-24
Replace boost/random with stdlib random

andrjohns:
#### Submission Checklist

#### Summary

Replaces all `boost/random` usages (distribution variate generators) with their stdlib `<random>` equivalents. This doesn't touch the `mixmax` RNG usage.

A big side-effect here is that results will be slightly different for the same seed (several test values also updated because of this)

#### Intended Effect

Reduce Boost dependency footprint

#### Side Effects

Results for existing seeds will no longer match

#### Documentation

N/A

#### Copyright and Licensing

### DIFF
--- a/src/stan/io/random_var_context.hpp
+++ b/src/stan/io/random_var_context.hpp
@@ -3,9 +3,9 @@
 
 #include <stan/io/var_context.hpp>
 #include <stan/io/validate_dims.hpp>
-#include <boost/random/uniform_real_distribution.hpp>
 #include <algorithm>
 #include <limits>
+#include <random>
 #include <string>
 #include <vector>
 
@@ -50,8 +50,7 @@ class random_var_context : public var_context {
       for (size_t n = 0; n < num_unconstrained_; ++n)
         unconstrained_params_[n] = 0.0;
     } else {
-      boost::random::uniform_real_distribution<double> unif(-init_radius,
-                                                            init_radius);
+      std::uniform_real_distribution<double> unif(-init_radius, init_radius);
       for (size_t n = 0; n < num_unconstrained_; ++n)
         unconstrained_params_[n] = unif(rng);
     }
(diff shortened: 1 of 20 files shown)


WardBrian (approved):

WardBrian:
@andrjohns this has been failing on develop: https://jenkins.flatironinstitute.org/blue/organizations/jenkins/Stan%2FStan/detail/develop/419/pipeline

it looks like maybe the std rngs aren’t really as cross-platform?

andrjohns:
> @andrjohns this has been failing on develop: https://jenkins.flatironinstitute.org/blue/organizations/jenkins/Stan%2FStan/detail/develop/419/pipeline
> 
> it looks like maybe the std rngs aren’t really as cross-platform?

Oh bizarre. I'll see if I can reproduce on mac and see where it's coming from

WardBrian:
@andrjohns since it's blocking downstream CI, thoughts on reverting while you explore?

andrjohns:
> @andrjohns since it's blocking downstream CI, thoughts on reverting while you explore?

No objections from me!

andrjohns:
Ah, this is a wrinkle. Looks like there's no guarantee that the `random` distributions will be identical between compilers.

  - [`std::normal_distribution`: g++](https://godbolt.org/z/fGjrMKn3j)
  - [`std::normal_distribution`: clang++](https://godbolt.org/z/1Wq911Pe4)

WardBrian:
Classic C++ "standardization"

andrjohns:
I think we still have an option though. Boost's [`uniform_01`](https://github.com/stan-dev/math/blob/develop/lib/boost_1.87.0/boost/random/uniform_01.hpp) just uses the standard approach for `uint64` RNGs:

```cpp
double factor = 1.0 / (double(rng.max() - rng.min()) + 1.0);
double result = 1;
while (result >= 1) {
  result = double(rng() - rng.min()) * factor;
}
```

Then we can use the `boost::math` quantile functions for the target distributions

PR #3418, opened 2026-08-28
Flush JSON after top level finishes 

### LINKED ISSUE #1349: Saving metric with large models produces truncated JSON files

jgabry:
#### Submission Checklist

#### Summary

https://github.com/stan-dev/cmdstan/issues/1349 has the context. 

Flush when the top level JSON object is finished. @WardBrian I guess this is slightly different than your suggestion, but I think equivalent for the metric JSON files. I wasn't sure if there are ever nested JSON objects in other contexts. 

#### Intended Effect

Make it so that metric JSON files are readable after warmup but before sampling finishes

#### Copyright and Licensing

Jonah Gabry

### DIFF
--- a/src/stan/callbacks/json_writer.hpp
+++ b/src/stan/callbacks/json_writer.hpp
@@ -24,7 +24,7 @@ namespace callbacks {
  * The writer doesn't try to validate the object's internal structure
  * or object completeness, only syntactic correctness.
  *
- * @tparam Stream A type with with a valid `operator<<(std::string)`
+ * @tparam Stream A type with a valid `operator<<(std::string)` and `flush()`
  * @tparam Deleter A class with a valid `operator()` method for deleting the
  * output stream
  */
@@ -227,6 +227,7 @@ class json_writer final : public structured_writer {
       record_element_needs_comma_ = true;
     } else {
       *output_ << "\n";
+      output_->flush();
     }
   }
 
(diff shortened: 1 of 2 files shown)


WardBrian:
only flushing at the top level is fine. I'm honestly not sure this kind of mocked counter test actually adds much value (for example, if it gets flushed more than once, do we really care?)

jgabry:
Yeah fair point, I can remove the test

WardBrian:
and did you confirm that this fixes the behavior you were seeing up in cmdstan?

jgabry:
In the process of testing it right now, will report back

jgabry:
Ok I built cmdstan using this stan branch and I can successfully read the JSON during sampling, so I think it's working as intended

WardBrian (approved):

PR #3429, opened 2026-09-03
Newton optimizer: handle flat directions without producing NaN

### LINKED ISSUE #3425: Newton optimizer produces NaN parameters for a target with a flat direction, no error raised
[`make_negative_definite_and_solve`](https://github.com/stan-dev/stan/blob/8ad5c989c850fc15aa1c365181272d4e4697d4e6/src/stan/optimization/newton.hpp#L18-L26) divides by `fabs(eigenvalues[i])` without checking for zeros (division at [line 24](https://github.com/stan-dev/stan/blob/8ad5c989c850fc15aa1c365181272d4e4697d4e6/src/stan/optimization/newton.hpp#L24)). For a parameter absent from the target, the Hessian is zero and the step is `0/0`:

```cpp
// model whose log_prob ignores its single parameter, returns 0.5
stan::optimization::newton_step<flat_model, false>(model, params_r, params_i);
// f0 = 0.5, params_r[0] = -nan
```

The caller loop accepts NaN (`while (f1 < f0)` is false for NaN) and the service [returns `error_codes::OK`](https://github.com/stan-dev/stan/blob/8ad5c989c850fc15aa1c365181272d4e4697d4e6/src/stan/services/optimize/newton.hpp#L138) without a finiteness check, so the run reports success with NaN parameters. Probably related to  #3306.

### Environment:
- Ubuntu 22.04.5 LTS (WSL2, kernel 6.18.33.2-microsoft-standard-WSL2), x86_64
- stan-dev/stan develop @ 8ad5c989c (stan::math develop @ d20b238b9c)
- g++ 11.4.0 (Ubuntu 22.04), -std=c++17

SteveBronder:
#### Submission Checklist

#### Summary

Fixes #3425.

`make_negative_definite_and_solve` divided the gradient projection by `fabs(eigenvalue)` with no guard against zero or near zero values. This could lead to some directions of the gradient and hessian being flat and causing NaN values to return. Now we check that the absolute of the eigen value is greater than a tolerance defined by the an epsilon scaled by the overall maximum eigenvalue. We reject non-finite step directions, candidate points, and objective values in newton_step instead of accepting them. And the newton service layer not returns `error_codes::SOFTWARE` with `TERM_LSFAIL` when the final log density or parameters are not finite.

Adds a `flat_target` test model plus unit tests at the solve, step, and service layers that reproduce the reported NaN.

#### Documentation

Updated docs for `make_negative_definite_and_solve` to reflect the change.
#### Copyright and Licensing

### DIFF
--- a/src/stan/optimization/newton.hpp
+++ b/src/stan/optimization/newton.hpp
@@ -4,6 +4,8 @@
 #include <stan/model/grad_hess_log_prob.hpp>
 #include <stan/model/log_prob_grad.hpp>
 #include <stan/math/prim/fun/Eigen.hpp>
+#include <cmath>
+#include <limits>
 #include <vector>
 
 namespace stan {
@@ -12,18 +14,53 @@ namespace optimization {
 typedef Eigen::Matrix<double, Eigen::Dynamic, Eigen::Dynamic> matrix_d;
 typedef Eigen::Matrix<double, Eigen::Dynamic, 1> vector_d;
 
-// Negates any positive eigenvalues in H so that H is negative
-// definite, and then solves Hu = g and stores the result into
-// g. Avoids problems due to non-log-concave distributions.
+/**
+ * Negates any positive eigenvalues in H so that H is negative
+ * definite, then solves Hu = g and stores the result into g.
+ * Avoids problems due to non-log-concave distributions.
+ *
+ * Eigenvalues whose magnitude is negligible relative to the largest
+ * eigenvalue are treated as zero and their directions are dropped
+ * from the solve, as in a pseudo-inverse. This keeps the step finite
+ * when the target is flat along some direction.
+ *
+ * @param[in] H Hessian of the log density
+ * @param[in, out] g gradient on input, Newton step direction on output
+ */
 inline void make_negative_definite_and_solve(matrix_d& H, vector_d& g) {
   Eigen::SelfAdjointEigenSolver<matrix_d> solver(H);
   matrix_d eigenvectors = solver.eigenvectors();
   vector_d eigenvalues = solver.eigenvalues();
   vector_d eigenprojections = eigenvectors.transpose() * g;
+  double max_abs_eigenvalue = eigenvalues.cwiseAbs().maxCoeff();
+  double tolerance
+      = max_abs_eigenvalue * H.rows() * std::numeric_limits<double>::epsilon();
   for (int i = 0; i < g.size(); i++) {
-    eigenprojections[i] = -eigenprojections[i] / fabs(eigenvalues[i]);
+    double abs_eigenvalue = std::fabs(eigenvalues[i]);
+    if (abs_eigenvalue <= tolerance) {
+      eigenprojections[i] = 0;
+    } else {
+      eigenprojections[i] = -eigenprojections[i] / abs_eigenvalue;
+    }
   }
   g = eigenvectors * eigenprojections;
+}
+
+/**
+ * Returns true if every element of the vector is finite.
+ *
+ * @tparam Vec vector type with size() and operator[]
+ * @param[in] v vector to check
+ * @return true if all elements are finite
+ */
+template <typename Vec>
+inline bool all_finite(const Vec& v) {
+  for (int i = 0; i < static_cast<int>(v.size()); ++i) {
+    if (!std::isfinite(v[i])) {
+      return false;
+    }
+  }
+  return true;
 }
 
 template <typename M, bool jacobian = false>
@@ -35,6 +72,9 @@ double newton_step(M& model, std::vector<double>& params_r,
 
   double f0 = stan::model::grad_hess_log_prob<true, jacobian>(
       model, params_r, params_i, gradient, hessian);
+  if (!std::isfinite(f0)) {
+    return f0;
+  }
   matrix_d H(params_r.size(), params_r.size());
   for (size_t i = 0; i < hessian.size(); i++) {
     H(i) = hessian[i];
@@ -43,7 +83,9 @@ double newton_step(M& model, std::vector<double>& params_r,
   for (size_t i = 0; i < gradient.size(); i++)
     g(i) = gradient[i];
   make_negative_definite_and_solve(H, g);
-  //         H.ldlt().solveInPlace(g);
+  if (!all_finite(g)) {
+    return f0;
+  }
 
   std::vector<double> new_params_r(params_r.size());
   double step_size = 2;
@@ -57,11 +99,18 @@ double newton_step(M& model, std::vector<double>& params_r,
 
     for (size_t i = 0; i < params_r.size(); i++)
       new_params_r[i] = params_r[i] - step_size * g[i];
+    if (!all_finite(new_params_r)) {
+      f1 = -1e100;
+      continue;
+    }
     try {
       f1 = stan::model::log_prob_grad<true, jacobian>(model, new_params_r,
                                                       params_i, gradient);
     } catch (std::domain_error& e) {
       // FIXME:  this is not a good way to handle a general exception
+      f1 = -1e100;
+    }
+    if (!std::isfinite(f1)) {
       f1 = -1e100;
     }
   }
--- a/src/stan/optimization/newton.hpp
+++ b/src/stan/optimization/newton.hpp
@@ -1,8 +1,8 @@
 #ifndef STAN_OPTIMIZATION_NEWTON_HPP
 #define STAN_OPTIMIZATION_NEWTON_HPP
 
-#include <stan/model/grad_hess_log_prob.hpp>
-#include <stan/model/log_prob_grad.hpp>
+#include <stan/math/rev.hpp>
+#include <stan/math/rev/functor/finite_diff_hessian_auto.hpp>
 #include <stan/math/prim/fun/Eigen.hpp>
 #include <cmath>
 #include <limits>
@@ -36,9 +36,10 @@ typedef Eigen::Matrix<double, Eigen::Dynamic, 1> vector_d;
  * @param[in] H Hessian of the log density
  * @param[in, out] g gradient on input, Newton step direction on output
  */
-inline void make_negative_definite_and_solve(matrix_d& H, vector_d& g) {
+template <typename VecG>
+inline void make_negative_definite_and_solve(matrix_d& H, VecG& g) {
   Eigen::SelfAdjointEigenSolver<matrix_d> solver(H);
-  matrix_d eigenvectors = solver.eigenvectors();
+  auto&& eigenvectors = solver.eigenvectors();
   vector_d eigenvalues = solver.eigenvalues();
   vector_d eigenprojections = eigenvectors.transpose() * g;
   const double sqrt_eps = std::sqrt(std::numeric_limits<double>::epsilon());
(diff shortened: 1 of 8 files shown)


WardBrian:
Seems similar to #3309 which @nhuurre had some thoughts on, might be a good reviewer

src/stan/optimization/newton.hpp
```diff
@@ -12,20 +14,55 @@ namespace optimization {
...
+ * @param[in, out] g gradient on input, Newton step direction on output
+ */
 inline void make_negative_definite_and_solve(matrix_d& H, vector_d& g) {
   Eigen::SelfAdjointEigenSolver<matrix_d> solver(H);
   matrix_d eigenvectors = solver.eigenvectors();
   vector_d eigenvalues = solver.eigenvalues();
   vector_d eigenprojections = eigenvectors.transpose() * g;
+  double max_abs_eigenvalue = eigenvalues.cwiseAbs().maxCoeff();
+  double tolerance
+      = max_abs_eigenvalue * H.rows() * std::numeric_limits<double>::epsilon();
```
nhuurre: Where does this number come from? Is machine epsilon the appropriate baseline here? Why does it depend on the number of rows? Maybe this is related to the numerical precision of the solver...

Also you should handle the case where _all_ eigenvalues are zero. Set some minimum absolute tolerance.

src/stan/optimization/newton.hpp
```diff
@@ -12,20 +14,55 @@ namespace optimization {
...
   for (int i = 0; i < g.size(); i++) {
-    eigenprojections[i] = -eigenprojections[i] / fabs(eigenvalues[i]);
+    double abs_eigenvalue = std::fabs(eigenvalues[i]);
+    if (abs_eigenvalue <= tolerance) {
+      eigenprojections[i] = 0;
```
nhuurre: If target is flat then gradient is zero and what you substitute for the eigenvalue doesn't matter, as long as it's finite. But it's also possible that the target is linear, and if so, the gradient is nonzero while the hessian is still zero. I think you'd want nonzero movement in that case. So instead of zero you should use inverse tolerance. Unlike pseudo-inverse, such "saturating inverse" is continuous.
```suggestion
      eigenprojections[i] = -eigenprojections[i] / tolerance;
```

src/stan/optimization/newton.hpp
```diff
@@ -12,20 +14,55 @@ namespace optimization {
...
 
+/**
+ * Returns true if every element of the vector is finite.
+ *
+ * @tparam Vec vector type with size() and operator[]
+ * @param[in] v vector to check
+ * @return true if all elements are finite
+ */
+template <typename Vec>
+inline bool all_finite(const Vec& v) {
```
nhuurre: This looks like a reimplementation of [`stan::math::is_scal_finite`](https://github.com/stan-dev/math/blob/develop/stan/math/prim/err/is_scal_finite.hpp)

nhuurre:
By the way, and this is a pre-existing issue, but even a simple model like
```stan
parameters {
  vector[3] x;
}
model {
  x ~ normal(0, [10,1,0.1]');
}
```
takes 100 iterations, which is completely unreasonable behavior for a Newton solver. Apparently step size is 1.90735e-06 for every iteration so something is very wrong.

And that something is `stan::model::grad_hess_log_prob`. I didn't look into the details, but I do know that a finite-difference algorithm should divide by epsilon, like, _at some point_, and this one never does.
If I change [these "multiply-by-`half_epsilon`"](https://github.com/stan-dev/stan/blob/e1d4d90014d7199e2da3543665922d7d28041346/src/stan/model/grad_hess_log_prob.hpp#L46-L48) to "divide-by-`half_epsilon`"  (which is the smallest change that makes the algorithm look like it _could_ be correct), stepsize recovers to 1 and the solver converges in "only" 37 iterations.

`stan::model::grad_hess_log_prob` is not used by anything else since Laplace sampler gets the Hessian from `stan::math::internal::finite_diff_hessian_auto`. (I would have thought that that `internal` namespace meant it's not used outside of `math` but whatever)
If I change Newton solver to use `stan::math::internal::finite_diff_hessian_auto` then it converges in 2 iterations, with stepsize 1 for both. This is how Newton _should_ behave on a multinormal target.

WardBrian:
> `stan::model::grad_hess_log_prob` is not used by anything else since Laplace sampler gets the Hessian from `stan::math::internal::finite_diff_hessian_auto`. (I would have thought that that `internal` namespace meant it's not used outside of `math` but whatever)

That [ship may have sailed](https://github.com/roualdes/bridgestan/blob/ddf067ea72007ef569a8ba3b17b297a8dac38bae/src/model.hpp#L474-L475)... Even within math, it's not used by any other code, so I'm not sure why it (or `finite_diff_hessian_times_vector_auto`, which was just following the pattern) is in `internal`

>  If I change Newton solver to use stan::math::internal::finite_diff_hessian_auto then it converges in 2 iterations, with stepsize 1 for both. This is how Newton should behave on a multinormal target.

This sounds like it would be worthy of it's own PR

SteveBronder:
>> `stan::model::grad_hess_log_prob` is not used by anything else since Laplace sampler gets the Hessian from `stan::math::internal::finite_diff_hessian_auto`. (I would have thought that that `internal` namespace meant it's not used outside of `math` but whatever)

> That [ship may have sailed](https://github.com/roualdes/bridgestan/blob/ddf067ea72007ef569a8ba3b17b297a8dac38bae/src/model.hpp#L474-L475)... Even within math, it's not used by any other code, so I'm not sure why it (or `finite_diff_hessian_times_vector_auto`, which was just following the pattern) is in `internal`

For the record, anything inside of math's internal namespace has zero API guarantee aka the math library can change these on a whim with no version notice. This is what Eigen does as well (and partly why upgrading Eigen is such a hassle for us)

WardBrian:
Yeah I understand that. These functions were added entirely on their own with no direct usages in `math`, so I suspect the `internal` was more due to a lack of confidence or something to discourage use rather than an API stability concern

src/stan/optimization/newton.hpp
SteveBronder: So I'm going to be updating this to be more inline with what Eigen has.

```cpp
double max_abs_eigenvalue = eigenvalues.cwiseAbs().maxCoeff();
double tolerance = std::fmax(max_abs_eigenvalue * 4 * H.rows()
                                 * std::numeric_limits<double>::epsilon(),
                             std::numeric_limits<double>::min());
```

I had claude do an explainer on where it got those values from and it gave me the below. I think it is easier to block quote claude here as I think it actually does a decent job of explaining where it got each piece of this.

-----

This is the same cutoff Eigen's rank-revealing decompositions use to decide
whether a pivot or singular value is numerically zero, written out in one
expression. Eigen computes it in two steps: a dimensionless `threshold()`,
and a per-call multiplication by the largest pivot or singular value.

**The bound.** Eigen master's `threshold()` in
[`Eigen/src/misc/RankRevealingBase.h`](https://gitlab.com/libeigen/eigen/-/blob/master/Eigen/src/misc/RankRevealingBase.h#L84)
reads:

```cpp
// Higham's backward error bound: ||ΔA||₂ ≤ c·min(m,n)·u·||A||₂.
// The factor of 4 covers the constant c.
return m_usePrescribedThreshold
           ? m_prescribedThreshold
           : NumTraits<Scalar>::epsilon() * RealScalar(4 * (std::min)(self().rows(), self().cols()));
```

That is the backward error bound for a floating-point factorization from
N. J. Higham, *Accuracy and Stability of Numerical Algorithms*, 2nd ed.,
SIAM, 2002. A perturbation of the input smaller than this bound is
indistinguishable from rounding error in the factorization itself, so a
pivot below it carries no information about rank. Eigen 3.4 uses the same
formula without the factor of 4; the comment there
(`Eigen/src/QR/ColPivHouseholderQR.h:382-383`) says it "turns out to be
identical to Higham's formula used already in LDLt."

**Piece by piece:**

| Term in our code | Term in the bound | Where Eigen does it |
|---|---|---|
| `max_abs_eigenvalue` | `‖A‖₂` | Eigen multiplies `threshold()` by the largest pivot or singular value at the point of use: `abs(m_maxpivot) * threshold()` at `Eigen/src/QR/ColPivHouseholderQR.h:259` and `Eigen/src/LU/FullPivLU.h:334`; `m_singularValues.coeff(0) * threshold()` at `Eigen/src/SVD/SVDBase.h:153`. For a symmetric matrix the 2-norm is the largest absolute eigenvalue, so `eigenvalues.cwiseAbs().maxCoeff()` is the same quantity. |
| `4` | the constant `c` | Eigen master, `RankRevealingBase.h`: "The factor of 4 covers the constant c." |
| `H.rows()` | `min(m, n)` | Eigen master uses `(std::min)(rows(), cols())`; 3.4 uses `m_qr.diagonalSize()`. The Hessian is square, so both equal `H.rows()`. |
| `std::numeric_limits<double>::epsilon()` | `u`, the unit roundoff | Eigen uses `NumTraits<Scalar>::epsilon()`, which for `double` is `std::numeric_limits<double>::epsilon()`. |
| `std::fmax(..., std::numeric_limits<double>::min())` | lower clamp | `Eigen/src/SVD/SVDBase.h:153`: `numext::maxi<RealScalar>(m_singularValues.coeff(0) * threshold(), (std::numeric_limits<RealScalar>::min)())`. This is what makes an all-zero Hessian produce a positive cutoff instead of zero. |
| `abs_eigenvalue <= tolerance` drops the direction | comparison | Eigen counts a pivot toward the rank only when `abs(pivot) > premultiplied_threshold` (`ColPivHouseholderQR.h:262`), so `<=` is treated as zero, matching ours. |

**Precedent outside Eigen.** The same relative cutoff, dimension times
epsilon times the largest singular value or eigenvalue, is the default in
[SciPy's `pinvh`](https://github.com/scipy/scipy/blob/main/scipy/linalg/_basic.py)
(`rtol = max(a.shape) * eps`, applied to the largest absolute eigenvalue,
which is the symmetric eigendecomposition case we have here),
[MATLAB's `pinv`](https://www.mathworks.com/help/matlab/ref/pinv.html)
(`max(size(A)) * eps(norm(A))`), and LAPACK's
[`DGELSD`](https://github.com/Reference-LAPACK/lapack/blob/master/SRC/dgelsd.f)
(singular values `S(i) <= RCOND * S(1)` are treated as zero, with machine
precision used when `RCOND < 0`).

-----

I wish there was an easy way to share claude sessions.

src/stan/optimization/newton.hpp
SteveBronder: Talking to claude I had it cite all of it's choices and explain things. I've cleanup up Claude's explanation below which I think makes sense. 

------

Your suggestion divides by tolerance, which at that point was `n * eps * max(abs(lambda))`. That gives a step along a flat direction of roughly 1e15 times the gradient. Nocedal and Wright discuss exactly this on pp. 49-50 of Numerical Optimization (2nd ed., §3.4, "Eigenvalue Modification"): replacing small eigenvalues with a $\delta$ near machine precision produces a step that "is nearly parallel to q3 ... and quite long. Although f decreases along the direction pk, its extreme length violates the spirit of Newton's method." Their suggested magnitude is $\delta = \sqrt u$.

So the code now floors every eigenvalue magnitude rather than branching, and the floor is $\sqrt u$-scaled:

```cpp
const double sqrt_eps = std::sqrt(std::numeric_limits<double>::epsilon());
double max_abs_eigenvalue = eigenvalues.cwiseAbs().maxCoeff();
double delta = std::fmax(sqrt_eps * max_abs_eigenvalue, sqrt_eps);
for (int i = 0; i < g.size(); i++) {
  eigenprojections[i]
      = -eigenprojections[i] / std::fmax(std::fabs(eigenvalues[i]), delta);
}
```

This is the `max(abs(pivot), delta)` floor from the modified Cholesky factorization, eq. 3.49 on p. 54 of the same book (credited there to Gill, Murray, and Wright, with Moré and Sorensen for the bounded condition number), applied in the eigenbasis. Written that way it is the "PT-inverse" of Paternain, Mokhtari, and Ribeiro, SIAM J. Optim. 29(1) 2019, Definition 2.1 (arXiv:1707.08028 (https://arxiv.org/abs/1707.08028)): $|\lambda_{ii}|$ if $|\lambda_{ii}| \ge m$, else m. It is continuous, as you say, and it also bounds the effective condition number of the solve by $1 / \sqrt u$ ~ 6.7e7, which the old cutoff never did.

On the value of the floor:

- The relative term follows the $\delta = \sqrt u$ on p. 49. R's nlm (the Dennis and Schnabel minimizer in src/appl/uncmin.c (https://raw.githubusercontent.com/wch/r-source/trunk/src/appl/uncmin.c)) uses the same rule: tol = sqrt(epsm) with pivots floored at tol*diagmx, the square root of epsilon times the largest diagonal entry.
- The absolute term is the minimum absolute tolerance you asked for in the other comment, so an all-zero Hessian gives a finite direction. I dropped the numeric_limits::min() clamp because dividing a moderate gradient by it overflows to infinity. Precedent for an absolute clamp on curvature: Ceres' `min_lm_diagonal = 1e-6 on diag(J'J)` (include/ceres/solver.h (https://raw.githubusercontent.com/ceres-solver/ceres-solver/master/include/ceres/solver.h)) and Ipopt's `min_hessian_perturbation` (options (https://coin-or.github.io/Ipopt/OPTIONS.html)). Using √u for the absolute term makes it agree with the relative term when the largest eigenvalue is 1, the "well scaled" assumption Nocedal and Wright use on p. 196.

-----

nhuurre: > Your suggestion divides by tolerance, which at that point was `n * eps * max(abs(lambda))`. That gives a step along a flat direction of roughly 1e15 times the gradient. 

Claude seems to forget that I had also questioned whether that tolerance was appropriate.
And 1e15 times zero ("flat gradient") is still zero.

But yes, this all makes sense.

> Precedent for an absolute clamp on curvature: Ceres' `min_lm_diagonal = 1e-6 on diag(J'J)` and Ipopt's `min_hessian_perturbation`.

Feels like this sentence cuts off early? Maybe should be

> and Ipopt's `min_hessian_perturbation = 1e-20` on the iteration matrix.

src/stan/optimization/newton.hpp
nhuurre: any comment?

SteveBronder: Sorry! yes I need to fixup a few things . We can totally remove this function

SteveBronder:
> By the way, and this is a pre-existing issue, but even a simple model like

If I'm being honest, there are a _lot_ of funky things in this newton optimizer

SteveBronder:
@WardBrian so the cmdstan test is failing, but imo I think we should make this change. 

The difference is that we now use `propto=false` in newton. This is because we only need the gradient and hessian calculation once, while most of the time is spent in a loop where we use doubles everywhere. If we do `propto=true` then the call with just doubles returns 0. So that means we need to use `var` types in the loop when we have no intention of looking at the gradient. I think we should change / remove the cmdstan test and leave `propto=false` here. Thoughts?

WardBrian:
It will be a bit odd that newton and the other optimizers wouldn't agree with each other on a model where they all find the same mode but report a different `lp__`. But I could get over that.

There is also the option of using var types but not actually calling gradient, which would still be better than what the current code does.

nhuurre:
> There is also the option of using var types but not actually calling gradient

That's what [`stan::model::log_prob_propto`](https://github.com/stan-dev/stan/blob/develop/src/stan/model/log_prob_propto.hpp) does.
I'm not sure how expensive calculating all the proportionality constants is compared to the extra cost of the autodiff forward pass. Likely depends on the model; e.g. gaussian constants are cheap, binomial quite expensive.

SteveBronder:
> I'm not sure how expensive calculating all the proportionality constants is compared to the extra cost of the autodiff forward pass. Likely depends on the model; e.g. gaussian constants are cheap, binomial quite expensive.

I think compared to the forward pass I would not imagine them to be very expensive. For instance [normal_lpdf](https://github.com/stan-dev/math/blob/develop/stan/math/prim/prob/normal_lpdf.hpp#L82) is pretty simple 

```
  if constexpr (include_summand<propto>::value) {
    logp += NEG_LOG_SQRT_TWO_PI * N;
  }
  // actually idt this one will even go off because of T_scale
  if constexpr (include_summand<propto, T_scale>::value) {
    logp -= sum(log(sigma_val)) * N / math::size(sigma);
  }
```

I think many of them are kind of like this.

PR #3406, opened 2026-08-14
Fix ambiguity with size_t on 32-bit platforms

andrjohns:
#### Submission Checklist

- [x] Run unit tests: `./runTests.py src/test/unit`
- [x] Run cpplint: `make cpplint`
- [x] Declare copyright holder and open-source license: see below

#### Summary

I'm seeing a compilation failure with some downstream packaging on 32-bit systems:

```
../inst/include/stan/callbacks/structured_writer.hpp:88:16: error: class member cannot be redeclared
   88 |   virtual void write(const std::string& key, unsigned int value) {}
      |                ^
../inst/include/stan/callbacks/structured_writer.hpp:72:16: note: previous definition is here
   72 |   virtual void write(const std::string& key, std::size_t value) {}
```

Which is because `std::size_t` is an `unsigned int` on 32 bit. This PR just changes the overloads to explicit sized-types

#### Intended Effect

Fix compilation on 32-bit

#### Copyright and Licensing

Please list the copyright holder for the work you are submitting (this will be you or your assignee, such as a university or company): Andrew Johnson

By submitting this pull request, the copyright holder is agreeing to license the submitted work under the following licenses:
- Code: BSD 3-clause (https://opensource.org/licenses/BSD-3-Clause)
- Documentation: CC-BY 4.0 (https://creativecommons.org/licenses/by/4.0/)

PR #3434, opened 2026-09-20
Fix further int-size ambiguities

andrjohns:
#### Submission Checklist

#### Summary

Follow-up fix to #3406, the `uint64_t` typedef can be inconsistent between platforms (e.g., `unsigned long` vs `unsigned long long`) which breaks some of the overload resolution in the stan writer, so this just adds the separate overloads to fix support across all

#### Intended Effect

Fix compilation under macos

#### Documentation

Fixed compilation errors under MacOS due to ambiguous int typedefs

#### Copyright and Licensing


src/stan/callbacks/json_writer.hpp
WardBrian, line 1: I agree with marking the overrides as such, but we need to do it for every function that overrides or none, or we get the `-Werror,-Winconsistent-missing-override` in CmdStan

